PR #32, opened 2016-10-05
add not visible matcher

doronpr:


talkol:
I added this code manually, including both exists matcher and not exists

PR #65, opened 2016-12-20
Added replaceText action

yedidyak:


LeoNatan:
@talkol Pls check. Is this something we want to expose in detox, if detox is supposed to simulate e2e, do we want an action not normally possible by the user using normal keyboard / device.

PR #117, opened 2017-04-24
Update README.md

dassir:
pure native android demo documentation  - updated to 'not supported'


dassir:
Agreed. Although, I'm not sure we should keep the 'how to' instructions. Might be confusing for some.

rotemmiz:
Agreed, you can delete the instructions

PR #132, opened 2017-05-19
Architecture overview

DanielMSchmidt:
Hey there, just added a bit of docs on how detox works under the hood, I hope it helps 👍👍


docs/img-src/action-sequence.mmd
```diff
@@ -0,0 +1,24 @@
+sequenceDiagram
+  participant mocha
+  participant Detox Server
+  participant Detox Client
+  participant Earl Grey
+
+  %% Note over mocha,Detox Server: Runs on your computer
+  %% Note over Detox Client,Earl Grey: Runs on the device
+
+  mocha->>Detox Server: element(by.label('Click Me')).tap();
```
rotemmiz: detox is unopinionated re test runner, change mocha to "test runner"
Detox server is actually something else. `detox-server` is a different module that runs a web socket server, letting both tester (detox logic inside a test runner) and testee (detox code inside a device) communicate with each other. Both tester and testee are websocket clients

docs/img-src/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
rotemmiz: I may be too picky, but the img output is not pretty enough. I imagine something slicker :)

docs/img-src/architecture.mmd
```diff
@@ -0,0 +1,11 @@
+%% Ca( for flowchart below
```
rotemmiz: This diagram doesn't make much sense, what's the addition on top of the bigger diagram ?
I think its a bit confusing

docs/img-src/action-sequence.mmd
DanielMSchmidt: Thanks for explaining 👍

docs/img-src/build.sh
DanielMSchmidt: Do you know a better alternative for mermaid? I would love to have something with a source code, because it makes the whole thing extensible to other developers. (in contrast to a PS file or something)

docs/img-src/architecture.mmd
DanielMSchmidt: For me it helped my understand which parts of the system even exist in detox and what their responsibilities are, but I can remove it if you don't like it

docs/img-src/build.sh
rotemmiz: Mermeid might be a good solution, but we need to either restructure the diagram, or at least remove the arrows going from one end point to itself, it makes the diagram much clearer IMO.

DanielMSchmidt: ok, I will try a few more ways and link them here so we can discuss it on a visual basis

DanielMSchmidt:
@rotemmiz I redid the diagram and move a lot of its complexity to text so it is more extensible. I hope to extend it in the future with other kinds of interactions like assertions or device rotation or other things.

docs/Guide.Architecture.md
```diff
@@ -0,0 +1,17 @@
+# Architecture Deep Dive
```
rotemmiz: This is less of an architecture overview and more of a "How Detox Works" document, it is also not really a guide.
After going through the docs again, I think it best fits as a subsection/example in Introduction.HowDetoxWorks.md

docs/img-src/build.sh
rotemmiz: The new way is much nicer!

docs/img-src/action-sequence.mmd
```diff
@@ -0,0 +1,19 @@
...
+  participant Detox(3)
+  participant Detox Server(5)
+  participant Detox Client(7)
+  participant Earl Grey(9)
+
+  %% Note over Test Runner(1),Detox Server(5): Runs on your computer
+  %% Note over Detox Client(7),Earl Grey(9): Runs on the device
+
+  Test Runner(1)->>Detox(3): assertion / action (2)
+  Detox(3)->>Detox Server(5): native code as JSON (4)
```
rotemmiz: I think we should only number the actions, not the participants.

docs/Guide.Contributing.md
```diff
@@ -94,4 +93,8 @@ To run the e2e tests, after the application was built.
 npm run e2e
 ```
 
+### Architecture
+
+For more details on how Detox works under the hood, please see [our Architecture Guide](Guide.Architecture.md)
```
rotemmiz: This relates to "How Detox Works", not a part of Contributing

docs/img-src/action-sequence.mmd
DanielMSchmidt: Can you explain that a bit? would you name Test Runner, Detox, etc only in the text? I would find that a bit confusing if you try to get a general overview of what exactly is happening 👍

DanielMSchmidt:
@rotemmiz Pushed the next iteration, hope it is better suited for merge than the previous one ;)

DanielMSchmidt:
🎉

Issue #131, opened 2017-05-18
[feature] Add support for Orientation Change


DanielMSchmidt:
Hi there,

I would love to test if my drawer behaves as expected also on horizontal mode. 👍 
[I think EarlGrey provides even a method for that.](https://github.com/google/EarlGrey/blob/master/docs/api.md#controlling-device-orientation)

PR #133, opened 2017-05-20
WIP | Device Orientation Manipulation 

DanielMSchmidt:
This PR adds device orientation change support to the `device` API (see #131).
Last thing to do is add some description to the docs, at least from my perspective.

@talkol @rotemmiz Can you give me feedback for the implementation? Is there anything I could do better next time? The docs will follow tomorrow or the day after 👍 

- [x] remove extension to `MethodInvocation.m`
- [x] use earl grey function directly in `setOrientation`
- [x] move JS implementation to `Simulator.js`
- [x] write unit tests
- [x] add docs

### DIFF
--- a/detox/ios/Detox/MethodInvocation.m
+++ b/detox/ios/Detox/MethodInvocation.m
@@ -80,6 +80,24 @@
         if (![value isKindOfClass:[NSDictionary class]]) return nil;
         return [MethodInvocation invoke:value onError:onError];
     }
+	if ([type isEqualToString:@"UIDeviceOrientation"])
+	{
+		if (value == nil) {
+			return nil;
+		}
+		
+		if ([value isEqualToString:@"landscape"])
+		{
+			return [NSNumber numberWithLong:UIDeviceOrientationLandscapeLeft];
+		}
+		
+		if ([value isEqualToString:@"portrait"])
+		{
+			return [NSNumber numberWithLong:UIDeviceOrientationPortrait];
+		}
+		
+		return nil;
+	}
     return nil;
 }
 
--- a/detox/ios/Detox/MethodInvocation.m
+++ b/detox/ios/Detox/MethodInvocation.m
@@ -80,24 +80,6 @@
         if (![value isKindOfClass:[NSDictionary class]]) return nil;
         return [MethodInvocation invoke:value onError:onError];
     }
-	if ([type isEqualToString:@"UIDeviceOrientation"])
-	{
-		if (value == nil) {
-			return nil;
-		}
-		
-		if ([value isEqualToString:@"landscape"])
-		{
-			return [NSNumber numberWithLong:UIDeviceOrientationLandscapeLeft];
-		}
-		
-		if ([value isEqualToString:@"portrait"])
-		{
-			return [NSNumber numberWithLong:UIDeviceOrientationPortrait];
-		}
-		
-		return nil;
-	}
     return nil;
 }
 
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const _ = require('lodash');
 const argparse = require('../utils/argparse');
+const invoke = require('../invoke');
 
 class Device {
   constructor(client, session, deviceConfig) {
@@ -40,6 +41,16 @@ class Device {
 
   async sendUserNotification(params) {
     await this.client.sendUserNotification(params);
+  }
+
+  async setOrientation(orientation) {
+    // orientation is 'landscape' (meaning left side portrait) or 'portrait' (non-reversed)
+    const call = invoke.call(
+      invoke.IOS.EarlGrey(''),
+      'rotateDeviceToOrientation:errorOrNil:',
+      invoke.IOS.UIDeviceOrientation(orientation)
+    );
+    await new invoke.InvocationManager(this.client).execute(call);
   }
 
   async shutdown() {
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -1,7 +1,6 @@
 const fs = require('fs');
 const _ = require('lodash');
 const argparse = require('../utils/argparse');
-const invoke = require('../invoke');
 
 class Device {
   constructor(client, session, deviceConfig) {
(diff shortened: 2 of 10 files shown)


detox/ios/Detox/MethodInvocation.m
```diff
@@ -80,6 +80,24 @@ + (id) getValue:(id)value withType:(id)type onError:(void (^)(NSString*))onError
         if (![value isKindOfClass:[NSDictionary class]]) return nil;
         return [MethodInvocation invoke:value onError:onError];
     }
+	if ([type isEqualToString:@"UIDeviceOrientation"])
```
rotemmiz: The main advantage of the remote invocation mechanism is that we don't need to add any other new call to the protocol

I believe we didn't make the invocation mechanism easily accessible for extension, and this is our fault.

detox/src/devices/Device.js
```diff
@@ -42,6 +43,16 @@ class Device {
     await this.client.sendUserNotification(params);
   }
 
+  async setOrientation(orientation) {
```
rotemmiz: `Device.js` is platform agnostic, it's a base class (will be the base class for Android devices as well). This is a simulator specific logic, hence needs to be in `Simulator.js`

detox/src/devices/Device.js
```diff
@@ -42,6 +43,16 @@ class Device {
...
+    // orientation is 'landscape' (meaning left side portrait) or 'portrait' (non-reversed)
+    const call = invoke.call(
```
rotemmiz: you can invoke the EarlGrey functions directly instead

detox/test/e2e/f-simulator.js
```diff
@@ -28,4 +28,28 @@ describe('Simulator', () => {
     await element(by.label('Say Hello')).tap();
     await expect(element(by.label('Hello!!!'))).toBeVisible();
   });
+
+  describe('device orientation', () => {
```
rotemmiz: awesome stuff!

detox/src/devices/Device.js
rotemmiz: I see there are no unit tests here, the build will fail on lack of coverage.

rotemmiz: and it did 
https://travis-ci.org/wix/detox/builds/235424799#L2820

detox/ios/Detox/MethodInvocation.m
DanielMSchmidt: Does this mean I could have written `invoke.IOS.UIDeviceOrientation('UIDeviceOrientationLandscapeLeft')` or what would the JS call need to look like in this case?

detox/src/devices/Device.js
DanielMSchmidt: Okay, I will move it to `Simulator.js` and add some unit tests

detox/src/devices/Device.js
DanielMSchmidt: So I would do 

```js
invoke.IOS.EarlGrey(
         'rotateDeviceToOrientation:errorOrNil:',
        invoke.IOS.UIDeviceOrientation(orientation)
);
```

or how would it have to look like?

rotemmiz: The ObjC code looks like this:
```ObjC
[EarlGrey rotateDeviceToOrientation:UIDeviceOrientationLandscapeLeft errorOrNil:nil];
```

JS code pass `UIDeviceOrientation` which is an ENUM, so I guess the easiest way is to pass NSInteger with the position of the desired orientation ENUM.

Looking at `UIDevice.h`, which holds these ENUMS

```ObjC
typedef NS_ENUM(NSInteger, UIDeviceOrientation) {
    UIDeviceOrientationUnknown,
    UIDeviceOrientationPortrait,            // Device oriented vertically, home button on the bottom
    UIDeviceOrientationPortraitUpsideDown,  // Device oriented vertically, home button on the top
    UIDeviceOrientationLandscapeLeft,       // Device oriented horizontally, home button on the right
    UIDeviceOrientationLandscapeRight,      // Device oriented horizontally, home button on the left
    UIDeviceOrientationFaceUp,              // Device oriented flat, face up
    UIDeviceOrientationFaceDown             // Device oriented flat, face down
} __TVOS_PROHIBITED;
```

For `UIDeviceOrientationPortrait` pass pass `invoke.IOS.NSInteger(1)` 
For `UIDeviceOrientationLandscapeLeft` pass `invoke.IOS.NSInteger(3)`

I think this should work ...
Of course, for readability, we'd like to keep this ENUM values in JS as well.

So it would look something like this:
```js
invoke.IOS.EarlGrey(
         'rotateDeviceToOrientation:errorOrNil:',
        invoke.IOS.NSInteger(orientationEnum)
);
```

I hope it'll work :)

DanielMSchmidt: Hey, thanks for the insight, got it working without any native extension (yay 🎉 ). Though the call you describe doesn't work, it leads to an object like

```js
{
    type: "EarlGrey",
    value: "rotateDeviceToOrientation:errorOrNil:",
    id: 5
}
```

This leads to a "target is invalid" error, which, to me, looks a bit like a "bug", right?

rotemmiz: I spent try my code, it's reasonable I mistaken.
What is the working syntax then?

DanielMSchmidt: Sure, I just wanted to state that calling 
```js
invoke.IOS.EarlGrey(
         'rotateDeviceToOrientation:errorOrNil:',
        invoke.IOS.NSInteger(orientationEnum)
);
``` 

would be the most readable and easiest way, I will try to make it work, the tests will tell me if I did something wrong then ;)

detox/src/devices/Device.js
```diff
@@ -44,11 +44,18 @@ class Device {
...
+    // keys are possible orientations
+    const orientationMapping = {
+      landscape: 3, // top at left side landscape
+      portrait: 1  // non-reversed portrait
+    };
+    if (!Object.keys(orientationMapping).includes(orientation)) {
+      throw new Error(`setOrientation failed: provided orientation ${orientation} is not part of supported orientations: ${Object.keys(orientationMapping)}`)
+    }
+
+    const call = invoke.EarlGrey.call(
```
rotemmiz: take a look here
```js
this._call = invoke.call(invoke.EarlGrey.instance, 'detox_selectElementWithMatcher:', matcher._call);
```
https://github.com/wix/detox/blob/master/detox/src/ios/expect.js#L229

This function calls the static method`detox_selectElementWithMatcher`  function in `EarlGrey`

detox/src/invoke/EarlGrey.js
```diff
@@ -3,6 +3,18 @@ const instance = {
   value: 'instance'
 };
 
+function call(method, ...args) {
```
rotemmiz: I don't think we need to add a special `call` function here, we are calling EarlGrey's static methods elsewhere without the need for this code.

+ take a look at https://github.com/wix/detox/blob/master/detox/src/invoke/Invoke.js#L25
It's a neat trick using Proxy objects that enables this cool syntax:

```js
invoke.IOS.CGPoint({x: 2, y: 1}))
```
to actually work with no `CGPoint` function ever existing

https://github.com/wix/detox/blob/master/detox/src/invoke.test.js#L19

DanielMSchmidt:
@rotemmiz Can you give me final feedback, this PR is done from my side (at least as far as I can tell 👍 )

PR #167, opened 2017-06-21
WIP: location support

rotemmiz:


detox/src/devices/Device.js
```diff
@@ -38,6 +38,14 @@ class Device {
     return await Promise.resolve('');
   }
 
+  async setLocationPermission(permission) {
```
DanielMSchmidt: could you add some lines in the `docs`?

PR #169, opened 2017-06-22
add takeScreenshot

chirag04:


chirag04:
i couldn't get e2e tests to work locally. some direction would be nice.

chirag04:
@LeoNatan i'm mostly looking to be abel to compare on screenshots. i couldn't find an api to compare screenshots in earlgrey. we can maybe use https://github.com/mapbox/pixelmatch or something? thoughts?

LeoNatan:
Hi,

Thank you for submitting this PR. 
We plan to add support screenshot taking, along with other collected artifacts, with the solution to #171.

I have several problems with the implementation in this PR. Screenshots are converted to base64 during runtime and passed on the web socket — this can be a very heavy task (imagine screenshots from iPad Pro, iPhone Plus devices). It would be preferred to save this screenshots and only obtain them at the end, in a zip file. Also, the Earl Grey screenshot / snapshot API is not good, and I don’t think it should be used to take snapshots. Apple’s  [`-[UIDevice snapshotViewAfterScreenUpdates]`](https://developer.apple.com/documentation/uikit/uiscreen/1617814-snapshotviewafterscreenupdates?language=objc) API is a lot more optimal and returns a much truer representation of what is displayed on screen (for instance, visual effect views preserve their effects).

At this point, I am inclined to close this PR. It will take us a little longer to implement, but we want to get it right for all artifact types.

LeoNatan:
Regarding image comparison, we plan to have analytics of tests over time. We could fit this functionality there.

PR #184, opened 2017-07-03
Switched to booting the device directly (related to https://github.co…

silyevsk:
…m/facebook/FBSimulatorControl/issues/423)


detox/src/devices/Fbsimctl.js
```diff
@@ -39,13 +40,37 @@ class Fbsimctl {
...
+      log.info(`The device ${udid} is already booted`);
+      return;
+    }
+    
+    if(initialState === 'Booting') {
+      log.info(`The device ${udid} is already booting`);
+    } else {
+      const launchBin = "bash -c '`xcode-select -p`/Applications/Simulator.app/Contents/MacOS/Simulator " +
+                        "--args -CurrentDeviceUDID EB9B3A90-621B-4F6E-815C-1E5C17ED5ABE -ConnectHardwareKeyboard 0 " +
+                        "-DeviceSetPath /Users/sergeyi/Library/Developer/CoreSimulator/Devices > /dev/null 2>&1 < /dev/null &'";
```
rotemmiz: Seems like you forgot to use a relative path

rotemmiz: and CurrentDeviceUDID

PR #189, opened 2017-07-08
Add support for tapAtPoint action

blankg:
Motivation for adding this api - we are migrating our mobile app from a hybrid Cordova app to react-native, our first milestone is moving to react-native, getting rid of Cordova but still staying in the WebView and then gradually move our UI into react-native.
I would like to use detox for a very basic e2e sanity test and without the tapAtPoint api we cannot do that while most of our app is in a WebView.
Also this API can be useful for other scenarios I guess (tapping on maps and such) otherwise it won't be in EarlGrey :)


rotemmiz:
Hey @blankg,
Thanks for the contribution!

We were actually having a debate regarding this PR. Although EarlGrey offers this API, we were not sure if it makes any sense when used on full screen elements, since the tap point will be change on different devices (it will even change  if you decide to change the font size in your system).
This means we can't trust the same test will pass on all devices.

How are you planning to use it in your tests ? have you already tried it ?

blankg:
Hi @rotemmiz, I figured this was the reason you didn't expose this api at the first place.
I would not build an extensive test suite based on such an API but what we are looking for is running a very basic e2e sanity test (just a couple of clicks to see that the app is not broken).
It seems to work fine with the tap at point api.
I think this api could be useful for others migrating from cordova to react-native, which will probably mean staying with the WebView in some parts for a while.

I could added some disclaimers in the api documentation mentioning what you have written above, what do you say?

rotemmiz:
Sounds good, let's do it.

blankg:
@rotemmiz, updated the documentation. Let me know if you have some different phrasing in mind.

rotemmiz:
PR is now merged, thanks @blankg !

PR #195, opened 2017-07-12
WIP: Location support

yogevbd:


rotemmiz:
@silyevsk , we need to investigate why our idling resource doesn't synchronize the async request for location from the app :
```js
 navigator.geolocation.getCurrentPosition(success.bind(this), error.bind(this));
```

silyevsk:
navigator.geolocation.getCurrentPosition is an sync function, the expectation in the was checked before there was a return from getCurrentPosition. I've changed the screen/test a bit to wrap it with async/await

PR #200, opened 2017-07-16
App logs

silyevsk:


detox/src/utils/ArtifactsPathsProvider.js
```diff
@@ -0,0 +1,36 @@
+const fs = require('fs');
+const _ = require('lodash');
+const log = require('npmlog');
+
+class ArtifactsPathsProvider {
+  constructor(destinationParent) {
+    if(!destinationParent) {
+      throw 'destinationParent should not be undefined';
```
rotemmiz: throw new Error(''). I think this error string is not descriptive enough.

detox/src/devices/Device.js
```diff
@@ -3,15 +3,26 @@ const path = require('path');
...
   constructor(deviceConfig, sessionConfig, deviceDriver) {
     this._deviceConfig = deviceConfig;
     this._sessionConfig = sessionConfig;
     this.deviceDriver = deviceDriver;
+    this._currentLaunchNumber = 0;
+    this._currentTestArtifactsDestination = undefined;
 
     this.deviceDriver.validateDeviceConfig(deviceConfig);
+
+    this.relaunchApp = this.relaunchApp.bind(this);
```
rotemmiz: Why do we need all those binds ?

detox/package.json
```diff
@@ -81,6 +82,10 @@
         "functions": 100,
         "lines": 100
       }
-    }
+    },
```
rotemmiz: Not sure excluding `src/index.js` is the right thing. This is something we can test, it's actually important that we do, till we have a sane type system, we want to make sure beforeEach and afterEach exist and that they don't crash the test if detox is undefined.

detox/src/devices/Device.js
```diff
@@ -23,7 +34,52 @@ class Device {
     await this.relaunchApp({delete: !argparse.getArgValue('reuse')});
   }
 
+  async _copyArtifacts() {
```
rotemmiz: I think this is not the job of `Device` to move artifacts around. Why not merge all the logic in an `ArtifactsHelper` as you previously suggested ?

detox/src/Detox.js
```diff
@@ -30,6 +31,18 @@ class Detox {
     this.userConfig = userConfig;
     this.client = null;
     this.device = null;
+    this._currentTestNumber = 0;
+    const artifactsLocation = argparse.getArgValue('artifacts-location');
+    if(artifactsLocation !== undefined) {
```
silyevsk: That's unsafe in JS in general due to implicit type conversions, so I think it's better to never use it.. Here for instance, if a user supplies an empty string, like `--artifacts-location ''`, then JS will convert `artifactsLocation` to `false`.

detox/src/devices/Device.js
silyevsk: I probably don't need all of them, only those whose this is incorrect and is called from outside this. I guess I can reduce the list, but some will have to stay there

detox/package.json
silyevsk: Oops, committed this by mistake, I actually added the the test for them, but forgot to remove the exclusion, will revert it.

detox/src/Detox.test.js
```diff
@@ -192,6 +193,47 @@ describe('Detox', () => {
...
+    Detox = require('./Detox');
+    detox = new Detox(schemes.validOneDeviceAndSession);
+    await detox.init();
+    await detox.afterEach();
+    expect(device.finalizeArtifacts).toHaveBeenCalledTimes(0);
+  });
+
+  it(`the constructor should catch exception from ArtifactsPathsProvider`, async () => {
+    mockCommandLineArgs({'artifacts-location': '/tmp'});
+    fs.mkdirSync = jest.fn(() => {throw 'Could not create artifacts root dir'});
```
silyevsk: changed

detox/src/utils/ArtifactsPathsProvider.js
silyevsk: changed

PR #217, opened 2017-07-31
device: add `resetContentAndSettings`

formatlos:
I added a new method to device to cleanup the device (like "Simulator > Reset Content and Settings" menu-item does), because I want to test specific screens which are only shown when the permissions are not set yet (e.g. a screen for explaining why the specific permission is needed)

### DIFF
--- a/detox/src/devices/Fbsimctl.js
+++ b/detox/src/devices/Fbsimctl.js
@@ -173,6 +173,14 @@ class Fbsimctl {
     await this._execFbsimctlCommand(options);
   }
 
+  async resetContentAndSettings(udid) {
+    await this.shutdown(udid);
+    const result = await exec.execWithRetriesAndLogs(`/usr/bin/xcrun simctl erase ${udid}`);
+    const resultCode = parseInt(result.stdout.trim().split(':')[1]);
+    await this.boot(udid);
+    return resultCode;
+  }
+
   async _execFbsimctlCommand(options, statusLogs, retries, interval) {
     const bin = `fbsimctl --json`;
     return await exec.execWithRetriesAndLogs(bin, options, statusLogs, retries, interval);
--- a/detox/test/e2e/f-device.js
+++ b/detox/test/e2e/f-device.js
@@ -1,4 +1,4 @@
-describe.only('Device', () => {
+describe('Device', () => {
   it('reloadReactNative - should tap successfully', async () => {
     await device.reloadReactNative();
     await element(by.label('Sanity')).tap();
--- a/detox/test/e2e/f-device.js
+++ b/detox/test/e2e/f-device.js
@@ -1,4 +1,4 @@
-describe('Device', () => {
+describe.only('Device', () => {
   it('reloadReactNative - should tap successfully', async () => {
     await device.reloadReactNative();
     await element(by.label('Sanity')).tap();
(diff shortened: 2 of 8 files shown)


detox/src/devices/Fbsimctl.js
```diff
@@ -173,6 +173,14 @@ class Fbsimctl {
     await this._execFbsimctlCommand(options);
   }
 
+  async resetContentAndSettings(udid) {
```
LeoNatan: I think this is better suited as an input param to [device.launchApp()](https://github.com/wix/detox/blob/master/docs/APIRef.DeviceObjectAPI.md#devicelaunchappparams), inline with other API we have.

LeoNatan: CC @rotemmiz

formatlos: how would you name that param? `clean`?

rotemmiz: Not sure it fits `launchApp` since it has no relation to the app, it's more of a device function (like `setLocation` or `setOrientation`)

rotemmiz: `clean` can be confusing, since a user might think this option cleans the app, not resets the entire simulator.

LeoNatan: `resetContentAndSettings` is fine for a name.
I don’t feel comfortable with having UDID of the sim in the test? Maybe it should be `device.resetContentAndSettings()` and that will reset the current device.

rotemmiz: that's what @formatlos suggested, the UDID is not a part of the API on any function, but is used internally (each  `Device.js` instance has it's own deviceId)

formatlos: yes it's basically the same than for example `setLocation`

LeoNatan: 👍

formatlos: so is it alright then or do you want me to change something?

formatlos: @LeoNatan just wanted to know if there are still concerns about the PR?

LeoNatan: I don't think so. Need @rotemmiz to approve it, as he is the JS master around here. 😄

detox/test/e2e/f-device.js
```diff
@@ -1,4 +1,4 @@
-describe('Device', () => {
+describe.only('Device', () => {
```
rotemmiz: don't forget to remove the `only`

detox/src/devices/Fbsimctl.js
rotemmiz: Never wished for this title :( ... 
Let's talk a bit about the API:
1. I don't this is not a good API for reseting a permission, I much prefer if we can add an unset state to permissions with `appleSimUtils` https://github.com/wix/AppleSimulatorUtils/issues/8
2. Can there still can be a valid test case for resetting a device?
3. Since we aim Detox to be multi-platform framework I prefer not to give this method a name which is strongly recognized with Apple's terms, let's try thinking of something more generic maybe, like `reset` or `clearData` (not sure if these are better names). WDYT ?

detox/test/e2e/f-device.js
formatlos: oops, good point

formatlos:
1. 
Well that's why I didn't call it `unsetPermission` whatever ... So it's not just about resetting the permissions, that was just one example. Even more important for me is to be sure to start with a clean state of the simulator. Once in a while my detox tests start to fail, because the simulator seems to be in a "dirty" state (after `resetContentAndSettings ` everything runs as expected again).
Another thing would be to clean up the bundle Documents folder, local DBs, etc. 

2. 
good question

3. 
In general I also don't like the name being tied to one specific Platform. I also thought about `reset` or something along the lines in the first place, but in the end decided against it for one specific reason: It's not just resetting this one app and not just cleaning the data, it's resetting the whole simulator to a clean state. 
Also there might not even be a similar thing in android, so I thought it's better to stick with the ios term and not try to introduce new nomenclature for something that doesn't exist platform independent anyway. 
Apart from this concerns I'm also fine with renaming it...

PR #218, opened 2017-07-31
fix waitFor

doronpr:
joint effort with @aarongreenwald

### DIFF
--- a/detox/src/ios/expect.js
+++ b/detox/src/ios/expect.js
@@ -212,8 +212,10 @@ class WaitForInteraction extends Interaction {
     if (this._notCondition) {
       _conditionCall = invoke.call(invoke.IOS.Class('GREYCondition'), 'detoxConditionForNotElementMatched:', this._element._call);
     }
+
     this._call = invoke.call(_conditionCall, 'waitWithTimeout:', invoke.IOS.CGFloat(timeout/1000));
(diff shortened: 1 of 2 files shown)


LeoNatan:
Build failed. https://travis-ci.org/wix/detox/builds/259291191#L2316 @doronpr @aarongreenwald

detox/src/ios/expect.js
```diff
@@ -212,8 +212,10 @@ class WaitForInteraction extends Interaction {
     if (this._notCondition) {
       _conditionCall = invoke.call(invoke.IOS.Class('GREYCondition'), 'detoxConditionForNotElementMatched:', this._element._call);
     }
+
     this._call = invoke.call(_conditionCall, 'waitWithTimeout:', invoke.IOS.CGFloat(timeout/1000));
     await this.execute();
+    await new MatcherAssertionInteraction(this._element, this._originalMatcher).execute();
   }
   whileElement(searchMatcher) {
```
rotemmiz: I believe `waitFor...whileElement` doesn't fail as well, it needs to be consistent with `waitFor...withTimeout`

rotemmiz: Also, `not` conditions are not handled properly. We'll need to fix them as well.

PR #219, opened 2017-08-01
convert setLocation params to string with comma as decimal separator

formatlos:
using numbers with decimals for lat/long in setLocation results in an error, it has to be passed as string with a comma as decimal separator to `fbsimctl`

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -140,8 +140,8 @@ class Device {
   }
 
   async setLocation(lat, lon) {
-    lat = (lat + '').replace('.', ',');
-    lon = (lon + '').replace('.', ',');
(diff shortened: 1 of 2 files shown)


detox/src/devices/Device.js
```diff
@@ -140,6 +140,8 @@ class Device {
   }
 
   async setLocation(lat, lon) {
+    lat = (lat + '').replace('.', ',');
+    lon = (lon + '').replace('.', ',');
```
isnifer:
```js
lat = String(lat).replace('.', ',');
lon = String(lon).replace('.', ',');
```

rotemmiz: Thanks for the contribution!
If possible, please use `String()` as @isnifer's suggests. We will merge it ASAP.

PR #233, opened 2017-08-15
Cleanup refactor

simonracz:
Main change that could affect iOS is that the cleanup command now sends a parameter `stopRunner`. iOS can simply ignore it.

PR to let Travis do its job.

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
@@ -11,6 +11,9 @@ import com.wix.detox.espresso.UiAutomatorHelper;
 import com.wix.detox.systeminfo.Environment;
 import com.wix.invoke.MethodInvocation;
 
+import org.json.JSONException;
+import org.json.JSONObject;
+
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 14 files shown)


detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
```diff
@@ -119,7 +122,16 @@ public void run() {
...
-                        stop();
+                        try {
+                            boolean stopRunner = new JSONObject(params).getBoolean("stopRunner");
+                            if (stopRunner) {
+                                stop();
+                            } else {
+                                ReactNativeSupport.removeEspressoIdlingResources(reactNativeHostHolder);
+                            }
+                        } catch (JSONException e) {
+                            Log.e(LOG_TAG, "cleanup cmd doesn't have stopRunner param");
```
d4vidi: @simonracz I'm working on refactoring this switch/case clause (almost done), and couldn't help but wonder whether there was a particular reason no to fallback to a `stopRunner=false` behavior in case `stopRunner` was not provided. It's easy to do so using `optBoolean("stopRunner", false)` (instead of `getBoolean()`). I'd appreciate your input :)

simonracz: Hi @d4vidi ,

The short answer is, I don't remember. :)

In general treat my code as POC. So, feel free to refactor it and double check my decisions.

For this specific case, I think it would be safe to go with `optBoolean` here.

Cheers,
Simon

d4vidi: Sounds about right :) so what's left now is to decided whether the default should be `true` or `false` :)

Issue #175, opened 2017-06-26
Add support for a more robust detox config


simonracz:
The feature is described in a google docs [here](https://docs.google.com/document/d/1z2ENgzqNRHxxUfh81Illhf9Thjjgt3tGurPFg77bmtg/edit?usp=sharing)

rumax:
Will it be possible to use system variables too?

For example
```
export DETOX_DEVICE="iPhone 5"
export DETOX_DEVICE_CI="iPhone 7"
```

DanielMSchmidt:
Do we want to support a config file too? That could be used to configure the behavior more dynamically.

rotemmiz:
the config can be loaded from anywhere, that's how we currently do it
```js
const config = require('../package.json').detox;
```

DanielMSchmidt:
So ENV variables could be used there to overwrite, right?

rotemmiz:
not sure what you mean, ENV vars from where ?

DanielMSchmidt:
Sorry 🙈  @rumax mentions using environment variables to overwrite parts of the config. This overwrite part could be used in the generated file that includes `const config = require('../package.json').detox;`

By the way, what would you think of using the [jest-config]() package. That way we would use a way that might feel known and we could reuse that module. But that could be done separate from changing the structure of the config. 

I would now go ahead and try to build a first draft of the config described in the doc and we can think about ways to configure in the meantime 👍

DanielMSchmidt:
After reading the proposal again I would like to add a `name` field on the topLevel, that way we could have better defaults for the binaryPaths. I would default the`name` with the name field from the package.json. That way we could have almost zero configuration (except simulators, but we could maybe default to the first one in the list there too, to be zero config by default)

DanielMSchmidt:
@simonracz Regarding preparing the `parallel` feature, would you suggest to run multiple simulators in parallel that run the same test on each simulator (e.g. testing iPad & iPhone side by side) or would you rather split up the tests to run the suite faster (e.g. two iPhones 5 side by side) Maybe we could open up a new issue to keep track of the ideas in this direction.

rotemmiz:
The first approach is pretty easy (testing the same on multiple devices), but that's not what I think will bring much value.
Being able to shard tests into multiple identical devices is a much more complex task (especially if we want to to keep the API as is), I think that its value is much bigger. We already have test suites that take over 10 minutes, I believe that parallelizing tests can slice that by more than half.

[...]

PR #268, opened 2017-09-08
Change to decoupled binary config format

DanielMSchmidt:
This allows us to have a less verbose config, tackling the first part of #175.
This will introduce a breaking change (`appName` field), we might want to add a warning first instead of an error.

### DIFF
--- a/detox/src/Detox.js
+++ b/detox/src/Detox.js
@@ -66,8 +66,13 @@ class Detox {
       throw new Error(`'${deviceConfig.type}' is not supported`);
     }
 
+    const appName = this.userConfig.appName;
+    if (!appName) {
+      configuration.throwOnEmptyAppName();
+    }
+
     const deviceDriver = new deviceClass(this.client);
-    this.device = new Device(deviceConfig, sessionConfig, deviceDriver);
+    this.device = new Device(deviceConfig, sessionConfig, deviceDriver, appName, this.userConfig.binary);
     await this.device.prepare(params);
     global.device = this.device;
   }
--- a/detox/src/Detox.js
+++ b/detox/src/Detox.js
@@ -67,21 +67,38 @@ class Detox {
       throw new Error(`'${deviceConfig.type}' is not supported`);
     }
     const deviceDriver = new deviceClass(this.client);
+    const binaryPath = await this._getBinaryPath(deviceConfig, deviceDriver);
 
-    let appName = this.userConfig.appName;
-    if (!appName) {
-      log.info("appName was not found in config, we will set it for you");
-      try {
-        appName = await appContext.getAppName(deviceDriver.getPlatform());
-        log.info(`Got the appName, its "${appName}". If this is wrong, please set the appName config property`);
-      } catch (e) {
-        throw new Error("You neither set the appName, nor could we find it anywhere. Please set it in your configuration.");
-      }
-    }
-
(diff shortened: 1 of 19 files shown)


rotemmiz:
This diff is huge, all the reformatting make it very hard to actually go over the code.
I don't mind the reformatting, but let's do it in a different PR (or at least different commit).

Couldn't conclude if there are any breaking changes, but if there are, let's introduce them all together and not bit by bit.

DanielMSchmidt:
Ups, sorry that was automatic reformatting, I will redo the commit without.
I am trying to extend the PR to not introduce breaking changes and to make
a step into the direction of zero config tooling, stay tuned ;)

DanielMSchmidt:
I just added a second commit which infers the appName if it's not set. This leads to
- a step into the direction of zero configuration
- hopefully no breaking change (if these functions work fine)

I will think about another way to refer to these fallbacks as late as possible so that we don't break detox for anyone, haven't got a clear idea yet.

DanielMSchmidt:
@rotemmiz This is a more final version, it should not break the usage for anyone, still lets not merge it yet. Still thinking about a clean way to support the current version without relying on the automatic finding of the app name to work

detox/src/Detox.js
```diff
@@ -65,13 +66,41 @@ class Detox {
...
+    }
+  }
+
+  async _getBinaryPath(deviceConfig, deviceDriver) {
+    if (deviceConfig.binaryPath) {
+      return deviceConfig.binaryPath;
+    }
+
+    const binaryPaths = this.userConfig.binary || {};
+    const defaultBinaryPathOverwrite = binaryPaths[deviceDriver.getPlatform()];
```
rotemmiz: Instead, why not use `deviceDriver.getBinaryPath()` ?

detox/src/utils/appContext.mock.js
```diff
@@ -0,0 +1,116 @@
...
+{
+	[RCTPushNotificationManager didReceiveLocalNotification:notification];
+}
+
+@end
+`;
+
+const androidFileContent = `
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.myAndroidApp"
```
rotemmiz: This is not necessarily the final package name of the app. The source of truth is in the app/build.gradle.

detox/src/Detox.js
```diff
@@ -65,13 +66,41 @@ class Detox {
     if (!deviceClass) {
       throw new Error(`'${deviceConfig.type}' is not supported`);
     }
-
     const deviceDriver = new deviceClass(this.client);
-    this.device = new Device(deviceConfig, sessionConfig, deviceDriver);
+    const binaryPath = await this._getBinaryPath(deviceConfig, deviceDriver);
```
rotemmiz: BinaryPath should be contained inside deviceDriver IMO

detox/src/Detox.js
```diff
@@ -65,13 +66,41 @@ class Detox {
...
+
+    this.device = new Device(deviceConfig, sessionConfig, deviceDriver, binaryPath);
     await this.device.prepare(params);
     global.device = this.device;
   }
 
+  async _getAppName(deviceDriver) {
```
rotemmiz: I have a feeling that trying to extract the appName from project files will get us in lots of trouble. There are endless possibilities for configuring projects (especially if there are multiple targets on iOS or multiple buildFlavors on Android, and `build.gradle` might be in custom directories (not necessarily in `/app/`

PR #293, opened 2017-09-26
[Android] label -> contentDescription

simonracz:
I rewrote `label` handling on Android to make it search by `contentDescription` instead of `text`.

This showed that the detox test suite is actually tied to iOS. Therefore, I had to touch nearly all screens and tests to make it cross-platform.

iOS vs Android example
A `TextView` has no default `contentDescription` on Android. From RN's perspective this means that you have to set `accessibilityLabel` directly on a `<Text>`. On iOS this works without this.

I also updated the docs and `detox init`'s default e2e file.

### DIFF
--- a/detox/test/android/app/src/main/AndroidManifest.xml
+++ b/detox/test/android/app/src/main/AndroidManifest.xml
@@ -5,6 +5,7 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
 
     <uses-sdk
         android:minSdkVersion="16"
--- a/detox/test/src/Screens/ActionsScreen.js
+++ b/detox/test/src/Screens/ActionsScreen.js
@@ -29,12 +29,12 @@ export default class ActionsScreen extends Component {
         <TouchableOpacity onPress={this.onButtonPress.bind(this, 'Tap Working')}
           onLongPress={this.onButtonPress.bind(this, 'Long Press Working')}
           >
-          <Text style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }} accessibilityLabel={'Tap Me'}>Tap Me</Text>
+          <Text style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }}>Tap Me</Text>
         </TouchableOpacity>
 
         <TouchableOpacity onPress={this.onLongTimeout.bind(this)}
           >
-          <Text testID='WhyDoAllTheTestIDsHaveTheseStrangeNames' style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }} accessibilityLabel={'Tap Me For Long Timeout'}>Tap Me For Long Timeout</Text>
+          <Text testID='WhyDoAllTheTestIDsHaveTheseStrangeNames' style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }}>Tap Me For Long Timeout</Text>
         </TouchableOpacity>
 
         <TouchableOpacity onPress={this.onMultiTapPress.bind(this)}>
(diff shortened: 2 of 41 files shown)


detox/test/src/Screens/MatchersScreen.js
```diff
@@ -21,11 +21,11 @@ export default class MatchersScreen extends Component {
       <View style={{flex: 1, paddingTop: 20, justifyContent: 'center', alignItems: 'center'}}>
 
         <TouchableOpacity onPress={this.onButtonPress.bind(this, 'Label Working')}>
-          <Text style={{color: 'blue', marginBottom: 20}}>Label</Text>
+          <Text style={{color: 'blue', marginBottom: 20}} accessibilityLabel={'Label'}>Label</Text>
         </TouchableOpacity>
 
         <TouchableOpacity onPress={this.onButtonPress.bind(this, 'ID Working')}>
-          <Text testID='UniqueId345' style={{color: 'blue', marginBottom: 20}}>ID</Text>
+          <Text testID='UniqueId345' style={{color: 'blue', marginBottom: 20}} accessibilityLabel={'ID'}>ID</Text>
```
isnifer: @simonracz explain, please, why `testID` and `accessibilityLabel` have different labels? I'm using [`react-native-testid`](https://github.com/isnifer/react-native-testid) at Tipsi. And we have both parameters the same

simonracz: AccessibilityLabel is user visible. TestId is not.

detox/test/src/Screens/ActionsScreen.js
```diff
@@ -29,7 +29,7 @@ export default class ActionsScreen extends Component {
         <TouchableOpacity onPress={this.onButtonPress.bind(this, 'Tap Working')}
           onLongPress={this.onButtonPress.bind(this, 'Long Press Working')}
           >
-          <Text style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }}>Tap Me</Text>
+          <Text style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }} acccessibilityLabel={'Tap Me'}>Tap Me</Text>
```
LeoNatan: I don’t like these changes.
Can you please explain why we need to require users to add `accessibilityLabel` on each element?

On iOS, `accessibilityLabel` is used by Voice Over, for example, and is not always linked 1:1 with the content. The developer can certainly change it if they feel they can provide a better label for the content.
How does this work on Android?

Perhaps we should change the tests on iOS to `by.text(‘’)`, but requiring users to add `accessibilityLabel` on each element is not something I can live with.

Also please note you misspelled “accessibilityLabel” here (“ccc”).

docs/APIRef.Expect.md
```diff
@@ -52,6 +53,13 @@ await expect(element(by.id('RandomJunk959'))).toNotExist();
 await expect(element(by.id('UniqueId204'))).toHaveText('I contain some text');
 ```
 
+### `toHaveLabel(label)`
+- Similar to `toHaveText(text)`, but searches by accessibilityLabel (iOS) or by contentDescription (Android)
```
LeoNatan: The two shouldn’t be linked in such nonchalant manner. On iOS, text and label can have multiple meanings, and they wouldn’t be similar in cases where users care about the accessibility of their application. The documentation should split those completely.

docs/APIRef.Matchers.md
```diff
@@ -42,14 +43,22 @@ Find an element by text, useful for text fields, buttons.
 ```js
 await element(by.text('Tap Me'));
 ```
+
+#### `by.label(label)`
+Find an element by accessibilityLabel(iOS) or contentDescription(Android), useful for text fields, buttons.
```
LeoNatan: Accessibility label is useful for almost any UI element on iOS. As [per documentation](https://developer.apple.com/documentation/objectivec/nsobject/1615181-accessibilitylabel?language=objc), this should be properly set to any element with custom appearance, such as images. The system helps by trying to set a value to this property be similar to the title of elements such as text fields or buttons, but also for other elements. The “useful” comment implies incorrect here.

detox/test/src/Screens/WaitForScreen.js
```diff
@@ -38,14 +38,14 @@ export default class WaitForScreen extends Component {
...
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text1'}>Text1</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text2'}>Text2</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text3'}>Text3</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text4'}>Text4</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text5'}>Text5</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text6'}>Text6</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text7'}>Text7</Text>
+            <Text style={{height: 30, backgroundColor: '#e8e8f8', padding: 5, margin: 10}} accessibilityLabel={'Text8'}>Text8</Text>
           </ScrollView>
```
LeoNatan: I still don’t understand why, if tests are now `by.text()`, why this label was added to all views?

detox/test/src/Screens/MatchersScreen.js
LeoNatan: @isnifer Bad idea.
https://github.com/isnifer/react-native-testid/issues/1

erezrokah:
Maybe you'll find the discussion(s) here helpful:

https://github.com/facebook/react-native/issues/7135
https://github.com/facebook/react-native/issues/9777

LeoNatan:
People abusing `accessibleLabel` makes me sad. There is so little awareness of what accessibility is, and why it is important. That’s true on the web, too, so there is no excuse.

docs/APIRef.Expect.md
simonracz: You probably read much more out of the "similar" word that was meant there. But, I agree, I'll rewrite it.

docs/APIRef.Matchers.md
simonracz: Ok. I'll remove the useful part.

detox/test/src/Screens/WaitForScreen.js
simonracz: Not all tests search by text, even after this commit.

I followed the following pattern:
For fixed elements, I set accessibilityLabel. For dynamic elements I didn't. (The reason behind this was that blind people could read out the structure of the app from the fixed elements.)

I can remove nearly all accessibilityLabels and search only by text if you prefer. (I'll leave the accessibilityLabel for the Matcher tests, where we explicitly test for accessibilityLabel.)

LeoNatan: I think this will be best. What do you think? It will also show best practice to users that wonder how to use the API.

docs/APIRef.Expect.md
LeoNatan: My issue is that some users l, particularly in RN ecosystem, are not familiar with iOS concepts, and learn half-truths because of imprecise text they see here and there. I’m sure it’s the same for Android. This is why I try to push for accuracy, so everyone gets the correct picture.

detox/test/src/Screens/WaitForScreen.js
simonracz: I think detox users should almost always search by `text`. And only use `label`, if they really want to search by accessibilityLabel/contentDescription.

This is why I found it important to also change the docs, sample codes and `detox init` e2e file.

LeoNatan: 100% agree. That’s why I think all the labels in the test are unnecessary. 

But the most accurate is to match by identifier. By text is problematic if you use localization.

simonracz:
I addressed the comments in the latest commit.

detox/test/android/app/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
```
mfrachet: Can't get the point of this line in this PR

simonracz: It was missing for an Android test. (Which is not triggered by CI yet.)

The whole point of the PR was to make the test suite cross platform.

PR #315, opened 2017-10-05
Docs: Jest usage

DanielMSchmidt:
Heavily inspired by the solution of @AlanFoster and with the help of @levito

### DIFF
--- a/docs/Guide.Jest.md
+++ b/docs/Guide.Jest.md
@@ -1,0 +1,63 @@
+# Jest as Test Runner
+
+## Usage
+
+### 0. Use the [Getting Started](Introduction.GettingStarted.md) Guide to set up detox
+
+Except that you need to skip the install mocha step.
+
+### 1. Install Jest
+
+```sh
+npm install --save-dev jest
+```
+
+### 2. Remove mocha specific files
+
+You should remove `e2e/mocha.opts`, you no longer need it.
+
+### 3. Write a detox setup file
+
+```js
+const detox = require('detox');
+const config = require('../package.json').detox;
+
+// Set the default timeout
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000;
+
+beforeAll(async () => {
+  await detox.init(config);
+});
+
+afterAll(async () => {
+  await detox.cleanup();
+});
+
+// optional, you may remove this part
+beforeEach(async () => {
+  await device.reloadReactNative();
+});
+```
+
+### 4. Run jest
+
+Add this part to your `package.json`:
+```json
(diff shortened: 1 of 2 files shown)


docs/Guide.Jest.md
```diff
@@ -0,0 +1,57 @@
...
+  await device.reloadReactNative();
+});
+```
+
+### 4. Run jest
+
+Add this part to your `package.json`:
+```json
+"scripts": {
+    "test:e2e": "jest e2e --setupTestFrameworkScriptFile=./jest/setup-e2e-tests.js"
```
chirag04: I like how you can define jest config in your package.json so you don't have to pass in `--setupTestFrameworkScriptFile`. Just making a note. either way is fine. 

```
"jest": {
    "bail": true,
    "setupTestFrameworkScriptFile": "<rootDir>/jest/setup.js"
  },
```

rotemmiz:
I would add a note re Detox that it exports expect, waitFor, element and by, so Jest users might find themselves using a different expect object implicitly.

rotemmiz:
BTW, we plan on adding another detox.init flavour/flag that will not export all that into globals, and users would need to explicitly import it. It would make lots of sense for Jest.

That might be an AWESOME opportunity for a PR (anyone? hint hint...)

docs/Guide.Jest.md
DanielMSchmidt: This is intentional because we wanted to make sure that they run side by side with your normal unit tests

chirag04: sure. either way is fine. you can push that logic in setup file too: https://github.com/grabbou/react-native-detox-example/blob/master/jest/setup.js#L10-L14

DanielMSchmidt:
@rotemmiz Noted the exposure of expect and so on
@chirag04 Thanks for the note, but this was intentional to have test run side by side with the unit tests

docs/Guide.Jest.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+We need the `--runInBand` as detox doesn't support parallelism yet.
+
+### Writing Tests
+
+There are some things you should notice:
+
+- Don't worry about mocks being used, detox works on the compiled version of your app.
+- Detox exposes it's primitives (`expect`, `device`, ...) globally
```
rotemmiz: Detox exposes it's primitives (`expect`, `device`, ...) globally, it will override Jest's global `expect` object.

DanielMSchmidt:
@rotemmiz I will merge this, implemented your wish. Feel free to edit and improve 👍

PR #316, opened 2017-10-05
Add documentation for the usage with Android

DanielMSchmidt:
This should enable other users to use android more
easily as the current status configuration can be
quite confusing.

Build with the help of @appsthatmatter

Relates to #96


simonracz:
Awesome work! 👍

PR #321, opened 2017-10-06
Fix Android atIndex matcher by fixing typo

pietropizzi:
There was a typo in the name of the java method. It was 'matherForAtIndex' instead of 'mat**c**herForAtIndex'


detox/src/android/expect.js
```diff
@@ -218,7 +218,7 @@ class Element {
   atIndex(index) {
     if (typeof index !== 'number') throw new Error(`Element atIndex argument must be a number, got ${typeof index}`);
     const matcher = this._originalMatcher;
-    this._originalMatcher._call = invoke.call(invoke.Android.Class(DetoxMatcher), 'matherForAtIndex', invoke.Android.Integer(index), matcher._call);
+    this._originalMatcher._call = invoke.call(invoke.Android.Class(DetoxMatcher), 'matcherForAtIndex', invoke.Android.Integer(index), matcher._call);
```
DanielMSchmidt: Great job 👍 @rotemmiz this is why generated code is superior 🏆 :trollface:

pietropizzi: One thing: I could not find a test for this that was failing before. Not sure if this should have been caught or not ¯\_(ツ)_/¯

PR #335, opened 2017-10-11
Basic support for Jest runner

Kureev:
With this PR merged in, users will be capable of using Jest test runner without significant problems:
- Now it is possible to use standard interface to run tests:
  ```sh
  detox test -r jest
  ``` 
  instead of 
  ```sh
  jest e2e --setupTestFrameworkScriptFile=./jest/setup/e2e-tests.js --runInBand
  ```
- Documentation reflects new test runner mechanism


docs/Guide.Jest.md
```diff
@@ -44,16 +43,14 @@ beforeEach(async () => {
...
+  "setupTestFrameworkScriptFile": "./e2e/init.js"
 },
 "scripts": {
-    "test:e2e": "jest __e2e__ --setupTestFrameworkScriptFile=./jest/setup/e2e-tests.js --runInBand",
+    "test:e2e": "detox test -r jest",
     "test:e2e:build": "detox build"
 }
 ```
 
-We need the `--runInBand` as detox doesn't support parallelism yet.
```
rotemmiz: Will running without the flag run tests serially?

detox/local-cli/detox-test.js
```diff
@@ -38,6 +38,9 @@ switch (program.runner) {
   case 'mocha':
     command = `node_modules/.bin/${program.runner} ${testFolder} --opts ${testFolder}/${program.runnerConfig} ${configuration} ${loglevel} ${cleanup} ${reuse} ${debugSynchronization} ${artifactsLocation}`;
     break;
+  case 'jest':
```
rotemmiz: What about passing detox command line args? Will those work?

Kureev: We can think of a way doing this. Although, in this version I wasn't aiming to provide this. If this is mandatory, we can discuss the approach

docs/Guide.Jest.md
Kureev: I added the flag [here](https://github.com/wix/detox/pull/335/files#diff-7ba9426c8fa0338c47632e8dbf8c11f3R42) so if you use `detox test -r jest` it is always there

detox/local-cli/detox-test.js
rotemmiz: I think it is, the feature is not 100% usable without it.

rotemmiz: `detox test --configuration ios.sim.release`
`detox test --loglevel verbose`
`detox test --debug-synchronization`

Kureev: Which flags are important? 
In the previous version there were no recommendations regarding flags so kept it out of scope. However, I'm fine adding important flags to make jest support even better 💪 .

Kureev: > detox test --configuration ios.sim.release
detox test --loglevel verbose
detox test --debug-synchronization

Will add these ones ASAP 💪

rotemmiz: We need all of them ;)

rotemmiz: I mean, let's try to add a way to make the detox cli a passthrough, and not manually add each command when we add it to Detox itself.
I know it's not how we do it with our Mocha support, but we need to up our game and create better code even in the cli tool.

Kureev: @rotemmiz what is the goal?
- tests can access CLI arguments
or
- detox can access CLI arguments
---
If we talk about the second one, then it already works like this (commander parses arguments before passing them to test runner)

rotemmiz: You're right, but we execute a child process and these argument are not being passed into it.
`cp.execSync(command, {stdio: 'inherit'});` 
We probably just need to pass them on...

Kureev: Oh, now I see. Indeed, we just need to pass arguments to the child process. Then there will be no point in passing all these flags in the command itself

Kureev: Seems like `jest` doesn't allow custom CLI parameters. The only approach I see is to use environment vars

DanielMSchmidt:
@rotemmiz just a question, would you add different runner to the test matrix?

hellogerard:
Totally tangential question, but here goes... Been playing around with detox and jest, and I noticed that if I have two test files (e.g. `firstTest.spec.js` and `secondTest.spec.js`), detox will terminate, uninstall, install and launch the app for each file.  It does _not_ do this with mocha. Any ideas?  Is this just because of the repeated call to `detox.init()`?   If so, where would a good place to put this so that it is only called once?

Kureev:
> detox will terminate, uninstall, install and launch the app for each file.
@hellogerard I believe it is related to --reuse flag. See (https://github.com/Kureev/detox/blob/8bf5abde5fcc2ff178244e4a411b1a72341c69a8/detox/local-cli/detox-test.js#L12)

rotemmiz:
I didn't use Detox with Jest yet, but from what I could see, there is no way to run a global `before` across files to create a one time setup (which is how `detox.init()` should be run), so this is indeed an issue.
reference: https://github.com/facebook/jest/issues/3832

rotemmiz:
@DanielMSchmidt Not sure, the only thing we need to test here are is how the test runner is triggered, so it's pretty easy to test with unit tests.

hellogerard:
> reference: facebook/jest#3832

@rotemmiz Thanks for that link. There's a PR that closes that issue. Once it's merged, it looks like it can solve the global setup issue. PR is https://github.com/facebook/jest/pull/4506

Kureev:
Quick update on the PR: we discussed a way to go with @rotemmiz so you may expect that I'll push my changes today-tomorrow 😉

hellogerard:
Hmm. I may have spoke too soon, or I don't understand the jest changes. I tried to use the new global setup by doing the following:

1. Created two test files (e.g. `first.spec.js` and `second.spec.js`).
1. Created a `jest-environment-detox.js` file:

    ```js
    require("babel-polyfill");

    const NodeEnvironment = require('jest-environment-node');

    const config = require('../package.json').detox;

    class DetoxEnvironment extends NodeEnvironment {
      async setup() {
        try {
          await detox.init(config, { launchApp: true });
        } catch (e) {
          console.log('e', e);
        }
        await super.setup()
      }

      async teardown() {
        try {
          await detox.cleanup();
        } catch (e) {
          console.log('e', e);
        }
        await super.teardown()
      }
    }

    module.exports = DetoxEnvironment;
    ```
1. Specifying the test environment by passing `--testEnvironment="./e2e/jest-environment-detox"` to the `jest` command.

But no dice. Not only is it running `detox.init` for each test file, it's not even loading right (i.e. getting `device is not defined`).  Am I doing this right?  😕

Kureev:
Hi @rotemmiz!

It seems like I have all my unit tests passing locally, but got some issues on CI 🤔 . Can you help me to figure why?

rotemmiz:
Seems like some of the params passed to mocha have changed, try running `scripts/ci.sh` and see if it passes locally

Kureev:
@rotemmiz seems that after the last fix CI dies due to timeout 😐 . Should I increase it a little bit?

rotemmiz:
I have a feeling that something has changed in Travis. We’ll need to debug it, and add better visibility to the build log.

Does everything work locally?

Kureev:
> Does everything work locally?

I have the same problem with timeout on the local machine, but I'll try to extend it (locally) and let you know if it actually works.

Kureev:
Ok, it gets stuck at 
```
detox verb 3: /usr/bin/xcrun simctl terminate 33E49BFA-ED99-4ACD-A865-5E18863556B7 org.reactjs.native.example.example
detox info 3: Terminating org.reactjs.native.example.example...
detox info 3: org.reactjs.native.example.example terminated
detox verb 4: /bin/cat /dev/null >$HOME/Library/Developer/CoreSimulator/Devices/33E49BFA-ED99-4ACD-A865-5E18863556B7/data/tmp/detox.last_launch_app_log.out 2>$HOME/Library/Developer/CoreSimulator/Devices/33E49BFA-ED99-4ACD-A865-5E18863556B7/data/tmp/detox.last_launch_app_log.err && SIMCTL_CHILD_DYLD_INSERT_LIBRARIES="/Users/xamd/Library/Detox/ios/a719a7fd8af96db5fefad3da20b783f61b168620/Detox.framework/Detox" /usr/bin/xcrun simctl launch --stdout=/tmp/detox.last_launch_app_log.out --stderr=/tmp/detox.last_launch_app_log.err 33E49BFA-ED99-4ACD-A865-5E18863556B7 org.reactjs.native.example.example --args -detoxServer ws://localhost:57195 -detoxSessionId adf47c1c-c956-e93c-7640-cf174d884a54
detox info 4: Launching org.reactjs.native.example.example...
```

Even if I run my emulator manually, still the same issue... No idea how to debug it. Any suggestions?

rotemmiz:
I’m not near a computer to verify that, but I think that —args is misplaced (check the last command vs Detox 5.8.2)

Kureev:
I found an issue in my local setup: seems for some reason my emulator that was used by tests got stuck (no idea, really). After resetting all the settings, everything worked. Will report once I finish all tests locally. 🤞

Kureev:
I can confirm that all tests (including e2e) are green.

rotemmiz:
Rerunning your build

rotemmiz:
I want to add a very important request to this feature: since users choose and setup their test runner probably stick with it for more than an hour, it is important to add an option to configure the test runner in the detox config inside `package.json` (just like we are able configure the test folder), and ditch the test runner selection as a command line argument.

This should be a very small addition.

rotemmiz:
Tested locally on both Xcode 8.3.3 and Xcode 9, could not reproduce the timeout we're having in CI. I'll have to check tomorrow.

Kureev:
> it is important to add an option to configure the test runner in the detox config inside package.json (just like we are able configure the test folder)

Will do. If you have any other requirements for this PR, it will be great if you can list them here so I can make one final batch of changes 😉

Kureev:
> Tested locally on both Xcode 8.3.3 and Xcode 9, could not reproduce the timeout we're having in CI. I'll have to check tomorrow.

If there is a way to reload your iOS emulator on Travis, I would do that first. For me it solved exactly the same problem on my local machine.

docs/Guide.Jest.md
LeoNatan: What is stopping detox from parallelism?

rotemmiz: a few things, mainly controlling multiple simulators. It's in our roadmap, shouldn't be too hard.

Kureev:
Ready for review, @rotemmiz

rotemmiz:
@Kureev, some of our users use Detox with mocha (or any other test runner for that matter) independently of detox-cli. The current change breaks their setup since they can not pass arguments anymore. 
When we spoke, I encouraged you use argv and merge them with env vars if needed. 
The fact that Jest doesn't support argv doesn't need to enforce other test runners to work with env vars.

Kureev:
> The current change breaks their setup since they can not pass arguments anymore.
When we spoke, I encouraged you use argv and merge them with env vars if needed.
The fact that Jest doesn't support argv doesn't need to enforce other test runners to work with env vars.

Not sure I follow. All arguments are respected 🤔

rotemmiz:
Detox core now only supports env vars, not argv. Try running 
`node_modules/.bin/mocha e2e --opts e2e/mocha.opts --configuration ios.sim.release --loglevel verbose`
It will not work.

Kureev:
> Detox core now only supports env vars, not argv. Try running
node_modules/.bin/mocha e2e --opts e2e/mocha.opts --configuration ios.sim.release --loglevel verbose
It will not work.

Sure, but this is not about Detox either. Detox should be run only by detox-cli and `detox test` command, isn't it correct?

rotemmiz:
I don't agree, Detox is test runner independent. 
What if people are using AVAJS now? They have no way of using detox-cli, and we're changing their API under their feet just to add support for other test runner in our CLI tool...
That will not be responsible of us.

In any case, using argv (where possible) is a better and responsible solution, it removes any possibility of environment var leaks.

Kureev:
> What if people are using AVAJS now?

I expect it to be not supported (yet). Otherwise I would expect an article in docs and/or other guidelines' describing how to setup Detox with your-faviourite-tool.

> In any case, using argv (where possible) is a better and responsible solution, it removes any possibility of environment var leaks.

I'm fine changing code to support `argv` on `argsparse` level, but the part about env var leaks is still not clear to me.

Kureev:
Should be good now. First we look for the flag in `argv` and if it isn't there (we use Jest), it falls back to env variable

detox/test/ios/example.xcodeproj/project.pbxproj
```diff
@@ -975,7 +975,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				DEAD_CODE_STRIPPING = NO;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = J966JLDEF9;
 				INFOPLIST_FILE = example/Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
```
rotemmiz: Was that added in purpose?

Kureev:
I reset `example.xcodeproj` (sorry, committed this change by mistake. I was running this code on my emulator by hand, so had to sign it with certificate ¯\\_(ツ)\_/¯) + it seems that tests are green again!

Kureev:
The last CI run was flaky 😒 I changed nothing (okok, I removed my cert pub hash from `xcodeproj`), but have failed specs. Can you please try to re-run this and let me know if there is anything else you would like me to take care of?

detox/local-cli/detox-test.js
```diff
@@ -21,26 +21,54 @@ program
...
+  console.log(command);
+  cp.execSync(command, {
+    stdio: 'inherit',
+    env: {
+      ...process.env,
+      configuration: program.configuration,
+      loglevel: program.loglevel,
+      cleanup: program.cleanup,
+      reuse: program.reuse,
+      debugSynchronization: program.debugSynchronization ? 3000 : '',
```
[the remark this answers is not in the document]
Kureev: Hey @rotemmiz! Nice catch, fixed! 😉

rotemmiz:
🎉 !!!
Thanks @Kureev !!

xfumihiro:
@hellogerard Jest's TestEnvironment is sandboxed (see [this](https://github.com/facebook/jest/pull/4713))
Here's the [PR](https://github.com/facebook/jest/pull/4716) for Global setup/teardown

PR #355, opened 2017-10-21
Update Jest-related docs

Kureev:
Scope of changes:
- Add `jest` to `detox --help` command
- Add missing flags for "test:e2e" script

### DIFF
--- a/docs/Guide.Jest.md
+++ b/docs/Guide.Jest.md
@@ -46,12 +46,14 @@ Add this part to your `package.json`:
   "setupTestFrameworkScriptFile": "./e2e/init.js"
 },
 "scripts": {
-    "test:e2e": "detox test",
+    "test:e2e": "detox test -c ios.sim.debug -a e2e",
     "test:e2e:build": "detox build"
+},
+"detox": {
+  "runner": "jest",
+  ...
 }
 ```
-
-In the `detox` part of your `package.json`, add `"runner": "jest"` to tell detox that you want to use jest runner instead of mocha.
 
 ### Writing Tests
 
--- a/docs/Guide.Jest.md
+++ b/docs/Guide.Jest.md
@@ -46,7 +46,7 @@ Add this part to your `package.json`:
   "setupTestFrameworkScriptFile": "./e2e/init.js"
 },
 "scripts": {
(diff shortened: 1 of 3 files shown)


docs/Guide.Jest.md
```diff
@@ -46,13 +46,15 @@ Add this part to your `package.json`:
   "setupTestFrameworkScriptFile": "./e2e/init.js"
 },
 "scripts": {
-    "test:e2e": "detox test",
+    "test:e2e": "detox test -c ios.sim.debug -a e2e",
```
rotemmiz: -a sets artifacts output location, what did you mean by that?

Kureev: for some reason, if you don't specify anything, `_getAbsolutePath` tries to create artifacts folder under `undefined/`

rotemmiz: That’s a different issue then, but we certainly don’t want them in our e2e tests library (or anywhere for that matter, it’s unrelated to this part of the docs). Let’s open an issue to fix that ASAP

Kureev: I think it is related to `process.cwd()` in child process. I'll follow up on this

Kureev: @rotemmiz ok, it seems to happen only if I use `xcode` config (to debug using xcode), because `binaryPath` is not specified. Should I make an exception here?

rotemmiz: Not sure I follow why `_getAbsolutePath` is needed in that case, but I'll take a look in the code later.
I still think the issue you're trying to debug is out of the scope of  this PR, so we can change the docs, and open a new issue with that.

Kureev: Sorry for miscommunication. `process.cwd()` has nothing to do with the issue. Yeah, I can remove artifacts part as far as it is only related to `xcode` debug configuration. I'll follow up on this in separated issue

Kureev:
@rotemmiz I updated this PR by removing `-a e2e`. I think it is good to go now.

Kureev:
Should we merge it in?

rotemmiz:
Still trying to figure out why e2e fails on Xcode 8.3.3, re ran the build, see you on the other side.

Issue #386, opened 2017-11-06
Unable to connect to Genymotion emulator


vasyas:
#### Description
Unable to connect to Genymotion emulator. Stock Android emulator works ok.
Deployment to Genymotion using react-native works also ok.

Error when running Detox:
```
~/workspace/crg/tests> mocha native --opts native/mocha.opts --configuration android.emu.debug --reuse --loglevel verbose


detox info 10:24:15: server listening on localhost:58283...
detox verb ws onOpen [object Object]
detox verb ws send: {"type":"login","params":{"sessionId":"ff60085d-e041-8510-063d-ba1e880e99e8","role":"tester"},"messageId":0}
detox verb ws onMessage: {"type":"loginSuccess","params":{"sessionId":"ff60085d-e041-8510-063d-ba1e880e99e8","role":"tester"},"messageId":0}
detox verb ws
detox verb 1: /Users/vasyas/Library/Android/sdk/tools/emulator -list-avds --verbose
detox verb 1: stdout: Nexus_5X_API_25_x86
detox verb 1: stdout:
  1) "before all" hook

  0 passing (141ms)
  1 failing

  1)  "before all" hook:
     Error: Can not boot Android Emulator with the name: '192.168.57.101:5555',
      make sure you choose one of the available emulators: Nexus_5X_API_25_x86
      at EmulatorDriver._callee3$ (node_modules/detox/lib/devices/EmulatorDriver.js:35:19)
      at tryCatch (/Users/vasyas/.config/yarn/global/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (/Users/vasyas/.config/yarn/global/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (/Users/vasyas/.config/yarn/global/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
      at step (node_modules/detox/lib/devices/EmulatorDriver.js:1:747)
      at node_modules/detox/lib/devices/EmulatorDriver.js:1:907
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

Running Genymotion:
```
~/workspace/crg/tests> adb devices -l
List of devices attached
192.168.57.101:5555    device product:vbox86p model:Google_Nexus_5X___7_1_0___API_25___1080x1920 device:vbox86p
```

[...]

PR #397, opened 2017-11-09
Implemented AttachedAndroidDriver device for connecting to Genymotion

vasyas:
As discussed in #386 and in https://github.com/wix/detox/issues/96#issuecomment-343156488 , I've implemented AttachedAndroidDriver (really a cut version of EmulatorDriver) that is able to connect to Genymotion and other adb-attached devices.

Also updated doc. Not sure if a test required.

### DIFF
--- a/detox/src/devices/AttachedAndroidDriver.js
+++ b/detox/src/devices/AttachedAndroidDriver.js
@@ -1,0 +1,38 @@
+const _ = require('lodash');
+const AAPT = require('./android/AAPT');
+const Emulator = require('./android/Emulator');
+const AndroidDriver = require('./AndroidDriver');
+
+class AttachedAndroidDriver extends AndroidDriver {
+
+  constructor(client) {
+    super(client);
+
+    this.aapt = new AAPT();
+    this.emulator = new Emulator();
+  }
+
+  async acquireFreeDevice(name) {
+    const adbDevices = await this.adb.devices();
+    const filteredDevices = _.filter(adbDevices, {name: name});
+
+    let adbName;
+    switch (filteredDevices.length) {
+      case 1:
+        const adbDevice = filteredDevices[0];
+        adbName = adbDevice.adbName;
+        break;
+      case 0:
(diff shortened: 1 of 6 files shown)


detox/src/devices/AttachedAndroidDriver.js
```diff
@@ -0,0 +1,38 @@
...
+const AAPT = require('./android/AAPT');
+const Emulator = require('./android/Emulator');
+const AndroidDriver = require('./AndroidDriver');
+
+class AttachedAndroidDriver extends AndroidDriver {
+
+  constructor(client) {
+    super(client);
+
+    this.aapt = new AAPT();
```
rotemmiz: I just found out AAPT was in EmulatorDriver and not in AndroidDriver, pushed a change, please remove it as well.

detox/src/devices/AttachedAndroidDriver.js
```diff
@@ -0,0 +1,38 @@
...
+    this.emulator = new Emulator();
+  }
+
+  async acquireFreeDevice(name) {
+    const adbDevices = await this.adb.devices();
```
rotemmiz: This is an almost exact copy of the logic in EmulatorDriver (and we hate duplicate code of course 😬 ) . Why not merge in into a utility function inside `AndroidDriver` ?

docs/More.AndroidSupportStatus.md
```diff
@@ -58,7 +58,19 @@ All Core APIs are 100% implemented.
 ## Emulator control
 1. **Emulators** are fully supported, to choose an emulator to run your tests on check `emulator -list-avds`. If none exist, create one.
 2. **Devices** - Coming soon!
-3. **Genymotion** -  Coming a bit later...
```
rotemmiz: This does not relate to Genymotion specifically, it can be used with connected device, I think it should be emphesized in the docs.

rotemmiz:
Thanks for this great PR!
please take a look at the notes, there are minor changes I think should be done before merging.

vasyas:
Updated, please review.

PR #404, opened 2017-11-13
Stop transpiling detox and detox-server, use sources and require node >=7.6

mrtnrst:
### Overview

Remove deprecated package and updated to latest suggest package per Babel docs.

image

### Discussion
I was noticing errors with a project that was not importing _babel-preset-latest_ as a dev dependency. Rather than importing deprecated code I went ahead and updated to current logic. 

_Example:_
```js
detox test
```
would return 
`Couldn't find preset "latest" relative to directory "/Users/arista/***/node_modules/detox"`

### DIFF
--- a/detox-server/package.json
+++ b/detox-server/package.json
@@ -35,17 +35,16 @@
     "babel-core": "^6.8.0",
     "babel-eslint": "^6.0.4",
     "babel-polyfill": "^6.8.0",
-    "babel-preset-latest": "^6.22.0",
+    "babel-preset-env": "^1.6.1",
     "babel-register": "^6.8.0"
   },
   "babel": {
-    "env": {
-      "test": {
-        "presets": [
-          "latest"
-        ],
-        "retainLines": true
-      }
-    }
+    "presets": [
+      ["env", {
+        "targets": {
+          "node": "current"
+        }
+      }]
+    ]
   }
 }
--- a/detox-server/package.json
+++ b/detox-server/package.json
@@ -14,8 +14,7 @@
   },
   "scripts": {
     "start": "node src/cli.js",
-    "test": ":",
-    "prepublish": "npm run build"
+    "test": ":"
   },
   "bugs": {
     "url": "https://github.com/wix/detox/issues"
--- a/detox-server/package.json
+++ b/detox-server/package.json
@@ -20,7 +20,7 @@
     "url": "https://github.com/wix/detox/issues"
   },
   "homepage": "https://github.com/wix/detox/detox-server",
-  "main": "lib/index.js",
+  "main": "./src/index.js",
   "author": "Tal Kol <talkol@gmail.com>",
   "license": "MIT",
   "dependencies": {
--- a/detox-server/package.json
+++ b/detox-server/package.json
@@ -13,7 +13,6 @@
     "detox-server": "src/cli.js"
   },
   "scripts": {
-    "build": "node src -d lib",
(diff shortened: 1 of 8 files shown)


dferguson08:
🙏 Really need this 🙏 Nice find/fix, @mrtnrst

rotemmiz:
What version of node do you guys use? 
I am thinking of dropping Babel altogether, node 7.6 handles detox and async await with no need to transpile any code.

mrtnrst:
We are currently using v8.6 LTS. I thought for legacy support but unsure how far back you wanted to go.

rotemmiz:
263 days since Node 7.6 was released. Good time to say bye to Babel :)

mrtnrst:
I can open another PR to remove Babel if you aren't on it already. 🙇

rotemmiz:
yes, please do 🤗

mrtnrst:
@rotemmiz I removed the extra babel packages but kept the build. Also added a min version for node engines which will introduce a breaking change, let me know your thoughts.

rotemmiz:
This is not enough, I want to remove all trace of babel.
1. remove babel in deps
2. remove babel in builds (`detox/build.sh`, `detox-server npm run build`)
3. change package.json main in both from `lib/index.js` to `src/index.js`

detox-server/package.json
```diff
@@ -10,11 +10,11 @@
     "url": "https://github.com/wix/detox.git"
   },
   "bin": {
-    "detox-server": "lib/cli.js"
+    "detox-server": "src/cli.js"
   },
   "scripts": {
-    "build": "BABEL_ENV=test babel src -d lib",
-    "start": "node lib/cli.js",
+    "build": "node src -d lib",
```
rotemmiz: we can drop `build` script altogether

detox/scripts/build.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash -e
 
 echo -e "\nTranspiling JavaScript sources"
```
rotemmiz: drop the echo

brunolemos:
Any quick way to point the package.json to this pull request?
Tried `"detox": "wix/detox#404/head",` but it gives an error `Package "undefined@undefined" doesn't have a "name".` because detox uses lerna and the root package.json is not the package I want.

rotemmiz:
hey @mrtnrst , any updates on your side ? If you need any help, please ping me.

mrtnrst:
@rotemmiz sorry I've had a busy week. Going to spend some time later today to try and get the CI issues resolved.

mrtnrst:
@rotemmiz so this was a lot easier than I originally thought! I just have to pay closer attention to my changes. 😁 🎉

brunolemos:
Tested the PR, it works if I run `npm run e2e:ios` inside the `detox/test` folder.
But, tested on a an empty react-native project with detox and it says this:

`Couldn't find preset "es2015" relative to directory "/Users/brunolemos/Projects/react-native-css-in-js-benchmarks/node_modules/detox/node_modules/child-process-promise"`

rotemmiz:
@mrtnrst , one last thing, in order to really test it works, let's point the test project to use Detox from main and not specifically from source.
https://github.com/wix/detox/blob/master/detox/test/e2e/init.js

change:
```js
const detox = require('../../src/index');
```
to 
```js
```

Then, if it passes CI, we're good.

Issue #410, opened 2017-11-15
Support long press duration for element


jhen0409:
We have a use case of two events for different long press duration, implement by `onPressIn` and `onPressOut`.

It would be great if we could support duration for [`longPress`](https://github.com/wix/detox/blob/a1ca6c35bdc2ed8264c4e880872b2bf384a4230c/detox/src/ios/expect.js#L60).

For iOS, we could use [`actionForLongPressWithDuration`](https://github.com/google/EarlGrey/blob/master/EarlGrey/Action/GREYActions.h#L54) of EarlGrey, but I don't know how to do for Android.

LeoNatan:
Good suggestion, thanks!

PRs are welcome. Perhaps someone can assist with Android.

LeoNatan:
Thanks for implementing!

PR #412, opened 2017-11-16
Add duration of element.longPress for iOS

jhen0409:
Related to #410. I also added the test case `Long Press Me 1.5s`.

### DIFF
--- a/detox/src/ios/expect.js
+++ b/detox/src/ios/expect.js
@@ -65,9 +65,13 @@ class TapAtPointAction extends Action {
 }
 
 class LongPressAction extends Action {
-  constructor() {
+  constructor(duration) {
     super();
-    this._call = invoke.callDirectly(GreyActions.actionForLongPress());
+    if (typeof duration !== 'number') {
+      this._call = invoke.callDirectly(GreyActions.actionForLongPress());
+    } else {
+      this._call = invoke.callDirectly(GreyActions.actionForLongPressWithDuration(duration / 1000));
+    }
   }
 }
 
@@ -268,8 +272,8 @@ class Element {
   async tapAtPoint(value) {
     return await new ActionInteraction(this, new TapAtPointAction(value)).execute();
   }
-  async longPress() {
-    return await new ActionInteraction(this, new LongPressAction()).execute();
+  async longPress(duration) {
+    return await new ActionInteraction(this, new LongPressAction(duration)).execute();
   }
   async multiTap(value) {
     return await new ActionInteraction(this, new MultiTapAction(value)).execute();
--- a/detox/test/e2e/c-actions.js
+++ b/detox/test/e2e/c-actions.js
@@ -17,13 +17,18 @@ describe('Actions', () => {
     await expect(element(by.text('Long Press Working!!!'))).toBeVisible();
   });
 
+  it(':ios: should long press with duration on an element', async () => {
+    await element(by.text('Long Press Me 1.5s')).longPress(1500);
+    await expect(element(by.text('Long Press With Duration Working!!!'))).toBeVisible();
+  });
+
   it('should multi tap on an element', async () => {
     await element(by.id('UniqueId819')).multiTap(3);
     await expect(element(by.id('UniqueId819'))).toHaveText('Taps: 3');
   });
 
   it('should tap on an element at point', async () => {
-    await element(by.id('View7990')).tapAtPoint({ x: 180, y: 140 });
+    await element(by.id('View7990')).tapAtPoint({ x: 180, y: 160 });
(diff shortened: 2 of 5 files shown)


detox/src/ios/expect.js
```diff
@@ -58,9 +58,13 @@ class TapAtPointAction extends Action {
 }
 
 class LongPressAction extends Action {
-  constructor() {
+  constructor(duration) {
     super();
-    this._call = invoke.callDirectly(GreyActions.actionForLongPress());
+    if (duration) {
```
rotemmiz: let's add default value, drop the condition, and use only `GreyActions.actionForLongPressWithDuration`

detox/test/e2e/c-actions.js
```diff
@@ -17,13 +17,18 @@ describe('Actions', () => {
     await expect(element(by.text('Long Press Working!!!'))).toBeVisible();
   });
 
+  it(':ios: should long press with duration on an element', async () => {
```
rotemmiz: I don't want to create more discrepancy between iOS and Android API, I'm willing to help you add Android support as well if you need help.

detox/test/e2e/c-actions.js
```diff
@@ -17,13 +17,18 @@ describe('Actions', () => {
...
+  });
+
   it('should multi tap on an element', async () => {
     await element(by.id('UniqueId819')).multiTap(3);
     await expect(element(by.id('UniqueId819'))).toHaveText('Taps: 3');
   });
 
   it('should tap on an element at point', async () => {
-    await element(by.id('View7990')).tapAtPoint({x:180, y:140});
+    await element(by.id('View7990')).tapAtPoint({x:180, y:160});
```
rotemmiz: Why was that changed ?

jhen0409: I added [a button for new test case](https://github.com/wix/detox/pull/412/files#diff-bed70c67a493bc438deb1689cadd32ee), so the `UniqueId819` button position was changed.

rotemmiz: 👍

detox/test/e2e/c-actions.js
jhen0409: Yeah I also really want this feature can be supported for Android, if you could help will be great. Thanks. :)

generation/earl-grey/index.js
```diff
@@ -47,6 +47,7 @@ if (
 
 // Constants
 const SUPPORTED_TYPES = [
+  "CFTimeInterval",
```
DanielMSchmidt: 🎉🎉🎉

jozan:
How is this progressing? Can I help somehow? We pretty much need this feature for both platforms. :)

jozan:
Alright, I started working on this feature base on this PR.

rotemmiz:
@jozan, for Android?

jozan:
For iOS, we need this feature eventually for Android as well but even
getting iOS to work would be very nice.

I tried to rebase to master but it did not work out of the box since lot
has changed past three months especially in generation dir. Also, after
rebase both longPress and longPress(duration) did not pass tests and failed
to error from this line:
https://github.com/wix/detox/blob/master/detox/ios/Detox/MethodInvocation.m#L225
where value i was 0.

I don’t have time to debug further today.

rotemmiz:
Go away stale bot! Sorry for the huge delay. We have intention to merge this in.

LeoNatan:
Any news on this PR? Let’s merge for iOS and then in the future add for Android?

rotemmiz:
Yes, let's do it

LeoNatan:
@jhen0409 Could you please merge master and see that tests are not broken?

jhen0409:
@LeoNatan 👌

LeoNatan:
Thanks!

LeoNatan:
@yershalom Jenkins is flaky, please look why.

detox/src/ios/expect.js
```diff
@@ -65,9 +65,9 @@ class TapAtPointAction extends Action {
 }
 
 class LongPressAction extends Action {
-  constructor() {
+  constructor(duration = 750) {
```
LeoNatan: Why is the default 750? Is this what Earl Grey uses in the long press action?

jhen0409: I forgot the reason, but it should be 1000 for now.

LeoNatan: Let's instead use the default. If no param is provided, let's call the current API, but if there is provided, let's call the new API.

jhen0409:
The some style changes caused by Prettier, I reverted it.

LeoNatan:
Cool. Let’s wait for the CI and I’ll merge. Thank you for your contribution!

PR #421, opened 2017-11-21
Minor documentation fix

plasticine:
What it says on the tin. 😎

### DIFF
--- a/docs/APIRef.MockingUserNotifications.md
+++ b/docs/APIRef.MockingUserNotifications.md
@@ -17,7 +17,7 @@ await device.relaunchApp({userNotification: notification});
 ```js
 describe('Background push notification', () => {
 	beforeEach(async () => {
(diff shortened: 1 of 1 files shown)


rotemmiz:
The current docs are correct, we don't use callbacks to notify mocha we're done, we use async-await instead.

plasticine:
@rotemmiz Hey there, hrm — perhaps I should have posted a more useful PR description after all!

I’m aware of the nature of the `done` callback and async/await with Mocha, however the issue if you look at the diff again is that you’ve mixed the usage;

```
describe('Background push notification', () => {
	beforeEach(async () => {
	  await device.relaunchApp({userNotification: userNotificationPushTrigger}, done) // <--- undefined
	});
});
```

plasticine:
@rotemmiz Don’t mention it! Thanks for your work on Detox — it’s fantastic, and really exciting.

PR #423, opened 2017-11-21
Detox CLI: Jest integration fixes 

rotemmiz:
Fixed jest integration issues


LeoNatan:
What is this breaking?

rotemmiz:
command line:
- runner option has been removed, will only be available from json config in package.json

json config:
- `test-runner` instead of `runner` 
- `test-runner` should be passed with relative path from root, will not join path with `specs` anymore.

Issue #419, opened 2017-11-19
jest-jasmine2 issues


sibelius:
I've got this running jest example https://github.com/wix/detox/tree/master/examples/demo-react-native-jest for android

`yarn test-android`

```jsx
console.error node_modules/jest-jasmine2/build/jasmine/Env.js:194
    Unhandled error

  console.error node_modules/jest-jasmine2/build/jasmine/Env.js:195
    Error: spawn adb ENOENT
        at _errnoException (util.js:1024:11)
        at Process.ChildProcess._handle.onexit (internal/child_process.js:190:19)
        at onErrorNT (internal/child_process.js:372:16)
        at _combinedTickCallback (internal/process/next_tick.js:138:11)
        at process._tickCallback (internal/process/next_tick.js:180:9)
```

rotemmiz:
Please be more specific about the error... What machine are you running on ? Do you have ANDROID_HOME env var ?

We are about to land a PR (#424) fixing lots of the issues with jest, but this doesn't seem to be related, it seems like there are issues with starting `adb`

sibelius:
mac os x sierra

`ANDROID_HOME = /Users/sibelius/Library/Android/sdk`

adb is working fine on my terminal

rotemmiz:
try running detox test in verbose mode
`detox test -c android.emu.debug -l verbose`

sibelius:
```
node_modules/.bin/jest e2e --runInBand
 rror: Could not create artifacts root dir: undefined/detox_artifacts.2017-11-21T21:50:52.099Z
 server listening on localhost:50032...
 onOpen [object Object]
 send: {"type":"login","params":{"sessionId":"6a044050-493a-0aca-33c9-4293d719fd7b","role":"tester"},"messageId":0}
 onMessage: {"type":"loginSuccess","params":{"sessionId":"6a044050-493a-0aca-33c9-4293d719fd7b","role":"tester"},"messageId":0}
  rbx
 : /Users/sibelius/Library/Android/sdk/tools/emulator -list-avds --verbose
 Nexus_4_API_25
 Nexus_5X_API_26
 Nexus_5_API_26
 O_Nexus_6_API_26_8.0
 Pixel_XL_API_27
 : stdout:
 emulator: Android virtual device file at: /Users/sibelius/.android/avd/Nexus_5X_API_26.ini
 emulator: virtual device content at /Users/sibelius/.android/avd/Nexus_5X_API_26.avd
[...]

Issue #362, opened 2017-10-24
Problems on Android using Jest test runner


SMJ93:
#### Description

We have switched test runners from Mocha to Jest using the instruction guide [here](https://github.com/wix/detox/blob/master/docs/Guide.Jest.md), but there seem to be a few issues:

1. **Verbose is always enabled**

With mocha we would add the flag for this to be enabled: `--loglevel verbose`

2. **When running android build and test the instrumentation is never created casing an error** 
```
detox build --configuration android.emu.debug && detox test --configuration android.emu.debug
```

```
mulator stdout:
 : /Users/user/Library/Android/sdk/platform-tools/adb  devices
 List of devices attached
 emulator-5554	device
 : stdout:
 : stdout:
3: /Users/user/Library/Android/sdk/platform-tools/adb -s emulator-5554 shell input  keyevent 82
4: /Users/user/Library/Android/sdk/platform-tools/adb -s emulator-5554 shell am force-stop com.appname.dev
adb -s emulator-5554 shell am instrument -w -r -e detoxServer ws://localhost:8099 -e detoxSessionId test -e debug false com.appname.dev.test/android.support.test.runner.AndroidJUnitRunner
 25344mentation spawned, childProcess.pid:
 send: {"type":"isReady","params":{},"messageId":-1000}
android.util.AndroidException: INSTRUMENTATION_FAILED: com.appname.dev.test/android.support.test.runner.AndroidJUnitRunner
nstrumentation stderr:
nstrumeat com.android.commands.am.Instrument.run(Instrument.java:410)
nstrumeat com.android.commands.am.Am.runInstrument(Am.java:232)
nstrumeat com.android.commands.am.Am.onRun(Am.java:125)
nstrumeat com.android.internal.os.BaseCommand.run(BaseCommand.java:54)
nstrumeat com.android.commands.am.Am.main(Am.java:95)
nstrumeat com.android.internal.os.RuntimeInit.nativeFinishInit(Native Method)
nstrumeat com.android.internal.os.RuntimeInit.main(RuntimeInit.java:284)
nstrumentation stderr:
nstrumentationProcess terminated due to receipt of signal null
```

When printing the instrumentation there is no bundle ID for the app we want to test - this used to show with mocha:

`adb shell pm list instrumentation`
```
instrumentation:com.android.emulator.smoketests/android.support.test.runner.AndroidJUnitRunner (target=com.android.emulator.smoketests)
[...]

Issue #407, opened 2017-11-14
Timeout before emulator loads + terminated due to receipt of signal null


paulbrittain:
Tests fail when it tries to sent the emulator a key event, but the emulator has only just been started and is booting up.  

**The error:**
```
"before all" hook: _callee:
 Error: Command failed: .../Library/Android/sdk/platform-tools/adb -s emulator-5554 shell input keyevent 82
 `.../Library/Android/sdk/platform-tools/adb -s emulator-5554 shell input keyevent 82` (exited with error code 137)
```

**The package.json:**
```
      "android.emu.debug": {
        "binaryPath": "android/app/build/outputs/apk/app-debug.apk",
        "build": "pushd android && ./gradlew assembleDebug assembleAndroidTest -DtestBuildType=debug && popd",
        "type": "android.emulator",
        "name": "Pixel_XL_API_26"
      }
```

**Longer log:**
```
detox verb Emulator stdout:  emulator: Starting QEMU main loop
detox verb Emulator stdout:  path /Users/myname/.android/avd/Pixel_XL_API_26.avd/system.img.qcow2
detox verb Emulator stdout:  path /Users/myname/.android/avd/Pixel_XL_API_26.avd/vendor.img.qcow2
detox verb Emulator stdout:  emulator: registered 'boot-properties' qemud service
detox verb Emulator stdout:  emulator: Adding boot property: 'ro.opengles.version' = '131072'
detox verb Emulator stdout:  emulator: Adding boot property: 'qemu.sf.fake_camera' = 'both'
detox verb Emulator stdout:  emulator: Adding boot property: 'dalvik.vm.heapsize' = '384m'
detox verb Emulator stdout:  emulator: Adding boot property: 'qemu.hw.mainkeys' = '0'
detox verb Emulator stdout:  emulator: Adding boot property: 'qemu.sf.lcd_density' = '560'
detox verb Emulator stdout:  emulator: goldfish_events.have-dpad: false
detox verb Emulator stdout:  emulator: goldfish_events.have-trackball: false
detox verb Emulator stdout:  emulator: goldfish_events.have-camera: true
detox verb Emulator stdout:  emulator: goldfish_events.have-keyboard: true
detox verb Emulator stdout:  emulator: goldfish_events.have-lidswitch: false
detox verb Emulator stdout:  emulator: goldfish_events.have-touch: false
detox verb Emulator stdout:  emulator: goldfish_events.have-multitouch: true
[...]

PR #434, opened 2017-11-27
Better Android emulator sync

rotemmiz:
1. Making sure all child proccesses are closed, fixing issues with Mocha 4 and Jest
2. Spawning a detached emulator process, gaining same behaviour as we have in iOS, which leaves simulator open when tests are done
3. Applying fix on broken emulator config.ini
4. Better sync through reading getprop dev.bootcomplete


rotemmiz:
Fixes #419, #407
Attempts to fix #362

PR #425, opened 2017-11-21
Add support for generating android matchers

DanielMSchmidt:
This restructures the generation to allow for multiple adapters, leaving this open for further extension. It only adds one generated action to the live code, so that the diff doesn't get any bigger.

Although it might look intimidating as a diff, most of it is just moving files around and extracting the core generation from the platform specifics.

PR #449, opened 2017-12-04
Add generation for tap at location

DanielMSchmidt:
As discussed in #425 this adds tap at location support (I thought there were more than this, this is why I split it up)

### DIFF
--- a/detox/src/android/expect.js
+++ b/detox/src/android/expect.js
@@ -38,7 +38,7 @@ class TapAction extends Action {
 class TapAtPointAction extends Action {
   constructor(value) {
     super();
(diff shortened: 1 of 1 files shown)


LeoNatan:
Why is this being added? Tap at location is not a good API and is prone to a lot of flakiness. We have blocked issues requesting this in the past as there are almost always more appropriate ways to handle these issues. What is the usecase here?

LeoNatan:
@rotemmiz  ^

DanielMSchmidt:
@LeoNatan It's already there, so I just swapped to generated code

Issue #164, opened 2017-06-12
Nested ScrollView cause "Multiple UI elements match"


rumax:
Nested ScrollView cause Multiple UI elements match

## Steps to reproduce
 - Create a nested ScrollView. For example the main ScrollView will be scrolled vertically, nested will be scrolled horizontally
 - Or check this example https://github.com/rumax/detox/commit/ce3596ca0978df60a0875c47170b0919bbfd4e3e
- Add tests for `scroll`, `swipe`
```
it('should be visible', async () => {
    await expect(element(by.id('vScrollView'))).toBeVisible();
});
it('should swipe', async () => {
    await element(by.id('vScrollView')).swipe('up', 'fast');
});

it('should scroll', async () => {
    await element(by.id('vScrollView')).scroll(500, 'down');
});
```

## Actual result:
```
detox info 5: com.wix.demo.react.native launched 
  Example
    ✓ should have welcome screen (793ms)
    ✓ should show hello screen after tap (364ms)
    ✓ should show world screen after tap (360ms)
    1) should be visible


  3 passing (15s)
  1 failing

  1) Example should be visible:
     Error: Multiple UI elements matched for given criteria.
Exception with Assertion: {
  "Assertion Criteria" : "assertWithMatcher: matcherForSufficientlyVisible(>=0.750000)",
  "Element Matcher" : "(((respondsToSelector(accessibilityIdentifier) && accessibilityID('vScrollView')) && !kindOfClass('RCTScrollView')) || (kindOfClass('UIScrollView') && ((kindOfClass('UIView') || respondsToSelector(accessibilityContainer)) && ancestorThatMatches(((respondsToSelector(accessibilityIdentifier) && accessibilityID('vScrollView')) && kindOfClass('RCTScrollView'))))))",
  "Recovery Suggestion" : "Create a more specific matcher to narrow matched element"
}

Error Trace: [
  {
    "Description" : "Multiple elements were matched: (
    "<RCTCustomScrollView:0x7f85bb091000; AX=N; AX.frame={{0, 0}, {414, 736}}; AX.activationPoint={207, 368}; AX.traits='UIAccessibilityTraitNone'; AX.focused='N'; frame={{0, 0}, {414, 736}}; opaque; alpha=1>",
[...]

PR #458, opened 2017-12-12
Nested Scrollview fix

Protothor:
Solution to issue #164 
Gets first scrollView in Element.scroll() to fix iOS nested scrollView problem


rotemmiz:
Can you please provide an e2e test that will prove this correction?

LeoNatan:
I don't think this is a correct fix. As I understand from the code, this will only scroll the first matched instance. But as the issue describes, if two scroll views exist, outer for vertical and inner for horizontal scrolls, scrolling vertically with this solution would fail.

This is a good workaround for people that need the type of solution here, but this is not a general case.

Correct me if I'm wrong.

Protothor:
Sorry about not getting back to you until now, but I live on the West Coast of the US (GMT -8)

Based on my testing it works, and in theory based on the code it should too. The "_extendToDescendantScrollViews" function gets all UIScrollViews inside of the RCTScrollView and the order they will appear in array will be from parent to child. Since the first one is always the first UIScrollView that is returned from that array will be the one we're targeting based on a testID the "atIndex(0)" will target the correct one. Unfortunately, I didn't have the time to attack the "waitFor().action().whileElement().scroll()" this time because it uses a great deal of native code to do the whileElement matcher, but this allows testing of nested ScrollViews in iOS in some fashion even if it's not as pretty.

Protothor:
So, I'm not sure what the problem with the 4th build.

The first time it froze on the "before all" hook
>   1) "before all" hook
>   0 passing (2m)
>   1 failing
>   1) "before all" hook:
>      Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

And the second time it timed out during the lerna set up
> $ ./scripts/demo-projects.ios.sh
>  ./scripts/bootstrap.sh 
> lerna info version 2.4.0
> lerna info Bootstrapping 9 packages
> lerna info lifecycle preinstall
> lerna info Installing external dependencies
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
> Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
> The build has been terminated

Otherwise the tests works, it's a simple nested scrollview that uses "swipe", "scrollTo" and "scroll" to scroll the both the parent and child scroll views.

Protothor:
Hello again, I didn't want to bother anyone around the end of the year holidays, but this passes, there are tests written, is there anything else I can do to complete this pull request?

rotemmiz:
Looks great, I just need you to make sure it doesn't break Android e2e, and since we still can't run it in CI please make sure[ it runs locally ](https://github.com/wix/detox/blob/master/docs/Guide.Contributing.md#android). When done mention it here and I'll merge

Protothor:
So, I tested android and the way that the deep linking worked there, it was getting stuck. So I just updated the test to launch a non-linked instance so that it wouldn't lock the app in after a device.reloadReactNative()

Protothor:
Hello, checking in again. Seeing where we're at with this. It's all passing. I've verified android locally. Anything else I can do to get this merged?

lorrainefox:
This update is very important to my project. Can we get an approval please?  I have some people on my team looking for another automated testing tool and I would rather stay with detox.

rotemmiz:
The best way to to have this PR merged is by adding an e2e test that proves its 
necessity. This e2e test will also make sure it doesn't break in the future.
This is where I'd add it:
E2e test:
https://github.com/wix/detox/blob/master/detox/test/e2e/c-actions.js

Test app:
https://github.com/wix/detox/blob/master/detox/test/src/Screens/ActionsScreen.js

Protothor:
Oh, I'm sorry if I placed the tests incorrectly. I added another test suite. (p-nested-scroll-view), linked it to the main page and ran it as a part of the suite.

detox/test/e2e/n-deep-links.js
```diff
@@ -1,4 +1,7 @@
 describe('Deep Links', () => {
+  after(async () => {
```
rotemmiz: Why was this added?

detox/test/e2e/p-nested-scroll-views.js
```diff
@@ -0,0 +1,63 @@
+describe('Nested Scroll Views', () => {
+  beforeEach(async () => {
+    await device.reloadReactNative();
+  });
+
+  beforeEach(async () => {
```
rotemmiz: You can merge both` beforeEach`s

detox/test/e2e/p-nested-scroll-views.js
```diff
@@ -0,0 +1,63 @@
...
+    await element(by.text('Nested ScrollView')).tap();
+  });
+
+  it('Should handle swipe', async () => {
```
rotemmiz: Nicee!!

detox/test/e2e/p-nested-scroll-views.js
```diff
@@ -0,0 +1,63 @@
...
+
+    await element(by.id('childScrollView')).scrollTo('right');
+    await expect(element(by.text('Right'))).toBeVisible();
+    await expect(element(by.text('Left'))).toBeNotVisible();
+    await element(by.id('childScrollView')).scrollTo('left');
+    await expect(element(by.text('Right'))).toBeNotVisible();
+    await expect(element(by.text('Left'))).toBeVisible();
+  });
+  it('Should handle scroll', async () => {
+    try {
```
rotemmiz: What are you catching here?

rotemmiz:
Ping ;)

heuism:
@Protothor does this work that @rotemmiz can merge in?

timea:
Can we please get this PR in? Would be very helpful.

PR #468, opened 2017-12-20
[BREAKING] Upgrade to gradle 4.1 and android gradle plugin 3

rotemmiz:
this is a breaking change and it requires Android users to update to gradle 4.1


fdnhkj:
Hi @rotemmiz , any idea when this will get published (detox 7) ?
I am facing an issue with `flavorDimensions` with latest detox version published (6.0.4).

Thanks!

PR #479, opened 2017-12-29
Generation: Move DetoxAction invocations to generated code

DanielMSchmidt:
To do this I had to add some mechanisms to allow
sanitizers to work with function names and arguments
instead of only with types as before.

Also some more types were enabled.

### DIFF
--- a/detox/src/android/espressoapi/DetoxAction.js
+++ b/detox/src/android/espressoapi/DetoxAction.js
@@ -5,7 +5,38 @@
 */
 
 
-
+function sanitize_android_edge(edge) {
+	switch (edge) {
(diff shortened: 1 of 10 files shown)


rotemmiz:
Looks great, I just need you to make sure it doesn't break Android e2e, and since we still can't run it in CI please make sure[ it runs locally ](https://github.com/wix/detox/blob/master/docs/Guide.Contributing.md#android). When done mention it here and I'll merge

DanielMSchmidt:
@rotemmiz tested locally with result ✅ 

```sh
 66 passing (3m)
  4 pending
```

PR #487, opened 2018-01-04
note on typing text and hardware keyboard

mbardauskas:


LeoNatan:
Why is this necessary? This is not a new behavior.

mbardauskas:
I haven't encountered it before, because I had detox tests running on a separate simulator somehow. Since I wasn't the only one encountering this lately, I thought the note might be helpful for a handful of people

LeoNatan:
Thanks. I added a note to the documentations.

PR #497, opened 2018-01-08
Update AndroidDriver.js

Crash--:
Fix an issue when the filter doesn't exist. 

Also added a verbose log with ADB attached devices since it's pretty difficult to debug on CI pipeline.


rotemmiz:
Great addition, i believe this should be log.error, not log.verbose though

Crash--:
Done

rotemmiz:
Sorry for the nag, but actually the whole string can be a part of the error string. It will have better readability, surely on CI

PR #500, opened 2018-01-08
Update top level package.json to make it valid

coderjonny:
#### Making the top-level `package.json` be valid. 
This will allow you to add detox package as a remote URL from github and use a specific commit or the latest `#master` branch.

Allows `yarn add https://github.com/wix/detox.git#master -D` to work.
Also allows adding `detox#<some specific commit>` as a dependency in a project's `package.json`: 
` "devDependencies": { "detox": "https://github.com/wix/detox.git#master", ... }`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,33 @@
 {
-  "scripts": {
-    "test:e2e:ios": "pushd detox/test && npm run e2e:ios && popd",
-    "test:e2e:android": "pushd detox/test && npm run e2e:android && popd",
-    "release": "node scripts/release.js"
-  }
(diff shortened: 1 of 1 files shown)


rotemmiz:
I'm not sure this will help you with installing via yarn.
Detox is a monorepo, the library you actually want is the one inside detox/detox, it's package.json is valid.
The external package.json has no real purpose apart from holding scripts...
Not sure yarn can handle monorepos, but if it does, it would be great if you could write a note on how to do it here.
In the meantime, I'm closing this PR.

PR #511, opened 2018-01-11
Fix waitFor text on Android

ygorbarboza:
TextMatcher assertion throws a AssertionError, not a EspressoException

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
@@ -1,6 +1,5 @@
 package com.wix.detox.espresso;
 
-import android.support.test.espresso.Espresso;
 import android.support.test.espresso.EspressoException;
 import android.support.test.espresso.ViewAction;
 import android.support.test.espresso.ViewInteraction;
@@ -9,7 +8,6 @@ import android.view.View;
 import junit.framework.AssertionFailedError;
 
 import org.hamcrest.Matcher;
-import org.hamcrest.Matchers;
 
 import static android.support.test.espresso.Espresso.onView;
 import static android.support.test.espresso.assertion.ViewAssertions.doesNotExist;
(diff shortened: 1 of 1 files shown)


rotemmiz:
Thanks for the contribution! good catch!
Can you please add a test to make sure this actually works ?

ygorbarboza:
Sure. Where you guys keep the instrumental tests?

simonracz:
Hey @ygorbarboza,

Thanks for the PR.

We test detox android in a sample project under `detox/test`.
The waitfor tests are here : https://github.com/wix/detox/blob/master/detox/test/e2e/e-waitfor.js
The relevant screen is here : https://github.com/wix/detox/blob/master/detox/test/src/Screens/WaitForScreen.js

What is this TextMatcher exactly? I made a quick search, but couldn't find it. We have a test that finds a view byText and waitFor and it passes, so I am a bit confused.

simonracz:
Another thing, if this is a correct fix, then please add it to the `waitForAssertMatcherWithSearchAction` too.

ygorbarboza:
The problem occur on the following situation:

You have a `waitFor` with `DetoxMatcher.matcherForText` and a timeout of 50 seconds and a view that will appear one second after i start the `waitFor`. The test will take 50 seconds to end because the first assertion throws a `AssertionError` on the Android side that is not handled and the `waitFor`will last until the timeout finish.

This does not happen with `matcherForTestId` because the exception is it throws is as EspressoException

simonracz:
Sorry, I still don't get it.

1. The scenario you described would mean, that an AssertionError would be thrown nearly instantenously. It would not wait for 50 seconds.

2. Also, this [test](https://github.com/wix/detox/blob/master/detox/test/e2e/e-waitfor.js#L27) uses a text matcher and it passes. It goes to a different waitfor function, but they are equvivalent related to this issue.

It would be great if you could provide a test, that fails without this patch.

ygorbarboza:
After digging a little more i found out that the exception assertion was taking 50 seconds was because i had multiple views with the same text and it throws an exception that make the `waitForAssertMatcher` loop only and for timeout

PR #480, opened 2017-12-29
Let Linting run before unit tests

DanielMSchmidt:
Closes #466 

## Commits

1. enable lint check in CI
2. Run `npm run lint --fix`
3. Add `lint-staged` as dependency, was missing previously
4. fix the eslint errors, by
    - changing code that was actually sub-optimal
    - disabling eslint checks which were in conflict with prettier
    - disabling eslint checks for patterns we use heavily or were super hard to fix (`consistent-return`)

PR #517, opened 2018-01-13
Run linting before Unit Tests and in CI

DanielMSchmidt:
This is the follow-up PR for #480 (I had to close this one due to too many merge conflicts)

@rotemmiz It would be great if we could merge this soon, this PR is just too big to be stable for a long time 🙈

### DIFF
--- a/detox/.eslintrc
+++ b/detox/.eslintrc
@@ -246,7 +246,7 @@
     ],
     "max-nested-callbacks": [
       "error",
-      4
+      5
     ],
     "max-params": [
       "error",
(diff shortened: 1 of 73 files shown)


DanielMSchmidt:
If this PR gets outdated we can remove `ba6e4c5` and run `npm run lint -- --fix` on master 👍

DanielMSchmidt:
@rotemmiz It would be super mega cool if you could prioritize the review on this one, because it will become stale in the blink of an eye. The TL;DR of this PR is "let's actually use our eslint" ;)

DanielMSchmidt:
@rotemmiz what do you think about reducing the eslint rules down to the set that we currently seem to follow & enabling three check in ci? That way we could add new rules alongside with refactorings step by step if we care about them

DanielMSchmidt:
I think this approach is unbearable, maybe we can simply remove every eslint rule we don't follow for now and re-add them step by step if we care 👍

Issue #526, opened 2018-01-18
Enable detox-server to bind to something different than localhost


mverardo:
Currently, detox allows us to determine the address to which the client should connect [with the 
 `server` key in `package.json`](https://github.com/wix/detox/blob/master/docs/APIRef.Configuration.md#server-configuration). 

The problem is it lacks an option to set the IP to which `detox-server`'s websocket should bind, [defaulting to `localhost`](https://github.com/wix/detox/blob/master/detox-server/src/DetoxServer.js#L11), which doesn't allow us to connect to this server through an external device on the network.

For context: we bumped into this problem when trying to use detox on our device farm, in which devices are not connected directly to the machine running `detox-server`.

The fix to this would be simple: add a new config to the `session` key on package.json to override the `localhost` default [here](https://github.com/wix/detox/blob/master/detox-server/src/DetoxServer.js#L11) (WebSocketServer accepts a `host` key).

* Detox:  6.0.4
* Node: irrelevant
* Device: irrelevant
* Xcode: irrelevant
* macOS:  irrelevant

LeoNatan:
This can have potential issues on iOS, as networking requires App Transport Security. Asking users to add exceptions to internal servers sounds sub-optimal to me. Any solution that deals with the issues brought here should also account for the ATS problem.

LeoNatan:
As we've been moving more and more to integrate the server itself with the test runner process, we will not be implementing this.

PR #527, opened 2018-01-18
Add serverHost and shouldStartServer config

Igor1201:
`serverHost` config should override `detox-server` host and port values, enabling it to bind to an IP other than localhost. `shouldStartServer` should force the server start when we specify a session.

See https://github.com/wix/detox/issues/526 for more info

### DIFF
--- a/detox-server/src/DetoxServer.js
+++ b/detox-server/src/DetoxServer.js
@@ -7,11 +7,11 @@ log.heading = 'detox-server';
 log.loglevel = 'wss';
 
 class DetoxServer {
-  constructor(port) {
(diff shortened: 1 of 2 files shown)


rotemmiz:
Hey Igor, 
I'll gladly accept this PR, but a few things are missing:
1. please rebase the PR with master, and let's see it passes CI
2. Add docs!
3. Add unit tests (which should now fail since coverage is below 100%

LeoNatan:
Before this is accepted, we should discuss how to solve real-world iOS issues with ATS:
https://github.com/wix/detox/issues/526#issuecomment-360540630

PR #532, opened 2018-01-22
set location should not use comma

hiaw:
using comma instead of dot will cause fbsimctl to fail.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -140,8 +140,8 @@ class Device {
   }
 
   async setLocation(lat, lon) {
-    lat = String(lat).replace('.', ',');
-    lon = String(lon).replace('.', ',');
+    lat = String(lat);
+    lon = String(lon);
     await this.deviceDriver.setLocation(this._deviceId, lat, lon);
   }
 
@@ -198,7 +198,7 @@ class Device {
     if(path.isAbsolute(appPath)) {
       return appPath;
     }
-    
+
     const absPath = path.join(process.cwd(), appPath);
     if (fs.existsSync(absPath)) {
       return absPath;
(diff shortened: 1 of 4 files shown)


rotemmiz:
CI should be green now, can you please rebase the PR and rerun the tests ?

LeoNatan:
Can you please explain what is going on here? What the issue is and what the fix is? I don't understand.

hiaw:
The current code will give error if giving geo coordinate with decimal points. The PR should fix it.

screen shot 2018-01-29 at 1 16 07 pm
screen shot 2018-01-29 at 1 16 30 pm

hiaw:
@rotemmiz this PR is rebased to the master branch. 

But CI will fail because the test project's detox based on the  npm version instead of building from this code change.

https://github.com/wix/detox/blob/master/detox/test/package.json#L19

rotemmiz:
Why do you suspect it uses the npm version?
We use lerna to control the project dependencies. Projects in the monorepo which are declared in lerna.json are being linked inside node_modules rather than being downloaded from npm.

hiaw:
ok. i'm not correct. 

The CI failed because of wrong username & password while trying to publish the doc? 
https://travis-ci.org/wix/detox/jobs/334522771

rotemmiz:
@DanielMSchmidt can you please help with this?

DanielMSchmidt:
@rotemmiz two fixes, one is in #544 so it doesn't even run for the PR.
The second one is we (someone with full access to travis 😉 ) need to set `$GIT_USER` and `$GIT_TOKEN` with valid credentials that allow a push to the repo (for the github-pages branch). For more details, [please see the documentation of Docusaurus](https://docusaurus.io/docs/en/publishing.html)

rotemmiz:
Hi again, 
Please rebase your branch so we can make sure tests pass.

hiaw:
ok. done. no conflict & all test passed

LeoNatan:
Thanks, merged

PR #544, opened 2018-01-28
Website build only run on master

### LINKED ISSUE #542: Fix doc generation to only run from master
Docs generation seem to run from PRs as well currently, we need to make sure it only runs from master (or even better, only on releases) (https://github.com/wix/detox/blob/master/scripts/publish.sh)

While we're at it, generated docs have found their way into master, they need to be purged and git ignored.

DanielMSchmidt:
This fixes #542

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,13 +56,7 @@ matrix:
         - NAME='Website - master'
       node_js: 8
       script:
-      - git config --global user.email "$GIT_USER@users.noreply.github.com"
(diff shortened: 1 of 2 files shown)


rotemmiz:
Do you actually know if this works ? or do we need to TDD (Travis Driven Development) ?

DanielMSchmidt:
I debugged it on my travis 🙈  should be more careful in the future, the last thing was more of a best guess according to the docs

Issue #522, opened 2018-01-16
Android test apk build path not aligned when using flavors


sdg9:
I've set my android binary path to `android/app/build/outputs/apk/pocPlayStore/debug/app-poc-playStore-debug.apk`

In my project when I run `./gradlew assembleAndroidTest` the apk is built to `/android/app/build/outputs/apk/androidTest/pocPlayStore/debug`

Using detox 7.0.0-alpha.0 when running detox tests I get the error

> Error: '/android/app/build/outputs/apk/pocPlayStore/androidTest/debug/app-poc-playStore-debug-androidTest.apk' could not be found, did you run './gradlew assembleAndroidTest' ?

### The Problem
Test APK is actually built to:  `../apk/androidTest/<flavor>/..`
Detox expects: `../apk/<flavor>/androidTest/..`

### Potential solution
In my case, replacing

https://github.com/wix/detox/blob/12df788cabc261ce4a25eb02b1b3e57f87174bd6/detox/src/devices/AndroidDriver.js#L38

with

```js
splitPath.splice(_.findIndex(splitPath, (name) => name === 'apk') + 1, 0, 'androidTest');
```

works but I'm not sure the best approach hence I haven't PRed it.  I assume apk will always be in the build path, although I'm not certain, and that the flavor should appear after this for everyone.

PR #554, opened 2018-02-02
Support testApk path for gradle builds with multiple flavor matrix

sdg9:
I initially created issue #522 as I wasn't familiar with this codebase to PR a solution.

Since then PR #537 was created to fix it.

Unfortunately I still have the problem.  In my real life app example my input SDK path is
```
~/android/app/build/outputs/apk/pocPlayStore/debug/app-poc-playStore-debug.apk
```
and the detox build physically builds to 
```
~/android/app/build/outputs/apk/androidTest/pocPlayStore/debug/app-poc-playStore-debug-androidTest.apk
```

I built off #537 adding my real world example to the tests and updated required logic to generate the expected result. 

I introduced a new strings util file as I needed to do some string manipulation.  Mainly converting `['poc', 'playStore']` to `pocPlayStore`.

### DIFF
--- a/detox/src/devices/android/APKPath.test.js
+++ b/detox/src/devices/android/APKPath.test.js
@@ -12,10 +12,9 @@ describe('APKPath', () => {
   });
 
   it(`path for a gradle build flavor`, async () => {
-    const inputApkPath = '~/somePath/build/outputs/apk/development/debug/app-development-debug.apk';
-    const expectedTestApkPath = '~/somePath/build/outputs/apk/androidTest/development/debug/app-development-debug-androidTest.apk';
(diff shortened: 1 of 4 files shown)


cpetzel:
My React Native application also uses build flavors and cannot execute any tests because of this...

detox/src/devices/android/APKPath.test.js
```diff
@@ -12,10 +12,9 @@ describe('APKPath', () => {
   });
 
   it(`path for a gradle build flavor`, async () => {
-    const inputApkPath = '~/somePath/build/outputs/apk/development/debug/app-development-debug.apk';
```
rotemmiz: Please add tests rather than replacing, we need to make sure all variations work.

sdg9: @rotemmiz good call, done

PR #557, opened 2018-02-04
Fix greeting from rendering and make sure it's matched, but now the test fails

jozan:
For some reason a test for long timeout after press was not working at
all. I noticed that it tested the wrong thing (is button visible)
instead of greeting text which should have been appeared after 4k ms of
the button press.

Greeting text didn’t appear since there was a bug — which was weirdly
enough a valid Javascript syntax — which is now fixed.

However I’ve got no idea how to fix the failing test. Any help?


LeoNatan:
I think the idea of this test has changed over time. Detox ignores JS timers larger than 1.5s, and so, if that button is visible, it means it has not waited for the timer to pass. I will make a change now to better reflect this functionality in the test suite.

jozan:
Alright, good to know. 🙃

Issue #563, opened 2018-02-06
Platform parity and uniform documentation


DanielMSchmidt:
These two topics normally don't mix that good, but please bare with me for a moment. I would like to introduce a change to the way we call native code from the tests, that IMHO enables us to clearer communicate where we are not on par with either platform, to document this and also to ensure new features are built consistently.

**Current State**: Currently we have a rough architecture like this:

```
Test Runner -> classes in expect.js / matchers.js -> invoke.call(Directly) -> WS
```

The part I would like to change is mainly the second one, the wrapper classes around the generated code. They had a reason to exist before we mostly ran on generated code, but nowadays there is no real reason behind having these, expect for that they expose our public API.

**Vision**: 

I would like us to write wrapper code like this:

```javascript
element.cmd("replaceText", {
  "documentation": "Replaces the text of the element",
  "example": "await element(by.id('textField')).replaceText('passcode again');",
  "implementations": {
    "android": ViewActionsApi.replaceText,
    "ios": GreyActions.actionForReplaceText,
  }
});
```

**Benefits**: 

- Super clean API with almost no cognitive overhead
- As this code can be invoked while generating the docs, we would have all the information we need to automate our documentation process for the API
- We can directly see where android / ios features are missing
- Once we have parity we can easily make our tests fail when one implementation is missing if we want to
- We could also "prepare" these commands for features we want to see and have first time contributors fill the gaps. We could show a button like "implement this now" in the docs if there is no implementation or feature parity is not there for an API.

LeoNatan:
Should solve a lot of boilerplate code. 👌
What I am missing is argument information. An example is fine, but sometimes it's hard to convey all inputs possible in one or two examples.

LeoNatan:
Couldn't the third part ("`invoke.call(Directly)`") also be auto-generated?

DanielMSchmidt:
Third part is `invoke.call` or `invoke.callDirectly` directly is for the generated code, the other one is what we assemble ourselves.

DanielMSchmidt:
@rotemmiz You mentioned in person that you have also some input on this, would you mind sharing? 😇

DanielMSchmidt:
I put some more thoughts and research into this I think the approach I proposed goes a bit far away from the current code base (especially the classes). With that in mind I would suggest a more classed based approach like this:

**expect.js**
```js
/**
 *	Some comments on how Element is supposed to work
 */
class Element {
  constructor(matcher) {
    this._originalMatcher = matcher;
[...]

PR #564, opened 2018-02-06
Add generation for GREYInteraction

DanielMSchmidt:
This PR moves most iOS calls over to generated code (`detox_selectElementWithMatcher` is the only missing one and it's a bit of a special case.

### DIFF
--- a/detox/src/ios/expect.js
+++ b/detox/src/ios/expect.js
@@ -12,11 +12,18 @@ const NotExistsMatcher = matchers.NotExistsMatcher;
 const TextMatcher = matchers.TextMatcher;
 const ValueMatcher = matchers.ValueMatcher;
 const GreyActions = require('./earlgreyapi/GREYActions');
+const GreyInteraction = require('./earlgreyapi/GREYInteraction');
+const GreyCondition = require('./earlgreyapi/GREYCondition');
+const GreyConditionDetox = require('./earlgreyapi/GREYConditionDetox');
 
 let invocationManager;
 
 function setInvocationManager(im) {
   invocationManager = im;
+}
+
+function callThunk(element) {
+  return typeof element._call === 'function' ? element._call() : element._call;
 }
 
 //// examples
@@ -41,7 +48,7 @@ detox.invoke.execute(_getInteraction2);
 
 */
 
-class Action {}
+class Action { }
 
 class TapAction extends Action {
   constructor() {
@@ -171,20 +178,15 @@ class Interaction {
 class ActionInteraction extends Interaction {
   constructor(element, action) {
     super();
-    //if (!(element instanceof Element)) throw new Error(`ActionInteraction ctor 1st argument must be a valid Element, got ${typeof element}`);
-    //if (!(action instanceof Action)) throw new Error(`ActionInteraction ctor 2nd argument must be a valid Action, got ${typeof action}`);
-    this._call = invoke.call(element._call, 'performAction:', action._call);
-    // TODO: move this.execute() here from the caller
+
+    this._call = GreyInteraction.performAction(callThunk(element), callThunk(action));
   }
 }
 
 class MatcherAssertionInteraction extends Interaction {
   constructor(element, matcher) {
     super();
-    //if (!(element instanceof Element)) throw new Error(`MatcherAssertionInteraction ctor 1st argument must be a valid Element, got ${typeof element}`);
-    //if (!(matcher instanceof Matcher)) throw new Error(`MatcherAssertionInteraction ctor 2nd argument must be a valid Matcher, got ${typeof matcher}`);
-    this._call = invoke.call(element._call, 'assertWithMatcher:', matcher._call);
-    // TODO: move this.execute() here from the caller
+    this._call = GreyInteraction.assertWithMatcher(callThunk(element), callThunk(matcher));
   }
 }
 
@@ -205,11 +207,15 @@ class WaitForInteraction extends Interaction {
   async withTimeout(timeout) {
     if (typeof timeout !== 'number') throw new Error(`WaitForInteraction withTimeout argument must be a number, got ${typeof timeout}`);
     if (timeout < 0) throw new Error('timeout must be larger than 0');
-    let _conditionCall = invoke.call(invoke.IOS.Class('GREYCondition'), 'detoxConditionForElementMatched:', this._element._call);
-    if (this._notCondition) {
-      _conditionCall = invoke.call(invoke.IOS.Class('GREYCondition'), 'detoxConditionForNotElementMatched:', this._element._call);
+
+    let _conditionCall;
+    if (!this._notCondition) {
+      _conditionCall = GreyConditionDetox.detoxConditionForElementMatched(callThunk(this._element));
+    } else {
+      _conditionCall = GreyConditionDetox.detoxConditionForNotElementMatched(callThunk(this._element));
     }
-    this._call = invoke.call(_conditionCall, 'waitWithTimeout:', invoke.IOS.CGFloat(timeout/1000));
+
+    this._call = GreyCondition.waitWithTimeout(_conditionCall, timeout / 1000)
     await this.execute();
   }
   whileElement(searchMatcher) {
@@ -227,10 +233,11 @@ class WaitForActionInteraction extends Interaction {
     this._originalMatcher = matcher;
     this._searchMatcher = searchMatcher;
   }
+
   async _execute(searchAction) {
-    //if (!searchAction instanceof Action) throw new Error(`WaitForActionInteraction _execute argument must be a valid Action, got ${typeof searchAction}`);
-    const _interactionCall = invoke.call(this._element._call, 'usingSearchAction:onElementWithMatcher:', searchAction._call, this._searchMatcher._call);
-    this._call = invoke.call(_interactionCall, 'assertWithMatcher:', this._originalMatcher._call);
+    const _interactionCall = GreyInteraction.usingSearchActionOnElementWithMatcher(callThunk(this._element), callThunk(searchAction), callThunk(this._searchMatcher));
+
+    this._call = GreyInteraction.assertWithMatcher(_interactionCall, callThunk(this._originalMatcher));
     await this.execute();
   }
   async scroll(amount, direction = 'down') {
@@ -296,7 +303,7 @@ class Element {
   }
 }
 
-class Expect {}
+class Expect { }
 
 class ExpectElement extends Expect {
   constructor(element) {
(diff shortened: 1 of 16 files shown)


rotemmiz:
Is this still a WIP?

DanielMSchmidt:
Yes, but close to being finished

DanielMSchmidt:
I now rebased this PR. The original goal was more ambitious, but I think incremental steps forward are a nicer way to work together 🙈 I think this might help with #618 maybe only by us having a conversation about what these instance methods mean for us in detox 👍

generation/__tests__/ios.js
```diff
@@ -333,6 +333,28 @@ describe("iOS generation", () => {
 		});
 	});
 
+	describe("instance methods", () => {
```
rotemmiz: Those tabs look awful in github :(

LeoNatan: They look great. Don’t switch to spaces.

generation/core/generator.js
```diff
@@ -7,13 +7,14 @@ const fs = require("fs");
 
 const { methodNameToSnakeCase } = require("../helpers");
 let globalFunctionUsage = {};
-module.exports = function({
+module.exports = function ({
```
rotemmiz: Could you please make it a bit more readable by creating a named function and exporting it, instead of exporting an anonymous one. Sorry for nitpicking, I just can't get used to this javascript jungle magic. :(

generation/core/generator.js
```diff
@@ -81,13 +89,22 @@ module.exports = function({
 	}
 
 	function createMethod(classJson, json) {
+		// For us this means that we don't set the class type, but let it be set from the outside
+		// This is an assumption based on our current experience with EarlGrey, we might need to rework this at some point
+		const isTranslucent = !json.static;
```
rotemmiz: what does `translucent` mean in this context ?

generation/core/generator.js
```diff
@@ -112,7 +129,7 @@ module.exports = function({
 		return json;
 	}
 
-	function createMethodBody(classJson, json) {
+	function createMethodBody(classJson, json, isTranslucent) {
```
rotemmiz: try not to pass booleans in function params, it's very confusing.
https://martinfowler.com/bliki/FlagArgument.html

rotemmiz: Instead, just create two functions with clear naming

generation/core/generator.js
```diff
@@ -163,24 +184,29 @@ module.exports = function({
...
-							t.objectProperty(
-								t.identifier("value"),
-								addArgumentContentSanitizerCall(arg, json.name)
-							)
-						])
+						t.objectProperty(
+							t.identifier("type"),
+							t.stringLiteral(addArgumentTypeSanitizer(arg))
+						),
+						t.objectProperty(
```
rotemmiz: Is this formatting the work of prettier ? It's very weird,
screen shot 2018-03-11 at 20 41 43

generation/core/generator.js
```diff
@@ -215,7 +243,7 @@ module.exports = function({
 			: typeCheckCreator(json);
 	}
 
-	return function(files) {
+	return function (files) {
```
rotemmiz: Again, javascript voodoo. I'm too old for this magic. ;(

generation/package.json
```diff
@@ -23,7 +23,6 @@
 		"jest": "^20.0.4",
 		"lerna": "2.0.0-rc.4",
```
rotemmiz: You don't need lerna here, it can be safely deleted

rotemmiz:
Of course this needs to pass e2e before merge (from travis log)
```
     Error: Error: target is invalid
      at Client.execute (/Users/travis/build/wix/detox/detox/src/client/Client.js:75:13)
      at <anonymous>
```

generation/core/generator.js
LeoNatan: What is going on here?

generation/__tests__/ios.js
rotemmiz: 🛒- i

generation/core/generator.js
DanielMSchmidt: sure, what would be a good name for a function that returns a code generator?

generation/core/generator.js
DanielMSchmidt: It means that we don

generation/core/generator.js
rotemmiz: getGenerator ?

generation/core/generator.js
rotemmiz: Not sure I got it, can you elaborate?

generation/package.json
DanielMSchmidt: sure 👍

generation/core/generator.js
DanielMSchmidt: Prettier being prettier 🤷‍♂️

generation/core/generator.js
DanielMSchmidt: Sorry, pressed enter too early, does this make more sense?

generation/core/generator.js
DanielMSchmidt: I moved this part down into the functions

DanielMSchmidt:
@rotemmiz I refactored the things you mentioned and got a working version up (let's see if CI agrees). I could only finish one of the calls, but left TODO items on the others. They "only" need a bit of fine tuning, but I thought this progress might already be helpful for you, especially in regards of #618

rotemmiz:
Awesome! Will we be able to generate non static invocations now ? Can't see an example in the new generated code, can you please point me to it ?

detox/src/ios/expect.js
```diff
@@ -164,20 +171,17 @@ class Interaction {
 class ActionInteraction extends Interaction {
   constructor(element, action) {
     super();
-    //if (!(element instanceof Element)) throw new Error(`ActionInteraction ctor 1st argument must be a valid Element, got ${typeof element}`);
-    //if (!(action instanceof Action)) throw new Error(`ActionInteraction ctor 2nd argument must be a valid Action, got ${typeof action}`);
-    this._call = invoke.call(element._call, 'performAction:', action._call);
-    // TODO: move this.execute() here from the caller
+
+    this._call = GreyInteraction.performAction(callThunk(element), callThunk(action));
```
DanielMSchmidt: @rotemmiz this is currently the only successful invocation of an instance method, I need to find out why the others fail, but I guess it's a call side problem

rotemmiz:
Hey @DanielMSchmidt , anything to be done on your side ?

detox/src/ios/expect.js
rotemmiz: Is this resolved ?

DanielMSchmidt:
Not yet, trying to get to it next weekend

DanielMSchmidt:
@rotemmiz This is done now 👍 Let's get it into master ;)

DanielMSchmidt:
@rotemmiz ping, could you review this PR? 🙏 😇

generation/core/generator.js
```diff
@@ -81,13 +89,19 @@ module.exports = function({
...
+
 		const m = t.classMethod(
 			"method",
 			t.identifier(methodNameToSnakeCase(json.name)),
-			json.args.map(({ name }) => t.identifier(name)),
+			args,
 			t.blockStatement(createMethodBody(classJson, json)),
 			false,
-			json.static
+			true
```
rotemmiz: will this mean every method is static now ?

generation/core/generator.js
```diff
@@ -189,10 +214,12 @@ module.exports = function({
 				t.objectProperty(
 					t.identifier("target"),
 					t.objectExpression([
-						t.objectProperty(t.identifier("type"), t.stringLiteral("Class")),
+						t.objectProperty(t.identifier("type"), t.stringLiteral(json.static ? "Class" : "Invocation")),
```
rotemmiz: ok, got it

generation/core/generator.js
rotemmiz: Then why passing this boolean anyway ? why not get rid of it ?

rotemmiz: https://martinfowler.com/bliki/FlagArgument.html

generation/core/type-checks.js
```diff
@@ -26,8 +26,33 @@ const isGreyMatcher = ({ name }) =>
     throw new Error('${name} should be a GREYMatcher, but got ' + JSON.stringify(ARG));
   }
 `)({
-		ARG: t.identifier(name)
-	});
+			ARG: t.identifier(name)
+		});
+const isGreyAction = ({ name }) =>
```
rotemmiz: What do you think about just defining functions instead of assigning functions to variables ?

```js
function isGreyAction({name}) {
}
```

generation/core/generator.js
DanielMSchmidt: This is a method of `babel-types`, I could write a wrapper like `createStaticClassMethod`, would that be more readable for you?

Every method is static, I choose to deal with instance methods a little different, I just let the target be passed in as the first argument. My point for this is that we have the class-like behavior more on the caller-side of this function and I don't want to duplicate it in any way. Also, the calls are more similar now than they would have been otherwise. If we want to change this somewhen I would propose we rethink this at a point in time when we get rid of these wrapper classes and access the generated code "directly", like in this proposal: #563

generation/core/type-checks.js
DanielMSchmidt: I like the idea 👍

DanielMSchmidt:
@rotemmiz Addressed your comments / replied. Thank you for taking the time to review <3

DanielMSchmidt:
Fun fact: This PR moves us from 47 to 40 open `.call` invocations

rotemmiz:
💪

rotemmiz:
@DanielMSchmidt , I'm getting a weird behaviour when running ci.ios.sh locally.
generation script generates something a bit different and breaks tests. Can you please take a look ?

rotemmiz:
I now get it, I switched to EarlGrey 1.13, there is some breakage, not sure what.

rotemmiz:
@LeoNatan, [NS_REFINED_FOR_SWIFT macro](https://github.com/google/EarlGrey/commit/499474e02cc486b65c360da362173ec77e7495ee#diff-ad85ec7ab42faaa4cd1a922503fca1caR18) was added to a few of EarlGrey's functions. 
What should we do if we want to invoke this code through `DTXMethodInvocation`? should we just ignore it in the method signature ?

LeoNatan:
Refined means there is a better alternative for Swift. It has no influence on ObjC.

rotemmiz:
So 
`method: "performAction:"`
not
`method: "performAction:NS_REFINED_FOR_SWIFT"`

right?

LeoNatan:
Yes, of course.

It's an attribute.
`# define CF_REFINED_FOR_SWIFT __attribute__((swift_private))`

PR #583, opened 2018-02-26
add functionality for scrolling date picker

DmitryPonomarenko:
add GreyAction for scrolling date picker iOS (changing value at defined column)


docs/APIRef.ActionsOnElement.md
```diff
@@ -107,3 +108,10 @@ await element(by.id('scrollView')).swipe('down');
 await element(by.id('scrollView')).swipe('down', 'fast');
 await element(by.id('scrollView')).swipe('down', 'fast', 0.5);
 ```
+
+### `scrollDatePickerAtColumnToValue(column,value)`
+Scroll Date Picker at defined column to required value
+
+```js
+await element(by.type('UIPickerView')).scrollDatePickerAtColumnToValue(0,'22');
```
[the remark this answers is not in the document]
rotemmiz: we encourage usage of `by.id`, please use testIDs even in docs.

rotemmiz: 1. Please add a valid test screen in the test project, add both date and time pickers, and add a test to verify the API works.
2. Android implementation needs to have a different API. We can not use same API for both, Android has only `date` or `time` pickers, iOS has `date`, `time` and `datetime` pickers, therefore:
iOS API should use `setValueAtColumn`
Android API should use `setDate(y, m, d)` and `setTime(h, m)`

Android resources:
  [DatePicker](https://developer.android.com/reference/android/widget/TimePicker.html)
  [TimePicker](https://developer.android.com/reference/android/widget/DatePicker.html)
Espresso API: 
  [PickerActions](https://developer.android.com/reference/android/support/test/espresso/contrib/PickerActions.html)
  [Example](https://stackoverflow.com/questions/44835094/check-datepicker-calendar-value-in-android-espresso-framework)

detox/src/ios/expect.js
```diff
@@ -284,6 +291,9 @@ class Element {
     this._selectElementWithMatcher(this._originalMatcher._avoidProblematicReactNativeElements());
     return await new ActionInteraction(this, new SwipeAction(direction, speed, percentage)).execute();
   }
+  async scrollDatePickerAtColumnToValue(column,value) {
```
rotemmiz: This might need a name change as well, so we won't get confused by arg order.
`setPickerColumnToValue` or something similar.

docs/APIRef.ActionsOnElement.md
LeoNatan: Let's make the Android API more consistent with iOS:

`setToDate(y, m, d)` and `setToTime(h, m)`

examples/demo-react-native/e2e/example.spec.js
```diff
@@ -6,6 +6,14 @@ describe('Example', () => {
   it('should have welcome screen', async () => {
     await expect(element(by.id('welcome'))).toBeVisible();
   });
+
+  it('open datePicker screen', async () => {
+    await element(by.id('datePickerIOS')).tap();
+    await expect(element(by.type('UIPickerView'))).toBeVisible();
+    await element(by.type('UIPickerView')).setColumnToValue(1,"6")
+    await element(by.type('UIPickerView')).setColumnToValue(2,"34")
```
LeoNatan: Let's add an `expect` here that checks that the correct date has been selected in the picker.

examples/demo-react-native/app.js
```diff
@@ -10,19 +10,43 @@ import {
...
     if (this.state.greeting) return this.renderAfterButton();
-    return (
+    return this.state.goToScreen === 'datePicker' ? (
+      <View style={{flex: 1, paddingTop: 20, justifyContent: 'center', alignItems: 'center'}}>
+        <DatePickerIOS testID = {'CustomDatePicker'}
+          style = {styles.datePicker}
+          date={this.state.chosenDate}
+          onDateChange={this._setDate}
+        />
+        <TouchableOpacity testID='back_button' onPress={() => {this.setState({goToScreen: false})}}  >
```
LeoNatan: I don't think we need a back button. We always call `reloadReactNative` after every test.

rotemmiz: This is not the place, move it to `detox/test` project

LeoNatan:
@DmitryPonomarenko Why did you close the PR?

PR #586, opened 2018-02-27
Update Example .travis.yml in CI Guide

mtmckenna:
Installing applesimutils seems to be required to have builds pass in Travis.

### DIFF
--- a/docs/Guide.RunningOnCI.md
+++ b/docs/Guide.RunningOnCI.md
@@ -68,6 +68,8 @@ env:
   - NODE_VERSION=stable
 
 install:
+- brew tap wix/brew
+- brew install --HEAD applesimutils
 - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.2/install.sh | bash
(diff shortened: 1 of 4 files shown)


docs/Guide.RunningOnCI.md
```diff
@@ -68,6 +68,8 @@ env:
   - NODE_VERSION=stable
 
 install:
+- brew tap wix/brew
+- brew install --HEAD applesimutils
```
LeoNatan: Please do no longer use `—HEAD`. There are proper versions now, which should be stable.

mtmckenna:
Based on feedback from @LeoNatan , I removed the `--HEAD` argument from the `brew` command. I also removed `--HEAD` from the same command I found elsewhere in the docs.

I left the `--HEAD` in `scripts/install.ios.sh` because I didn't want this pull request to modify any code. However, perhaps it makes sense to make the change to `.install.ios.sh` as well in a different commit?

docs/Guide.RunningOnCI.md
mtmckenna: Thank you! Updated the pull request.

mtmckenna:
Hello! Anything I should do to help get this one merged in? Thank you!

PR #587, opened 2018-02-27
[WIP] Website: show instructions side by side

DanielMSchmidt:
This PR aims to show the install instructions side by side, similar to the react-native docs.

- [X] improve tooling
- [X] build component
- [x] style component
- [ ] use component

### DIFF
--- a/website/.babelrc
+++ b/website/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["env", "react"]
+}
--- a/website/.storybook/addons.js
+++ b/website/.storybook/addons.js
@@ -1,0 +1,2 @@
+import "@storybook/addon-actions/register";
+import "@storybook/addon-links/register";
--- a/website/.storybook/config.js
+++ b/website/.storybook/config.js
@@ -1,0 +1,9 @@
+import { configure } from "@storybook/react";
+
+// automatically import all files ending in *.stories.js
+const req = require.context("../stories", true, /.stories.js$/);
+function loadStories() {
+  req.keys().forEach(filename => req(filename));
+}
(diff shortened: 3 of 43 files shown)


DanielMSchmidt:
Update from my side: I will be working on it this evening

DanielMSchmidt:
I am continuing the work on this right now 👍

DanielMSchmidt:
Status: Still need to use the component, but it looks okay-ish in the storybook :)

DanielMSchmidt:
Oh man, this is harder then it seemed -.- But we are getting there, had to remove styled components as they didn't work well with docusaurus. Now somehow the included component is not dynamic, it doesn't do anything on click. This needs more investigation on my side.

DanielMSchmidt:
Uh nice, who is it? :trollface: Thank you for your patience

DanielMSchmidt:
Asking for help is our best option here, I wrote a demo, will keep you posted: https://github.com/facebook/Docusaurus/issues/584

DanielMSchmidt:
@rotemmiz Outcome of the ticket on Docusaurus was that they don't rehydrate react, so they don't support interactivity in their documentations. We can go different routes from here:

1. Build the rehydrating on our own (You can load a script into docusaurus and we could use normal script tags to inline the rehydration logic) - should be feasible, might be hacky.
2. Open up an upstream PR for adding interactivity - we need to discuss that first with the maintainers, there might be good reasons against it in a general case that we are not aware of currently.
3. Drop the side-by-side functionality and anything interactive in the docs

DanielMSchmidt:
Closing this for now

PR #598, opened 2018-03-04
test sharding/parallelization 

doronpr:
+ moved e2e tests to jesr


LeoNatan:
#601 is blocking this. Should be fixed first before merging this.

doronpr:
@LeoNatan why blocker?
there's going to be a bug where PN isn't ready for parallelization but we can fix it when we get to it. No?

LeoNatan:
I don't think that is the right approach. I think we should fix all known issues before merging this large PR. 
in the case of #601, it's a very thing to fix.

doronpr:
closing in favour of https://github.com/wix/detox/pull/609

PR #600, opened 2018-03-05
Update build.gradle

joegoodall1:
Needs to match package in AndroidManifest.xml


rotemmiz:
Why is that needed?

rotemmiz:
Gradle overrides these values in manifest.

PR #604, opened 2018-03-06
Improve iOS notifications dispatch timing

### LINKED ISSUE #590: Starting application from background with notification happens in foreground (iOS)
When trying to test opening an application from the background with a push notification with this code:

```js
await device.sendToHome();
await device.relaunchApp({ newInstance: false, userNotification });
```

The notification is sent to the application but only after the native `UIApplicationState` is already changed to `UIApplicationStateActive`. This makes the application behave as if the notification was opened when the app was already in the foreground.

The same behaviour happens when trying like this:

```js
await device.sendToHome();
await device.relaunchApp({newInstance: false});
```

@LeoNatan

LeoNatan:
Closes #590


rotemmiz:
shipit 
...
...
...
...
...





:(


@LeoNatan, get us that cool bot!

PR #605, opened 2018-03-06
Functionality for testing picker ios

DmitryPonomarenko:
Add test screen with pickerView(iOS)
Add tests to e2e folder for this pickerView


detox/test/e2e/p-datePicker.js
```diff
@@ -0,0 +1,17 @@
...
+    beforeEach(async () => {
+      await device.reloadReactNative();
+    });
+  
+    beforeEach(async () => {
+      await element(by.text('DatePicker')).tap();
+    });
+    
+    it('check and scroll datePicker', async () => {
+        await expect(element(by.type('UIPickerView'))).toBeVisible();
```
LeoNatan: Why by type and not by id? Does it not work?

detox/test/ios/example.xcodeproj/project.pbxproj
```diff
@@ -491,10 +519,14 @@
 				39A34C5D1E30ED3600BEBB59 /* libcxxreact.a */,
 				39A34C5F1E30ED3600BEBB59 /* libjschelpers.a */,
 				39A34C611E30ED3600BEBB59 /* libjschelpers.a */,
+				D58447B0204E9C6000248543 /* libjsinspector.a */,
```
LeoNatan: What is this? What will happen if old RN is used? In CI, we have an RN matrix that changes RN versions on the fly. Currently we are testing RN 51 and 53.3

detox/test/src/Screens/DatePickerScreen.js
```diff
@@ -0,0 +1,34 @@
...
+export default class DatePickerScreen extends Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    return (
+      <View style={{flex: 1, paddingTop: 20, justifyContent: 'center', alignItems: 'center'}}>
+        <DatePickerIOS testID = {'CustomDatePicker'}
```
LeoNatan: If there is `testID` why find by type in the test?

DmitryPonomarenko: UIPickerView have not accessibility ID, testID = {'CustomDatePicker'} need to be deleted

detox/test/e2e/p-datePicker.js
DmitryPonomarenko: UIPickerView have not accessibility ID, testID = {'CustomDatePicker'} need to be deleted

LeoNatan:
Please do not open new PRs, just push in to previous.

From the previous PR discussion:
>Android implementation needs to have a different API. We can not use same API for both, Android has only date or time pickers, iOS has date, time and datetime pickers, therefore:
iOS API should use setColumnToValue
Android API should use setToDate(y, m, d) and setToTime(h, m)

> Android resources:
  [DatePicker](https://developer.android.com/reference/android/widget/TimePicker.html)
  [TimePicker](https://developer.android.com/reference/android/widget/DatePicker.html)
Espresso API: 
  [PickerActions](https://developer.android.com/reference/android/support/test/espresso/contrib/PickerActions.html)
  [Example](https://stackoverflow.com/questions/44835094/check-datepicker-calendar-value-in-android-espresso-framework)

detox/test/e2e/p-datePicker.js
LeoNatan: This is a problem. If user has more than one pickers, the API will break

LeoNatan: `UIPickerView` is a `UIView`, so it can have an accessibility ID. Are you seeing that `UIDatePicker` does not forward that ID?

LeoNatan:
You are missing the `docs/APIRef.ActionsOnElement.md` information in this PR.

detox/test/e2e/p-datePicker.js
DmitryPonomarenko: I checked that UI element via AcessebilityInspector (from Xcode Instruments), and now I found only one way to get element by type

LeoNatan: OK, thanks

detox/test/e2e/p-datePicker.js
```diff
@@ -0,0 +1,17 @@
...
+        await expect(element(by.type('UIPickerView')).setColumnToValue(1,"6"));
+        await expect(element(by.type('UIPickerView')).setColumnToValue(2,"34"));
```
LeoNatan: Expect should not be on actions.
You should have a label and see that the correct value is shown there.

Should be something like 
```
await element(by.type('UIPickerView')).setColumnToValue(2,"34");
await expect(element(by.id('time'))).toHaveText('6:34');
```

timsawtell-sportsbet:
I was working on a PR of my own to fix this issue and should have checked here first! Very keen to see this enhancement get in, as it's a blocker for our tests at the moment.

Would it be possible to get this merged with support for iOS only? I notice the implementation for Android is missing, and I myself haven't attempted to implement the Android version because we're not using it. Surely motivated Android devs could implement the GreyAction for Android should they have a desire to get it working. *crosses fingers*

LeoNatan:
@timsawtell-sportsbet We discusses this internally and finallly decided we will merge it before the Android support lands. Once the PR is satisfactory (very soon), we will merge.

LeoNatan:
Why is this stale for so long? @DmitryPonomarenko 
Now there are conflicts in the branch. Please merge mast first and then make the changes as needed.

detox/test/e2e/p-datePicker.js
skv-headless: It won't work on android

LeoNatan: We know. We want to hone down iOS support first, then we will implement Android as well.

generation/package.json
```diff
@@ -1,52 +1,47 @@
...
-    }
-  },
-  "dependencies": {
-    "babel-generate-guard-clauses": "^2.0.0",
-    "babel-template": "^6.26.0",
-    "java-method-parser": "^0.4.5"
-  }
-}
\ No newline at end of file
+	"name": "generation",
```
LeoNatan: What changed in this file? If nothing, please revert this file.

docs/APIRef.ActionsOnElement.md
```diff
@@ -107,3 +108,13 @@ await element(by.id('scrollView')).swipe('down');
 await element(by.id('scrollView')).swipe('down', 'fast');
 await element(by.id('scrollView')).swipe('down', 'fast', 0.5);
 ```
+### `setColumnToValue(column,value)`
```
LeoNatan: This should be explicitly marked as iOS only in the documentation.

detox/test/ios/example.xcodeproj/project.pbxproj
```diff
@@ -1145,7 +1201,7 @@
 					"-ObjC",
 					"-lc++",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)";
+				PRODUCT_BUNDLE_IDENTIFIER = org.reactjs.native.example.example;
```
LeoNatan: Why is this change necessary?

detox/test/e2e/p-datePicker.js
LeoNatan: This test should be wrapped with `if(iOS)`.

detox/test/e2e/p-datePicker.js
```diff
@@ -0,0 +1,17 @@
...
+    });
+  
+    beforeEach(async () => {
+      await element(by.text('DatePicker')).tap();
+    });
+    
+    it('check and scroll datePicker', async () => {
+        await expect(element(by.type('UIPickerView'))).toBeVisible();
+        await element(by.type('UIPickerView')).setColumnToValue(1,"6");
+        await element(by.type('UIPickerView')).setColumnToValue(2,"34");
```
LeoNatan: Right now, this test checks nothing. There should be a label in the screen, and on date picker change value, it should update. You should then assert that the label has the correct value.

DmitryPonomarenko: Checking label with text will be always true, because this label always on date picker, I think right solution is checking property of datePicker current value or something like that, now I work in this direction, add this functionality asap

LeoNatan: I don't understand. The label is updated with the value set in the picker. So if you select 6:41, it won't say "6:34" in the label. If you assert that `label.text == "6:34"`, you will test that the picker really changed value. That's what we want to test in the testing suite.

DmitryPonomarenko: In my example if element (UIPickerView) can't find variable with current column or value test will be failed

LeoNatan: But that is not what we test in the suite. We test that the native system behaves as expected following Detox changes.

LeoNatan:
As we discussed, also please test if Earl Grey has API to assert on the picker value at column.

detox/test/e2e/s-datePicker.js
```diff
@@ -0,0 +1,22 @@
+describe('DatePicker', () => {
+    beforeEach(async () => {
+      await device.reloadReactNative();
+    });
+  
+    beforeEach(async () => {
+      await element(by.text('DatePicker')).tap();
+    });
+    
+    it('datePicker and dateLabel should be visible', async () => {
```
LeoNatan: This is not necessary.

detox/test/ios/example.xcodeproj/project.pbxproj
```diff
@@ -1137,15 +1137,15 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				DEAD_CODE_STRIPPING = NO;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = S3GLW74Y8N;
```
LeoNatan: I don't understand why these changes are going in?
Please revert it. The entire project,pbxproj file changes seem to me like something that should not have been committed.

generation/package.json
LeoNatan: This still remains open. Why did the file change?

PR #614, opened 2018-03-09
Clarify google() repo needs to be added twice.

markhu:


paulbrittain:
Nice, necessary addition. I don't think the last sentence `Remember you need to add it twice, once to each section.` is needed though.

rotemmiz:
Hey there, thanks for the clarification, but I still think its not clear enough. If you can, please add example of the missing configuration. I agree with the previous comment of @paulbrittain

Issue #607, opened 2018-03-06
Introduce complementary UiAutomator API on Android  


guyca:
Motivation:
While Espresso is sufficient for most use cases, interacting with UI outside of the application is a common requirement.
Work has already been done in this regard [here](https://github.com/wix/detox/blob/master/detox/android/detox/src/main/java/com/wix/detox/uiautomator/UiAutomator.java)

rotemmiz:
It can be very useful to generate UIAutomator's js API. @danielmschmidt

https://developer.android.com/reference/android/support/test/uiautomator/UiDevice.html

d4vidi:
@omerwin please associated this ticket with your PR 🙏

PR #618, opened 2018-03-11
[WIP] Generate UIDevice.java API for JS invocation

rotemmiz:
@DanielMSchmidt, I tried generating UIDevice.java's API, but most of its functions are non-static, and I'm not sure how to access them this way.
This is how we currently access the UIDevice instance:
https://github.com/wix/detox/blob/master/detox/src/devices/AndroidDriver.js#L100

Any Ideas ?


rotemmiz:
This relates to #607

rotemmiz:
My suggestion: 
For non static java functions the generator should create a json with `Invocation` instead of `Class`, and get the object instance as a parameter. the js function should still be static.

DanielMSchmidt:
@rotemmiz I think I did this in this PR: https://github.com/wix/detox/pull/564/files#diff-fce821d0f871b8c551c11b2428f7cec7

The API also resolved around non-static instances (see createMethod in `generation/core/generator.js`). I am not 100% sure on which track I was there, but I will follow up on my thoughts there, rebase the PR and try to get it into master first, as a Proof of Concept for these non-static functions

rotemmiz:
done in #743

PR #623, opened 2018-03-14
iOS: Add support for userActivity API

### LINKED ISSUE #601: `createPushNotificationJson` creates a notification.json file under a constant path
This is no good if we want parallelization.
Path should either be generated using `mktemp` (and cleaned up after use), or it should contain enough process information to be unique per instance.

### LINKED ISSUE #622: Support UserActivity
Enable associating external links to an app so it can trigger the opening/ be associated with it and triggered via an external link. For example and invite link to an app.

LeoNatan:
At first will only include associated domain user activity, but will be extensible.

For this, I merged `openURL()` and `sendUserNotification()` internally to `deliverPayload()`. The new `userActivity()` API will use that same mechanism.

Closes #622

Also fixed #601 in this branch


detox/test/e2e/k-user-notifications.js
```diff
@@ -23,7 +23,7 @@ describe(':ios: User Notifications', () => {
     await expect(element(by.text('From calendar'))).toBeVisible();
   });
 
-  it('Foreground push notifications', async () => {
+  it.only('Foreground push notifications', async () => {
```
rotemmiz: you forgot this here

LeoNatan: Yes

LeoNatan:
Hm, something failed for RN53. Will look romorrow.

detox/src/devices/Device.js
```diff
@@ -123,6 +113,20 @@ class Device {
     return !params.delete && !params.newInstance && this._processes[_bundleId];
   }
 
+  _assertHasSingleParam(singleParams, params) {
```
LeoNatan: Isn't the "assert" moniker incorrect if it has the side-effect of returning the count? Should be "count" and the `throw` should be in the calling method.

detox/src/devices/Device.js
```diff
@@ -180,15 +184,10 @@ class Device {
...
-  //Do not remove this, it is very important for our customers!
-  async __debug_sleep(ms) {
-    return new Promise(resolve => setTimeout(() => resolve(), ms));
-  }
-
   async _sendPayload(key, params) {
     const payloadFilePath = this.deviceDriver.createPayloadFile(params);
     let payload = {};
-    //JS does not support {key: "asd"} JSON generation where the `key` is a variable 🤦‍♂️
+    //JS does not support {key: "asd"} JSON generation where the `key` is a variable
```
LeoNatan: 🤦‍♂️

LeoNatan:
If the tests fail, just restart the specific build that failed. I've been doing that for a few times. Travis…

detox/src/devices/Device.js
rotemmiz: Not done yet...
+ There's Android client to align to the protocol changes

rotemmiz:
Android is all good. When CI is green, let's merge

detox/src/devices/Device.js
```diff
@@ -187,7 +184,6 @@ class Device {
   async _sendPayload(key, params) {
     const payloadFilePath = this.deviceDriver.createPayloadFile(params);
     let payload = {};
-    //JS does not support {key: "asd"} JSON generation where the `key` is a variable
```
LeoNatan: Why was this removed? This line is important, otherwise I might be confused like before. There is nothing logical about JS.

rotemmiz: Then let's write a manifest about everything that's broken in JavaScript, it's not gonna be a short one...

The tests protect this specific thing from a wrong change.

LeoNatan: Why is this comment bothering you?

PR #626, opened 2018-03-16
Support spaces in iOS app name

ssg-luke:
Ran into some issues getting detox to work on the first project I used it with. This allow apps with spaces in the name to run correctly.


rotemmiz:
I think it's gonna pass once you fix unit tests

PR #628, opened 2018-03-19
Running on windows support

simonbuchan:
Just a few issues I ran into running on windows. Boils down to:
- npm scripts (that might run on windows) should be either:
  - `node some/script.js` (what I've done) or
  - add `cross-env` dev dep, and use `cross-env some/script` where there is a `some/script` with a `#!` and `some/script.bat` written for windows `cmd.exe`. This one is no fun.
- A few cases of `exec('node_modules/.bin/foo'...)` which doesn't work in cmd.exe (it tries to run `node_modules` with `/.bin/foo`). Not quite sure about my fixes for these, I'm happy to change things.
- several cases of `.split('\n')`, which broke at least starting the emulator (looking for `foo`, not finding `foo\r`) - fixed this at the source, by replacing `\r\n` on windows.

I also noticed `detox/scripts/build.sh android` was a little out of date while porting, is that code still used?

Otherwise, things seem to work quite well! Haven't confirmed I haven't broken linux / mac, so


simonbuchan:
Also I didn't touch `detox-cli` - I'm not quite sure what the point is? `yarn <bin>` and `npx <bin>` obsolete packages like that. I'm happy to port it to node or batch and test if you like?

DanielMSchmidt:
@simonbuchan Thank you for taking care of this <3 👍 
Would you mind adding some docs? Do you know if we can test some of these things in our CI? Do we need another solution there?

simonbuchan:
There's a lot more rough edges I'd like to iron out first, e.g. when using `android.emulator` it gets stuck 
trying to start the emulator again on the second run (not sure what's happening there), and I'm not sure how to approach the `build` config step yet, but these get rid of the big blockers for the basic -> `detox test` flow.

I've got a couple of hacks that could make all the tests pass on windows (all just path separators, AFAICT), but I don't know if that means they actually work right! At least until I've gotten on top of those I'm not sure where a windows CI would help, nor how that would work. You might have noticed I had to ignore the windows branch in exec for coverage to pass, and I'm not sure how that would work cross-platform (collect on all platforms and merge in a final step? Sounds complex!)

So far not sure what docs would change? "Windows will probably mostly work" isn't a great line! And I'd like to see some other users experience with this first before promising more. The big difference would probably be whatever happens with `build`.

Ideally we will be using this cross-platform quite a bit more in the future, and I'll be happy to follow this up with more support on request (though keep in mind I'm in UTC+12:00).

simonbuchan:
Talking of CI, actually, I noticed there's a pretty complex `eslint` config that is not being used or followed at all? What's up with that, and should I be following it in `scripts`?

simonbuchan:
Let me know if the above fix is OK in theory, it seems to be required for windows `android.emulator` (vs `.attached`).

rotemmiz:
This is something we'll have to test locally, there's no test in CI that verifies that a second e2e works, or is there....  
1. The demo-projects CI job runs two subsequent tests on two different projects, with the same devices. 
2. but there still is no Android E2E in CI, we're currently working on moving to our own CI solution (Jenkins with Mac virtual machines capable of running x86 emulators for those who ask)

simonbuchan:
It seems like it does, since it still has 100% coverage? I might be misunderstanding branch coverage.

rotemmiz:
I will be reviewing this soon, sorry for the delay.

tachtevrenidis:
is this going to be merged soon? :-)

detox/local-cli/detox-run-server.js
```diff
@@ -2,14 +2,11 @@
 
 const program = require('commander');
 const cp = require('child_process');
-const fs = require('fs');
 const path = require('path');
 
 program.parse(process.argv);
 
-if (fs.existsSync(path.join(process.cwd(), 'node_modules/.bin/detox-server'))) {
```
rotemmiz: npm and yarn install detox-server at different locations.
it might be either it `node_modules` or in `node_modules/detox/node_modules`

I am not familiar with `require.resolve('detox-server/package.json');` trick. Will it cover both cases ?

detox/scripts/postinstall.js
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
+
+// Javascript for windows support.
```
rotemmiz: Unnecessary comment

rotemmiz:
Sorry for the delay in reviewing, but it is now done :)

Unfortunately, we have no capacity to run Detox on Windows in CI, therefore I cannot guarantee this will continue working without breaking.  We will need to add a disclaimer in the docs, and rely on the community to report issues and fix them. :(
Not ideal, but this is the best we can do right now.

detox/src/devices/EmulatorDriver.js
```diff
@@ -52,10 +52,17 @@ class EmulatorDriver extends AndroidDriver {
...
-    const filteredDevices = _.filter(adbDevices, {type: 'emulator', name: name});
+    let adbDevices = await this.adb.devices();
+    let filteredDevices = _.filter(adbDevices, {type: 'emulator', name: name});
+
+    // If it's not already running, start it now.
+    if (!filteredDevices.length) {
+      await this.emulator.boot(name);
+    }
+
+    adbDevices = await this.adb.devices();
```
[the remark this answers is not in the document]
simonbuchan: Oops, should be inside the `if` here - it's refreshing `adbDevices` after the requested name is booted. See the comments above as to why this is happening, but roughly: on windows trying to start the emulator the way it currently is being done when it's already running causes issues for some reason.

detox/local-cli/detox-run-server.js
simonbuchan: Yep, it's the standard hack to find the location of an NPM package the same way node does, since every NPM package has a `package.json`. Remember:
- `require / require.resolve('foo/bar')` looks up `foo` in node_modules, then adds `bar` to that path
- without the `.../bar` it would use the `main` field, which is likely to be in a nested subdir
- `package.json` always exists.

detox/scripts/postinstall.js
simonbuchan: More clearly `Using JS instead of shell script to avoid breaking on windows`, but I'm happy to remove it!

simonbuchan:
If you want to look into CI later, the standard I've seen in open source is https://www.appveyor.com/

simonbuchan:
### CI failure

Hmm, not sure what's up: only one config, so seems like it flaked?

### CI

Blorg, never mind, CircleCI is mac/linux only.

Looks like:
- Typescript doesn't run CI on windows? They have CircleCI and Travis running, but both are linux.
- Same for babel
- react-native only have circleci.

Weird. Seems like AppVeyor do have decent-ish [node support](https://www.appveyor.com/docs/lang/nodejs-iojs/), even if the docs are old. I could look into setting this up, I think the main thing would be installing the android SDK on windows reliably.

### Tests

Note that the tests do *not* all pass on windows yet - here's an example, all 4 failures are trivial path things like this:

```
  ● APKPath › simple path

    expect(received).toEqual(expected)
    
    Expected value to equal:
      "~/somePath/build/outputs/apk/androidTest/debug/app-debug-androidTest.apk"
    Received:
      "androidTest\\debug\\app-debug-androidTest.apk"

       9 |     const inputApkPath = '~/somePath/build/outputs/apk/debug/app-debug.apk';
      10 |     const expectedTestApkPath = '~/somePath/build/outputs/apk/androidTest/debug/app-debug-androidTest.apk';
    > 11 |     expect(APKPath.getTestApkPath(inputApkPath)).toEqual(expectedTestApkPath);
      12 |   });
      13 | 
      14 |   it(`path for a gradle build flavor`, async () => {
      
      at Object.it (src/devices/android/APKPath.test.js:11:50)
```

I'm not sure what you want to do here, some options are:
- switch app and tests to consistently using [`require('path').posix`](https://nodejs.org/api/path.html#path_path_posix) - these paths normally work fine on windows (except for the executable, hence the `binPath` changes)
- have the tests `.replace(/\\\\/g, '/')`, and get rid of `~` usage.
- separate windows and posix path tests.
- input / expected paths based on platform
- leave it how it is for now, follow up on it later

I do have a Mac Mini for iOS dev now (it's absolutely terrible to work on, but it works 🤷), so I can be more confident I'm not breaking things, I guess.

rotemmiz:
Re tests, it sounds reasonable to just use Node's compatibility layer.
BTW, we intend to move our CI to a [home brewed Jenkins server](https://jenkins-oss.wixpress.com/) on top of MacStadium infrastructure next week, no Windows support is planned though.

simonbuchan:
> Re tests, it sounds reasonable to just use Node's compatibility layer.

Are you referring to `require('path').posix`? (It's not really a compatibility layer, but the implementation of `path` used by default on POSIX, there's a `.win32` also) - that would be pretty easy, but it's not really correct so it might break stuff. In particular it means the win32 absolute path `C:\foo\bar\baz\file.apk` gets treated as a relative path, and other bad stuff due to `\` not being a path separator, but that's solvable by replacing `\` with `/` on user input.

Another, cleaner answer would be to change the tests so they look something like:

```js
const inputApkPath = path.join(_dirname, 'build/outputs/apk/debug/app-debug.apk');
const expectedTestApkPath = path.join(_dirname, 'build/outputs/apk/androidTest/debug/app-debug-androidTest.apk');
```

but this means the actual diffs depend on the repo working directory path.

simonbuchan:
OK updated tests - should be safe and straightforward changes.

Were there specific things you wanted in terms of doc changes? At the moment, all I can think of is something like "windows should probably work, but be aware that `binaryPath` and `build` might not work cross-platform" - eg. `cd android && ./gradlew assembleDebug assembleAndroidTest` will *not* work on windows (it needs to be `gradlew`, no `./`) That doesn't seem enough for a whole page, where would you put that?

Other than that I think this is good to go modulo follow-up reviews.

simonbuchan:
While writing some docs noticed I hadn't gotten around to `detox-cli` (which I don't use myself, because `yarn/npx detox` works fine for me).

Let me know if these docs are OK, I'm not 100% on the tone or placement.

Hopefully that CI failure was just a flake? Not sure how I could have broken it.

simonbuchan:
Incidentally, while testing `build` hit the fact that react-native's `:ReactAndroid` gradle project doesn't seem to even configure on windows, meaning `detox/test` can't be built. 🙁

rotemmiz:
We have two flavors for the test app, fromBin and fromSource. The thing is, even if you build from bin, the fact that the settings.gradle includes `ReactAndroid` source, it will evaluate its build scripts.

You can remark it in `settings.gradle` and `build.gradle`, and build the `fromBin` flavour.

Tell me if it works.

simonbuchan:
Yeah. that seems to work fine, and the tests pass. (It was complaining about a missing `repositoryUrl` on the `:ReactAndroid` project, don't suppose that makes any sense to you?)
Meh, It's only a problem if you want Windows CI at some point.

detox/src/devices/Device.test.js
```diff
@@ -547,11 +548,11 @@ describe('Device', () => {
...
   it(`should accept absolute path for binary`, async () => {
     const actualPath = await launchAndTestBinaryPath('absolutePath');
-    expect(actualPath).toEqual('/tmp/abcdef/123');
+    expect(actualPath).toEqual(process.platform === 'win32' ? 'C:\\Temp\\abcdef\\123' : '/tmp/abcdef/123');
   });
 
   it(`should accept relative path for binary`, async () => {
     const actualPath = await launchAndTestBinaryPath('relativePath');
-    expect(actualPath).toEqual(`${process.cwd()}/abcdef/123`);
+    expect(actualPath).toEqual(path.join(process.cwd(), 'abcdef/123'));
```
rotemmiz: Shouldn't this be split to `path.join(process.cwd(), 'abcdef', '123')` ?

detox/src/devices/EmulatorDriver.js
```diff
@@ -52,10 +52,18 @@ class EmulatorDriver extends AndroidDriver {
     }
 
     await this._fixEmulatorConfigIniSkinName(name);
-    await this.emulator.boot(name);
```
rotemmiz: What was the issue here ? boot should return a valid device, either it's shutdown or already booted. If it doesn't we may have issue with `Emulator.js`

docs/More.Roadmap.md
```diff
@@ -11,6 +11,14 @@ The current API supports addition of multiple platforms, Android is next. This i
 ### iOS physical device support
 Currently detox only supports running on iOS simulators, we plan on adding support for running on devices as well.
 
+### Windows support
+There is some work done for running detox on Windows, but it's still fairly untested. Please open issues for anything you run into, but be aware of these limitations:
```
rotemmiz: Detox, not detox

docs/More.Roadmap.md
```diff
@@ -11,6 +11,14 @@ The current API supports addition of multiple platforms, Android is next. This i
...
+
+- Apple doesn't support iOS apps on Windows, obviously, so you're limited to the in-progress Android support.
+- `binaryPath` can be left as a relative path with `/`, or use `\\` if you don't need cross-platform support.
+- `build` should not use `./gradlew ...`, but simply `gradlew ...` - you may prefer scripting the build outside of
+  detox if you want to maintain cross-platform support - or simply have two configurations!
+
```
rotemmiz: 👌

detox/src/devices/EmulatorDriver.js
rotemmiz: I now see the comments [here](https://github.com/wix/detox/pull/628#commitcomment-28173251).

Not sure how to proceed here. This fix is in bad taste IMO. 
So the real issue is creating and removing those emulator log files ?

simonbuchan: I could experiment a bit, I suspect redirecting to different files should fix it, but this was a less user visible change (eg. .gitignore)
Do you not find it logical to only start an emulator if it's not already running?

detox/src/devices/Device.test.js
simonbuchan: Path will normalise separators, and windows in fine with either except for binary paths (where they are ambiguous with switches)

detox/src/devices/EmulatorDriver.js
simonbuchan: Nope, gave it a try and it still hangs on startup with no output to the log file even if it's a new file each time.
Killing the first emulator then will hang it, killing it causes the second to start up and detox to continue successfully.
Seems like there's something lower level specific to the android emulator going on, possibly needs to have a separate AVD directory?

docs/More.Roadmap.md
```diff
@@ -11,6 +11,14 @@ The current API supports addition of multiple platforms, Android is next. This i
 ### iOS physical device support
 Currently detox only supports running on iOS simulators, we plan on adding support for running on devices as well.
 
+### Windows support
+There is some work done for running Detox on Windows, but it's still fairly untested. Please open issues for anything you run into, but be aware of these limitations:
+
+- Apple doesn't support iOS apps on Windows, obviously, so you're limited to the in-progress Android support.
+- `binaryPath` can be left as a relative path with `/`, or use `\\` if you don't need cross-platform support.
+- `build` should not use `./gradlew ...`, but simply `gradlew ...` - you may prefer scripting the build outside of
+  detox if you want to maintain cross-platform support - or simply have two configurations!
```
LeoNatan: Why is there a line break here?

docs/More.Roadmap.md
```diff
@@ -11,6 +11,14 @@ The current API supports addition of multiple platforms, Android is next. This i
 ### iOS physical device support
 Currently detox only supports running on iOS simulators, we plan on adding support for running on devices as well.
 
+### Windows support
+There is some work done for running Detox on Windows, but it's still fairly untested. Please open issues for anything you run into, but be aware of these limitations:
+
+- Apple doesn't support iOS apps on Windows, obviously, so you're limited to the in-progress Android support.
```
LeoNatan: Let’s please remove the “obviously”.

docs/More.Roadmap.md
LeoNatan: Also, should be “Detox”.

simonbuchan:
Gah. 🤦‍♂️ Thanks for the reviews guys!
@rotemmiz If the emulator boot is an issue, I could only skip the second boot on win32? I suspect passing a temp dir to [`-datadir`](https://developer.android.com/studio/run/emulator-commandline#data-filedir) would also work, but that will very likely break people (but would be a neat option to know you're getting a fresh OS install each time / run in parallel?)

simonbuchan:
Hah, just looked the jenkins log and saw a big fat crash stack at the end... then noticed it was the app crash test 😅

simonbuchan:
🎉🕊 Thanks so much for all the discussion and feedback! This turned out a lot bigger and gnarlier than I expected at the start, but isn't that always the way!

LeoNatan:
Thank you for all the good work!

PR #631, opened 2018-03-20
fix: post install script failure on windows

eladav:
To allow developers to `npm i` a project that has detox as a dependency on Windows we changed the post install script from a bash script to js script and run using node.


LeoNatan:
Hi,

There is a parallel effort to make Detox work on Windows (#628) so I am closing this one.

Thanks

PR #636, opened 2018-03-22
Create workaround for nested scrollviews to be scrollable

raphaeleidus:
right now the `.scroll` method on ios automatically overrides the users element selection and will always break if they are trying to scroll a scrollview with a scrollview/webview inside it.

This PR introduces a new argument that allows this functionality to be disabled.

The reason this exists in the firstplace is to automatically fix a bad selector
```javascript
<View testID="notAScrollview">
  <Scrollview>
   ....
  </Scrollview>
</View>
......
element(by.id('notAScrollview')).scroll(500, 'down')
```
The above will just work today because detox will just find a child scrollview inside the element instead of making the dev change their selector to something like,
`by.type('RCTCustomScrollView').withAncestor(by.id('notAScrollview'))`

This PR will not break that functionality but if a user needs to disable it, they have the ability.
On android this functionality was commented out and didn't exist. This PR brings it back in but behind a flag that is set to false by default.


raphaeleidus:
This pull request provides a workaround for #164 but does not immediately solve the problem for those users, but it will unblock them by giving them the control to work around the issue.

akread:
Yes please! Right now we are hiding some web views to work around this issue and continue our E2E tests on a project.

raphaeleidus:
@rotemmiz can you review this?

moQuez:
this would be very helpful!

raphaeleidus:
PLEASE!!! #164 is still relevant and this still addresses the issue. PLEASE comment on this or something. It is not stale It just needs a project maintainer to review it. @rotemmiz is Detox no longer maintained?

LeoNatan:
This will be solved soon in a different PR.

LeoNatan:
See #896

PR #639, opened 2018-03-26
Automatically filter platform tests by inferring from config

rotemmiz:


rotemmiz:
@guyca, this is for you :)

PR #643, opened 2018-03-27
IOS: Set project requirement to Xcode 9 and above [WIP]

LeoNatan:
Remove legacy local and remote notification API
Remove legacy RN support


LeoNatan:
Tests are broken because the demo project uses legacy API. It will be removed, and I need to decide how to pass user notifications to the internal project.

LeoNatan:
Maybe I will fake them with "openURL", like I did with user activity.

LeoNatan:
React Native dropped support for Xcode 8: https://github.com/facebook/react-native/issues/18944
I want to merge this, Notifications be damned. Let the new owner finally move to the new API.

PR #652, opened 2018-03-28
Android: Support RN50-51 changes

rotemmiz:
This PR aims to:
1. Add support for changes made in [NativeAnimatedModule](https://github.com/facebook/react-native/commit/2b4ff6ea19dc674cf035ee419daa132cde8d1f5e#diff-a735317509988b7133cb51f6ea6cbd1f) 
2. Temp fix for `waitForReactNativeLoad` by adding Thread.sleep() on RN>=50, not ideal, but will enable usage of Detox on RN51 projects even before we find the true culprit of this sync defect.

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/ReactNativeCompat.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/ReactNativeCompat.java
@@ -1,0 +1,36 @@
+package com.wix.detox;
+
+import org.joor.Reflect;
+
+import java.util.Map;
+
+public class ReactNativeCompat {
+    static Map<String, Object> VERSION;
+
+    static {
+        try {
+            Class reactNativeVersion = Class.forName("com.facebook.react.modules.systeminfo.ReactNativeVersion");
+            VERSION = Reflect.on(reactNativeVersion).field("VERSION").get();
+        } catch (ClassNotFoundException e) {
+        }
+
+    }
+
+    public static int getMinor() {
+        return (Integer) VERSION.get("minor");
+    }
+
+    public static void waitForReactNativeLoad(Object reactNativeHostHolder) {
+        if (getMinor() >= 50) {
+            ReactNativeSupport.waitForReactNativeLoad(reactNativeHostHolder);
+            try {
+                //TODO- Temp hack to make Detox usable for RN>=50 till we find a better sync solution.
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        } else {
+            ReactNativeSupport.waitForReactNativeLoad(reactNativeHostHolder);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


detox/android/detox/src/main/java/com/wix/detox/ReactNativeCompat.java
```diff
@@ -0,0 +1,36 @@
...
+import org.joor.Reflect;
+
+import java.util.Map;
+
+public class ReactNativeCompat {
+    static Map<String, Object> VERSION;
+
+    static {
+        try {
+            Class reactNativeVersion = Class.forName("com.facebook.react.modules.systeminfo.ReactNativeVersion");
```
simonracz: This class is present only in RN 50 and above.

Later, the getMinor call could fail on these versions. As we don't differentiate on older versions, I think we could just set it to 49.5.0 in the catch as a default version.

detox/android/detox/src/main/java/com/wix/detox/espresso/AnimatedModuleIdlingResource.java
```diff
@@ -82,91 +83,102 @@ public boolean isIdleNow() {
...
 
+    private boolean isIdleRN51(Object animModuleClass) {
+        Object animModule = Reflect.on(reactContext).call(METHOD_GET_NATIVE_MODULE, animModuleClass).get();
+        Object nodesManager = Reflect.on(animModule).call("getNodesManager").get();
+        boolean hasActiveAnimations = Reflect.on(nodesManager).call("hasActiveAnimations").get();
+        if (!hasActiveAnimations) {
+            if (callback != null) {
+                callback.onTransitionToIdle();
+            }
+            Log.i(LOG_TAG, "AnimatedModule is idle, no operations");
```
simonracz: I think we shouldn't log out idle states. Only busy states.

Originally, I also logged both states, but after the initial debugging it become overwhelming.

rotemmiz: Agreed

simonracz: I looked at this [diff.](https://github.com/facebook/react-native/commit/2b4ff6ea19dc674cf035ee419daa132cde8d1f5e#diff-a735317509988b7133cb51f6ea6cbd1f)

I am not even sure that we need an animatedIdlingResource at all on later RN versions. It seems, they delegate everything to the UIModule. I don't fully understand the new flow in that code though.

detox/android/detox/src/main/java/com/wix/detox/ReactNativeCompat.java
rotemmiz: Good catch, thanks

detox/android/detox/src/main/java/com/wix/detox/espresso/AnimatedModuleIdlingResource.java
rotemmiz: Well, it was a trial and error kinda fix :)
But I think that there's a small gap between the time when the UIBlock is posted and when it starts executing.

rotemmiz:
Done deal

simonracz:
Awesome! 👍

Issue #251, opened 2017-08-27
Call supported EarlGrey and Espresso matcher APIs in runtime


DanielZlotin:
#### Description
Instead of manually adding any and all matcher APIs like [this](https://github.com/wix/detox/issues/246) request, maybe we can add functionality to detox API to call any supported matchers from JS in runtime.
For example, something like `expect(...).toMatch('earlgrey_isClickable')`.

isnifer:
@DanielZlotin already done here - https://github.com/wix/detox/pull/178

rotemmiz:
@isnifer, #178 only exposes an easy way to access `GREYActions` from JS. this is not exposed to the users, it's just a convenient way for Detox developers to add actions to the API (not the API itself).

What @DanielZlotin is asking is a way to use native **[matchers](https://github.com/wix/detox/blob/master/detox/src/ios/matchers.js)**, not **[actions](https://github.com/wix/detox/blob/master/detox/src/ios/expect.js#L46)**, I actually think this is a great idea. Your comment may confuse others.

isnifer:
@rotemmiz ok, removed my previous comment

LeoNatan:
@isnifer I have noticed that you are very active in our issues. While we appreciate your enthusiasm, please try to limit your comments only to cases where you have concrete solution to the issue. Too much chatter in the issue discussion makes it more difficult for us to identify the issue source and assist. Thanks for your understanding.

isnifer:
OK, no problem :)

On Sep 6, 2017 01:08, "Leo Natan" <notifications@github.com> wrote:

@isnifer <https://github.com/isnifer> I have noticed that you are very
active in our issues. While we appreciate your enthusiasm, please try to
limit your comments only to cases where you have concrete solution to the
issue. Too much chatter in the issue discussion makes it more difficult for
us to identify the issue source and assist. Thanks for your understanding.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/wix/detox/issues/251#issuecomment-327317788>, or mute
the thread

LeoNatan:
I think this is out of scope. Closing for now.

PR #668, opened 2018-04-11
expose EarlGrey's matcherForInteractable

kkhaidukov:
Will expose EarlGrey's matcherForInteractable in `by` for iOS.

### DIFF
--- a/detox/src/ios/expect.js
+++ b/detox/src/ios/expect.js
@@ -11,6 +11,7 @@ const ExistsMatcher = matchers.ExistsMatcher;
 const NotExistsMatcher = matchers.NotExistsMatcher;
 const TextMatcher = matchers.TextMatcher;
 const ValueMatcher = matchers.ValueMatcher;
+const InteractableMatcher = matchers.InteractableMatcher;
 const GreyActions = require('./earlgreyapi/GREYActions');
 
 let invocationManager;
@@ -381,7 +382,8 @@ const by = {
   type: (value) => new TypeMatcher(value),
   traits: (value) => new TraitsMatcher(value),
   value: (value) => new ValueMatcher(value),
-  text: (value) => new TextMatcher(value)
+  text: (value) => new TextMatcher(value),
+  interactable: () => new InteractableMatcher()
 };
 
 const exportGlobals = () => {
--- a/detox/src/ios/expect.test.js
+++ b/detox/src/ios/expect.test.js
@@ -44,6 +44,10 @@ describe('expect', async () => {
     await e.expect(e.element(e.by.traits(['startsMedia', 'adjustable', 'allowsDirectInteraction', 'pageTurn']))).toBeNotVisible();
   });
 
+  it(`element by interactable`, async () => {
+    await e.expect(e.element(e.by.interactable())).toExist();
+  });
+
   it(`matcher helpers`, async () => {
     await e.expect(e.element(e.by.id('test').withAncestor(e.by.id('ancestor')))).toBeVisible();
     await e.expect(e.element(e.by.id('test').withDescendant(e.by.id('descendant')))).toBeVisible();
--- a/detox/src/ios/matchers.js
+++ b/detox/src/ios/matchers.js
@@ -105,6 +105,13 @@ class ValueMatcher extends Matcher {
   }
 }
 
+class InteractableMatcher extends Matcher {
+  constructor() {
+    super();
+    this._call = invoke.callDirectly(GreyMatchers.matcherForInteractable());
+  }
+}
+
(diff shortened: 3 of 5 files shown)


kkhaidukov:
@rotemmiz I need help with this failed build here -- https://travis-ci.org/wix/detox/jobs/367176057
locally I'm not having this failure

rotemmiz:
Travis will be Travis 😒, restarted, green

kkhaidukov:
@rotemmiz the e2e test is added, please have another look

rotemmiz:
I am trying to understand the usecase for using this API.
I'm trying to understand the test you added [here](https://github.com/wix/detox/pull/668/files#diff-3f98db3bae3363a915d7d36d0436bba2R71). If you're just trying to test that the view can be interacted upon, why not use expect to be visible ? it may be even more conservative since it has 75% visibility threshold.
BTW, EarlGrey validates on most actions if the view is intractable before performing, tap included, so the test you wrote seems redundant.

kkhaidukov:
@rotemmiz We have an app with tabbed interface, and two (or more) tabs may have elements with the same testIDs, because when the tab is switched, a new screen is pushed onto the stack, overlaying the one beneath it (which is not destroyed and its elements are therefore locatable). In this case, detox cannot distinguish between the two elements when trying to e.g. tap.

Adding the `interactable` matcher solves this problem -- if I remove `.and(by.interactable())` from my test, it will fail because multiple elements are matched.

rotemmiz:
What prevents you from adding testID to each tab ?

rotemmiz:
`interactable` means that this view has 10 at least visible pixels. does this solve your issue ? how ?

BTW, I think [this proposition](https://github.com/wix/detox/issues/251) will solve your problem and much more.

kkhaidukov:
> What prevents you from adding testID to each tab ?
would that help? `withAncestor` still matches only immediate parents, or am I doing something wrong

> does this solve your issue ? how ?
it does, it ensures that only the elements belonging to the uppermost screen will be matched

> BTW, I think this proposition will solve your problem and much more.
it might, but it doesn't look like it's getting any action

rotemmiz:
So is everything mounted to the screen (even things which are not in the active tab?)
Do you use react native navigation?

kkhaidukov:
> So is everything mounted to the screen (even things which are not in the active tab?)

yes

> Do you use react native navigation?

if you mean `wix/react-native-navigation`, then no, we don't

rotemmiz:
In sake of your app's performance, try detaching those views from the hierarchy. This will also solve your problem.

kkhaidukov:
@rotemmiz any other changes needed to get this PR merged?

rotemmiz:
Hey, sorry for the delayed response here, but I don't think this PR is a good idea, along with what I wrote previously. 
A solution that can solve your problems (and was already kinda [implemented on Android](https://github.com/wix/detox/commit/b9a832f665606f130a37c7d924c6cfc5ef014e41)), is making sure non-interactable views won't be matched. 

A PR of this type makes much more sense IMO.

rotemmiz:
Detox is being used in React Native Navigation v2, and we had the same issue of multiple views with the same testID on multiple tabs on its  Android playground app (the app we test RNNv2 on).
Adding the same functionality on iOS matchers should be added here: https://github.com/wix/detox/blob/master/detox/ios/Detox/GREYMatchers+Detox.m

You'll probably need to add `matcherForInteractable` to `grey_allOf()` on all matchers.

kkhaidukov:
Currently I'm trying to switch IdMatcher from calling GreyMatchers.matcherForAccessibilityID to a method that I created in GREYMatchers+Detox.m that looks like this:
```
+ (id<GREYMatcher>)detoxMatcherForAccessibilityID:(NSString *)accessibilityID {
    if (![ReactNativeSupport isReactNativeApp])
    {
        return  [self matcherForAccessibilityID:accessibilityID];
    }
    else
    {
        Class RCTTextViewClass = NSClassFromString(@"RCTText") ?: NSClassFromString(@"RCTTextView");
        return grey_allOf(grey_sufficientlyVisible(),
                          grey_accessibilityID(accessibilityID),
                          grey_not(grey_descendant(grey_allOf(grey_kindOfClass(RCTTextViewClass),
                                                              grey_accessibilityID(accessibilityID),
                                                              nil))),
                          nil);
    }    
}
```
But the problem is, the tests get very noticeably slower, when there is an extra grey_ matcher inside of grey_allOf. I've tried with `grey_sufficientlyVisible` and `grey_interactable`.
Is it the expected behavior or am I just doing it wrong?
@rotemmiz @LeoNatan

PR #675, opened 2018-04-18
[Android] Enable usage of custom instrumentation test runners

wiyarmir:
- Spawn adb shell and ask package manager for all runners
- Regex the one for the package under test and use it


wiyarmir:
closing and reopening to trigger another build

detox/src/devices/AndroidDriver.js
```diff
@@ -72,6 +79,8 @@ class AndroidDriver extends DeviceDriverBase {
       return this.instrumentationProcess.pid;
     }
 
+    const testRunner = await this.adb.getInstrumentationRunner(deviceId, bundleId)
```
rotemmiz: ;

rotemmiz: ; missing

detox/src/devices/AndroidDriver.test.js
```diff
@@ -0,0 +1,37 @@
+describe("AndroidDriver", () => {
```
rotemmiz: Drivers are not unit tested on purpose. The mostly touch command line tools and are just wrappers. Unit tests don't give much value. They are being tested in the e2e suite.

detox/src/devices/android/ADB.test.js
```diff
@@ -55,5 +55,10 @@ xdescribe('ADB', () => {
     await adb.unlockScreen('deviceId');
     expect(exec).toHaveBeenCalledTimes(1);
   });
+
```
rotemmiz: I guess we should consider dropping unit tests on ADB as well, but for now, don't delete this test

detox/src/devices/AndroidDriver.test.js
wiyarmir: This function is doing some regex voodoo so that's why I thought about adding one at least, but I'm ok with removing it.

wiyarmir:
Took the freedom of fixing ADB tests a bit. Probably `Parse 'adb device' output` and `getInstrumentationRunner` are the only ones adding value at the moment, but I let them be. Have a look @rotemmiz

PR #679, opened 2018-04-20
Add check for ungenerated code

DanielMSchmidt:
We had at times updates of the source for code generation not being reflected in the adapters as the code was not generated again. Let's add a check for this here 👍

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,8 @@
+{
+	"useTabs": false,
+	"tabWidth": 2,
+	"printWidth": 140,
+	"singleQuote": true,
+	"arrowParens": "always",
+	"trailingComma": "none"
+}
--- a/detox/src/ios/earlgreyapi/GREYActions.js
+++ b/detox/src/ios/earlgreyapi/GREYActions.js
@@ -7,36 +7,32 @@
 
 function sanitize_greyDirection(action) {
 	switch (action) {
-		case "left":
+		case 'left':
 			return 1;
-		case "right":
+		case 'right':
 			return 2;
-		case "up":
+		case 'up':
 			return 3;
-		case "down":
+		case 'down':
 			return 4;
 
 		default:
-			throw new Error(
-				`GREYAction.GREYDirection must be a 'left'/'right'/'up'/'down', got ${action}`
-			);
+			throw new Error(`GREYAction.GREYDirection must be a 'left'/'right'/'up'/'down', got ${action}`);
 	}
 } 
 function sanitize_greyContentEdge(action) {
 	switch (action) {
-		case "left":
+		case 'left':
 			return 0;
-		case "right":
+		case 'right':
 			return 1;
-		case "top":
+		case 'top':
 			return 2;
-		case "bottom":
+		case 'bottom':
 			return 3;
 
 		default:
-			throw new Error(
-				`GREYAction.GREYContentEdge must be a 'left'/'right'/'top'/'bottom', got ${action}`
-			);
+			throw new Error(`GREYAction.GREYContentEdge must be a 'left'/'right'/'top'/'bottom', got ${action}`);
 	}
 } 
 class GREYActions {
--- a/generation/check-for-git-changes.sh
+++ b/generation/check-for-git-changes.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+git stash
+npm run build
+
+if [[ $TRAVIS && $(git diff) ]]; then
+    printf "\n\n"
+    echo "There seem to be changes after running the code generation."
+    echo "This might be because you updated a native dependency and forgot to run the code generation."
+    echo "If this is the case run 'cd generation && npm run build'"
+    printf "\n"
+    echo "Another source of this error might be that you have changed generated code and checked it in."
+    echo "You should not be doing this, please change the source of the generated code and not the code itself."
+    printf "\n"
+    echo "$(git diff)"
+    git stash pop
+    exit 1
+else
+    git stash pop
+fi
--- a/detox/src/android/espressoapi/DetoxAction.js
+++ b/detox/src/android/espressoapi/DetoxAction.js
@@ -7,34 +7,30 @@
 
 function sanitize_android_edge(edge) {
 	switch (edge) {
-		case "left":
+		case 'left':
 			return 1;
-		case "right":
+		case 'right':
 			return 2;
-		case "top":
+		case 'top':
 			return 3;
-		case "bottom":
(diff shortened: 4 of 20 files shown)


generation/check-for-git-changes.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+if [[ $(git diff) ]]; then
+    echo "Git shows that there are changes."
```
rotemmiz: Please be more explicit in these logs. 
1. What happened ?
2. Why did it happen ?
3. What should the user do now in order to resolve it ?

DanielMSchmidt:
@rotemmiz added error message, could you review it again?

DanielMSchmidt:
@rotemmiz I understood that we wanted to use prettier in our project, so I thought adding it to the generated code too makes sense. I think we are doing a suboptimal job enforcing it, I would offer to help with that. The [mockserver](https://github.com/mesosphere/mockserver) I am currently writing enforces it with the help of jest, which I like very much, you can see [the config here](https://github.com/mesosphere/mockserver/blob/4e5c7659d6279f0949f73dcd16918ff6b3447c0d/package.json#L64-L74).

I found this quote on [a podcast I listen to](https://softskills.audio/2018/03/31/episode-103-team-dynamics-and-bad-code/) (freely quoted as I don't remember when they said it):

> Prettier reformats your code to look worse, but at the same time it makes your co-workers code much more readable.

This resonated a lot with me, I think everyone has an opinion on how code should be looking like for whatever (totally) reasons. prettier frees us from having to have conversations about how it should look like, it just always looks the same, freeing up brain cycles for more important tasks 😇

rotemmiz:
You're right but I still really don't like these formatting choices. It makes the code longer (in lines), and very weirdly formatted.
I would want to explore it there are different "flavors" tho this formatting before running it on the whole project. I do like ESLint's defaults much better.
Sorry for delaying this for too long, but these formatting chocices made by prettier really hurt my eyes.

DanielMSchmidt:
Prettier has [a nice playground](https://prettier.io/playground/) and at the bottom left corner there is a "Show Options" button where you can play with the options until you like the output ;)

detox/src/ios/earlgreyapi/GREYInteraction.js
```diff
@@ -95,9 +136,20 @@ performAction:grey_tap()] // This should be separately called for the action.
 @param matcher The matcher to be evaluated on the @c element.
 
 @return The provided GREYInteraction instance with a matcher to be evaluated on an element.
-*/static assertWithMatcher(element, matcher) {
-    if (typeof matcher !== "object" || matcher.type !== "Invocation" || typeof matcher.value !== "object" || typeof matcher.value.target !== "object" || matcher.value.target.value !== "GREYMatchers") {
-      throw new Error('matcher should be a GREYMatcher, but got ' + JSON.stringify(matcher));
+*/ static assertWithMatcher(
```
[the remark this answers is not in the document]
rotemmiz: Hey @DanielMSchmidt , can you please take a look at this?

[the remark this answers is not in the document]
DanielMSchmidt: I'm on it, thank you for pinging 👍

DanielMSchmidt:
@rotemmiz I think we are there now, please take another look 🙏 

Might make sense to go over it commit by commit as the changeset is quite big due to prettier being unleashed on the generated code

rotemmiz:
I hate this argument as well, but please use spaces. Tabs make it much harder to read on GitHub, and we spend a lot of time reviewing here

detox/src/ios/earlgreyapi/GREYActions.js
```diff
@@ -4,92 +4,95 @@
...
   }
 
   /*@return A GREYAction that performs multiple taps of a specified @c count at a specified
 @c point.
-*/static actionForMultipleTapsWithCountAtPoint(count, point) {
-    if (typeof count !== "number") throw new Error("count should be a number, but got " + (count + (" (" + (typeof count + ")"))));
-    if (typeof point !== "object") throw new Error("point should be a object, but got " + (point + (" (" + (typeof point + ")"))));
-    if (typeof point.x !== "number") throw new Error("point.x should be a number, but got " + (point.x + (" (" + (typeof point.x + ")"))));
-    if (typeof point.y !== "number") throw new Error("point.y should be a number, but got " + (point.y + (" (" + (typeof point.y + ")"))));
+*/ static actionForMultipleTapsWithCountAtPoint(
```
rotemmiz: still here!
separate `*/ static actionForMultipleTapsWithCountAtPoint(` to 
```
*/ 
static actionForMultipleTapsWithCountAtPoint(
```

DanielMSchmidt: Can't do much about it, this would be something we would need to do in prettier itself, providing an upstream PR. Do you really want me to spend the time there? Would be fine with that, just wanted to clarify.

rotemmiz: It's not related to prettier, it's the code generator

DanielMSchmidt:
Unfortunately, I can't do much about it. I add a `leadingComments` entry for the comment in babel, which means "insert the comment before this node", leading to the output you see. So my options are either 

1) Go into babel and "fix" the `leadingComments` behaviour
2) Go into prettier and enable it to prettify this situation
3) Go with a hack solution (String replacement of the final output)
4) Let it be like this

rotemmiz:
4

DanielMSchmidt:
Okay :)

DanielMSchmidt:
@rotemmiz Could you give a final review here? 😇

DanielMSchmidt:
@rotemmiz I removed prettier from the generated code (only applied the new style to the code generation). I opened an issue at prettier/prettier#4465 let's see if we can get any help there. I might investigate if there is a quick fix within prettier that doesn't take too long.

.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+  "useTabs": true,
```
rotemmiz: useTabs:false!!

DanielMSchmidt: Sorry, forgot to squash one of the commits I dropped

rotemmiz:
@DanielMSchmidt can you please take a look, seems like the build is broken due to this change.
https://travis-ci.org/wix/detox/jobs/378118106#L657

DanielMSchmidt:
Yay, the check works. But damn, need to fix them 😂 will do that asap and ping you :)

PR #681, opened 2018-04-20
[Android] Ensure main thread when doing getInstanceManager

wiyarmir:
On brownfield apps, it could happen that the first Activity you launch (e.g. `MainActivity`) is not a React Native one, so the `ReactInstanceManager` has not been initialised yet. 

In that case, React Native will instantiate it, creating instances for all its dependencies as well. Unfortunately, at least one of such dependencies (`ReactChoreographer`) requires main thread ( https://github.com/facebook/react-native/blob/v0.55.3/ReactAndroid/src/main/java/com/facebook/react/modules/core/ReactChoreographer.java#L68 ) and because Detox runs in instrumentation, which is a different thread, this will crash.

This PR checks wether we are on main thread before attempting to reflect and changes to it if not.


rotemmiz:
good catch, but this code needs rewrite ASAP :(

Issue #594, opened 2018-02-28
Running Android tests on Linux


chrmod:
#### Description

Running Detox on Linux fails before emulator is able to start and install the app. 

#### Steps to Reproduce

Try to start `detox test` for any of example projects on a Linux machine. We have problems doing that and have to update `demo-react-native` to do that https://github.com/wix/detox/pull/593

On MacOS the build and tests works correctly and all tests are green.

On Linux, same code base, start the emulator, and runs the tests, but tests start to fast and fail. 
(When emulator finally finishes the initialization, the apk is not installed)

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 7.1.0
* Node: 8.9.4
* Device: Android Emulator 22/25/26  x86/arm 
* Linux: 4.13

#### Device and verbose Detox logs

```
node_modules/.bin/jest e2e --config=e2e/config.json --runInBand
 server listening on localhost:33293...
 onOpen [object Object]
 send: {"type":"login","params":{"sessionId":"f01bf5bd-4da5-61d2-39ff-27239855f5ef","r
ole":"tester"},"messageId":0}
 onMessage: {"type":"loginSuccess","params":{"sessionId":"f01bf5bd-4da5-61d2-39ff-2723
9855f5ef","role":"tester"},"messageId":0}
  rbx
 : /home/chrmod/Android/Sdk/tools/emulator -list-avds --verbose
 Nexus_5X_API_22
 Nexus_5X_API_24
 Nexus_5X_API_26
 Pixel_XL_API_26
 : stdout:
 -verbose -gpu host -no-audio @Nexus_5X_API_22
 : /home/chrmod/Android/Sdk/platform-tools/adb  devices
 List of devices attached
 emulator-5554  device
 : stdout:
 : stdout:
 FAIL  e2e/firstTest.spec.js
  Example
    ✕ should have urlbar (3ms)
  ● Example › should have urlbar
    response not received
      at setTimeout (../node_modules/telnet-client/lib/index.js:150:47)
      at Timeout.callback [as _onTimeout] (../node_modules/jsdom/lib/jsdom/browser/Win
dow.js:633:19)
```

chrmod:
PS. are you aware of any project tested of Detox running on Travis CI? I would assume that if it runs on Travis it will work on most Linux boxes. 
I'm willing to help with updating example projects accordingly and Travis CI setup.

rotemmiz:
[...]

PR #687, opened 2018-04-24
Android bugfix: support unauthenticated adb connection

chrmod:
In case `.emulator_console_auth_token` is missing or empty, `auth` command fails silently, but its error (or rather info of incorrect usage) polluts the output of subsequent command (in this case https://github.com/wix/detox/blob/master/detox/src/devices/EmulatorDriver.js#L57 - more specifically https://github.com/wix/detox/blob/master/detox/src/devices/android/ADB.js#L30).

This results in unexpected values being returned by `exec` (https://github.com/wix/detox/blob/master/detox/src/devices/android/EmulatorTelnet.js#L28).

In such situation detox tests start running before the device is booted, I've reported this problem before in  https://github.com/wix/detox/issues/594

### DIFF
--- a/detox/src/devices/android/EmulatorTelnet.js
+++ b/detox/src/devices/android/EmulatorTelnet.js
@@ -22,7 +22,9 @@ class EmulatorTelnet {
 
     await this.connection.connect(params);
     const auth = await fs.readFile(path.join(os.homedir(), '.emulator_console_auth_token'), 'utf8');
-    await this.exec(`auth ${auth}`);
+    if (auth) {
+      await this.exec(`auth ${auth}`);
+    }
   }
 
   async exec(command) {


rotemmiz:
In which cases emulator_console_auth_token does not exist?

chrmod:
@rotemmiz It does no exist on my linux system. Don't have enough Android knowledge to tell when that file should be created. My current assumption is that it may not exist on linux at all. 

BTW. I've tried to create an empty file, but it result with exactly same problem - empty value make `auth` not recognize the second argument and "throws" error message.

noomorph:
This is a quite curious case, and I'm about to rewrite parts in AndroidEmulator class.
I also don't know where does this file come from, so I posted a question to StackOverflow:

https://stackoverflow.com/questions/50061003/where-does-emulator-console-auth-token-come-from

rotemmiz:
https://developer.android.com/studio/run/emulator-console

> After the console displays OK, enter the auth auth_token command.
> 
> Before you can enter console commands, the emulator console requires authentication. auth_token must match the contents of the .emulator_console_auth_token file in your home directory.
> 
> If that file doesn't exist, the telnet localhost console-port command creates the file, which contains a randomly generated authentication token. To disable authentication, delete the token from the .emulator_console_auth_token file or create an empty file if it doesn't exist.


The file should be there right after you log in using telnet. I suspect you have a very old version of `emulator` binary. Please check it.

noomorph:
@rotemmiz , thanks for finding this piece of information! Oddly, I did not notice it while googling.

Anyway, I get an impression that the PR's way to fix the issue is not sound.

It is indeed not nice if Detox did not provide @chrmod with the error information and failed silently. If that is the case, we ought to improve logging in `exec()` method.

Secondly, some people believe that this is insecure, and they might be right. See the comment to https://stackoverflow.com/a/46834684/3472184:

screen shot 2018-04-27 at 14 09 03

If so, then insecure behavior is not something we want to turn on automatically on a failure condition, unbeknownst to the user. If we can avoid this, then we should avoid. If there are cases where we cannot avoid disabling security, then it should be done with an explicit flag, similarly to `--disable-web-security` in Chrome.

If older versions of Android SDK indeed do not create this file, then the PR should include the appropriate check for SDK version (before it skips) instead of unconditional skipping.

If the problem lies in file permissions for @chrmod 's home folder, then we might include this hint to troubleshooting section or to the error output.

In any case, I can confirm that Detox + Android works on my local workstation (Ubuntu 17.10), so the issue https://github.com/wix/detox/issues/594 is specific to configuration, project or environment.

PR #690, opened 2018-04-25
Add no window option for running android emulator with no window

yershalom:
Tested locally and passed

### DIFF
--- a/detox/local-cli/detox-test.js
+++ b/detox/local-cli/detox-test.js
@@ -34,7 +34,7 @@ program
     + 'e.g test with substring \':ios:\' in its name will not run when passing \'--platform android\'')
   .option('-f, --file [path]',
     'Specify test file to run')
-  .option('-n, --noWindow',
+  .option('-H, --headless',
     '[Android Only] Launch Emulator in headless mode. Useful when running on CI.')
   .parse(process.argv);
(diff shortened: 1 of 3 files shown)


detox/local-cli/detox-test.js
```diff
@@ -34,6 +34,8 @@ program
     + 'e.g test with substring \':ios:\' in its name will not run when passing \'--platform android\'')
   .option('-f, --file [path]',
     'Specify test file to run')
+  .option('-W, --nowindow',
+    'Run Android without window')
```
rotemmiz: 1.Be as expressive as you can, this is documentation. Text suggestion: "[Android Only] Launch Emulator in headless mode. Useful when running on CI."
2. Update docs (cli section)
3. Why capital W? 
4. Let's use same flag as emulator --no-window.
 WDYT?

yershalom: 1. Ok, fixed.
2. Done.
3. Right, changed this to -n
4. --no-window is problematic for how we are doing the flags parsing (as you said). I changed it to --noWindow

rotemmiz:
This is inconsistent with other methods. Look at debug-synchronization, it can still be parsed correctly to debugSynchronization

PR #693, opened 2018-04-25
Move GREYConfiguration to generated code

DanielMSchmidt:
### Goal

The goal of this PR is to move the self-written wrapper around GREYConfiguration over to generated code. This has the side-effect that I need to allow id as a type, which enables a lot of different APIs we haven't had before (which are relatively unguarded from a generation perspective, but that's due their nature).

### Commits

1. Generate the APIs needed and translate GREYCondition code
2. "Inline" GREYCondition code, so that it directly uses the generated files & remove the hand-written GREYCondition file
3. Rename GREYConditionApi to GREYCondition


DanielMSchmidt:
@rotemmiz This refactoring is ready now 👍

PR #699, opened 2018-04-29
Update java-method-parser to AST-based version

DanielMSchmidt:
This version uses an AST instead of regex-based parsing.

## How to review

- checkout the branch and go to `generation`
- run `npm install && npm run build`
- read through the [PR in `java-method-parser` which introduces the AST-based approach](https://github.com/DanielMSchmidt/java-method-parser/pull/45)

## Trade-Offs

- You will experience a longer generation time (~ 3-5 minutes instead of a few seconds). We would need to decide if the possibly increased stability is worth it or not :) 

## ToDos (after review)

- [ ] Publish official major version if we want to go with the AST-based approach. If this gets approved I will do the bump and update the version accordingly before merging.

### DIFF
--- a/generation/package.json
+++ b/generation/package.json
@@ -13,7 +13,10 @@
 	"author": "DanielMSchmidt <danielmschmidt92@gmail.com>",
 	"license": "MIT",
 	"lint-staged": {
-		"*.{js,json,css,md}": ["prettier --write", "git add"]
+		"*.{js,json,css,md}": [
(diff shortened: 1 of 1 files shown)


DanielMSchmidt:
@rotemmiz I added some review instructions for you 👍

DanielMSchmidt:
@rotemmiz and I decided to not go forward with this as the performance is not the best currently :)

PR #700, opened 2018-04-29
Update Android emulator path to support SDK Tools >=25.3.0

elyalvarado:
Since Android SDK Tools 25.3.0 (March 2017) The Android Emulator was removed from the SDK Tools package and moved to a different SDK directory: [Android Emulator v15.3.0 release notes](https://developer.android.com/studio/releases/emulator#25-3)

Also, according to [this issue in the google issue tracker](https://issuetracker.google.com/issues/66886035) the new path is also the preferred way of invoking the emulator, and the old path will at some point be deprecated / removed.

For recent API versions (and depending on the OS architecture) running the old tools/emulator can cause errors because the different architecture versions of the emulator won't be found in the tools folder of the Android SDK. By example, I'm using the Android SDK v27 build tools and the latest versions of all the Android tools, and when running the emulator with the command that detox uses I get the following errors:

```
usr/local/share/android-sdk/tools/emulator -verbose -gpu host -no-audio @Nexus_5_API_27
emulator:Android emulator version 26.0.3.0 (build_id 3965150)
emulator:Found AVD name 'Nexus_5_API_27'
emulator:Found AVD target architecture: x86
emulator:argv[0]: '/usr/local/share/android-sdk/tools/emulator'; program directory: '/usr/local/share/android-sdk/tools'
emulator:  Found directory: /usr/local/share/android-sdk/system-images/android-27/google_apis/x86/

emulator:Probing for /usr/local/share/android-sdk/system-images/android-27/google_apis/x86//kernel-ranchu: file missing
emulator:Auto-config: -engine classic (based on configuration)
emulator:  Found directory: /usr/local/share/android-sdk/system-images/android-27/google_apis/x86/

emulator:try dir /usr/local/share/android-sdk/tools
emulator:Looking for emulator-x86 to emulate 'x86' CPU
emulator:Probing program: /usr/local/share/android-sdk/tools/emulator64-x86
emulator:Probing program: /usr/local/share/android-sdk/tools/emulator-x86
PANIC: Missing emulator engine program for 'x86' CPU.
```

Given that from August this year the minimum target version required for submitting apps to the Play store will be API 26, the path of the emulator should be updated.

### DIFF
--- a/detox/src/devices/android/Emulator.js
+++ b/detox/src/devices/android/Emulator.js
@@ -10,7 +10,9 @@ const argparse = require('../../utils/argparse');
 
 class Emulator {
   constructor() {
-    this.emulatorBin = path.join(Environment.getAndroidSDKPath(), 'emulator', 'emulator');
+    const newEmulatorPath = path.join(Environment.getAndroidSDKPath(), 'emulator', 'emulator');
+    const oldEmulatorPath = path.join(Environment.getAndroidSDKPath(), 'tools', 'emulator');
+    this.emulatorBin = fs.existsSync(newEmulatorPath) ? newEmulatorPath : oldEmulatorPath;
   }
 
   async listAvds() {
--- a/detox/src/devices/android/Emulator.js
+++ b/detox/src/devices/android/Emulator.js
@@ -10,7 +10,7 @@ const argparse = require('../../utils/argparse');
 
 class Emulator {
   constructor() {
-    this.emulatorBin = path.join(Environment.getAndroidSDKPath(), 'tools', 'emulator');
+    this.emulatorBin = path.join(Environment.getAndroidSDKPath(), 'emulator', 'emulator');
   }
 
   async listAvds() {


noomorph:
I think it would be better to support both paths in order of preference than just drastically switch to another location.

Anyway your observation is very nice. To my embarrassment, I also had this problem, but didn’t figure out the reason behind it, so I’ve been fiddling with folders and files in my SDK till it worked. Now I know, thanks!

elyalvarado:
@noomorph you're right. I've updated the pull request to check first if the new emulator path exists before using it, and otherwise falling back to the previous path.

noomorph:
@rotemmiz , actually this is a useful pull request 👍  . Today I was lazy to get AVD with API 26 on the other MacBook, so I tried to run the project on Nexus API 27 and only with the change in PR it works.

What do you think? Maybe we can fix a couple of things in PR if there's something you don't like and merge? Why not?

rotemmiz:
Sure! Lets do it!

PR #701, opened 2018-04-30
Migrate from Travis CI to Jenkins

yershalom:


scripts/demo-projects.ios.sh
```diff
@@ -2,15 +2,25 @@
 
 source $(dirname "$0")/travis_logger.sh
 
-run_f "lerna bootstrap"
+if [ ! -z ${REACT_NATIVE_VERSION} ]; then
+  node scripts/change_react_native_version.js "examples/demo-react-native" ${REACT_NATIVE_VERSION}
+fi
+
+lerna bootstrap
```
rotemmiz: No need to bootstrap twice, let's just change RN version on both and run bootstrap once.

No need for "runf" anymore?

yershalom: 1. ok.
2. I'll keep the "run_f" until we'll migrate from travis

PR #702, opened 2018-04-30
Emulator wrapper fixes 

noomorph:
1. fix: emulator should always get non-empty command line arguments (regression from https://github.com/wix/detox/pull/690)
2. fix: detox fails to start when another emulator is running
3. feat: logging emulator output on errors

Regarding (2), there was an issue (likely on Linux, but not sure if it is the reason) with a race condition.

`tail.on("line", function(data) {` was called too late, when the promise has already rejected.
 Since the function was returning `return promise;` without catch (e.g. `return promise.catch(`), the `promise._cpResolve` call was happening too late when the consumer of `boot` logic went into rejection conditional branch.

### DIFF
--- a/detox/src/devices/android/Emulator.js
+++ b/detox/src/devices/android/Emulator.js
@@ -24,8 +24,15 @@ class Emulator {
   }
 
   async boot(emulatorName) {
-    const headless = argparse.getArgValue('headless') ? '-no-window' : '';
-    const cmd = `-verbose -gpu host -no-audio ${headless} @${emulatorName}`;
+    const cmd = _.compact([
+        '-verbose',
+        '-gpu',
+        'host',
+        '-no-audio',
+        argparse.getArgValue('headless') ? '-no-window' : '',
+        `@${emulatorName}`
+    ]).join(' ');
+
     log.verbose(this.emulatorBin, cmd);
     const tempLog = `./${emulatorName}.log`;
     const stdout = fs.openSync(tempLog, 'a');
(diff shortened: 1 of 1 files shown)


detox/src/devices/android/Emulator.js
```diff
@@ -24,8 +24,15 @@ class Emulator {
   }
 
   async boot(emulatorName) {
-    const headless = argparse.getArgValue('headless') ? '-no-window' : '';
-    const cmd = `-verbose -gpu host -no-audio ${headless} @${emulatorName}`;
+    const cmd = _.compact([
+        '-verbose',
+        '-gpu',
+        'host',
```
rotemmiz: Change to `gpu auto` for better support. Current command breaks on some machines

noomorph:
@rotemmiz , done.

noomorph:
@rotemmiz , the build is green.

detox/src/devices/android/Emulator.js
```diff
@@ -24,45 +24,55 @@ class Emulator {
...
-      promise._cpResolve();
+      fs.unlink(tempLog, _.noop);
     }
 
-    return promise;
+    log.verbose(this.emulatorBin, ...emulatorArgs);
+    const childProcessPromise = spawn(this.emulatorBin, emulatorArgs, { detached: true, stdio: ['ignore', stdout, stderr] });
+    childProcessPromise.childProcess.unref();
+
+    return childProcessPromise.catch((err) => {
```
rotemmiz: can't this be separated into two declarations ?

```js
childProcessPromise.catch((err) => {
}
....
return childProcessPromise;
```

noomorph: But this is completely different logic then. What do you mean?
screenshot from 2018-05-02 15-14-50

PR #711, opened 2018-05-02
Enable linting on CI

DanielMSchmidt:
Last try (#517) I decided to fix the "errors", but this time I choose to just change the eslint config to just match the way we write code. Once this is merged we can decide if we want to re-enable easy wins (`eqeqeq`, `no-lonely-if`, etc), but we can do that one by one, which should lead to a lot smaller PRs 👍

PR #722, opened 2018-05-10
Bump requirements to Node 8.x and activate minimal eslint configuration

noomorph:
Continuation of https://github.com/wix/detox/pull/711 ...

/cc @DanielMSchmidt


rotemmiz:
We should be more explicit in the `engines` section. spread operator works from `node@8.3.0`, not 8.0.0.
ON the other hand, the docs should state lower than 8.11, 8.3.0 is enough.

rotemmiz:
Unless there's something else I'm missng

Issue #627, opened 2018-03-18
scripts/postinstall.sh not found


NazarYablonskiy:
Hi,
I'm getting following
(CI, Linux, detox@7.2.0)

```
...
error An unexpected error occurred: "/builds/xxx/xxx/xxx/mobile/node_modules/detox: Command failed.
  Exit code: 127
  Command: sh
  Arguments: -c scripts/postinstall.sh
  Directory: /builds/xxx/xxx/xxx/mobile/node_modules/detox
  Output:
  sh: scripts/postinstall.sh: not found".
...
```

Let me know if you need more details.
Thanks!

rotemmiz:
Where is this output from ? 
Please be more descriptive in the first place, and follow the template.

Stanok3694:
Hi. I had the same issue now, when I tried to install detox into my RN 0.50.3 project...

PS C:\Aigh\hb-aigh-main-fe> npm i detox --save-dev

> detox@7.3.4 postinstall C:\Aigh\hb-aigh-main-fe\node_modules\detox
> scripts/postinstall.sh

and after this I have got following error:
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! detox@7.3.4 postinstall: `scripts/postinstall.sh`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the detox@7.3.4 postinstall script.

What does it mean?

rotemmiz:
It means that Detox doesn't run on Windows currently.

LeoNatan:
@jhen0409 Could you please open a new issue? Also, would appreciate a PR :-)

PR #723, opened 2018-05-11
Fix postinstall script of detox on `sh`

jhen0409:
Related to https://github.com/wix/detox/issues/627#issuecomment-388379155.

We can reproduce the issue by the following script:

```bash
$ sh -c "touch script.sh && chmod +x script.sh && script.sh"
# sh: script.sh: command not found
```

### DIFF
--- a/detox/scripts/postinstall.sh
+++ b/detox/scripts/postinstall.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/sh -e
 
 if [ `uname` == "Darwin" ]; then
   source "$(dirname ${0})/build_framework.ios.sh"
--- a/detox/package.json
+++ b/detox/package.json
@@ -27,7 +27,7 @@
     "test": "npm run unit",
     "unit:watch": "jest --watch",
     "prepublish": "npm run build",
-    "postinstall": "scripts/postinstall.sh"
+    "postinstall": "./scripts/postinstall.sh"
   },
   "devDependencies": {
     "eslint": "^4.11.0",


detox/scripts/postinstall.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/sh
```
rotemmiz: hey @jhen0409 
Is there a reason you removed the `-e` flag ?

jhen0409: It may unnecessary for the current case, but I'll add back it to be consistent. 👍

detox/scripts/postinstall.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/sh -e
```
LeoNatan: Why was this changed to `sh` from `bash`? Not saying  bash is absolutely needed here, but I don't think there is a linux installation that comes without bash?

jhen0409: Alpine Linux doesn't have bash by default (it is still installable), that's why it launch scripts by sh. We use it because its image size.

rotemmiz: postinstall now uses node script to support Windows among Linux and macOS.
Give it a try (as soon as a new release is published), if you still have an issue, please report/submit a new PR

rotemmiz:
postinstall now uses node script to support Windows as well as Linux and macOS.
Please check if this solves your issue (as soon as a new release is published), If you still have problems, please report an issue/submit a PR.
In the meantime, I am closing this PR.

PR #727, opened 2018-05-12
Move iOS Device rotation to generated code

### LINKED ISSUE #719: Move IOS Driver to generated code
As a step in the direction of purely relying on generated code for the native adapters, we need to move the iOS Driver file `detox/src/devices/IosDriver.js` to use generated code. It still uses not generated code to trigger the device rotation.

DanielMSchmidt:


detox/src/ios/earlgreyapi/EarlGrey.js
```diff
@@ -0,0 +1,83 @@
+/**
+
+	This code is generated.
+	For more information see generation/README.md.
+*/
+
+
+function sanitize_uiDeviceOrientation(value) {
+	const orientationMapping = {
+		landscape: 3, // top at left side landscape
```
LeoNatan: Any reason why not expose all options?

DanielMSchmidt: It's just a refactoring, I would rather improve the situation a different PR :)

DanielMSchmidt: We have and this is the actual mapping we are currently using: https://github.com/wix/detox/blob/master/detox/src/devices/IosDriver.js#L59-L62

DanielMSchmidt: It was my first bigger PR in detox 😻

LeoNatan: Eh I confused rotate with shake. 
In either case, what blocks the full enum being exposed here?

PR #732, opened 2018-05-15
Add publish website to jenkins also

yershalom:


rotemmiz:
What will be the result of this PR? publishing on both CIs ? or is Jenkins still not running this script ?
We might need to disable one of those options.

Issue #571, opened 2018-02-14
Detox configuration should allow ability to specify app apk path and androidTest apk path


hoda0013:
#### Description
Currently Detox configuration asks for the file location of the apk file to be tested. It then uses convention to determine where the androidTest apk is located based on the filename it finds at the specified location. It would be nice to allow both paths to be specified in case the developer doesn't want to follow the file naming / hierarchy conventions.

rotemmiz:
I agree. A PR is most welcome!

savelichalex:
Yeah, that really frustrating. I just have started to work with `detox` and `apk` for `androidTest` is not that cli is expecting, though I followed docs

savelichalex:
I added new property to configuration in this [commit](https://github.com/savelichalex/detox/commit/d97883a9e9a81d95f4be610b79bced6d17b76b72). What do you think about it? Is it ready for PR or I need to add some sort of tests?

asztal:
Should be resolved by #1007 (not yet in a release though)

rotemmiz:
Closed with #1007. Out with detox@9.0.7

PR #733, opened 2018-05-15
Add `testBinaryPath` property to android configuration

savelichalex:
https://github.com/wix/detox/issues/571 that explain this PR


yershalom:
@savelichalex Don't forget to add `testBinaryPath` to `validOneDeviceAndSession` in `configurations.mock.js` so the jest coverage will pass

savelichalex:
@yershalom Can you please explain suggested fix? I took a look and saw configurations only for ios, but actually this new property work only on android (as I understood ios don't have any "test binary")

savelichalex:
@yershalom I made the change and the coverage magically return to 100%, I didn't know how - pure magic for me, thanks for suggestion!

PR #746, opened 2018-05-19
Move expect.js to generated code

### LINKED ISSUE #726: Move Android Interactions to generated code
As a step in the direction of purely relying on generated code for the native adapters, we need to move the Android Interactions located at `detox/src/android/expect.js` to use generated code.  None of the interactions are moved to generated code, so this might potentially be a bigger task, given that CI does not work with android currently.

DanielMSchmidt:
Closes #726

### DIFF
--- a/detox/src/android/espressoapi/DetoxAssertion.js
+++ b/detox/src/android/espressoapi/DetoxAssertion.js
@@ -1,0 +1,105 @@
+/**
+
+	This code is generated.
+	For more information see generation/README.md.
+*/
+
+
+function sanitize_matcher(matcher) {
+  if (!matcher._call) {
+    return matcher;
+  }
+
+  const originalMatcher = typeof matcher._call === 'function' ? matcher._call() : matcher._call;
+  return originalMatcher.type ? originalMatcher.value : originalMatcher;
+} 
+class DetoxAssertion {
+  static assertMatcher(i, m) {
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.espresso.DetoxAssertion"
+      },
+      method: "assertMatcher",
+      args: [{
+        type: "Invocation",
+        value: i
+      }, {
+        type: "Invocation",
+        value: sanitize_matcher(m)
+      }]
+    };
+  }
+
+  static assertNotVisible(i) {
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.espresso.DetoxAssertion"
+      },
+      method: "assertNotVisible",
+      args: [{
+        type: "Invocation",
+        value: i
+      }]
+    };
+  }
+
+  static assertNotExists(i) {
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.espresso.DetoxAssertion"
+      },
+      method: "assertNotExists",
+      args: [{
+        type: "Invocation",
+        value: i
+      }]
+    };
+  }
+
+  static waitForAssertMatcher(i, m, timeoutSeconds) {
+    if (typeof timeoutSeconds !== "number") throw new Error("timeoutSeconds should be a number, but got " + (timeoutSeconds + (" (" + (typeof timeoutSeconds + ")"))));
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.espresso.DetoxAssertion"
+      },
+      method: "waitForAssertMatcher",
+      args: [{
+        type: "Invocation",
+        value: i
+      }, {
+        type: "Invocation",
+        value: sanitize_matcher(m)
+      }, {
+        type: "Double",
+        value: timeoutSeconds
+      }]
+    };
+  }
+
+  static waitForAssertMatcherWithSearchAction(i, m, searchAction, searchMatcher) {
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.espresso.DetoxAssertion"
+      },
+      method: "waitForAssertMatcherWithSearchAction",
+      args: [{
+        type: "Invocation",
+        value: i
+      }, {
+        type: "Invocation",
+        value: sanitize_matcher(m)
+      }, searchAction, {
+        type: "Invocation",
+        value: sanitize_matcher(searchMatcher)
+      }]
+    };
+  }
+
+}
+
+module.exports = DetoxAssertion;
--- a/detox/src/android/espressoapi/UIDevice.js
+++ b/detox/src/android/espressoapi/UIDevice.js
@@ -151,31 +151,41 @@ class UiDevice {
   }
 
   static pressKeyCode(element, keyCode) {
-    if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
-    return {
-      target: element,
-      method: "pressKeyCode",
-      args: [{
-        type: "Integer",
-        value: keyCode
-      }]
-    };
-  }
+    function pressKeyCode1(keyCode) {
+      if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
+      return {
+        target: element,
+        method: "pressKeyCode",
+        args: [{
+          type: "Integer",
+          value: keyCode
+        }]
+      };
+    }
 
-  static pressKeyCode(element, keyCode, metaState) {
-    if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
-    if (typeof metaState !== "number") throw new Error("metaState should be a number, but got " + (metaState + (" (" + (typeof metaState + ")"))));
-    return {
-      target: element,
-      method: "pressKeyCode",
-      args: [{
-        type: "Integer",
-        value: keyCode
-      }, {
-        type: "Integer",
-        value: metaState
-      }]
-    };
+    function pressKeyCode2(keyCode, metaState) {
+      if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
+      if (typeof metaState !== "number") throw new Error("metaState should be a number, but got " + (metaState + (" (" + (typeof metaState + ")"))));
+      return {
+        target: element,
+        method: "pressKeyCode",
+        args: [{
+          type: "Integer",
+          value: keyCode
+        }, {
+          type: "Integer",
+          value: metaState
+        }]
+      };
+    }
+
+    if (arguments.length === 1) {
+      return pressKeyCode1.apply(null, arguments);
+    }
+
+    if (arguments.length === 2) {
+      return pressKeyCode2.apply(null, arguments);
+    }
   }
 
   static pressRecentApps(element) {
--- a/detox/src/android/espressoapi/ViewActions.js
+++ b/detox/src/android/espressoapi/ViewActions.js
@@ -5,6 +5,14 @@
 */
 
 
+function sanitize_matcher(matcher) {
+  if (!matcher._call) {
+    return matcher;
+  }
+
+  const originalMatcher = typeof matcher._call === 'function' ? matcher._call() : matcher._call;
+  return originalMatcher.type ? originalMatcher.value : originalMatcher;
+} 
 class ViewActions {
   static clearGlobalAssertions() {
     return {
@@ -14,6 +22,17 @@ class ViewActions {
       },
       method: "clearGlobalAssertions",
       args: []
+    };
+  }
+
+  static actionWithAssertions(viewAction) {
+    return {
+      target: {
+        type: "Class",
+        value: "android.support.test.espresso.action.ViewActions"
+      },
+      method: "actionWithAssertions",
+      args: [viewAction]
     };
   }
 
@@ -29,14 +48,35 @@ class ViewActions {
   }
 
   static click() {
-    return {
-      target: {
-        type: "Class",
-        value: "com.wix.detox.espresso.DetoxViewActions"
-      },
-      method: "click",
-      args: []
-    };
+    function click0() {
+      return {
+        target: {
+          type: "Class",
+          value: "android.support.test.espresso.action.ViewActions"
+        },
+        method: "click",
+        args: []
+      };
+    }
+
+    function click1(rollbackAction) {
+      return {
+        target: {
+          type: "Class",
+          value: "android.support.test.espresso.action.ViewActions"
+        },
+        method: "click",
+        args: [rollbackAction]
+      };
+    }
+
+    if (arguments.length === 0) {
+      return click0.apply(null, arguments);
+    }
+
+    if (arguments.length === 1) {
+      return click1.apply(null, arguments);
+    }
   }
 
   static swipeLeft() {
@@ -232,6 +272,24 @@ class ViewActions {
       },
       method: "openLinkWithUri",
       args: [uri]
+    };
+  }
+
+  static repeatedlyUntil(action, desiredStateMatcher, maxAttempts) {
+    if (typeof maxAttempts !== "number") throw new Error("maxAttempts should be a number, but got " + (maxAttempts + (" (" + (typeof maxAttempts + ")"))));
+    return {
+      target: {
+        type: "Class",
+        value: "android.support.test.espresso.action.ViewActions"
+      },
+      method: "repeatedlyUntil",
+      args: [action, {
+        type: "Invocation",
+        value: sanitize_matcher(desiredStateMatcher)
+      }, {
+        type: "Integer",
+        value: maxAttempts
+      }]
     };
   }
 
--- a/detox/src/ios/earlgreyapi/EarlGrey.js
+++ b/detox/src/ios/earlgreyapi/EarlGrey.js
@@ -1,0 +1,11 @@
+/**
+
+	This code is generated.
+	For more information see generation/README.md.
+*/
+
+
+
+class EarlGrey {}
+
+module.exports = EarlGrey;
--- a/generation/__tests__/android.js
+++ b/generation/__tests__/android.js
@@ -65,21 +65,6 @@ describe('Android generation', () => {
 
   describe('validation', () => {
     describe('Matcher<View>', () => {
-      it('should fail getting no object', () => {
-        expect(() => {
-          ExampleClass.matcherForAnd('I am a string', 'I am one too');
-        }).toThrowErrorMatchingSnapshot();
-      });
-
-      it('should fail with a wrong class', () => {
-        class AnotherClass {}
-        const x = new AnotherClass();
-
-        expect(() => {
-          ExampleClass.matcherForAnd(x, x);
-        }).toThrowErrorMatchingSnapshot();
-      });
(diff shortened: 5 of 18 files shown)


rotemmiz:
getting multiple errors in CI
```
m should be an instance of Matcher, got "[object Object]", it appears that it has a wrong class name: "Object"
```

1. maybe add a `toString` to that object
2. unit tests are failing...

DanielMSchmidt:
Will check that out 👍 worked on my machine TM

noomorph:
E2E tests don't pass on Android, almost everything fails with the same error:

```
  ● Async and Callbacks › should handle async await

    Error: Unhandled arg typeViewInteraction

      71 |     // when this test run fails, we want a stack trace from up here where the
      72 |     // $callee is still available, and not inside the catch block where it isn't
    > 73 |     const potentialError = new Error()
      74 | 
      75 |     try {
      76 |       await this.sendAction(new actions.Invoke(invocation));
      
      at Client.execute (../../src/client/Client.js:73:28)
      at InvocationManager.execute (../../src/invoke.js:11:33)
      at MatcherAssertionInteraction.execute (../../src/android/expect.js:112:29)
      at ExpectElement.toBeVisible (../../src/android/expect.js:244:87)
      at Object.it (10.async-and-callbacks.test.js:17:47)
```

@DanielMSchmidt , could you take a look please? Try out locally and report what you're getting.
Thanks in advance!

noomorph:
And I confirm that still I can see the `console.error` stuff @rotemmiz reported above:

```
m should be an instance of Matcher, got "[object Object]", it appears that it has a wrong class name: "Object"
```

It was printed 3-4 times during the test run, is it okay?

DanielMSchmidt:
@noomorph What @rotemmiz saw was an error being thrown, we now have the console.errors, because I could not find the root cause of the check going wrong

DanielMSchmidt:
@noomorph Just checked to be sure and the E2E tests are green and fine 👍

DanielMSchmidt:
@noomorph I added the logger to the generated code, please take a look and merge at will 👍

noomorph:
@DanielMSchmidt , thanks much. I'll check it out soon.

generation/core/generator.js
```diff
@@ -219,15 +220,21 @@ module.exports = function getGenerator({
...
       const ast = t.program([createClass(json), createExport(json)]);
       const output = generate(ast);
 
       const commentBefore = '/**\n\n\tThis code is generated.\n\tFor more information see generation/README.md.\n*/\n\n';
 
+      pathFragments.pop(); // remove filename
+      const outputPath = pathFragments.join("/")
+      const absoluteUtilsPath = path.resolve('../detox/src/utils');
+      const relativeUtilsPath = path.relative(outputPath, absoluteUtilsPath);
+      const logImport = `const log = require('${relativeUtilsPath}/logger');`;
```
noomorph: Change to `const log = require('${relativeUtilsPath}/logger').child({ __filename });`, so we can track from which files the log messages come.

noomorph: Also, could you extract this logic to a separate function, similar to a function `createTypeCheck(...)` above? I see that `generator` function looks already loong enough, why make it even bigger? :-)

generation/__tests__/android.js
```diff
@@ -2,6 +2,11 @@ const fs = require('fs');
 const remove = require('remove');
 const androidGenerator = require('../adapters/android');
 
+const mockErrorLog = jest.fn();
+jest.mock("../../detox/src/utils/logger", () => ({
```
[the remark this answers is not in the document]
DanielMSchmidt: I removed my own mock and used the one defined in `__mocks__/logger.js` Unfortunately we have two separate jest installations/contexts, this means I need to do a bit of glue code, but no biggie. I wonder what might happen if we would make more use of a central jest installation with multiple configs with the [jest multi-project runners](https://jestjs.io/blog/2017/05/06/jest-20-delightful-testing-multi-project-runner) 

Might give it a try if I have some spare time 🤔

generation/core/generator.js
DanielMSchmidt: Sure, no problem 👍 I think there are some parts I would like to slim down a bit and structure better, will go on with this soonish. My first goal was to get it working,so that I can see every edge case I ran into and make a better layer of abstraction later on ;)

DanielMSchmidt:
@noomorph Could you take another look 🙏

generation/__tests__/android.js
```diff
@@ -66,27 +71,25 @@ describe('Android generation', () => {
...
       it('should log that it gets no object', () => {
-        const spy = jest.spyOn(console, "error");
         expect(() => {
           ExampleClass.matcherForAnd('I am a string', 'I am one too');
         }).not.toThrow();
 
         
-        expect(spy).toHaveBeenCalled();
-        spy.mockRestore();
+        expect(mockErrorLog).toHaveBeenCalled();
```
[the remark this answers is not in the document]
DanielMSchmidt: 🤷‍♂️ I don't really see a point in having the `expect.any(String)`, but yeah it might sometime in the future catch something, so I'll add it :)

noomorph:
Looks good the way it is already. If builds pass, more or less, we can try to merge.

P.S. Out of curiosity, why do you do log.error. mockRestore manually?

DanielMSchmidt:
@noomorph are the reset by default?

noomorph:
@DanielMSchmidt , have you ever seen these messages on `npm run unit` in `detox`? Any ideas why? Of course, we can mock `.../logger.js`, but I would like to understand why do we have them in the first place? Thanks in advance.

screenshot from 2018-08-13 09-48-45

noomorph:
> @noomorph are the reset by default?

Yes, works fine. Anyway you already have auto-reset in your Jest configuration:
```json
	"jest": {
		"resetMocks": true,
		"resetModules": true,
```

So, if I remove these lines, it works perfectly fine.

```diff
diff --git a/generation/__tests__/android.js b/generation/__tests__/android.js
index 157a7558..d38a0f02 100644
--- a/generation/__tests__/android.js
+++ b/generation/__tests__/android.js
@@ -76,7 +76,6 @@ describe('Android generation', () => {
 
         
         expect(mockLogger.error).toHaveBeenCalledWith(expect.any(String));
-        mockLogger.error.mockRestore();
       });
 
       it('should log with a wrong class', () => {
@@ -88,7 +87,6 @@ describe('Android generation', () => {
         }).not.toThrow();
 
         expect(mockLogger.error).toHaveBeenCalledWith(expect.any(String));
-        mockLogger.error.mockRestore();
       });
 
       it("should succeed with the 'right' class", () => {
```

detox/src/android/espressoapi/DetoxAssertion.js
```diff
@@ -0,0 +1,133 @@
...
+
+  const originalMatcher = typeof matcher._call === 'function' ? matcher._call() : matcher._call;
+  return originalMatcher.type ? originalMatcher.value : originalMatcher;
+} 
+class DetoxAssertion {
+  static assertMatcher(i, m) {
+    if (typeof m !== 'object' || typeof m.constructor !== 'function' || m.constructor.name.indexOf('Matcher') === -1) {
+      const isObject = typeof m === 'object';
+      const additionalErrorInfo = isObject ? typeof m.constructor === 'object' ? 'the constructor is no object' : 'it has a wrong class name: "' + m.constructor.name + '"' : 'it is no object';
+      log.error('m should be an instance of Matcher, got "' + m + '", it appears that ' + additionalErrorInfo);
```
noomorph: Please, try to sort it out - why does it happen in unit tests for `detox` project? Maybe there are outdated inputs/mocks that need updating?

screenshot from 2018-08-13 09-48-45

detox/src/android/espressoapi/DetoxAssertion.js
```diff
@@ -0,0 +1,133 @@
...
+        value: i
+      }]
+    };
+  }
+
+  static waitForAssertMatcher(i, m, timeoutSeconds) {
+    if (typeof m !== 'object' || typeof m.constructor !== 'function' || m.constructor.name.indexOf('Matcher') === -1) {
+      const isObject = typeof m === 'object';
+      const additionalErrorInfo = isObject ? typeof m.constructor === 'object' ? 'the constructor is no object' : 'it has a wrong class name: "' + m.constructor.name + '"' : 'it is no object';
+      log.error('m should be an instance of Matcher, got "' + m + '", it appears that ' + additionalErrorInfo);
```
noomorph: Also, I see that you have almost or 100% identical error messages here, on lines 22, 73,  100. It's not easy to understand where the error messages come from. I suggest adding extra fields to the log message, e.g.:

screenshot from 2018-08-13 10-37-55

Below is a draft of what I think it is nice to have. It may be quick and dirty, and lacking generating `const util = require('util')` in the header

```diff
diff --git a/generation/core/generator.js b/generation/core/generator.js
index 86770566..ab97efe2 100644
--- a/generation/core/generator.js
+++ b/generation/core/generator.js
@@ -205,9 +205,9 @@ module.exports = function getGenerator({
     const typeCheckCreator = typeCheckInterfaces[type];
     const isListOfChecks = typeCheckCreator instanceof Array;
     return isListOfChecks
-      ? typeCheckCreator.map((singleCheck) => singleCheck(json))
+      ? typeCheckCreator.map((singleCheck) => singleCheck(json, functionName))
       : typeof typeCheckCreator === 'function'
-        ? typeCheckCreator(json)
+        ? typeCheckCreator(json, functionName)
         : t.emptyStatement();
   }
 
diff --git a/generation/core/type-checks.js b/generation/core/type-checks.js
index ccbe7e4a..764904f4 100644
--- a/generation/core/type-checks.js
+++ b/generation/core/type-checks.js
@@ -69,7 +69,7 @@ if (
 }
 
 function isOfClass(className) {
-  return ({ name }) =>
+  return ({ name }, functionName) =>
     template(`
 	if (
 		typeof ARG !== 'object' ||
@@ -79,10 +79,11 @@ function isOfClass(className) {
 		const isObject = typeof ARG === 'object';
 		const additionalErrorInfo = isObject ? (typeof ARG.constructor === 'object' ? 'the constructor is no object' : 'it has a wrong class name: "' + ARG.constructor.name +'"') : 'it is no object';
 
-		log.error('${name} should be an instance of ${className}, got "' + ARG + '", it appears that ' + additionalErrorInfo);
+		log.error({ event: EVENT, err: util.inspect(ARG) }, '${name} should be an instance of ${className}, got "' + ARG + '", it appears that ' + additionalErrorInfo);
 	}
 	`)({
-      ARG: t.identifier(name)
+      ARG: t.identifier(name),
+      EVENT: t.stringLiteral(functionName),
     });
 }
 
```

@DanielMSchmidt , I don't insist on it, however, but I'd be happy to know what you think about it.

detox/src/android/espressoapi/DetoxAssertion.js
DanielMSchmidt: @noomorph The problem is the way that we assert inputs vs. the way that we use it in the tests. I couldn't find the exact source, but in this test suite we are using either a plain object or a constructor with a different name, but the  assertion tests on the name of the constructor. 

I wanted to have a bit of guidance for our users, that they run into real errors or at least warnings when they are misusing an API, but it might be just the wrong thing or a wrong way. I'm happy to discuss solutions here.

noomorph: >  but in this test suite we are using either a plain object or a constructor with a different name, but the assertion tests on the name of the constructor.

So we just need to fix that in tests, right?

DanielMSchmidt: Yes. I would be glad if you could give it a try. Just replace the log.error with a throw and get the tests green 🙈

detox/src/android/espressoapi/DetoxAssertion.js
DanielMSchmidt: I'll give it a try. I like it better then before, but I would actually love to have an error there. But I think for that we would need to really check the integrity of the structure passed, to be sure we don't throw too eagerly. I will tackle this in a separate PR. For now, I'll change it to the better console.log

DanielMSchmidt:
@noomorph Changed this accordingly, hope it's fine now 👍

generation/core/global-functions.js
```diff
@@ -127,6 +127,10 @@ function sanitize_uiAccessibilityTraits(value) {
 } // END sanitize_uiAccessibilityTraits
 
 function sanitize_matcher(matcher) {
+  if (!matcher._call) {
```
noomorph: Looks something new to me, can you explain a bit? Is that a fix to the failing unit tests back then?

generation/adapters/android.js
```diff
@@ -2,15 +2,17 @@ const t = require('babel-types');
 const generator = require('../core/generator');
 const { callGlobal } = require('../helpers');
 
-const { isNumber, isString, isBoolean, isOfClass, isArray } = require('../core/type-checks');
+const { isNumber, isString, isBoolean, isArray, isDefined } = require('../core/type-checks');
 
 const typeCheckInterfaces = {
-  Integer: isNumber,
+  'ArrayList<String>': isArray,
+  'Matcher<View>': null, // isOfClass('Matcher'),
```
noomorph: @DanielMSchmidt , maybe let's put `isDefined` check here? Or there are special cases when the matcher is undefined?

DanielMSchmidt:
@noomorph Fixed that (the is defined check had a bug so it didn't occur previously, but that is fixed now) and the tests run through 👍

detox/src/android/espressoapi/ViewActions.js
```diff
@@ -40,6 +65,24 @@ class ViewActions {
     };
   }
 
+  static click(rollbackAction) {
+    if (!rollbackAction) {
```
noomorph: @DanielMSchmidt , I see unit test failing on this (`src/android/expect.js`). Is this a reason why you hesitated to add `isThruthy` check?

DanielMSchmidt: I will take another look at the unit tests 👍 No, I hesitated because I don't see so much value 🤷‍♂️

DanielMSchmidt: I should have hesitated because of this. The problem is it is optional and not required and we have no way to check if an argument is required or not. So all checks we do should test if the thing that is passed is what it promises to be if something is passed. If null is passed we won't be able to guard us on this side of the fence. I will do a quick research if there is any hypothetical way for us to know if sth is optional and I will let you know.

DanielMSchmidt:
@noomorph Unit should be fixed now, got weird android simulator issues locally, let
s see what the CI says 😇

noomorph:
Daniel, looks like it is still broken on Android. At least that's what CI tells.

DanielMSchmidt:
Will take a look 🙈

DanielMSchmidt:
@noomorph Could not reproduce, ran it three times on android and one time on iOS and everything went fine.

DanielMSchmidt:
So, this branch is now up to date 🎉 Locally it worked (before the massive rebasing I just did), so let's see what CI says 👍

rotemmiz:
Looks like the branch is not on par with master, there was an artifacts snapshot test we disabled but still runs here, can you make sure to merge current master to this branch?

DanielMSchmidt:
@rotemmiz That's strange, I rebased it from master before I pushed, so I am a bit confused. I can't see where I enabled it looking through the diff.

DanielMSchmidt:
I tried to see if sth is different on the JSON part send to the WS for the failing test, but they are identical (besides message id). So I think your hint into enabling a disabled functionality seems to be right.

DanielMSchmidt:
This is fixed and green on CI. @noomorph and @rotemmiz please review 😇

detox/src/android/espressoapi/UIDevice.js
```diff
@@ -151,31 +151,41 @@ class UiDevice {
...
-      method: "pressKeyCode",
-      args: [{
-        type: "Integer",
-        value: keyCode
-      }, {
-        type: "Integer",
-        value: metaState
-      }]
-    };
+    function pressKeyCode1(keyCode) {
```
rotemmiz: What happened here? 
I guess there's problem with Java overloads, did you handle it somewhere else in a different way?
Seems like the original code just has `metaState=0` as default.
https://android.googlesource.com/platform/frameworks/testing/+/79693ed/uiautomator/library/src/com/android/uiautomator/core/UiDevice.java#285

detox/src/android/espressoapi/UIDevice.js
```diff
@@ -151,31 +151,41 @@ class UiDevice {
...
+          type: "Integer",
+          value: keyCode
+        }, {
+          type: "Integer",
+          value: metaState
+        }]
+      };
+    }
+
+    if (arguments.length === 1) {
```
rotemmiz: is this generated code?

detox/src/android/espressoapi/ViewActions.js
```diff
@@ -17,28 +25,60 @@ class ViewActions {
...
+        value: "android.support.test.espresso.action.ViewActions"
       },
-      method: "click",
+      method: "clearText",
       args: []
     };
   }
 
+  static click() {
+    function click0() {
```
rotemmiz: Again here, why are there two clicks?
I see only one in UIDevice API

detox/src/ios/earlgreyapi/EarlGrey.js
```diff
@@ -0,0 +1,11 @@
+/**
```
rotemmiz: We can probably throw this away, as this class is empty. Everything is in EarlGreyImpl

generation/index.js
```diff
@@ -1,68 +1,57 @@
...
-		"../detox/src/ios/earlgreyapi/GREYConditionDetox.js",
-	"../detox/ios/EarlGrey/EarlGrey/Synchronization/GREYCondition.h":
-		"../detox/src/ios/earlgreyapi/GREYCondition.js",
-	"../detox/ios/Detox/GREYConfiguration+Detox.h":
-		"../detox/src/ios/earlgreyapi/GREYConfigurationDetox.js",
-	"../detox/ios/EarlGrey/EarlGrey/Common/GREYConfiguration.h":
-		"../detox/src/ios/earlgreyapi/GREYConfiguration.js",
-  "../detox/ios/EarlGrey/EarlGrey/Core/EarlGreyImpl.h":
-    "../detox/src/ios/earlgreyapi/EarlGreyImpl.js"
+  '../detox/ios/EarlGrey/EarlGrey/Action/GREYActions.h': '../detox/src/ios/earlgreyapi/GREYActions.js',
```
rotemmiz: better!

detox/src/android/espressoapi/UIDevice.js
DanielMSchmidt: yes

detox/src/android/espressoapi/ViewActions.js
DanielMSchmidt: There are two because the method is overloaded.

detox/src/android/espressoapi/UIDevice.js
DanielMSchmidt: The easiest way to handle overloading is to just check for the number of arguments. There are for sure better, more reliable ways, but this was the fastest. (I have checks in place that make sure we get some console output if we mess up here)

Issue #712, opened 2018-05-03
Detox test overwrites mocha grep option with --grep :android: --invert 


Ljo2017:
It seems that when running detox test, the grep options in mocha.opts gets overwritten by detox test and therefore not recognised.

Detox command used:
detox test -c ios.sim.release

Log:
node_modules/.bin/mocha e2e --opts e2e/mocha.opts --configuration ios.sim.release     **--grep :android: --invert** 

detox info 15:09:15: server listening on localhost:60573...
detox info 1: Searching for device matching iPhone 7... 

It looks as if --grep :android: --invert overwrites the grep option in mocha.opts.

If I run the command below, the grep option in mocha.opts is recognised:
node_modules/.bin/mocha e2e --opts e2e/mocha.opts --configuration ios.sim.release

Can we have this automatic platform deduction logic removed? Or is there any way not to have the mocha.opts option overwritten?

* Detox: 7.3.4
* React Native: .55
* Node: 8.9.4
* Device: iOS simulator
* Xcode: 9.2

rotemmiz:
For mocha you don't need to use Detox command line tool.
You can actually just use mocha's command line directly:
`mocha e2e --opts e2e/mocha.opts --configuration ios.sim.release`

Every detox cli param is passed almost with no change.

noomorph:
@Ljo2017  if you provide a function covered with integration tests that is able to merge `:platform` with an arbitrary user input, I think we could actually use it for mocha in next releases.

Ljo2017:
@noomorph we've made changes to merge the platform grep expression with the existing grep options in mocha.opts file and tried to make a pull request but got a permission denied error.

ERROR: Permission to wix/detox.git denied to Ljo2017.
fatal: Could not read from remote repository.

Could you give me access to the repository to make a pull request?

Ljo2017:
Thanks @noomorph.
Created a PR https://github.com/wix/detox/pull/748.

PR #748, opened 2018-05-22
Merging platform grep expression with existing grep options from mocha.opts

Ljo2017:
This PR is for the issue https://github.com/wix/detox/issues/712.

### DIFF
--- a/detox/local-cli/detox-test.js
+++ b/detox/local-cli/detox-test.js
@@ -2,6 +2,7 @@
 
 const program = require('commander');
 const path = require('path');
+const fs = require('fs');
 const cp = require('child_process');
 const _ = require('lodash');
 const CustomError = require('../src/errors/CustomError');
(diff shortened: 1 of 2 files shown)


rotemmiz:
I understand that the current implementation causes issue to you, and I can see it happening for us as well. I'm not sure this solution is a good one though, as it is not general enough (no handling of other test runners, other param overrides, etc).

I do however want to engage in a conversation, and try to find a good and general solution for problems of this type.
We can start a new discussion in an issue, linking the current issues and this PR, and let's talk.

Ljo2017:
That's ok and I understand it's not a trivial change.

However, the platform option change in the latest detox version broke the existing functionality and now we don't have the flexibility to add our own options.

This change probably is the minimum change we could make to fix this issue at this point in time until we come up with a better solution.

PR #755, opened 2018-05-29
Update Getting started with Jest

magrinj:


noomorph:
@magrinj, could you check the current document and decide if it actually needs these additions?
https://github.com/wix/detox/blob/master/docs/Introduction.GettingStarted.md

Currently, your PR looks to me out of sync.

magrinj:
@noomorph You're right, apparently my PR is not needed anymore

PR #762, opened 2018-05-30
Move generation to babel 7

DanielMSchmidt:
This allows us to have nicer comments in our generated code (and we need to do it anyways at some point).

### DIFF
--- a/detox/src/ios/earlgreyapi/EarlGreyImpl.js
+++ b/detox/src/ios/earlgreyapi/EarlGreyImpl.js
@@ -15,14 +15,14 @@ function sanitize_uiDeviceOrientation(value) {
 } 
 class EarlGreyImpl {
   /*Provides the file name and line number of the code that is calling into EarlGrey.
-In case of a failure, the information is used to tell XCTest the exact line which caused
-the failure so it can be highlighted in the IDE.
-
-@param fileName   The name of the file where the failing code exists.
-@param lineNumber The line number of the failing code.
-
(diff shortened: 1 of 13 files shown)


detox/src/devices/IosDriver.js
```diff
@@ -3,9 +3,9 @@ const fs = require('fs');
 const DeviceDriverBase = require('./DeviceDriverBase');
 const InvocationManager = require('../invoke').InvocationManager;
 const invoke = require('../invoke');
-const GREYConfigurationApi = require('./../ios/earlgreyapi/GREYConfiguration');
-const GREYConfigurationDetox = require('./../ios/earlgreyapi/GREYConfigurationDetox');
-const EarlyGrey = require('./../ios/earlgreyapi/EarlGrey');
+const GREYConfigurationApi = require('./../ios/earlgreyapi/GREYConfiguration'); // eslint-disable-line node/no-unpublished-require
```
rotemmiz: Please try to understand why this eslint doesn't like those lines. It makes no sense.

DanielMSchmidt: I can remove it, should be red on master too, right 🤷‍♂️ The explanation with the npmignore makes kinda sense though

rotemmiz:
@DanielMSchmidt, can we go forward with this PR? Can you please rerun it on the updated code?

DanielMSchmidt:
Sure, it was just out of focus previously 👍  Will put it on my list to rerun the generation again.

rotemmiz:
So is this ready?

PR #763, opened 2018-05-31
Follow-up windows changes

simonbuchan:
A couple quick fixes for `detox test --platform` and detox self unit tests running on windows.

### DIFF
--- a/detox/local-cli/detox-test.js
+++ b/detox/local-cli/detox-test.js
@@ -117,7 +117,7 @@ function runMocha() {
 function runJest() {
   const configFile = runnerConfig ? `--config=${runnerConfig}` : '';
 
-  const platformString = platform ? `--testNamePattern='^((?!${getPlatformSpecificString(platform)}).)*$'` : '';
+  const platformString = platform ? shellQuote(`--testNamePattern=^((?!${getPlatformSpecificString(platform)}).)*$`) : '';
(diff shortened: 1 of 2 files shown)


noomorph:
@simonbuchan , any updates so far? By the way, can we use a dedicated package for quoting? https://github.com/substack/node-shell-quote

I'd love to get unit test fixes for Win32 to the suite. Hope to see them merged :)

simonbuchan:
Nope, that's bash only, and not even particularly good quoting at that (see the issues list) - which is surprising to me, substack's stuff tends to be better than that. Been a while since I looked, but I don't remember finding anything that good.

In this case it doesn't actually matter, as the input isn't user controllable, and wouldn't give them any elevation of privilege if it were, a fully general quoting library is overkill. I'd only bother if we replaced all calls going through shell with an auto quoting library (eg all the ADB calls, some of which use user input) but that's a different PR.

Otherwise, is anything blocking this? PR failures look like they're just flaking.

noomorph:
@simonbuchan , good to hear from you. I think @rotemmiz was approaching you in Wix OSS Slack, could you check PM from him?

noomorph:
I took a look at the PR on Windows 10 64 Bit (CMD.EXE), Ubuntu 18.04 (Bash), MacOS 10.12 (Bash) — works solid, no objections.

PR #783, opened 2018-06-10
Make doc generation more robust

DanielMSchmidt:
There seem to be flakiness issues on CI, this is an approach to dampen the effect on us while investigating the root cause.


DanielMSchmidt:
@rotemmiz Could you take a look at this "fix". I would like us to less worry about the flakiness issues around our website deployment, therefore I implemented a simple retry strategy.

PR #785, opened 2018-06-11
Add Yarn workspace support to detox-cli

### LINKED ISSUE #779: detox-cli doesn't work on Yarn workspaces without nohoist
#### Description
Running the Detox CLI inside a subfolder within a Yarn workspace throws a `detox is not installed in this directory` error. This is because it's probably looking for `node_modules/detox` in the current directory, rather than in the hoisted `node_modules` at the monorepo root.

Adding `**/detox**` to the `nohoist` config prop in `package.json` works around this issue, but it'd be helpful if we didn't have to do this.

#### Steps to Reproduce
1. Create a [Yarn workspace](https://yarnpkg.com/lang/en/docs/workspaces/)
2. From within one of the workspace's sub-projects, run `yarn add --dev detox`. Note that it gets installed in `../node_modules` instead of `node_modules` because of Yarn's hoisting behavior.
3. Run `detox init`

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 7.4.0
* Node: 8.11.2
* macOS: 10.13.5

Zacqary:
Fixes #779 

If `node_modules/detox/package.json` doesn't exist, checks to see if `../package.json` has the cwd in its `workspaces.packages`, and if so, checks if `../node_modules/detox/package.json` exists.

### DIFF
--- a/detox-cli/cli.js
+++ b/detox-cli/cli.js
@@ -6,7 +6,36 @@ const fs = require('fs');
 const detoxPath = path.join(process.cwd(), 'node_modules/detox');
 const detoxPackageJsonPath = path.join(detoxPath, 'package.json');
 
-if (fs.existsSync(detoxPackageJsonPath)) {
+const isDetoxInstalled = (() => {
+  if (fs.existsSync(detoxPackageJsonPath)) {
+    return true;
+  }
+  // Check if this repo is within a Yarn workspace
+  const monorepoPackageJsonPath = path.join(
+    path.resolve(process.cwd(), '..'),
+    'package.json'
+  );
+  if (!fs.existsSync(monorepoPackageJsonPath)) {
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


DanielMSchmidt:
Seems solid, great work 👍

rotemmiz:
I'm a bit reluctant regarding this change. I am not sure it is detox-cli's job to understand all the possible project structures. This is an added complexity to a script that should be as simple, harmless, passthrough as possible.

Is there any other project that does the same to support yarn workspaces ?

noomorph:
$0.02 from me.

1. I agree it's nice to hear if:
> Is there any other project that does the same to support yarn workspaces ?

2. On the other hand, although this complexity is not something unmanageable, still there's quite a lot of additions to the main file. That's why I'd like to see a thorough **integration tests suite** against the Detox CLI before merging such changes.

PR #812, opened 2018-07-03
Bump Lodash dependency to ^4.17.5

noomorph:


rotemmiz:
Not that I have any objections, but semver calculation will resolve the same version for both ^4.14 and ^4.17, which is 4.x.x, so this PR doesn't really change anything in the user's bode_module dir.

noomorph:
Weird... 😕 
I checked https://github.com/npm/node-semver#caret-ranges-123-025-004
It says ^1.2.x := >=1.2.0 <2.0.0
So it should bump Lodash reqs at least to 4.17.x
Or am I wrong?

PR #820, opened 2018-07-10
Move generation to babel-template for better readability

### LINKED ISSUE #744: Move generator code to babel-template
Our current generator code is messy because building ASTs with `babel-types` is very verbose. If we switch to `babel-templates` we will have a lot less code and it's way easier to read and understand.

### Current implementation

```js
function createReturnStatement() {
  return t.returnStatement(
    t.objectExpression([
        t.objectProperty(
            t.identifier('target'),
            t.objectExpression([
                t.objectProperty(t.identifier('type'), t.stringLiteral('Invocation')),
                t.objectProperty(t.identifier('value'), t.identifier('element'))
            ])
        ),
        t.objectProperty(t.identifier('method'), t.stringLiteral(json.name)),
        t.objectProperty(t.identifier('args'), t.arrayExpression(args))
    ])
  );
}
```

### New implementation with `babel-template`

```js
const createReturnStatement = template(`
  return {
    target: {
      type: 'Invocation',
      value: element
    },
    method: '${json.name}',
    args: ARGS
  };
`)({
  ARGS: t.arrayExpression(args)
});
```

DanielMSchmidt:
Closes #744 and turns out that we actually can't move as much as I would like to, just because it is not possible with babel-template.

### DIFF
--- a/generation/core/generator.js
+++ b/generation/core/generator.js
@@ -181,21 +181,16 @@ module.exports = function getGenerator({
           : addArgumentContentSanitizerCall(arg, json.name)
     );
 
-    return t.returnStatement(
-      t.objectExpression([
-        t.objectProperty(
-          t.identifier('target'),
-          json.static
-            ? t.objectExpression([
-                t.objectProperty(t.identifier('type'), t.stringLiteral('Class')),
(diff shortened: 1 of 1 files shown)


DanielMSchmidt:
I think this is not really of help, let's close it instead

PR #854, opened 2018-07-24
Added com.android.support:support-annotations

jamespearson:
Added com.android.support:support-annotations to the dependency instructions, as the sample ```DetoxTest.java``` uses annotations.


rotemmiz:
ping

PR #858, opened 2018-07-25
fix: adds grep fallback for win32

noomorph:
Last two pull requests have broken Win32 compatibility by piping some of the commands' outputs to `grep` which is absent by default in the Windows environment.

This pull request makes it possible to switch to `findstr` command fallback if `process.platform === 'win32'`.


detox/src/utils/pipeCommands.test.js
```diff
@@ -0,0 +1,102 @@
...
+        const bundleId = 'com.company.bundle';
+        expect(escape(bundleId)).toBe('com\\.company\\.bundle');
+      });
+    });
+  });
+
+  describe('search by regexp pipe command', () => {
+    const regexp = '^Hello\\s ".*"$';
+
+    describe('on win32', () => {
```
rotemmiz: These tests seem to test nothing. we already know what `pipeUtils.search.regexp.win32` is. There's probably no need to rewrite the same logic in the test.

PR #860, opened 2018-07-26
Add Android back button (copy of PR#505)

noomorph:


noomorph:
@rotemmiz , looks good to me now.

PR #873, opened 2018-08-03
Implement the capability to change the language between tests

luisnaranjo733:
This PR implements the capability to change the language between tests for iOS by adding flags to xcrun. It looks like this is straightforward for Android as well. I didn't make that change because I'm not set up to verify it works on Android at the moment. I'm hoping someone else can make a change in my branch, or can just make another PR later.

Basically I just added a new optional language field to the params object that gets passed to device.launchApp().

Example usage:

init.js
```
beforeAll(async () => {
  await detox.init(config, {launchApp: false});
});
```

test.spec.js
```
const LOCALES = ["es-MX", "fr-FR"];

LOCALES.forEach(locale => {

  describe("Run tests in various languages", () => {

    beforeAll(async () => {
      await device.launchApp({
        newInstance: true,
        language: locale
      });
    });

    beforeEach(async () => {
    });

    it(`Example test in ${locale}`, doStuff());

  });


});
```


LeoNatan:
Thanks for this PR.

Locale is something different from language and it is incorrect to tie them together in all cases. I think it would be better to give an option for providing a different locale than the language, and if no locale is provided, use the system one (nothing passed, like it is now for language).

detox/src/devices/AndroidDriver.js
```diff
@@ -86,7 +86,7 @@ class AndroidDriver extends DeviceDriverBase {
     }
   }
 
-  async launch(deviceId, bundleId, launchArgs) {
+  async launch(deviceId, bundleId, launchArgs, language) {
```
LeoNatan: Please add a locale option as well.

rotemmiz:
Thanks for taking the time to submit this PR! 

There are a few things that need to be resolved before we can continue:
1. Regarding the API, Android must be taken into account. It seems to not be straightforward (as far as I tested). I'm not sure these params can be passed at launch. There are a few resources regarding changing locale during tests
https://developer.android.com/guide/topics/resources/localization#testing
https://stackoverflow.com/questions/16760194/locale-during-unit-test-on-android
2. Every feature **must be tested with the test app**. This means, that we need to add language support to the test app, and use the new API to test it actually triggers language/locale change in the app.
3. @LeoNatan's comment is very important, and on top of that, the two platforms seem to construct the locale/language/country in a slightly different way.
Android: `Locale locale = new Locale(language, country);` "de", "DE"
[iOS](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/TestingYourInternationalApp/TestingYourInternationalApp.html):  `-AppleLanguages "(de)"` or `-AppleLocale "fr_FR"`

LeoNatan:
Testing on the test suite can be very easy to handle: Just put the app language and app locales to text fields and test that they are correct.

LeoNatan:
@rotemmiz Regarding API, I am not sure I understand. 

In iOS settings, setting the country is really setting the locale. So en_GB, en_CA, en_US, de_DE, etc. Setting the language changes the interface language, but it's also given in the same notation, so en_GB, en_CA, en_US, de_DE, etc. It might also accept shorter notation ("de") and that is mapped to a default country.

In Android, is there no way to define the locale different than device/app language? Seems like a big design flaw as the two are really unrelated.

luisnaranjo733:
Hey, just thought I'd pop in and give an update - 

Basically I haven't had any bandwidth to work on this in a while. I'm hoping I'll have some time in 1-2 weeks. If anyone else is free to work on this, feel free to fork my fork or work directly in my fork (I'll grant anyone access that asks me).

I still plan on doing my best to address all the requested changes and feedback, just haven't had any time :(

luisnaranjo733:
Here is one approach we could take for implementing this on Android

1) Check [this](https://play.google.com/store/apps/details?id=net.sanapeli.adbchangelanguage) apk into the Detox source code. Document its usage and provide credit to the author
2) Use adb to install this APK and grant it CHANGE_CONFIGURATION permission when this language API is invoked
```
adb install path/to/adb-change-language.apk
adb shell pm grant net.sanapeli.adbchangelanguage android.permission.CHANGE_CONFIGURATION
```

3) Change the language by deeplinking to this 3rd party app via adb
```
adb shell am start -n net.sanapeli.adbchangelanguage/.AdbChangeLanguage -e language pt-rBR
```
4) The API would need to be slightly different for Android. The deep-linking syntax for this language switching app takes a single 'language' parameter in <language>:<locale> syntax (e.g. "es-MX"). As @LeoNatan pointed out, the language/locale system in Android is not as robust as iOS.

I still think this should be done in a separate PR, but I just wanted to share how I think this could be achieved :)

luisnaranjo733:
Looks like the CI results went from passing to failing when I closed and re-opened the PR.
Some of the e2e animations started failing. Anyone have any idea why? Seems completely unrelated.

22:21:21     ✕ should detect loops with final number of iterations (driver: Native) (481822ms)
22:21:21     ✕ should not wait during delays longer than 1.5s (driver: Native) (960341ms)
22:21:21     ✕ :ios: should wait during delays shorter than 1.5s (driver: Native) (960301ms)

LeoNatan:
I am OK with this change. @luisnaranjo733 Could you please update documentation with the new API? Thanks

luisnaranjo733:
Awesome! I just merged the latest from master into my branch, and confirmed that unit and e2e tests are still passing.

I've also added some doc changes to the device API page.

LeoNatan:
Thank you!

PR #880, opened 2018-08-11
[WIP] Generate documentation from JsDoc strings

DanielMSchmidt:
This is a first try of getting our documentation in a shape where you can easily spot differences and where it is very maintainable by moving it to the code.

It should lead to the resolution of #563


LeoNatan:
Let's keep the branch, but closing the PR for now.

PR #884, opened 2018-08-13
[Android] allow config extension

matteo-hertel:
👋 Wix!

First of all thanks for making detox, is really amazing 😄 

As per the title, I've created this super small PR to allow the android build file to inherit config from the root project, that comes in really handy if the project pulling in detox needs different configs.

I've ~stole~ borrowed the names from [https://github.com/react-native-community/react-native-linear-gradient/blob/master/android/build.gradle](https://github.com/react-native-community/react-native-linear-gradient/blob/master/android/build.gradle) given it's from the react native community it should be ok

I didn't open an issue for this, do you need me to?

### DIFF
--- a/detox/android/detox/build.gradle
+++ b/detox/android/detox/build.gradle
@@ -1,11 +1,18 @@
 apply plugin: 'com.android.library'
 
+def _ext = rootProject.ext;
+
+def _compileSdkVersion = _ext.has('compileSdkVersion') ? _ext.compileSdkVersion : 25;
+def _buildToolsVersion = _ext.has('buildToolsVersion') ? _ext.buildToolsVersion : '26.0.2';
+def _minSdkVersion = _ext.has('minSdkVersion') ? _ext.minSdkVersion : 18;
+def _targetSdkVersion = _ext.has('targetSdkVersion') ? _ext.targetSdkVersion : 25;
+
 android {
-    compileSdkVersion 25
-    buildToolsVersion '26.0.2'
+    compileSdkVersion _compileSdkVersion
+    buildToolsVersion _buildToolsVersion
     defaultConfig {
-        minSdkVersion 18
-        targetSdkVersion 25
+        minSdkVersion _minSdkVersion
+        targetSdkVersion _targetSdkVersion
         versionCode 1
         versionName "1.0"
     }


noomorph:
@matteo-hertel, thanks for submitting the PR and sorry for the delay. Actually, I think you'll have to wait even a bit more, till @rotemmiz returns from his vacation, that's in a few days.

Don't consider the next thing as a request, but I'd be happy to learn (Gradle is not my thing yet, still have to grasp it) - what is the use case that is addressed by this inheritance - does it solve a version conflict in your Gradle builds, speeds them up or anything else?

matteo-hertel:
Hey @noomorph thanks for your reply

The aim here is to have the android configurations in one place (the root project) and have the Detox build file using those config if provided, I have an example of a project that uses Detox in which the `compileSdkVersion` is 26 but Detox has version 25 hardcoded, and it errors out during the project build

It's also standard for android projects to allow this kind of extension, so win win 😄

noomorph:
@rotemmiz, you're more familiar with this area, please tell if this change benefits some project configurations. I don't see proofs like before the change VS after the change (e.g. project fails to build -> project builds successfully), but maybe it's just me and the benefit is already obvious for any person who's good with Gradle. :confused:  Techincally, I have no objections regarding the merge as the build passes on Gradle 4.10.

rotemmiz:
First, thanks for this PR and sorry for the super long response time.
I have come back to this PR for a few times now, and I never know how to react to it :(
On one hand, I have never seen and Android project library uses this trick, I never though it is a problem since it can always be overriden in the application/project scope.
On the other hand, in the limited case, where Detox is the only library compiled from source (and no other library needs to be overriden), this trick is actually useful, since users don't need to change anything in their setup.

The other part of the solution can come from a different approach, and that is consuming detox from a binary instead of from sources. There's an open PR made by the members of the team, adding aar artifacts to Detox's npm package. This way, you won't have to go through a (potentially) tedious compilation process and adaptations in your project.

matteo-hertel:
Hi @rotemmiz thanks for getting back to me, my use case was covering the use case of Detox being the only library compiled from source, if the open PR that adds binary to npm covers this use case as well then is a non issue anymore, feel free to close this PR I was happy to contribute to the project 😄

PR #898, opened 2018-08-22
Remove use of legacy notifications API [WIP]

LeoNatan:
Remove legacy local and remote notification API
Remove legacy RN support

React Native also dropped support for Xcode 8: facebook/react-native#18944


abarisic86:
Any reason why this isn't reviewed and merged?

LeoNatan:
Internal reasons and backlog.

Is there something specifically that is blocking you?

Liampronan:
@LeoNatan not sure if this PR would fix this issue, but it seems like there is an issue related to Xcode 10: https://github.com/wix/Detox/issues/911#issuecomment-424621863. i may give this branch a try in a bit to see if it helps with that issue

LeoNatan:
All issues with Xcode 10 building have been resolved, as far as I know. This branch updates the minimum version requirement of app deployment target and cleans up several uses of deprecated APIs. If implemented correctly, it should give no signs to the end user that something changed (other than require the end user to update their toolchain and their apps).

Liampronan:
got it, thanks for clarifying, @LeoNatan! I just saw that issue and posted here cause I thought it might help with context. After going through some more issues, I was able to solve my problem, so yeah feel free to ignore above post and thank you for getting back!

noomorph:
@LeoNatan, what's up with this branch? Anything useful that is still unmerged and we could leverage?

LeoNatan:
The whole branch.

LeoNatan:
Cannot be merged because of One App.

LeoNatan:
I see now that internally, we will start using the new iOS notifications API soon. In that case, I will fix conflicts and merge once that happens.

LeoNatan:
I'll close this and start a new one.

Issue #897, opened 2018-08-21
Postinstall script fails when just command line tools are installed


rye:
#### Description

When I install `detox` on macOS with just the XCode command line tools installed, `detox`'s postinstall script fails with the following output:

```console
⑆ npm i -g detox
/usr/local/bin/detox -> /usr/local/lib/node_modules/detox/local-cli/detox.js

> dtrace-provider@0.8.7 install /usr/local/lib/node_modules/detox/node_modules/dtrace-provider
> node-gyp rebuild || node suppress-error.js

  ACTION binding_gyp_ndtp_target_build_ndtp .
  TOUCH Release/obj.target/ndtp.stamp

> detox@8.1.6 postinstall /usr/local/lib/node_modules/detox
> node scripts/postinstall.js

+++ dirname /usr/local/lib/node_modules/detox/scripts/build_framework.ios.sh
++ dirname /usr/local/lib/node_modules/detox/scripts
+ detoxRootPath=/usr/local/lib/node_modules/detox
++ node -p 'require('\''/usr/local/lib/node_modules/detox/package.json'\'').version'
+ detoxVersion=8.1.6
++ echo 8.1.6
++ xcodebuild -version
++ shasum
++ awk '{print $1}'
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
+ sha1=aba5c8f25bf328b40be1e29d2d653101f6ecc89d
+ detoxFrameworkDirPath=/Users/kristofer/Library/Detox/ios/aba5c8f25bf328b40be1e29d2d653101f6ecc89d
+ detoxFrameworkPath=/Users/kristofer/Library/Detox/ios/aba5c8f25bf328b40be1e29d2d653101f6ecc89d/Detox.framework
+ main
+ '[' -d /Users/kristofer/Library/Detox/ios/aba5c8f25bf328b40be1e29d2d653101f6ecc89d ']'
+ '[' '!' -d /Users/kristofer/Library/Detox/ios/aba5c8f25bf328b40be1e29d2d653101f6ecc89d/Detox.framework ']'
+ echo '/Users/kristofer/Library/Detox/ios/aba5c8f25bf328b40be1e29d2d653101f6ecc89d was found, but could not find Detox.framework inside it. This means that the Detox framework build process was interrupted.
[...]

PR #900, opened 2018-08-26
build_framework.ios.sh: echo/exit from current shell

rye:
The `()` syntax spawns a subshell and executes the commands. Hence, the "successful" `exit 0` is trapped and is effectively a no-op. This change instead moves the execution of the `echo` statement to the parent shell which causes the `exit 0` to actually kill the process where it is.

This is part of the fixes to the `build_framework.ios.sh` shell script that I had requested in #897.

This PR changes from subprocess execution to in-shell execution. The biggest change that this makes is that it causes the `exit 0` to actually stop the shell script.


rotemmiz:
Thanks for this!

PR #903, opened 2018-08-28
Fix links in detox object docs

fvonhoven:
* Noticed the links to the methods for detox `object` didn't work, so went ahead an fixed them up.
* And apparently my editor cleaned up some whitespace 🤷‍♂️ 

##### Before
detox-object-broken

##### After

detox-object-fix


rotemmiz:
Thanks!

PR #904, opened 2018-08-29
feat: device.takeScreenshot(name)

### LINKED ISSUE #807: Provide API for taking screenshots on demand
The artifacts API in its current form enables screenshots to be easily taken before and after each test.
What if the test wants to take a screenshot at some arbitrary point in the middle of the test?

A few scenarios where this would be valuable:

- Using detox to take screenshots of various screens for QA purposes. If the tests could control when screenshots are taken, they could navigate, take a screenshot, and then repeat the cycle without having to restart the test to manually trigger an afterEach screenshot capture, thus saving time and effort.
- A test encounters an unexpected state in the middle of the test, and wants to capture the intermediary screen (that the afterEach screenshot may be too late to capture)
- A test just wants to capture a screenshot of some valid intermediary state for human sanity checking purposes that is not available at the end of the test.

Perhaps the API could look something like this:
```
// detox could append ".png" and put this in the same folder as beforeEach.png and afterEach.png
await device.takeScreenshot("exampleScreenShotName");
```

Would love to see what you guys think about this :)

noomorph:
Resolves #807

Things to complete:

- [x] fix folder issue
- [x] make sure .untrackArtifact() is called
- [x] save the remainder of screenshots between `before` and `beforeEach`, `afterEach` and `afterAll`
- [x] documentation


noomorph:
@rotemmiz , if you'll get some time to look, do you agree with this implementation approach in general?

LeoNatan:
I think I want us to discuss this first. I think exposing a `takeScreenshot()` API is not a good approach, or at least not good in the cases I am thinking this functionality is needed.

So first, I will state the case that has the most need for this functionality in my view: when a test fails. User wants a screenshot of the device at the time of failure. (Out of the three cases pointed in #807, this is the most important and most useful to users in my opinion.)

However an explicit approach will almost never work for this case. The reason is asynchronicity. It takes an amount of time from Detox failing to notifying the user to them calling `takeScreenshot()`. However short that amount is, it is without doubt already too late and the UI has most likely changed. Even if the testing suite (Detox) implements a mechanism to automatically take a screenshot on a failure, it would still be too late.

Screenshots on failure should be captured in native, at the exact time of failure. This will ensure as accurate screenshot as possible. On iOS, this is easy—there is API which can be called that will create a snapshot of the screen.

If we decide to add this approach (in addition to the `takeScreenshot()` API that is already implemented), we should design a mechanism to transfer these screenshots from the device. One way could be to transmit the base64 data of the screenshot back to the server/tester and have that save it where appropriate and with whatever name it deems fit. Another is to provide an artifacts folder to the device and come up with a universal naming scheme, so that device can save the screenshots on its own.

haswalt:
The main benefit of having this api exposed to us allows us to implement detox as our main deployment workflow. Without this we have to implement something like fastlane tools which means writing uitest scripts for ios aching means never use detox as it’s duplcating stuff really at this point. If detox would allow us to take screenshots we can remove all other test frameworks from our pipeline and rely on tests and you capture from one concise suite / set of files. 

I know this ends up being a little bit out of the remit for detox, however I feel it’s a strong use case particularly with the view to integrate react native more deeply with detox / become the defacto test method.

elyalvarado:
This sounds very interesting, as @haswalt  commented such an API will be highly beneficial for documentation purposes and to avoid using other screen capture tools in the CI pipeline. Having it integrated right into detox allows for deduplication of code for capturing screenshots.

It is important to note that detox is already being used for purposes other than testing, as an example I'm already using the video capture option of detox for generating artifacts that are reviewed by designers, so having it take screenshots explicitly through an API will make detox complete the whole suite of tools I need to generate tools for reviewing UI/UX.

This feature could also later be extended for automated comparison of images by using computer vision, kind of snapshots of the UI as rendered.

fjtrujy:
Hello,
What is going to happen with this PR? 
Do we have plans to merge it in the future?
Do we have still work to do over here?

I see as well a very good useful feature, so if it is possible I'm more than happy to help as much as I can to have this feature included.

Thanks

bomkhec:
I see this PR not yet merge. Why?
Currently I need API takeScreenShot to add Attachment screenshot for reporter (ex: Allure framework).
Do you know another way to add attachment for reporter? Plz help me. Tks.

noomorph:
@elyalvarado, thanks for the words of support. Right now the problem with this PR is that I had to disengage from contributing to Detox in September (involuntarily), and there's still some other job to do before I can resume on Detox. At best, it might be another 2 or 3 weeks to wait, but it's only a cautious assesment.

henrymoulton:
It'd be great to have this feature inside Detox.

One reason we'd value this feature is for taking screenshots of our screens in various states and comparing to a design .PNG (Otherwise known as Visual or Design QA).

Once we have the ability to take the screenshot we can compare it to another image using something like [Jest Image Snapshot](https://github.com/americanexpress/jest-image-snapshot/). 

We can also build the ability to upload the screenshots somewhere to sit alongside designed screens which would help designers, PMs and even engineers understand the progress UI engineers are making for during a project.

Perhaps there's another tool for doing this, but after familiarising myself with Storybook's [Automated Visual Testing](https://storybook.js.org/testing/automated-visual-testing/) guide I think this might be a missing piece in the React Native ecosystem that this feature could provide (the web has this ability with headless Chrome).

@noomorph I assume life has got in the way of this PR?

noomorph:
@henrymoulton, oh yeah, I'm in the middle of finishing an urgent assignment. 😿 Detox team wants me to return already, but I have two commitments on me with a higher priority. And again, the cautious ETA is the end of this month. 🤞

inanc-g:
Any News on this?

noomorph:
@inanc-g , yes, there are news. I'm currently working on this sequence of pull requests:

* #1109 - Switch CLI to yargs (I'm in the final stages, testing on operating systems, minor fixes and docs)
* #1165 - Detox Instruments Integration (by the way, this is a new artifact plugin)
* … - not totally sure, if I'll have some things more prioritized assinged to me, maybe nothing
* #904 - add device.takeScreenshot() function (this is this PR)

inanc-g:
@noomorph thank you for the update, looking foward to this functionality.

LeoNatan:
Before implementing, let's discuss the approach. See my comments in https://github.com/wix/Detox/pull/904#issuecomment-417913887

noomorph:
Guys, looks like we are almost done here. The only missing thing is documentation. I plan to add it tomorrow and we can release.

noomorph:
To everyone waiting, interested, concerned or involved:

* @AntaninaKaltsova
* @BrettWalkerWork
* @LeoNatan
* @bomkhec
* @elyalvarado
* @falquaddoomi
* @fjtrujy
* @hasanen
* @haswalt
* @henrymoulton
* @inanc-g
* @jmatsushita
* @noamco 
* @pwfcurry 
* @rotemmiz
* @rszalski
* @transitive-bullshit
* @vincenzovitale
* @vkatushenok

The pull request is landing onto `master`.
There are two caveats, however, so please mind:

* See @LeoNatan's comment: https://github.com/wix/Detox/pull/904#issuecomment-417913887 (tl;dr - in the current implementation, screenshots are taken asynchronously via CLI commands, so exact timings are not guaranteed if you seek for precise screenshots immediately after errors or whatever like that)
* Right now you won't be able to use `device.takeScreenshot` in `--take-screenshots failing` mode due to a bit simplified implementation of screenshots plugin lifecycle. This shortcoming will be fixed in one of the next minor versions.

elyalvarado:
Excellent work @noomorph, I look forward to using this feature in the next release and to all the exciting possibilities this will allow.

Issue #901, opened 2018-08-27
point to a single configuration file


oreporan:
#### Description
Currently it seems that detox allows configuring only in the `package.json` of the project. But using a mono-repo (lerna), we want to run detox in specific packages. However I don't want to duplicate our detox config in each package.json
I've seen a discussion in https://github.com/wix/detox/issues/175 but it doesn't mention creating the file outside the `package.json`
Possible solution:

`detox-config.json`

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 8.0.0
* Node: 8.9.4

tomatobrown:
you can do this in the current build
```
const config = require('../config').detox;
await detox.init(config);
```

oreporan:
Well wouldn't that be just for the `test` command?
I'm talking about 

`detox build --configuration`

oreporan:
Unless I put a detox and configuration object in my `package.json`, I still get:

```js
 if (_.size(config.configurations) === 1) {
                    ^
TypeError: Cannot read property 'configurations' of undefined
    at getDefaultConfiguration (/Users/ore/env/dev/repos/x/node_modules/detox/local-cli/detox-test.js:207:21)
    at Object.<anonymous> (/Users/ore/env/dev/repos/x/node_modules/detox/local-cli/detox-test.js:23:151)
```js

**Comment by oreporan** [comment]
This is due to 

detox/local-cli.js
`const config = require(path.join(process.cwd(), 'package.json')).detox;`

**Comment by rotemmiz** [comment]
You're right, we should give an opportunity for users to get an external config file, not only in package.json. 
Logic may try getting config from package.json, and failover to a config file (or vice versa).

A PR is most welcome!

**Comment by oreporan** [comment]
Trying to run `scripts/ci.ios.sh` in the repo results in error:

```
> detox-test@0.0.1 build:ios /Users/ore/env/dev/open_source/detox/detox/test
> detox build --configuration ios.sim.release

/Users/ore/env/dev/open_source/detox/detox/local-cli/detox-build.js:28
  throw new Error(`Could not find build script in detox.configurations["${program.configuration}"]`);
  ^

Error: Could not find build script in detox.configurations["ios.sim.release"]
```

noomorph:
@opherg , check what's inside of your `detox/tests/package.json`. Maybe you modified something?

oreporan:
Sorry, I found the issue and removed the comment :)

oreporan:
Made a PR
[...]

PR #906, opened 2018-08-29
Allow giving detox config file as param, with fallbacks

oreporan:
Solves issue #901 

I couldn't really add specific tests for this, maybe you can help me out on where.
Because the e2e is basically one project, so its already taking the config from the `tests/package.json` and I'm not sure adding another test just for this is justified


oreporan:
Anyone help me out on why this is failing?
Looks unrelated to me

haswalt:
Android tests always fail that I can see. This PR is about fixing that #910 so yeah unrelated.

detox/local-cli/detox-build.js
```diff
@@ -27,3 +29,15 @@ if (buildScript) {
 } else {
   throw new Error(`Could not find build script in detox.configurations["${program.configuration}"]`);
 }
+
+function getConfigurationFile(configPath) {
+  let config;
+  if (configPath) config = require(path.join(process.cwd(), configPath));
```
LeoNatan: What would happen if configPath is not a relative path, say like `/Users/me/myConfig.json`?

detox/local-cli/detox-test.js
```diff
@@ -204,11 +206,24 @@ function clearDeviceRegistryLockFile() {
...
 function getDefaultConfiguration() {
-  if (_.size(config.configurations) === 1) {
-    return _.keys(config.configurations)[0];
+  const file = getConfigurationFile();
+  if (file && _.size(file.configurations) === 1) {
+    return _.keys(file.configurations)[0];
   }
 }
 
+function getConfigurationFile(configPath) {
```
LeoNatan: Why the need for duplication? Can't you use the same function?

LeoNatan:
Thanks for this PR!
A few questions from me.

oreporan:
?

noomorph:
@oreporan, firstly, thanks for submitting the PR. :+1: I'd also love to see this feature in Detox. However, from my perspective, it still lacks a few things to get it merged, including but not limited to the updates to Detox CLI documentation.

I'll review it more thoroughly within the next few days, and get back to you with a list of TODOs. Cheers!

oreporan:
Let me know what to do, I'd like to get this in, in the near future 
It will really cleanup our package.jsons

oreporan:
Anyone had a chance to look at this? PR has been open for over a month now

rotemmiz:
Hey @oreporan, I have a few nitpicks (naturally) :)

1. Let's change the file name from `.detoxrc.json` to `.detoxrc`
2. Can you please add documentation in the relevant sections ? (added links)
https://github.com/wix/Detox/blob/master/docs/Introduction.GettingStarted.md#3-add-detox-config-to-packagejson
https://github.com/wix/Detox/blob/master/docs/APIRef.Configuration.md#device-configuration
3. Ideally I would like to test this somehow, but the cli code is not tested at all... our bad.

Thanks!

oreporan:
@rotemmiz 
Thanks, made the changes. Not sure what change is needed in 
https://github.com/wix/Detox/blob/master/docs/APIRef.Configuration.md#device-configuration

However I did add it to now to ` docs/APIRef.DetoxCLI.md`

docs/APIRef.DetoxCLI.md
```diff
@@ -72,7 +72,7 @@ Run a command defined in 'configuration.build'
 | --- | --- |
 | -h, --help                            |  output usage information |
 | -c, --configuration \<device config\> |  Select a device configuration from your defined configurations,if not supplied, and there's only one configuration, detox will default to it |
-| --config-path \<configPath\> |  Select a device config-file path, if not supplied, detox will default to the package.json, and if not found there, detox will fallback to .detoxrc.json |
```
oreporan: 👍

detox/src/utils/getConfigurationFile.js
```diff
@@ -0,0 +1,18 @@
+const path = require('path');
+const packageJson = 'package.json';
+const detoxRc = '.detoxrc';
+
+function getConfigurationFile(configPath) {
+    let config;
+    if (configPath) config = require(path.isAbsolute(configPath) ? configPath : path.join(process.cwd(), configPath));
```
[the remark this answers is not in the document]
noomorph: Sorry, missed your `path.isAbsolute()` check. But anyway, the necessity of integration tests here still remains.

detox/src/utils/getConfigurationFile.js
```diff
@@ -0,0 +1,18 @@
...
+    if (!config) config = require(path.join(process.cwd(), packageJson)).detox;
```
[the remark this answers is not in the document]
noomorph: Also, I don't like these tricks with `path.isAbsolute() ? ... : ...`. I think that this method is more appropriate: [path.resolve([...paths]) - Path | Node.js v10.11.0 Documentation](https://nodejs.org/api/path.html#path_path_resolve_paths)

detox/src/utils/getConfigurationFile.js
```diff
@@ -0,0 +1,18 @@
...
+    if (!config) {
+      try {
+        config = require(path.join(process.cwd(), detoxRc));
+      } catch (error) {}
+    }
+    return config
+  }
+
+  module.exports = getConfigurationFile;
```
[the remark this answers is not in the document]
oreporan: Not sure how I can test these changes other configurations. As it seems that there is 1 app which is being tested, and it has its config in the `package.json`

oreporan:
trying to re-run the tests locally fails in:

```

❌  error: Build input file cannot be found: '/Users/ore/env/dev/open_source/detox/detox/test/node_modules/react-native/Libraries/WebSocket/libfishhook.a'
```

I've recently updated to xcode 10

noomorph:
@oreporan , it's not Detox issue AFAI can see. Have you tried to google it?
That's what I found recently on this: https://github.com/facebook/react-native/issues/19569

detox/local-cli/detox-build.js
```diff
@@ -2,14 +2,17 @@
...
+const {getConfigurationFile, getDefaultConfigurationFile} = require('../src/utils/getConfigurationFile');
 const path = require('path');
 const cp = require('child_process');
 program.description(`[convenience method] run the command defined in 'configuration.build'`)
-       .option('-c, --configuration [device configuration]', 'Select a device configuration from your defined configurations,'
-                                                             + 'if not supplied, and there\'s only one configuration, detox will default to it')
+  .option('--config-path [configPath]',
+    'Select a device config-file path, if not supplied, detox will default to the package.json, and if not found there, detox will fallback to .detoxrc')
+  .option('-c, --configuration [device configuration]', 'Select a device configuration from your defined configurations,' +
+    'if not supplied, and there\'s only one configuration, detox will default to it', getDefaultConfigurationFile())
```
noomorph: @oreporan , why `getDefaultConfigurationFile()`? Come on, it is a string like `android.emu.release`, `ios.sim.debug`, why require file? It is the second time I write it to you.

detox/src/utils/getConfigurationFile.js
```diff
@@ -0,0 +1,26 @@
+const path = require('path');
+const packageJson = 'package.json';
+const detoxRc = '.detoxrc';
+
+function requireWithRoot(suffix) {
+  return require(path.resolve(suffix))
+}
+
+function getConfigurationFile(configPath) {
```
noomorph: Regarding the testing question, I struggle to see what's the issue here :confused: especially if you rewrite it in an OOP way, something like this:

```js
class ConfigurationResolver {
  constructor(cwd = process.cwd()) {
    this.cwd = cwd;
  }

  getDetoxConfiguration(configPath) {
    if (configPath) {
      return this.loadConfiguration(configPath);
    }

    const { detox } = this.loadConfiguration('package.json');
    if (detox) {
      return detox;
    }

    return this.loadConfiguration('.detoxrc');
  }

  loadConfiguration(configPath) {
    return require(path.resolve(this.cwd, configPath));
  }
}

ConfigurationResolver.default = new ConfigurationResolver();

module.exports = ConfigurationResolver;
```

You can instantiate it with any `cwd` directory and test against it.

detox/local-cli/detox-build.js
```diff
@@ -2,14 +2,17 @@
...
 const _ = require('lodash');
 const program = require('commander');
+const {getConfigurationFile, getDefaultConfigurationFile} = require('../src/utils/getConfigurationFile');
 const path = require('path');
 const cp = require('child_process');
 program.description(`[convenience method] run the command defined in 'configuration.build'`)
-       .option('-c, --configuration [device configuration]', 'Select a device configuration from your defined configurations,'
-                                                             + 'if not supplied, and there\'s only one configuration, detox will default to it')
+  .option('--config-path [configPath]',
+    'Select a device config-file path, if not supplied, detox will default to the package.json, and if not found there, detox will fallback to .detoxrc')
```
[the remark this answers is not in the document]
oreporan: Currently the logic I made was `config-file -> package.json -> detoxrc`
You're suggesting to change it to
`config-file ->  detoxrc -> package.json` ?

detox/src/utils/getConfigurationFile.js
oreporan: For the tests, I was referring to using the current testing framework which initializes detox (using `package.json` config) and runs the tests against it.
You're speaking about a unit test?

oreporan:
> @oreporan , you also need to think about the fact the in most of existing Detox configurations there's a critical place where still users load a configuration from `package.json`. See an example:
> 
> https://github.com/wix/Detox/blob/master/examples/demo-react-native/e2e/init.js#L2

Yes I'm aware of that, when testing locally I had to change that as well.
The issue here as I see it as that you are "initializing" detox in 2 different locations. Once in CLI and once in code

noomorph:
@oreporan, exactly! And you need to solve this problem too in your pull request.

oreporan:
> @oreporan, exactly! And you need to solve this problem too in your pull request.

In the fork I've been using I just wrote the path to the correct config. Isn't that enough? You want to implement the same fallback logic in the `init()` method you mean?

noomorph:
@oreporan , yes, something like that. Maybe, at the end of the day, detox does not need arguments in `detox.init()` at all, if you are able to figure out that from the launch args (see `detox/src/utils/argparse.js`)

oreporan:
@noomorph 
I think removing the param from `init()` would be a breaking change wouldn't it?
Because if a user *does* choose to put the param, should it be stronger than the CLI command? should it throw an error?

Maybe this is a separate PR?

detox/src/utils/ConfigurationResolver.test.js
```diff
@@ -0,0 +1,57 @@
+const fs = require('fs');
```
noomorph: You know... since the number of possible run environments is limited to 4, it would make sense to create four folders in `__mocks__/configuration-resolver-it/`, e.g.:

* `__mocks__/configuration-resolver-it/detoxrc-and-package/`
* `__mocks__/configuration-resolver-it/detoxrc/`
* `__mocks__/configuration-resolver-it/package/`
* `__mocks__/configuration-resolver-it/blank/`

I think tests would be more readable then, with less hassle about creating the environment dynamically.
Also, there are no tests for unhappy paths when a resolved file is not found or is malformed.

detox/src/utils/ConfigurationResolver.js
```diff
@@ -0,0 +1,39 @@
...
+
+    const { detox } = this.loadConfiguration(packageJson);
+    if (detox) {
+      return detox;
+    }
+    return this.loadDetoxrcConfiguration()
+  }
+
+  loadDetoxrcConfiguration() {
+    var data = fs.readFileSync(this.resolvePath(detoxRc)).toString();
```
noomorph: Is there a specific consideration of why you want to limit `.detoxrc` to JSON format only?
Personally, I'd love to be able to evaluate it as JS, I'm confident that some people would find how to leverage that for their purposes.

noomorph:
@oreporan , I think if you make this place less stricter...

https://github.com/wix/Detox/blob/master/detox/src/index.js#L41

...then it won't quite be a breaking change, rather a new feature.

You can use `argparse.getArgValue('config-path')` and if it is not there, there is still `ConfigurationResolver` who is smart enough to figure out it needs to be `package.json`, since you have already implemented in that class the fallback logic to load `.detoxrc` or `package.json`.

noomorph:
@oreporan, because, otherwise... what is the sense in this pull request - without config resolution on the test runner side? I don't see scenarios where it can be helpful while being half-baked... if I am correct, then it would just always get overridden by what's in `package.json.`

As for the situation where the `detox.init()` is called with a hardcoded config, it should be more or less trivial - you load a configuration specified in `--config-path` and if it is not equal to `config` passed into the `detox.init(config)`, you print an error that helps the user to figure out that if they want to use `.detoxrc` or `--config-path` feature, then they should remove their `detox.init(packageJson.detox, settings)` in favor of `detox.init(null, settings)` to enable automatic config resolution.

detox/local-cli/detox-build.js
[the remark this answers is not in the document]
rotemmiz: @noomorph, the order you suggest seems reasonable, seems like it follows the standard (AFAIK), [Babel for example](https://babeljs.io/docs/en/config-files).
I agree with your statement, it will probably (hopefully) won't break anything.

detox/src/utils/ConfigurationResolver.js
rotemmiz: So what do you suggest ? this can (and should) be separated and added in a later PR IMO.

detox/local-cli/detox-build.js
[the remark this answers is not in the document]
noomorph: @oreporan, go for it.

detox/src/utils/ConfigurationResolver.js
noomorph: @rotemmiz , I suggest using plain `require()` instead on `JSON.parse(fs.readFileSync...`

rotemmiz:
Hey! Can we wake this PR up? I want to see it passes CI so we can accept it! There's so much work here, would be shame if it goes to waste.

noomorph:
@rotemmiz, I like this PR as well, but it's not finished. There's still detox.init() rewrite pending.

I'm wondering if I might continue the work to complete it. :confused:

oreporan:
sorry, been busy - didn't have much time to work on it

jgreen210:
Improving detox's configurability would be very useful. Having to hard code things in package.json is often inconvenient.

rotemmiz:
Reopening.
@oreporan do you need help finishing this?

LeoNatan:
This seems dead and there is another PR (#1238) on this subject. Closing

Issue #615, opened 2018-03-09
detox.init doesn't resolve in hybrid apps with initial native page (until you manually navigate to a react native page)


etaiklein:
# Description

While running detox in my hybrid app, detox gets stuck on init before timing out. [last log is `detox verb ws send: {"type":"isReady","params":{},"messageId":-1000}`]

The hybrid app has native login, but most of the app is in react native. If I rush through login and get to the react native part of the app then Detox runs smoothly.

# Setup

Setup the same as described in https://github.com/wix/detox/blob/master/docs/Introduction.GettingStarted.md

### init.js

    const detox = require('detox');
    const config = require('../package.json').detox;
    
    before(async () => {
      console.log("************************* CONSOLE LOG ASYNC BEFORE *************************");
      await detox.init(config);
      console.log("************************* CONSOLE LOG ASYNC END BEFORE *************************");
    });
    
    after(async () => {
      await detox.cleanup();
    });

As you can see in the log, it never reaches 'CONSOLE LOG ASYNC END BEFORE' (unless I skip through the ios section of the app quickly)

### Package.json

      "dependencies": {
        ...
        "mocha": "5.0.4",
        "detox": "7.1.0",
        ...
      },
      "detox": {
        "configurations": {
          "ios.sim.debug": {
            "binaryPath": "member/ios/build/Build/Products/Debug-iphonesimulator/MyappMember.app",
            "build": "xcodebuild -workspace member/ios/MyappMember.xcworkspace -scheme MyappMember -configuration Debug  -sdk iphonesimulator -derivedDataPath member/ios/build",
            "type": "ios.simulator",
            "name": "iPhone 7"
          }
        }
      },

# Detox, Node, Device, Xcode and macOS Versions

* Detox: 7.1.0
* Node: 8.9.4
* Device: Iphone 7 simulator
* Xcode: 9.2
* macOS: 10.12.6

# Device and verbose Detox logs

### verbose Detox Logs

    Bug Crusher:mobile etai$ detox test -l verbose
    node_modules/.bin/mocha e2e --opts e2e/mocha.opts  --loglevel verbose
    
    
    ************************* CONSOLE LOG ASYNC BEFORE *************************
    detox-server info 16:50:24: server listening on localhost:57324...
    detox verb ws onOpen [object Object]
    detox verb ws send: {"type":"login","params":{"sessionId":"417aebb8-870c-5798-827c-de0301e7cf17","role":"tester"},"messageId":0}
[...]

PR #931, opened 2018-09-10
Report iOS app readiness right after launch

Ilushkanama:
When app launches, do not wait for app to become idle to report that app is ready. Readiness does not mean that app should be idle.

### DIFF
--- a/detox/ios/Detox/DetoxManager.m
+++ b/detox/ios/Detox/DetoxManager.m
@@ -101,10 +101,8 @@ static void detoxConditionalInit()
 
 - (void)_appDidLaunch:(NSNotification*)note
 {
-	[EarlGrey detox_safeExecuteSync:^{
 		self.isReady = YES;
 		[self _sendGeneralReadyMessage];
-	}];
 }
 
 - (void)_sendGeneralReadyMessage


LeoNatan:
I don’t understand, why is this change submitted? From out investigation, this was necessary, and why we have it.

Ilushkanama:
@LeoNatan having a hybrid app with native first screen, we don't instantiate RN context immediately. That means `WXRNLoadIdlingResource` doesn't become idle until we show screen with RN content, and the app doesn't report that it's ready until that. But we have no chance to advance from the native start screen to the RN screen because we can't run any test until the app is ready. It's like a deadlock.
And in the mentioned case, the app is ready to accept messages from Detox server, just it's `currentStatus` is not "idle". Anything that requires app to be idle will still not be executed until the app actually becomes idle.

LeoNatan:
This doesn’t sound right to me. `WXRNLoadIdlingResource` only becomes busy if a bundle starts loading. You can check the implementation yourself. I fear your problem lies elsewhere. How did you determine it comes from `WXRNLoadIdlingResource`?

LeoNatan:
I’m closing this PR. If you see an issue in idling resources, please open an issue.

Ilushkanama:
I see that `grey_busyResourcesReturnEarly:` returns set containing `WXRNLoadIdlingResource`. Changing notification name in `waitForReactNativeLoadWithCompletionHandler:` from `RCTContentDidAppearNotification` to `RCTJavaScriptDidLoadNotification` resolves the issue also, as long as we load RN bundle on start but don't display any RN content on first screen. Straightforward readiness report without waiting for app to become idle (which I submitted in this PR) works too.
I think the same problem was described in #615, but that issue is closed and the problem seems to remain relevant. I also think that I should've opened another issue instead of PR submission.
@LeoNatan what do you think about the two possible fixes I described? Maybe it's a good idea to choose RN notification name in `ReactNativeSupport` depending on type of application (we can specify app type in `package.json`).

Issue #401, opened 2017-11-13
Cannot click 'Done' on keyboard


joszo:
#### Description
Cannot click 'Done' on keyboard

#### Steps to Reproduce
After running tests I cannot move further because I can't click 'Done' button
I've tried https://github.com/wix/detox/issues/209
`\n` opened keyboard with pause / wait option. 

```
 it('I want to Login', async () => {
        profileTabButton = element(by.id('profileTabButton'));
        singUp = element(by.text('Get started'));
        countryDropDown = element(by.id('country-dropdown-content'));
        countryChoose = element(by.text('+47'));
        phoneNumberField = element(by.type('RCTTextField'));

        await expect(profileTabButton).toBeVisible();
        await profileTabButton.tap();
        await expect(singUp).toBeVisible();
        await singUp.tap();
        await expect(countryDropDown).toBeVisible();
        await countryDropDown.tap();
        await countryChoose.tap();
        await phoneNumberField.tap();
        await phoneNumberField.typeText('10001234');
        await phoneNumberField.typeText('\n');
    });
})

```
#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 5.93
* Node:  7.10.1
* Device: Iphone 6
* Xcode: 9.0.0
* macOS: 10.12.6
#### Logs:

```
     Error: Error: Cannot find UI element.
Exception with Action: {
  "Action Name" : "Type '
'",
  "Element Matcher" : "((kindOfClass('RCTTextField') && !(kindOfClass('RCTScrollView'))) || (kindOfClass('UIScrollView') && ((kindOfClass('UIView') || respondsToSelector(accessibilityContainer)) && ancestorThatMatches((kindOfClass('RCTTextField') && kindOfClass('RCTScrollView'))))))",
  "Recovery Suggestion" : "Check if the element exists in the UI hierarchy printed below. If it exists, adjust the matcher so that it accurately matches element."
}

Error Trace: [
  {
    "Description" : "Failed to type string '
', because key [K] could not be found on the keyboard.",
    "Description Glossary" :     {
      "K" : "more, letters"
    },
    "Error Domain" : "com.google.earlgrey.ElementInteractionErrorDomain",
    "Error Code" : "0",
    "File Name" : "GREYKeyboard.m",
    "Function Name" : "+[GREYKeyboard grey_setErrorForkeyNotFoundWithAccessibilityLabel:forTypingString:error:]",
    "Line" : "494"
  }
]
[...]

Issue #209, opened 2017-07-21
Is it possible to send "Done" or "Go" to the keyboard?


dtuite:
Is it possible to use `typeText` to send keyboard events other than typing letters into forms?

rotemmiz:
Detox uses EarlGrey to interact with the UI (among other things). Take a look at these examples here, https://github.com/google/EarlGrey/issues/248

In brief, you can typeText `\n` or match `by.label('return')` and tap on the matched view.

rotemmiz:
Closed by mistake, please close if this answers your question.

dtuite:
That's great! Thank you.

heuism:
@rotemmiz, sorry for reopen this

I just got a slight difficulty for pressing enter on a numpad keyboard. It doesn't have Enter keyboard so when I wanna press Enter, it just tell me that
`GREYKeyboard: No known SHIFT key was found in the hierarchy.`

Thanks for helping out.

krazykira:
If you have a login screen then use `"\n"` at the end of the field you want the keyboard to close.

e.g:

```
    await element(by.id('UsernameTestId')).typeText('username');
    await element(by.id('PasswordTestId')).typeText('password\n');
    await element(by.id('LoginTestId')).tap();
```

PR #956, opened 2018-09-28
feat: expose dismissKeyboard API from EarlGrey

karanjthakkar:
The EarlGrey implementation has had the api to dismiss keyboard available for a very long time: https://github.com/google/EarlGrey/pull/475

This should provide an alternative for folks using detox to dismiss keyboard instead of these alternatives: [Use `by.label('Done')`](https://github.com/wix/Detox/issues/401#issuecomment-344209097), [Type `\n` in last text field](https://github.com/wix/Detox/issues/209#issuecomment-317201130 )

PS: Have been using it in our fork here: https://github.com/Skyscanner/detox/commit/e1bb5414f8766dbe5992f9f22fb280c3bb0fcdfb for a few days and it seems to be working fine 🙂 If that helps.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -166,6 +166,10 @@ class Device {
     await this.deviceDriver.setOrientation(this._deviceId, orientation);
   }
 
+  async dismissKeyboard() {
+    await this.deviceDriver.dismissKeyboard(this._deviceId);
+  }
+
   async setLocation(lat, lon) {
     lat = String(lat);
     lon = String(lon);
--- a/detox/src/devices/Device.test.js
+++ b/detox/src/devices/Device.test.js
@@ -324,6 +324,13 @@ describe('Device', () => {
     expect(device.deviceDriver.setOrientation).toHaveBeenCalledWith(device._deviceId, 'param');
   });
 
+  it(`dismissKeyboard() should pass to device driver`, async () => {
+    device = validDevice();
+    await device.dismissKeyboard();
+
+    expect(device.deviceDriver.dismissKeyboard).toHaveBeenCalledWith(device._deviceId);
+  });
+
   it(`sendUserNotification() should pass to device driver`, async () => {
     device = validDevice();
     await device.sendUserNotification('notif');
--- a/detox/src/devices/drivers/DeviceDriverBase.js
+++ b/detox/src/devices/drivers/DeviceDriverBase.js
@@ -99,6 +99,10 @@ class DeviceDriverBase {
     return await Promise.resolve('');
   }
 
+  async dismissKeyboard(deviceId) {
+    return await Promise.resolve('');
+  }
+
   async setURLBlacklist(urlList) {
     return await Promise.resolve('');
   }
--- a/detox/src/devices/drivers/IosDriver.js
+++ b/detox/src/devices/drivers/IosDriver.js
@@ -61,6 +61,11 @@ class IosDriver extends DeviceDriverBase {
     await this.client.execute(call);
   }
 
+  async dismissKeyboard(deviceId) {
+    const call = EarlyGreyImpl.dismissKeyboardWithError(invoke.EarlGrey.instance);
+    await this.client.execute(call);
+  }
+
   defaultLaunchArgsPrefix() {
     return '-';
   }
--- a/detox/src/devices/ios/AppleSimUtils.js
+++ b/detox/src/devices/ios/AppleSimUtils.js
@@ -106,7 +106,7 @@ class AppleSimUtils {
       trying: `Installing ${absPath}...`,
       successful: `${absPath} installed`
     };
-    await this._execSimctl({ cmd: `install ${udid} "${absPath}"`, statusLogs });
+    await this._execSimctl({ cmd: `install ${udid} "${absPath}"`, statusLogs, retries: 2 });
   }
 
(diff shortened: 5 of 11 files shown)


rotemmiz:
Hey,

Thanks for the contribution!

I am wondering about this API, what is the equivalent action a user can make to dismiss a keyboard? The action imagine are: Pressing done, defocussing a text field or moving to the next line.

Also, you're missing an e2e test suite that will actually test this feature. Tell us if you need any help with it.

LeoNatan:
I too have doubts whether such an API should exist. It is best to test by dismissing the way you expect the user to dismiss the keyboard, rather than dismissing it programatically.

LeoNatan:
OK, let's have the option.

LeoNatan:
One thing to see how the API behaves in the context of Detox when a keyboard is not shown and users calls the API.

karanjthakkar:
@LeoNatan This is the error I get when I try to dismiss a hidden keyboard:

screen shot 2018-09-28 at 15 11 14

Looks good?

LeoNatan:
Yes, looks good.

karanjthakkar:
@LeoNatan I've added tests and updated docs. I'm not sure when the CI runs the tests for this PR but they run successfully locally. Let me know if you want me to make any more changes 🙂

yershalom:
Hi @karanjthakkar ,
Can you please push some dummy commit to trigger the CI?

karanjthakkar:
@yershalom done 🙂 was the failing test previously a fluke or did I miss something? 🤔

yershalom:
If you want to set test only for specific test just add `:ios:` in the test description.
For example: https://github.com/wix/Detox/blob/master/detox/test/e2e/02.matchers.test.js#L33

karanjthakkar:
Can someone pls check the CI? There seem to be really unrelated failures going on everytime I run it.

rotemmiz:
I am still wondering regarding this API, as it doesn't really represent a user action, but it seems like I'm in the minority group here, so let's go ahead and add it. 
There are couple of things though I still see failing (rightfully) in CI:

1. The last test which fails on all builds, is running sanity on the creation of artifacts. Detox test suite is set to generate artifacts for every test, and if a test is added, the snapshot must be updated to include the artifacts from the new test.
2. the specific e2e test actually fails in the latest build on [RN51 iOS](https://jenkins-oss.wixpress.com/job/detox-ios-51-pr/629/console) although it passes on [RN56 iOS](https://jenkins-oss.wixpress.com/job/detox-ios-56-pr/629/console). This probably needs attention.
```
17:02:36 FAIL ./20.device-dismissKeyboard.test.js (8.58s)
17:02:36   ● Device DismissKeyboard › :ios: should dismiss keyboard
17:02:36 
17:02:36     Error: Cannot perform action due to constraint(s) failure.
17:02:36     Exception with Action: {
17:02:36       "Action Name":  "Tap",
17:02:36       "Element Description":  "<RCTUITextField:0x7fe7e9005600; AX=Y; AX.id='testInput'; AX.value='Test Input'; AX.frame={{187.66666666666666, 60}, {0, 17}}; AX.activationPoint={187.66666666666666, 68.5}; AX.traits='UIAccessibilityTraitNone'; AX.focused='N'; frame={{0, 0}, {0, 17}}; alpha=1; text='Test Input'>",
17:02:36       "Failed Constraint(s)":  "interactable",
17:02:36       "All Constraint(s)":  "(!(isSystemAlertViewShown) && ((respondsToSelector(isAccessibilityElement) && isAccessibilityElement) || kindOfClass('UIView')) && (enabled && !(((kindOfClass('UIView') || respondsToSelector(accessibilityContainer)) && ancestorThatMatches(!(enabled))))) && interactable)",
17:02:36       "Recovery Suggestion":  "Adjust element properties so that it matches the failed constraint(s)."
17:02:36     }
17:02:36     
17:02:36     
17:02:36     [
17:02:36       {
17:02:36         "Description":  "Cannot perform action due to constraint(s) failure.",
17:02:36         "Error Domain":  "com.google.earlgrey.ElementInteractionErrorDomain",
17:02:36         "Error Code":  "1",
17:02:36         "File Name":  "GREYBaseAction.m",
17:02:36         "Function Name":  "-[GREYBaseAction satisfiesConstraintsForElement:error:]",
17:02:36         "Line":  "66"
17:02:36       }
17:02:36     ]
```

The best way of running it locally is just changing the RN version on the test project to 51 and running the e2e suite again. This script can handle this + changing to the correct react version:
`node scripts/change_react_native_version.js "detox/test" ${REACT_NATIVE_VERSION}`

LeoNatan:
After internal discussion, this is not a direction we would like to take. Such an API encourages writing tests that diverge from actual user interaction flows. If a user should tap on the screen to dismiss, the test should be `element(by.id('background')).tap()` or pressing the return key on the keyboard.

I agree that `by.label('Done')` and typing `\n` in text fields aren't perfect, and should be improved. Would you like to discuss this and perhaps help us with a PR? I was thinking of perhaps an API such as `await element(by.id('PasswordTestId')).tapEnterKey();` or something similar that would internally type `\n` to dismiss the keyboard. What do you think?

noomorph:
Just $0.02 from me, on Android, I think it is technically possible to dismiss keyboard as a user. Sorry for commenting into the closed thread.

LeoNatan:
Is this standard or is it vendor dependent?

PR #975, opened 2018-10-09
Change regex to allow matching whitespace of adb list

bltavares:


noomorph:
@bltavares , what is the use case? How can I reproduce the issue?

bltavares:
@noomorph Hi, this was a PR opened by mistake. The issue was regarding having an extra `^M` whitespace after the bundle id, which caused the end of line matcher to fail. This seems to be fixed already on the latest version, but we haven't updated yet.

My intention was to open a patch fix for an older version on our internal repo, while we can't update the dependency. Sorry to bother with this notification

PR #982, opened 2018-10-13
Fix #981 device.launchApp launchArgs not passed to MainActivity on Android

### LINKED ISSUE #981: device.launchApp launchArgs not passed to MainActivity on Android
#### Description
When using `device.launchApp` passing `launchArgs` this are received by the Instrumentation Activity (the DetoxTest) but not passed down to the actual Android app being run (the MainActivity).

I understand the rationale behind closing Issue #530. However, this is a completely different thing. Using iOS the launchArgs are received by the App being tested, but in Android detox swallows the launchArgs on the test runner and are not being received by the MainActivity, in which even if you have the logic to handle command line arguments they won't show up.

#### Steps to Reproduce
Create a test that launches the app:

```
  it('Receives command line arguments', async () => {
    await device.launchApp({
      newInstance: true,
      launchArgs: { name: 'world' },
    })
    await expect(element(by.text('Hello world!!!'))).toBeVisible()
  })
```

Handle the command line arguments on the MainActivity (in this case pass as initialProps):

```
package com.test.detoxCommandLineArgs;

import android.os.Bundle;
import com.facebook.react.ReactActivity;
import com.facebook.react.ReactActivityDelegate;

public class MainActivity extends ReactActivity {

    /**
     * Sets the launch options for the React View
     */
    @Override
    protected ReactActivityDelegate createReactActivityDelegate() {
        return new ReactActivityDelegate(this, getMainComponentName()) {
            @Override
            protected Bundle getLaunchOptions() {
                return getIntent().getExtras();
            }
        };
    }

    /**
     * Returns the name of the main component registered from JavaScript.
     * This is used to schedule rendering of the component.
     */
    @Override
    protected String getMainComponentName() {
        return "DetoxTestCommandLineArguments";
    }
}
```

#### Expected behavior
In the example above the app should receive the launchArgs as initial props. It works when launching the app from the command line passing the values as extras, but it doesn't when running the test.

elyalvarado:
Fix example DetoxTest.java to pass launch arguments to activity being tested by cloning the command line arguments used when calling the test to the MainActivity Intent created in the test.

### DIFF
--- a/detox/test/ios/example/AppDelegate.m
+++ b/detox/test/ios/example/AppDelegate.m
@@ -136,10 +136,21 @@ RCT_EXPORT_MODULE();
 #else
 	jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
 #endif
-	
+
+  // Parse process arguments into a launchArgs initialProp that is passed to the react root view
+  NSMutableArray *arguments   = [[NSMutableArray alloc]initWithArray:[[NSProcessInfo processInfo] arguments]];
+  [arguments removeObjectAtIndex:0]; // Remove executable path
+  NSMutableDictionary *initialProps = [NSMutableDictionary dictionary];
+  if([arguments count] > 0) {
+    NSString *allArgs = [arguments componentsJoinedByString:@" "];
+    initialProps[@"launchArgs"] = allArgs;
+  } else {
+    initialProps = NULL;
+  }
+
 	RCTRootView *rootView = [[RCTRootView alloc] initWithBundleURL:jsCodeLocation
 														moduleName:@"example"
-												 initialProperties:nil
+												 initialProperties:initialProps
 													 launchOptions:launchOptions];
 	rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];
 	
--- a/detox/test/android/app/src/androidTest/java/com/example/DetoxTest.java
+++ b/detox/test/android/app/src/androidTest/java/com/example/DetoxTest.java
@@ -1,5 +1,8 @@
 package com.example;
 
+import android.content.Context;
+import android.content.Intent;
+import android.support.test.InstrumentationRegistry;
 import android.support.test.filters.LargeTest;
 import android.support.test.rule.ActivityTestRule;
 import android.support.test.runner.AndroidJUnit4;
@@ -19,7 +22,19 @@ import org.junit.runner.RunWith;
 public class DetoxTest {
 
     @Rule
-    public ActivityTestRule<MainActivity> mActivityRule = new ActivityTestRule<>(MainActivity.class, false, false);
+    public ActivityTestRule<MainActivity> mActivityRule =
+        new ActivityTestRule<MainActivity>(MainActivity.class, false, false) {
+            @Override
+            protected Intent getActivityIntent() {
+                Context targetContext = InstrumentationRegistry.getInstrumentation().getTargetContext();
+                Intent result = new Intent(targetContext, MainActivity.class);
+
+                // Get the arguments passed to the Test Runner, and forward them to the MainActivity Intent
+                // so that they can be received by the app
+                result.putExtras(InstrumentationRegistry.getArguments());
+                return result;
+            }
+        };
 
     @Test
     public void runDetoxTests() {
--- a/detox/test/android/app/src/main/java/com/example/MainActivity.java
+++ b/detox/test/android/app/src/main/java/com/example/MainActivity.java
@@ -1,8 +1,28 @@
 package com.example;
 
+import android.os.Bundle;
 import com.facebook.react.ReactActivity;
+import com.facebook.react.ReactActivityDelegate;
 
 public class MainActivity extends ReactActivity {
+
+    /**
+     * Sets the launch options for the React View
+     */
+    @Override
+    protected ReactActivityDelegate createReactActivityDelegate() {
+        return new ReactActivityDelegate(this, getMainComponentName()) {
+            @Override
+            protected Bundle getLaunchOptions() {
+                String launchArgs = getLaunchArgs();
+                Bundle bundle = new Bundle();
+                if(launchArgs != null) {
+                    bundle.putString("launchArgs", launchArgs);
+                }
+                return bundle;
+            }
+        };
+    }
 
     /**
      * Returns the name of the main component registered from JavaScript.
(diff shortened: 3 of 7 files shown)


noomorph:
@elyalvarado, thanks for the pull request. Can you add a failing e2e test that gets fixed with your change?

elyalvarado:
@noomorph I'll try to add a failing test, it'll require some changes on the MainActivity and on the AppDelegate to handle the launch args, but I think is worth it because it'll also serve as an example case for people asking on how to handle launch args for both platforms.

rotemmiz:
@elyalvarado, do you think you're able to make sure the Android  launch args e2e test passes here?

elyalvarado:
@rotemmiz, yes of course, if there is an existing test just point me to it and I'll make it pass with this PR. I just haven't had the time to create a test myself, but I'm using this PR in my code to use launch args to successfully pass args during e2e tests in both platforms, like:

- logout the currently logged user
- login a specific user
- skip intro screens
- others...

rotemmiz:
I went over the test suite and found out that we actually do not have tests for this, not even for iOS :(
It won't be fair to burden you with completing untested scenario for iOS, but it would be super if you could add an e2e test at least for Android. Tests should probably go here:
https://github.com/wix/Detox/blob/master/detox/test/e2e/06.device.test.js

How do you consume the launch args in your code? via a custom native module?

elyalvarado:
I parse them on the native side as a string and pass them to the React side as an initial prop.

I'll amend this commit with the e2e tests for both, as its kind of the same test and on iOS is trivial as it works out-of-the-box. But It'll also require some changes to the `AppDelegate` code in iOS and the `MainActivity` on the Java side to pass the args as an initial prop.

elyalvarado:
Added the e2e tests for `launchArgs`, some notes here:

- In order to be able to test that the app is receiving the arguments, I'm forwarding the received arguments as a `launchArgs` initial prop to the react native app.
- The point above required changing both the `AppDelegate` and the `MainActivity`
- As expected iOS and Android handle arguments in different ways, so in order to be consistent I'm transforming the `extras` passed to the `MainActivity` in Android to a plain string adding a `-` (dash) for each argument name, followed by the argument. This is kind of cheating as it is an implementation detail best left to each particular app, but otherwise the feature can't be tested.
- The original change on the `DetoxTest.java` is still there, and is basically forwarding the arguments as extras to the MainActivity Intent. I decided on using `extras` as it is what can be also used from the command line according to the docs [(adb IntentSpec)](https://developer.android.com/studio/command-line/adb#IntentSpec) .
- For the e2e tests I added a new screen to the app and modified the `renderScreenButton` to allow passing props down to any screen.

Let me know if you need any further changes, corrections, docs, or anything else regarding this change.

detox/test/ios/example/AppDelegate.m
```diff
@@ -136,10 +136,21 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    NSString *allArgs = [arguments componentsJoinedByString:@" "];
+    initialProps[@"launchArgs"] = allArgs;
+  } else {
+    initialProps = NULL;
+  }
+
 	RCTRootView *rootView = [[RCTRootView alloc] initWithBundleURL:jsCodeLocation
 														moduleName:@"example"
-												 initialProperties:nil
+												 initialProperties:initialProps
```
LeoNatan: I am sorry, I don't understand this change. Launch arguments are not initial props. Why was this code added here? 

Launch arguments are process launch argument.

LeoNatan:
I don't understand the changes to the iOS portion. Launch arguments is not something that should be passed as props to JS. Adding that logic to AppDelegate makes no sense, as normally on iOS, launch arguments are controlled by the operating system and the user has no way of passing custom arguments. I have a feeling it is the same in Android.

LeoNatan:
Also, requiring the user to do this as part of their AppDelegate/<whatever analogous in Android> just for Detox doesn't make sense to me. The framework should do this automatically, if indeed we agree that this information should be passed to the app.

elyalvarado:
@LeoNatan both of your observations are true. Launch arguments are not initial props, and it'll be wrong to require the user to do it as part of AppDelegate, or in general app initialization, as part of using Detox. 

That is why the changes are to the example app used for testing Detox itself, and as the testing framework supports launching with initial arguments, an implementation is needed to test it. And even though the equivalence between launch arguments and initial props might be discussed, I couldn't find any other primitive to pass data between native and react without resorting to writing a native module.  If you have any suggestions on how to test it, I'll be eager to hear about it.

Keep in mind that the initial issue referenced here ( #981 ) is that the arguments are not being passed to the main activity on Android, which is an inconsistency with what happens in iOS, and this are the changes to the `DetoxTest.java` file, which is also just a reference implementation being pointed to in the documentation, so the users are free to change it to suit their needs.

In any case, the changes to the `MainActivity.java` and `AppDelegate.m` files are just there to pass down the arguments as initial props for testing purposes, and are just one of many ways to implement argument passing to the react side. This files are not pointed to anywhere in the docs as a step to add Detox to a project, nor Detox itself will stop working without these changes.

LeoNatan:
> Keep in mind that the initial issue referenced here ( #981 ) is that the arguments are not being passed to the main activity on Android, which is an inconsistency with what happens in iOS, 

We don't pass launch arguments to React Native on iOS. I'm not sure why you say this. One difference might be that iOS has support for launch arguments through its Unix heritage, which Android does not expose from the process itself, so this has to be passed from the test case.

Now correct me if I'm wrong, but both changes are to the example projects and are not a real solutions for Android. I think it's better to have those documented in the documentation rather than have code parsing all launch arguments. We use the launch arguments to pass information to Detox, and such parsing could cause parse issues which shouldn't exist in the first place in the demo projects.

I also do not think launch arguments is the correct way to communicate with your React Native side. [Mocking](https://github.com/wix/Detox/blob/master/docs/Guide.Mocking.md) is a better solution. On native, at least on iOS, the parsing system is much more robust than `_parseLaunchArgs` function supplied with the PR.

On iOS (or macOS), you would normally never access the `arguments` array directly, but instead use either the system's parsed arguments (through `NSUserDefaults` API) or use a dedicated parser. I don't think this is something we want to support here, as Mocking is just a better method.

I am inclined to close this PR. @rotemmiz

LeoNatan:
The leaking itself is not the issue, but the parsing responsibility. It's a complexity I am not sure we should have in our demo project.

I agree that not everything is absolute, and sometimes launch arguments are necessary. But it should be up to the developer to implement it in their own app, taking the risks. That's why I prefer the documentation way, with a statement added to warn users that the launch arguments include Detox-internal key-values and should be parsed with care.

elyalvarado:
I like both suggestions and the guidance on where to make the changes is spot on. I'll take a look at the proposed changes as soon as I have some time.

> I still think this PR is an important one, and I would like to see it land in master, as this functionality has many valid use cases (and we ought to test it properly as well).
> 
> I still think though, that the implementation should be improved. Added some thoughts in comments throughout the code.
> 
> Sorry for the delayed response on this one :(

rotemmiz:
Hey @elyalvarado, do you think you'll be able to complete this one?

elyalvarado:
Hi @rotemmiz I saw there were some changes in the Android code for Detox v11, and I haven't upgraded yet because there are some issues still with React Native 0.59. I have some free time next week, and I'll take a look and let you know if I can tackle this.

LeoNatan:
Closing this PR as no longer relevant. See #1385

PR #986, opened 2018-10-17
Update mocking guide for RN >= 0.57

elyalvarado:
Metro versions >= v0.43 have a new configuration format. By default RN >= 0.57 need to use this new configuration for rn-cli.config.js otherwise it won't work.

This PR updates the mocking guide with this new format which is needed to use e2e.js files in RN from version 0.57

### DIFF
--- a/docs/Guide.Mocking.md
+++ b/docs/Guide.Mocking.md
@@ -26,17 +26,7 @@ This replacement mechanism provides a lot of flexibility to change implementatio
 
 #### Configuration
 0. For RN < 0.55, setup `react-native-repackager` in your library.
-1. Configure Metro by creating `rn-cli.config.js` to root dir and setting `getSourceExts()` to prioritize any given source extension over the default one, by triggering each of these commands the bundler will take `e2e.js` over `.js`
-
-    ```js
-    module.exports = {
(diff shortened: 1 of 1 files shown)


docs/Guide.Mocking.md
```diff
@@ -36,6 +36,19 @@ This replacement mechanism provides a lot of flexibility to change implementatio
 
     ```
 
+    or use the new format for metro configuration for RN >= 0.57 (or if you've metro >= 0.43)
```
LeoNatan: "if you've" is not a valid English phrase when "have" has a possessive meaning.

LeoNatan: It doesn't really matter. It will be squashed and merged at the end. Also, please capitalize "Metro".

I also think that going forward, we should have the new configuration be first, rather than "or use this new one".

elyalvarado: Ok, I'll make all the corrections including changing the new configuration to be first

LeoNatan: Thanks. Let's wait for @rotemmiz to review

rotemmiz:
Thanks!

PR #993, opened 2018-10-23
Added required step to example readme

pwfcurry:


rotemmiz:
That shouldn't be the case. The example project is independent from Detox source, it should just get the latest version from npmjs. Did you use lerna at some point in the process? It might have contaminated your env.

pwfcurry:
Looks like we're not doing that:

https://github.com/wix/Detox/blob/master/examples/demo-native-ios/e2e/init.js#L2

I can send a pull request to fix.

rotemmiz:
Good catch, thanks

PR #994, opened 2018-10-23
Reverted relative import of detox

pwfcurry:
Depending on the npm package means we don't need to `npm install` the root detox/ project.


pwfcurry:
Reverts change made in this commit https://github.com/wix/Detox/commit/1e2d9e3e0d2bfceadfb43118b484df9e955bb347#diff-c9dd5f01ac053ec6d63653b49e13f4a8R2

rotemmiz:
Thanks!

Issue #917, opened 2018-08-31
Detox hangs on detox config step with expo iOS


adam-stasiak:
Hi, 
I am running Detox with expo (app standalone, project contains native ios and android modules).
I face with scenario where app is correctly installed but during debugging I discovered that Detox does not go further than await detox.init(config); from init.js file. 
Environment:
```
mac 10.13
=> npm version:
6.1.0
=> node version:
v10.6.0
=> carthage version:
Please update to the latest Carthage version: 0.30.1. You currently are on 0.29.0
0.29.0
=> pod version:
1.5.3
=> exp version:
55.0.5
=> XCode version:
Xcode 9.4.1
Build version 9F2000
```

logs:

```
MacBook-Pro-Adam:mobile-project-final adamstasiak$ detox test -l verbose
node_modules/.bin/mocha e2e --opts e2e/mocha.opts --configuration ios.sim.debug --loglevel verbose     --grep :android: --invert     --artifacts-location "artifacts/ios.sim.debug.2018-08-31 12-19-59Z" 


detox[80418] INFO:  [DetoxServer.js] server listening on localhost:53225...
detox[80418] DEBUG: [AsyncWebSocket.js/WEBSOCKET_OPEN] opened web socket to: ws://localhost:53225
detox[80418] DEBUG: [DetoxServer.js/LOGIN] role=tester, sessionId=8896fd7f-270d-e893-f776-b590aac5fccb
detox[80418] DEBUG: [DetoxServer.js/LOGIN_SUCCESS] role=tester, sessionId=8896fd7f-270d-e893-f776-b590aac5fccb
detox[80418] DEBUG: [exec.js/EXEC_CMD, #0] /usr/bin/xcrun simctl list -j
detox[80418] DEBUG: [exec.js/EXEC_CMD, #1] applesimutils --list --byType "iPhone 8" --byOS "11.4"
detox[80418] DEBUG: [exec.js/EXEC_TRY, #1] Searching for device matching iPhone 8...
detox[80418] DEBUG: [exec.js/EXEC_CMD, #2] applesimutils --list --byId "14D76F42-23F6-4737-9F51-CCD34FD024C4"
detox[80418] DEBUG: [exec.js/EXEC_CMD, #3] /usr/bin/xcrun simctl uninstall 14D76F42-23F6-4737-9F51-CCD34FD024C4 io.project
detox[80418] DEBUG: [exec.js/EXEC_TRY, #3] Uninstalling io.project...
[...]

PR #999, opened 2018-10-26
append instead of clobber SIM_CTL_DYLD_INSERT_LIBRARIES

quinlanj:
Whatever the user specifies in `SIM_CTL_DYLD_INSERT_LIBRARIES` should propagate through and be reflected in the `xcrun simctl launch` call that AppleSimUtils makes.

My use case: I am passing in my own dylibs that I specify in `SIM_CTL_DYLD_INSERT_LIBRARIES`, and I would like this dylib to get used by my app. This will also help me fix https://github.com/wix/detox/issues/917

I append the detox framework path to the env var, separated by a colon, as this is the format expected specified by `man dyld`: 

```
DYLD_INSERT_LIBRARIES
              This is a colon separated list of dynamic libraries to load before the ones specified in the program.  This lets you test new modules of
              existing dynamic shared libraries that are used in flat-namespace images by loading a temporary dynamic shared library with just the new
              modules.   Note  that  this  has  no  effect  on  images  built  a  two-level  namespace  images  using  a dynamic shared library unless
              DYLD_FORCE_FLAT_NAMESPACE is also used.
```

### DIFF
--- a/detox/src/devices/ios/AppleSimUtils.js
+++ b/detox/src/devices/ios/AppleSimUtils.js
@@ -262,7 +262,7 @@ class AppleSimUtils {
       dylibs = `${process.env.SIMCTL_CHILD_DYLD_INSERT_LIBRARIES}:${dylibs}`;
     }
 
-    const launchBin = `/bin/cat /dev/null >${logsInfo.absStdout} 2>${logsInfo.absStderr} && ` +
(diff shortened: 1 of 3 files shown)


LeoNatan:
Thank you for your contributions! Really appreciated!

PR #1000, opened 2018-10-27
add blacklistRegex to GREYConfiguration on init

quinlanj:
Allow the user to set the blacklist for EarlGrey upon initialization of the `DetoxManager`. This will also help fix https://github.com/wix/Detox/issues/917

# Why
When `device.launchApp` is called, it waits for the device to be ready. In order for the device to be ready, all the tracked Earl Grey resources must be idle. Because Earl Grey tracks all network requests by default, if there are any network calls happening, the app will be considered 'busy'. 

Consider the case where an app is making many network calls to an URL we wish to be blacklisted. First, the app must be launched via `device.launchApp`. Next, the blacklist needs to be set via `device.setURLBlacklist`. However, because there is currently no way to set the blacklist upon initialization, Earl Grey will always consider the app to be 'busy' and the call to `device.launchApp` will hang. 

This can be solved if the user is able to pass in their own blacklist array in `launchApp` like this:
```
  await device.launchApp({
    ...
    newInstance: true,
    launchArgs: { detoxURLBlacklistRegex: ' \\("http://192.168.1.253:19001/onchange","https://e.crashlytics.com/spi/v2/events"\\)' },
  }); 
```

### DIFF
--- a/detox/ios/Detox/DetoxManager.m
+++ b/detox/ios/Detox/DetoxManager.m
@@ -45,6 +45,11 @@ static void detoxConditionalInit()
 	
 	NSUserDefaults* options = [NSUserDefaults standardUserDefaults];
 	
+	NSArray *blacklistRegex = [options arrayForKey:@"detoxURLBlacklistRegex"];
+	if (blacklistRegex){
+		[[GREYConfiguration sharedInstance] setValue:blacklistRegex forConfigKey:kGREYConfigKeyURLBlacklistRegex];
+	}
+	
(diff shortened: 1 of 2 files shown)


LeoNatan:
This sounds like a good idea. Thanks for this PR!

LeoNatan:
Do you mind updating the documentation of `launchApp` and `setURLBlacklist` to mention this new feature? Thanks

quinlanj:
@LeoNatan added some docs, lmk if they are ok!

LeoNatan:
Looks great, thank you!

Issue #99, opened 2017-03-16
Add detox definitions


Gongreg:
Having IDE support would really help writing tests. Since we have a lot of people liking typescript, something like DefinitelyTyped would help.

rotemmiz:
Great suggestion! I love it. We'll put it in our plan. 
Alternatively, you can submit a PR, that would be very much appreciated.

TareqElMasri:
I made the definitions, it's documented with examples, I'm waiting for the pull request to get approved, when it gets published, it will be used via `npm install @types/detox`
https://github.com/TareqElMasri/DefinitelyTyped/tree/master/types/detox

tachtevrenidis:
did this PR ever make it in?

DanielMSchmidt:
@tachtevrenidis There are still review comments, but it looks great otherwise 👍

TareqElMasri:
Goodnews everyone, the PR made it successfully, now the maintainers need to create npm package for `@types/detox` using  [Types publisher tool](https://github.com/Microsoft/types-publisher)
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/25005#issuecomment-381774947

rotemmiz:
Hooray! Let's add this to docs and close this issue.

TareqElMasri:
Fellas, it's up and running
https://www.npmjs.com/package/@types/detox

DanielMSchmidt:
Awesome 👍  Let's close this issue then <3

zibs:
Super awesome! Thanks for setting this up;

My only question is how to prevent the detox types from over-riding my jest types? Since detox overrides the jest object, it seems like when I'm trying to just write normal jest tests, TS still thinks the jest object is a detox one and so it complains for all the normal jest matchers/expects etc:

`[ts] Property 'toEqual' does not exist on type 'Expect<any>'.`

It's only small issue, so I don't feel like opening a new issue for it, but just curious if anyone has any ideas.

PR #1001, opened 2018-10-27
Add guide for integrating TypeScript with Detox + Jest

solkaz:
Adds a page in the `docs` folder with instructions on converting a Detox + Jest project to allow for writing the tests with TypeScript. Would love feedback on the guide!

I do have a couple of thoughts:
- Should this just be an extension of the Jest guide, since this is specific to that setup?
- There should probably be a Mocha counterpart for this (I don't use Mocha, however, so this would need to come from the community)

### DIFF
--- a/docs/Guide.TypeScriptIntegrationJest.md
+++ b/docs/Guide.TypeScriptIntegrationJest.md
@@ -84,6 +84,8 @@ Add typings for Detox, Jest, and Jasmine (the latter two are used in `init.ts`),
 npm install --save-dev @types/detox @types/jest @types/jasmine
 ```
 
+Note: [`@types/detox`](https://www.npmjs.com/package/@types/detox) is maintained by the community and not by Wix.
+
 You should now be able to run your Detox tests, written in TypeScript! If you're not writing your unit tests with TypeScript and Jest, you can skip the next section.
 
 ### 5. Unit test collisions
--- a/docs/Guide.TypeScriptIntegrationJest.md
+++ b/docs/Guide.TypeScriptIntegrationJest.md
@@ -1,0 +1,111 @@
+---
+id: Guide.TypeScriptJest
+title: Integrating TypeScript with Detox + Jest
+---
+
+## Usage
+
+### 0. Initialize Detox in your project
+
+Follow the [Getting Started](Introduction.GettingStarted.md) Guide to set up Detox with [Jest as your test runner](Guide.Jest.md) (optionally, set up Detox to [test Android](Introduction.Android.md) as well). When you are finished with that, we will make some changes to your project:
+
+- Remove Detox from `package.json` at the root level
+
+- Create a `package.json` in the `e2e` folder and add Detox and Jest to the project dependencies:
+
+```sh
+npm uninstall detox
+cd e2e
+npm init -y
+npm install -D detox jest
+```
+
+You should then move the settings for Detox from the root-level `package.json` to the `e2e/package.json`, but make sure that you update the `build` and `binaryPath` values for your Detox configurations:
+
+```json
+{
+  // ...
+  "detox": {
+    "configurations": {
+      "ios.sim.debug": {
+        "binaryPath": "../ios/build/Build/Products/Debug-iphonesimulator/YourAppName.app",
+        "build": "xcodebuild -project ../ios/YourAppName.xcodeproj -scheme YourAppName -configuration Debug -sdk iphonesimulator -derivedDataPath ../ios/build",
+        "type": "ios.simulator",
+        "name": "iPhone X"
+      },
+      "android.emu.debug": {
+        "binaryPath": "../android/app/build/outputs/apk/debug/app-debug.apk",
+        "build": "cd ../android; ./gradlew assembleDebug assembleAndroidTest -DtestBuildType=debug; cd ../e2e",
+        "type": "android.emulator",
+        "name": "Pixel_2_API_27"
+      },
+      "android.emu.release": {
+        "binaryPath": "../android/app/build/outputs/apk/release/app-release.apk",
+        "build": "cd ../android; ./gradlew assembleRelease assembleAndroidTest -DtestBuildType=release; cd ../e2e",
+        "type": "android.emulator",
+        "name": "Pixel_2_API_27"
+      }
+    }
+  }
+}
+```
+
+Before moving on, make sure that you can still successfully run your Detox tests.
+
+### 1. Add TypeScript + `ts-jest` to `e2e/package.json`
+
+```sh
+npm install --save-dev typescript ts-jest
+```
+
+### 2. Configure Jest to use `ts-jest`
+
+Modify your Jest configuration (`e2e/config.json` by default) to include the following properties
+
+```json
+{
+  "preset": "ts-jest",
+  "testEnvironment": "node",
+  "setupTestFrameworkScriptFile": "./init.ts"
+}
+```
+
+NB: this is mostly the same output of running `ts-jest config:init`, with `setupTestFrameworkScriptFile` being the only added property.
+
+### 3. `.js` -> `.ts`
+
+Convert all files in the `e2e` directory ending in `.js` to `.ts`
+
+### 4. Add typings for Detox, Jest, and Jasmine
+
+Add typings for Detox, Jest, and Jasmine (the latter two are used in `init.ts`), as well as for other modules that you use in your Detox tests.
+
+```sh
+npm install --save-dev @types/detox @types/jest @types/jasmine
+```
+
+You should now be able to run your Detox tests, written in TypeScript! If you're not writing your unit tests with TypeScript and Jest, you can skip the next section.
+
+### 5. Unit test collisions
+
+Detox overrides certain global members that were defined by Jest (such as `expect`), which means that their typings collide and will cause compiler errors if you try to use them in the same project. This means that we need to set up TypeScript to ignore Detox when compiling our unit tests, which isn't straightforward since the default type-lookup behavior of TypeScript [will automatically include typings from `node_modules` under the root directory](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types) (i.e. `e2e/node_modules`). Therefore, we'll set the main `tsconfig.json` to ignore typings within the `e2e/node_modules/@types` directory, and only consider typings in the root-level `node_modules`, while `e2e/tsconfig.json` will include types from both the `e2e` and root directory `node_modules`:
+
+```json
+{
+  // tsconfig.json
+  "compilerOptions": {
+    "typeRoots": ["node_modules/@types"]
+  }
+}
+```
+
+```json
+{
+  // e2e/tsconfig.json
+  "compilerOptions": {
+    "typeRoots": ["node_modules/@types", "../node_modules/@types"]
+  }
+}
+```
+
+With those `typeRoots` settings, you should now be able to run both your unit tests and Detox tests successfully and with no compiler errors.
--- a/docs/Guide.TypeScriptIntegrationJest.md
+++ b/docs/Guide.TypeScriptIntegrationJest.md
@@ -7,10 +7,9 @@ title: Integrating TypeScript with Detox + Jest
 
 ### 0. Initialize Detox in your project
 
-Follow the [Getting Started](Introduction.GettingStarted.md) Guide to set up Detox with [Jest as your test runner](Guide.Jest.md) (optionally, set up Detox to [test Android](Introduction.Android.md) as well). When you are finished with that, we will make some changes to your project:
+After following the [Getting Started](Introduction.GettingStarted.md) Guide to set up Detox + Jest, we will make some changes to your project:
 
 - Remove Detox from `package.json` at the root level
-
 - Create a `package.json` in the `e2e` folder and add Detox and Jest to the project dependencies:
 
 ```sh
@@ -20,7 +19,7 @@ npm init -y
 npm install -D detox jest
 ```
 
-You should then move the settings for Detox from the root-level `package.json` to the `e2e/package.json`, but make sure that you update the `build` and `binaryPath` values for your Detox configurations:
+Creating the `e2e` directory will be needed to prevent a typing collision issue with Jest and Detox (see step 5). You should then move the settings for Detox from the root-level `package.json` to the `e2e/package.json`, but make sure that you update the `build` and `binaryPath` values for your Detox configurations:
 
 ```json
 {
--- a/docs/Guide.TypeScriptIntegrationJest.md
+++ b/docs/Guide.TypeScriptIntegrationJest.md
@@ -9,22 +9,22 @@ title: Integrating TypeScript with Detox + Jest
 
 After following the [Getting Started](Introduction.GettingStarted.md) Guide to set up Detox + Jest, we will make some changes to your project:
 
-- Remove Detox from `package.json` at the root level
+- Remove Detox from `package.json` at the root level: `npm uninstall detox`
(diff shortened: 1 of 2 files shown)


docs/Guide.TypeScriptIntegrationJest.md
```diff
@@ -0,0 +1,111 @@
+---
+id: Guide.TypeScriptJest
+title: Integrating TypeScript with Detox + Jest
+---
+
+## Usage
+
+### 0. Initialize Detox in your project
```
rotemmiz: 0 is probably unnecessary. It's already mentioned in other guides

docs/Guide.TypeScriptIntegrationJest.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+NB: this is mostly the same output of running `ts-jest config:init`, with `setupTestFrameworkScriptFile` being the only added property.
+
+### 3. `.js` -> `.ts`
+
+Convert all files in the `e2e` directory ending in `.js` to `.ts`
+
+### 4. Add typings for Detox, Jest, and Jasmine
```
rotemmiz: Please mention @types/detox is maintained by the community, not the owners of this repo.

docs/Guide.TypeScriptIntegrationJest.md
```diff
@@ -0,0 +1,111 @@
...
+
+Add typings for Detox, Jest, and Jasmine (the latter two are used in `init.ts`), as well as for other modules that you use in your Detox tests.
+
+```sh
+npm install --save-dev @types/detox @types/jest @types/jasmine
+```
+
+You should now be able to run your Detox tests, written in TypeScript! If you're not writing your unit tests with TypeScript and Jest, you can skip the next section.
+
+### 5. Unit test collisions
```
rotemmiz: It be easier and more straightforward to just not export global functions with Detox on a TS project, and require them explicitly.

docs/Guide.TypeScriptIntegrationJest.md
solkaz: While I agree, I think it needs to be in there since there are a few steps that need to be taken after following those guides. I'll trim that section, however.

docs/Guide.TypeScriptIntegrationJest.md
solkaz: Sorry, I'm not sure that I understand your comment? The issue mostly deals with the current typings for Detox since it declares `expect` as a global constant when Jest also does. For unit tests (that aren't utilizing Detox), `expect` should be defined as a `jest.Expect` whereas they should be `Detox.Expect` for e2e tests (otherwise, the TS compiler gets confused, per [this comment](https://github.com/wix/detox/issues/99#issuecomment-402206854))

rotemmiz: What if you setup Detox not to export Expect as as global?
https://github.com/wix/Detox/blob/master/docs/APIRef.DetoxObjectAPI.md#explicit-imports-during-initialization
Then you'll have to import everything manually like so:

```js
const {device, expect, element, by, waitFor} = require('detox');
```
This solves the collision issue, doesn't it?

solkaz: I hadn't realized that was possible before, but I would prefer to do that. Unfortunately, the current Detox typings are limiting us from doing that because `expect`, `device` et al. are declared as global constants, so you'll get a compiler error when you `require` them from Detox. For that snippet to work, we'd need to change the typings so that they aren't defined as global constants (which makes sense, because it's possible that they might not be globally defined); that would break some user's setups, however, because then they must declare those global constants in their code.

cc-ing @tareqelmasri and @stevechun as they are maintainers for the typings

solkaz: Another issue I noticed was that using `import ... from "detox"` will cause a compiler error, because `detox` isn't a module according to its typings file. `require`-ing it is fine because `require` returns `any`, which is okay as a work-around, but since I'm using TypeScript, I'd prefer to have those imports strongly typed, and I'd prefer to use import statements over `require` as that's what I normally use.

solkaz: I did, the error I got was that `File '~/myProject/node_modules/@types/detox/index.d.ts' is not a module.`, because Detox has only been defined as a namespace with some certain typings, and not also as a module that exports certain members. When I added the following, the error went away:

```ts
declare module detox {
  const detox: Detox.Detox;
  const device: Detox.Device;
  const expect: Detox.Expect;
  // Define the rest of the exports here.
  export {
    detox,
    device,
    expect,
    // rest of the exports
  ]
}
```

Do you mean for one definition file to support both styles? I think if possible, it won't be ergonomic; if we assume `expect` to be typed as `Detox.Expect | undefined`, then the compiler will complain if you try to use it: `Cannot invoke an object which is possibly 'undefined'.`. You'd need to explicitly cast it as `Detox.Expect` or sandwich `!` between `expect` and the opening parenthesis (e.g. `expect!()`) to tell the compiler that you know that it's defined. Also, whether those constants are defined globally can't be known until runtime, which is why I'm more preferential to leaving it up to users to define those global constants in their code, even if the default behavior of Detox is to define them globally.

docs/Guide.TypeScriptIntegrationJest.md
solkaz: Added a note below

docs/Guide.TypeScriptIntegrationJest.md
```diff
@@ -0,0 +1,113 @@
...
+- Create a `package.json` in the `e2e` folder and add Detox and Jest to the project dependencies:
+
+```sh
+npm uninstall detox
+cd e2e
+npm init -y
+npm install -D detox jest
+```
+
+You should then move the settings for Detox from the root-level `package.json` to the `e2e/package.json`, but make sure that you update the `build` and `binaryPath` values for your Detox configurations:
```
luisnaranjo733: Just wondering - what is the value in moving your detox config to the e2e package?

luisnaranjo733: Ah I see, you are working around the type collision

luisnaranjo733: Might be good to add that as a justification for why you need to move stuff to e2e/

docs/Guide.TypeScriptIntegrationJest.md
solkaz: I've cut out some of the content here and added a sentence saying why it is necessary to have the tests in the e2e folder

docs/Guide.TypeScriptIntegrationJest.md
solkaz: 👍 I added a sentence in the first section that briefly explains it.

blefebvre:
@solkaz, thanks so much for your work on this guide!

I am working through it now and noted a couple of extra steps I had to take at the end of step 0 in order to get my existing tests to run successfully:

1. In `e2e/package.json`, "detox" property, I had to set `runner-config` to prevent it from looking for it's config file in the e2e/ directory (the default):
```
"runner-config": "./config.json"
```
2. In `e2e/init.js`, I had to update the package.json require to point at e2e's package.json (instead of the root project):
```
const config = require('./package.json').detox;
```

After these changes my tests again run as expected.

solkaz:
@blefebvre thanks for the feedback! I've updated my PR with your notes.

ScreamZ:
@solkaz I gave up and i'm now trying your method (because I'm running out of time) and now I have :

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    error TS2468: Cannot find global value 'Promise'.
    init.ts:7:11 - error TS2705: An async function or method in ES5/ES3 requires the 'Promise' constructor.  Make sure you have a declaration for the 'Promise' constructor or include 'ES2015' in your `--lib` option.

Even if I added to --lib the es2015
:(

solkaz:
Could you show me your `tsconfig.json` that you are using?

ScreamZ:
```
{
  "lib": ["es2015", "esnext"],
  "compilerOptions": {
    "typeRoots": ["node_modules/@types", "../../node_modules/@types"]
  }
}
```
```
{
  "setupTestFrameworkScriptFile": "./init.ts",
  "preset": "ts-jest",
  "testEnvironment": "node",
  "testMatch": ["**/specs/**/?(*.)+(spec|test).ts"],
  "moduleNameMapper": {
    "^@testId/(.*)$": "<rootDir>/identifiers/$1",
    "^@testHelpers/(.*)$": "<rootDir>/helpers/$1"
  },
  "globals": {
    "ts-jest": {
      "tsConfig": "./tsconfig.json"
    }
  }
}
```

solkaz:
Hm, not sure what the issue is from a quick glance, I'll try to take a closer look later. If you'd like an example to follow, please check my example project at https://github.com/solkaz/ts-detox-example

docs/Guide.TypeScriptIntegrationJest.md
```diff
@@ -0,0 +1,120 @@
...
+### 0. Initialize Detox in your project
+
+After following the [Getting Started](Introduction.GettingStarted.md) Guide to set up Detox + Jest, we will make some changes to your project:
+
+- Remove Detox from `package.json` at the root level: `npm uninstall detox`
+- Create a `package.json` in the `e2e` folder and add Detox and Jest to the project dependencies:
+
+```sh
+cd e2e
+npm init -y
```
rotemmiz: I am very confused about this step, why would the users need to uninstall Detox from the main project and init a new project inside e2e dir? This looks like a hack :(
Is it only because of the detox globals?

rotemmiz: I think this is a hacky solution and can not be documented as an official solution. There must be a better way.

solkaz: Can you elaborate on why it's a hacky solution, and how it can be improved?

rotemmiz: Setting Detox as a devDependency of a subproject inside e2e folder, running npm install twice is not a very common, or even wanted practice, it can and will have grave consequences on projects maintenance, especially if they're big, or include many developers.
I'm not an expert in typescript configuration, but this solution, as it is a bit hacky, can not be part of official documentation.
We should strive to solve the real problem, not to document workarounds.

Now, how can we solve this?

solkaz: I don't seeing any consequences of having multiple `package.json`s since there shouldn't be overlap between e2e code and implementation code. As I pointed out above, I do have an example project that has only one `package.json` [here](https://github.com/solkaz/ts-detox-example/tree/one-package-json-to-rule-them-all) and mimics the usual Detox setup (I think). If you're okay with that setup, then I'll update the PR to reflect that.

blefebvre: > Now, how can we solve this?

It seems like the 2nd package.json in e2e/ is only really necessary to avoid typing conflicts with `expect`, `describe` and other globals that Jest and Detox share. Knowing this, could a potential solution be that we are required to import the globals from their respective lib before using them? For example, `import { expect, describe, it, element } from "detox";`

I realize the scope of this change is large, and would have to be on an opt-in basis to avoid affecting existing Detox test suites. I am sure that I am missing some background as to why the global approach was taken, since I feel like most React Native devs would be comfortable with importing their dependencies.

> I do have an example project that has only one package.json here

@solkaz, this looks good to me! One less `npm install` to run, one less `node_modules` directory full of dependencies.

blefebvre: Dug a bit more and discovered that this already exists! https://github.com/wix/Detox/blob/master/docs/APIRef.DetoxObjectAPI.md#detoxinit

In init.js|ts:
```await detox.init(config, {initGlobals: false});```

At the top of each .spec.ts:
```const {device, expect, element, by, waitFor} = require('detox');```

Cool!

Comment (rotemmiz) [reply docs/Guide.TypeScriptIntegrationJest.md]: This was already [mentioned in the thread ](https://github.com/wix/Detox/pull/1001#discussion_r228734569) but was said to still suffer from typing issues.

Comment (solkaz) [reply docs/Guide.TypeScriptIntegrationJest.md]: @blefebvre to expand on what @rotemmiz said, once [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30410) goes through, you would be able to import the constants from `detox`:

```ts
const { device, expect, element, by, waitFor } from "detox";
```

TypeScript will recognize `expect` to be a `Detox.Expect` so your files would then compile. Currently this isn't possible because Detox isn't declared as a module, and `require`-ing doesn't seem to work because `expect` has already been declared.

docs/Guide.TypeScriptIntegrationJest.md
solkaz: DefinitelyTyped/DefinitelyTyped#30410 has been merged, so I will update my PR and example projects to import Detox constants.

TareqElMasri:
Guys, we had an issue with Jest because it was exporting `expect` method globally which conflicted the same `expect` method in `Detox` namespace, recently there was an effort from @tegnno to fix that, he thankfully made a pull request which I approved later to fix this issue
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32248

Currently we're just waiting for DefinitelyTyped guys to accept the pull request, that will fix the `expect` issue, the pull request is expected to be merged soon and I will notify you once that happens.

If anyone still needs a markdown file/ a medium article or an example project of how to get successfully those three to work together I will be happy to help

TareqElMasri:
@solkaz The pull request is merged right now and the global `expect` issue is fixed for good

docs/Guide.TypeScriptIntegrationJest.md
solkaz: @rotemmiz I've made changes to the document that recommend disabling `initGlobals` and using import statements for the Detox methods.

blefebvre:
I've worked through the steps again recently and ran into the following issue, in init.ts:

```
TypeError: adapter.afterAll is not a function
```

I was able to get past this error by changing the Jest adapter import line (in init.ts) to omit the `* as`:

```
import adapter from "detox/runners/jest/adapter";
```

This guide has been super helpful for me - looking forward to seeing it as part of the official docs!

solkaz:
@blefebvre Could you provide an example repo? Both forms (`import adapter` and `import * as adapter`) are working for me at [my example project](https://github.com/solkaz/ts-detox-example).

Technically, the `detox` module doesn't declare a default export as it uses [`module.exports`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32927#pullrequestreview-203883558), so I tried to go with a more ["correct"](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32927#discussion_r256963409) approach. It does work if you set enable `allowSyntheticDefaultImports` in your `tsconfig.json` (which isn't the default).

blefebvre:
@solkaz, I was not able to reproduce this issue with my open source project, so I suspect that it is a configuration issue on my side. Sorry for the noise!

rotemmiz:
Hey there,

I appreciate the hard work poured here (and on other repos as well, to make Detox with with TS.

We're a small team working on core Detox maintenance and new features. This guide, as good as it will be, can never be maintained by us, moreover, it doesn't have to be, since most of the work has been done by @solkaz and the community anyway.
I think that it would be best if this guide will be published as a blog post, or gist, or any other format, outside the official docs (as we have to commit to the official docs, and as said, we don't have the capacity to handle it all).
This goes the same for flow support or even running Detox on Expo apps, the current doc is outdated and unmaintained, and should have probably kept out of the repo.

I am closing this PR, and I suggest you make an external guide out of it (I would clap if you put it on Medium).

PR #1007, opened 2018-10-29
Added testBinaryPath as an optional config parameter

reime005:
At a specific point in an app configuration the 'testBinaryPath' was not correct and I had the need to configure it manually. Please let me know if that does not make sense.

Tested on Android emulator, with and without the testBinaryPath parameter.


chips5k:
+1 for this - my team need this feature as the conventions guessed path is completely wrong in our setup. the only work around at present is to rename and shift our files to match what its expecting...

rotemmiz:
While the PR code is good, I have an issue with the documentation. Since it's optional, I think testBinaryPath should be removed from the introduction example.

reime005:
sure, done

rotemmiz:
I want to merge this but CI on this PR is still remember, can you please check?

rotemmiz:
A unit test still fails.
`✕ installApp() with a custom app path should use custom app path`

rotemmiz:
DTXLoggingInfra submodule has changed its reference, can you please revert that?

rotemmiz:
There is still one conflict :(

reime005:
> There is still one conflict :(

should be fine now

PR #1017, opened 2018-11-03
update expo usage guides

quinlanj:
Expo apps didn't play well with detox 7 and above. I've since made some fixes for this, and it works for detox v9.0.6 now. Updating docs to reflect this change.


rotemmiz:
Thank you!
We can work on adding Detox Android support to expo app if you'd like.

PR #1025, opened 2018-11-05
Add --bail option to detox-test

EdwardDrapkin:
It's helpful to use --bail sometimes in either Jest or Mocha, so I exposed it and forwarded it in.


rotemmiz:
I think this can be passed using `-- --bail` to both test runners

noomorph:
@rotemmiz , no, we have no `--` passthrough logic in CLI yet.

rotemmiz:
Then this is probably what we need to add instead of exposing flags individually.
This is a very limited solution.

rotemmiz:
I would rather add a mechanism to pass args to test runners, kinda like npm does.
https://github.com/npm/npm/pull/5518/files#diff-c8ef191c95255343393f2a3256d29483R127

EdwardDrapkin:
I updated this PR to allow generic pass through test runner arguments.

I ran into two issues with the `-- <extra args>` approach: launching detox via yarn/npm (`yarn detox test ...`) swallowed the `--` before commander could parse it, and commander itself swallows the `--` before launching a subcommand like `detox-test.js`.  So the approach of `detox test ... -- --bail --forceExit` seemed out, so I considered passing the extra arguments as a flag but `detox test ... --extraArgs='--bail --forceExit'` seemed really awkward, so I settled on an environment variable, `DETOX_TEST_RUNNER_FLAGS`.  `DETOX_TEST_RUNNER_FLAGS='--bail --forceExit' yarn detox test ...` isn't the worst thing in the world, and using an environment variable means that people (like myself) can just set that environment variable in CI and get the bailout behavior across all tests without having to modify every invocation of detox.

EdwardDrapkin:
@rotemmiz is this approach acceptable or do you have an alternative you'd prefer?

rotemmiz:
I'm against setting an env var to cli if is not absolutely needed, it makes the usage cumbersome and error prone in my opinion. 
If it's not possible to add via `-- `, let's use `--args ` instead.

EdwardDrapkin:
I can't use `args` because `program.args` is reserved by commander and it doesn't seem that it will parse `--args="--bail --forceExit"` correctly.  Was trying with this short script:

```

program
    .allowUnknownOption()
    .option('-f [args]', 'foo')
    .parse(process.argv);

console.log(program);
```

It seems that the best I could do with args is `--runner-args=bail,forceExit` which means it's not possible to pass parameters to runner flags, or `--runner-args="bail forceExit watchIgnorePattern='123'"` which would allow them to be passed, but either approach seems much cludgier and error prone than the env var.  It's looking more and more like the only solution that will work with commander's parser is the env var :S

rotemmiz:
Do you believe that this can be handled properly by [`argparse`](https://www.npmjs.com/package/argparse)? 
I have been intending to switch for quite a while as commander has very critical and unhandled parsing errors.

noomorph:
@rotemmiz, @EdwardDrapkin , guys, I struggle a bit to see what's the problem here:

```javascript
function splitPassthrough(argv) {
  const position = argv.indexOf('--');

  if (position === -1) {
    return [argv, []];
  }

  return [argv.slice(0, position), argv.slice(position + 1)];
}


const [argv, pargv] = splitPassthrough(process.argv);

console.log('argv', argv); // use with commander
console.log('pargv', pargv); // pass as-is
```

Why can't we make a helper to split argv and use with commander in a way it used to be and add a routine to append `-- ... passthrough args` to the final command?

EdwardDrapkin:
Commander rewrites process.argv for subcommand invocations.  

Although I guess we could pass `[...process.argv]` instead of `process.argv`

EdwardDrapkin:
I spent some time digging in commander and it turns out it does parse the unknown args out and just drops them on the floor, but I was able to call the parser methods myself.  It appears to work that extra flags just get passed clean through to the test runner.

rotemmiz:
This is great stuff!! Thanks for the stubburn investigation!
Seems like this branch needs a rebase/merge from master? Can you please do it? I want to make sure CI is green before it gets merged.

noomorph:
Edward, thank you very much!

PR #1032, opened 2018-11-08
Update Introduction.Android.md

carbonjesse:
Small update to step 3 of android intro.

### DIFF
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -74,7 +74,7 @@ buildscript {
 
 ### 3. Create Android Test class
 
-You need to add the file `android/app/src/androidTest/java/com/[your.package]/DetoxTest.java` and fill it like [this](../detox/test/android/app/src/androidTest/java/com/example/DetoxTest.java), except that you need to change the package to your projects name.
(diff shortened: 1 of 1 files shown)


rotemmiz:
The value in packageName in `AndroidManifest.xml` doesn't matter as gradle overrides it.
Where did you encounter an issue?

carbonjesse:
When first going through the setup process for android, i ran into this issue which was resolved by adding the correct package name to that file.

https://stackoverflow.com/a/31563293/766115

Now that i'm passed that part and the tests run, switching the package name back to example.com in that file causes the tests to not run and the app not install.

rotemmiz:
Do you have an applicationId configured in your `build.gradle`?

like here: https://developer.android.com/studio/build/application-id

carbonjesse:
yes

rotemmiz:
I see, but if there's misconfiguration you'll get a `java.lang.ClassNotFoundException` whether you run tests with Detox or opening the app normally. Am I missing something?

carbonjesse:
I'm not too familiar with building android apps. Just thought this was helpful for the setup as I believe it would have saved me a few minutes.  

But if it's not clear what I'm intending feel free to close without merge.

PR #1035, opened 2018-11-10
Add getViewportSize for device

cbrevik:
### Motivation

Wanting to use the same code to scroll relative to height of device on both iOS and Android meant that we have to know the viewport size. So, in order to avoid depending on hard coded values, I've added a `device.getViewportSize`-function to retrieve the width and height of the current device. We use this internally at [ice](https://ice.no) for our Detox-tests.

### Discussion

On Android it is pretty straightforward to retrieve the viewport size through `adb shell wm size`.

On iOS I could not find a similar command, so I hack around it by checking the size of a screenshot taken on the device. Since taking a screenshot is functionality Detox already had, I thought maybe this implementation was acceptable. Also added the `image-size` npm package to help with retrieving the size of the screenshot. If there are alternative ways of solving this on iOS I'm all ears though.

Also added a E2E-test for iOS and Android each checking the device sizes for the E2E devices.

Hope this is found to be useful!

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -42,6 +42,7 @@
     "commander": "^2.15.1",
     "fs-extra": "^4.0.2",
     "get-port": "^2.1.0",
+    "image-size": "^0.6.3",
     "ini": "^1.3.4",
     "lodash": "^4.17.5",
     "minimist": "^1.2.0",
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -208,6 +208,10 @@ class Device {
     return this.deviceDriver.getPlatform(this._deviceId);
   }
 
+  async getViewportSize() {
+    return await this.deviceDriver.getViewportSize(this._deviceId);
+  }
+
   async _cleanup() {
     await this.deviceDriver.cleanup(this._deviceId, this._bundleId);
   }
--- a/detox/src/devices/drivers/AndroidDriver.js
+++ b/detox/src/devices/drivers/AndroidDriver.js
@@ -62,6 +62,10 @@ class AndroidDriver extends DeviceDriverBase {
     await this.invocationManager.execute(call);
   }
 
+  async getViewportSize(deviceId) {
+    return await this.adb.getViewportSize(deviceId);
+  }
+
   getTestApkPath(originalApkPath) {
     const testApkPath = APKPath.getTestApkPath(originalApkPath);
 
--- a/detox/src/devices/drivers/SimulatorDriver.js
+++ b/detox/src/devices/drivers/SimulatorDriver.js
@@ -2,6 +2,8 @@ const exec = require('child-process-promise').exec;
 const path = require('path');
 const fs = require('fs');
 const _ = require('lodash');
+const tempfile = require('tempfile');
+const sizeOf = require('image-size');
 const IosDriver = require('./IosDriver');
 const AppleSimUtils = require('../ios/AppleSimUtils');
 const configuration = require('../../configuration');
@@ -103,6 +105,20 @@ class SimulatorDriver extends IosDriver {
   async shutdown(deviceId) {
     await this._applesimutils.shutdown(deviceId);
     await this.emitter.emit('shutdownDevice', { deviceId });
+  }
+
+  async getViewportSize(udid) {
+    const temporaryFilePath = tempfile('.png');
+
+    await this._applesimutils.takeScreenshot(udid, temporaryFilePath)
+
+    var dimensions = sizeOf(temporaryFilePath);
+
+    fs.unlinkSync(temporaryFilePath);
+    
+    delete dimensions.type
+
+    return dimensions
   }
 
   async setLocation(deviceId, lat, lon) {
--- a/detox/src/devices/android/ADB.js
+++ b/detox/src/devices/android/ADB.js
@@ -149,6 +149,26 @@ class ADB {
     return Number(stdout.slice(0, stdout.indexOf(' ')));
   }
 
+  async getViewportSize(deviceId) {
+    const { stdout } = await this.adbCmd(deviceId, 'shell wm size').catch(e => e);
+  
+    const sizeRegex = /(\d+)x(\d+)/i
+  
+    const matches = stdout && stdout.match(sizeRegex)
+  
+    if (matches && matches.length === 3) {
+      return {
+        width: +matches[1],
+        height: +matches [2]
+      }
+    } else {
+      return {
+        width: undefined,
+        height: undefined
+      }
+    }
+  }
+
   async isBootComplete(deviceId) {
     try {
       const bootComplete = await this.shell(deviceId, `getprop dev.bootcomplete`, { silent: true });
--- a/detox/test/e2e/06.device.test.js
+++ b/detox/test/e2e/06.device.test.js
@@ -78,6 +78,34 @@ describe('Device', () => {
     await expect(element(by.text('Shaken, not stirred'))).toBeVisible();
   });
 
+  it(':ios: iPhone X viewport should equal 1125x2436', async () => {
+    const size = await device.getViewportSize();
+
+    if (!size) {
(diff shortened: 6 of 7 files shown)


rotemmiz:
Hey @cbrevik, thanks for the contribution! 
Before accepting this PR, I would like to discuss its necessity. 
In the example you added `device.getViewportSize()` helps scrolling to a certain position on the screen, but it doesn't solve a cases where the view is out of screen bounds (which is very common when rendering a screen on multiple screens with different size).

I want to argue that the following API is more robust and will work on any device regardless if the view is in the viewport or not (or even if it wasn't even rendered (!!), a common case on lists).

```js
await waitFor(element(by.text('notVisibleAtFirst'))).toBeVisible().whileElement(by.id('listView')).scroll(100, 'down');
await expect(element(by.text('notVisibleAtFirst'))).toBeVisible();
```

cbrevik:
Hi @rotemmiz!

Thanks for picking up the PR. You are right of course, my example was to simplistic and does not demonstrate why I find this to be necessary.

We actually do have something like your code example when trying to find an item in a `ScrollView`:
```js
const scrollTo = async (scrollViewId, scrollToRef) => {
    const size = await device.getViewportSize()
    await waitFor(scrollToRef)
        .toBeVisible()
        .whileElement(by.id(scrollViewId))
        .scroll(size.height / 4, "down")
}
```

I just didn't want to bloat the example in the docs. But can update that example if you want me to.

The problem for us is that we have a couple of screens with some very long "feeds" to scroll through. So if we just set a low scroll pixel value like `100`, then scrolling to find the item takes too long. It makes our E2E-tests runtime much longer than what we would like.

But on the other hand, setting some arbitrary higher value actually makes our tests flaky across Android/iOS and different devices. Say if I set scroll pixel length to `400`. That might work fine on some devices, but not as well on some others. The flakiness comes from the code sometimes scrolling past the element we want to find.

If the element is medium-sized and takes up a lot of the viewport, when you first see the element it might not be 75% visible for `toBeVisible` to trigger. But on the following scroll the code might go too far past it to be 75% visible on some devices.

We found that trying to find some magic unflaky value for different devices/platform was difficult. So calculating some relative scroll pixel value from the device height made our tests more robust in this regard.

rotemmiz:
This sounds much more reasonable now, but I still think the implementation is lacking, and in some cases wrong.
Computing these values externally from the app will in many cases produce wrong values, what if the app uses safe area on iOS, or immersive mode on Android? Whay about notification/navigation bar? The app viewport will include/exclude a hefty part of the screen.

A better approach would be to return the viewport size from native somehow, either when the device sends `isReady` to tester, or via a designated device API call.

But even so, this still doesn't satisfy all the usecases. Maybe the best approach would be to send the matched view size as part of the response from the device.

cbrevik:
I grant this approach is naive. It works for our use cases, but will probably not satisfy all as you say. 

If you could point me in the right direction I could take a look if you’d like. I started trying to retrieve it natively, but soon got a bit bogged down by the layers of the framework.

LeoNatan:
What's a viewport? That's not a term that I am familiar with from the native world. Do you mean screen size? If so, it should be called that.

> On iOS I could not find a similar command, so I hack around it by checking the size of a screenshot taken on the device. Since taking a screenshot is functionality Detox already had, I thought maybe this implementation was acceptable. Also added the image-size npm package to help with retrieving the size of the screenshot. If there are alternative ways of solving this on iOS I'm all ears though.

That's an incorrect hack which I don't want. A screenshot will not give you the expected size. The right thing to do is ask the native what dimensions the screen has. This is something that shouldn't even be cached, because under iOS, screen size can change under certain conditions, such as for accessibility reasons, iPad multitasking, etc. In case of multitasking on the iPad, a screenshot will give you the entire screen, but that's incorrect. I'm sure Android has the same problem, even by asking adb the emulator size.

Further, taking a screenshot of the simulator depends on how it is rendering its content. An iOS simulator doesn't always render 1:1 (to save performance), further complicating your solution.

detox/package.json
```diff
@@ -42,6 +42,7 @@
     "commander": "^2.15.1",
     "fs-extra": "^4.0.2",
     "get-port": "^2.1.0",
+    "image-size": "^0.6.3",
```
LeoNatan: As discussed in the comments, this should come from native and this is unnecessary.

detox/src/devices/Device.js
```diff
@@ -208,6 +208,10 @@ class Device {
     return this.deviceDriver.getPlatform(this._deviceId);
   }
 
+  async getViewportSize() {
```
LeoNatan: should be `screenSize`

detox/src/devices/drivers/AndroidDriver.js
```diff
@@ -62,6 +62,10 @@ class AndroidDriver extends DeviceDriverBase {
     await this.invocationManager.execute(call);
   }
 
+  async getViewportSize(deviceId) {
+    return await this.adb.getViewportSize(deviceId);
```
LeoNatan: As discussed with the iOS case, I don;t think this is correct.

detox/src/devices/drivers/SimulatorDriver.js
```diff
@@ -105,6 +107,20 @@ class SimulatorDriver extends IosDriver {
     await this.emitter.emit('shutdownDevice', { deviceId });
   }
 
+  async getViewportSize(udid) {
+    const temporaryFilePath = tempfile('.png');
+
+    await this._applesimutils.takeScreenshot(udid, temporaryFilePath)
```
LeoNatan: As discussed in the comments, this is a bad way of calculated results. It should come from native.

LeoNatan:
Sorry didn't read whole conversation. But requesting bounds without safe area is not something that is possible to do easily. Yes, screen without notch is easy, but safe area is a complex topic on iOS. It's an additive mechanism, where each view controller can add more safe areas to inner controllers. If you need the safe area for a specific view, that could require a changes to our infra, as we don't have mechanisms for pulling information on elements.

cbrevik:
Hi @LeoNatan!

"Viewport" is a term from web development, but if it is an unfamiliar term to native developers then it might as well be called "screen".

Totally see your point about the screenshot, especially considering multitasking on iPad. I was hoping to start a discussion like this by opening the PR. 👍 

Just wondering, does it matter if navigation bars/immersive mode/safe area/accessibility are a factor in terms of the size of the "content" area? 

When developers are writing Detox-tests, and are using features like that (safe area, etc), would they not take that into account? And still find value in knowing the total size of the screen of the device, and not just the content area size?

Might be a naive question, since our concrete use case is pretty simple 🤔

LeoNatan:
(Not sure what "immersive mode" means)

Safe areas are important. I still don't fully understand the use-case here, since scrolling can be performed in small chunks. If precision is important (like for example, paging of content), then yes, safe areas are very important because the usable scrollable area is the area available within the sum of all safe area insets. It's calculated by the system, and available to developers to inspect at runtime, so they **should** be taking it into account. React Native makes this much harder with the SafeAreaView being just padding. But that's beside the point. If you want to precisely scroll, you'd need that value for a specific scroll view.

I agree that screen size is independent from safe area insets.

cbrevik:
[Immersive mode on Android](https://github.com/wix/Detox/pull/1035#issuecomment-437593743) was mentioned above here as a potential problem as well. 

As for use case, it has more to do with test execution time on scrolling through long feeds, and trying to optimize how quickly we can scroll without sacrificing precision. I think I explained our use case pretty well in [this comment](https://github.com/wix/Detox/pull/1035#issuecomment-437592030), but if you think it’s not useful for others as part of the Detox API, that’s cool too 👍

For our part we wanted to avoid having to depend upon different hard coded scroll pixel values depending on which device we’re testing on

LeoNatan:
Well, you mention that setting a larger scroll offset causes issues. That is an independent something we should look at. Could you please open an issue, stating what and how you are doing it? I’d like to take a look.

PR #1036, opened 2018-11-11
iOS: Synchronization over waitForBackground and waitForActive states 

LeoNatan:
Add support for waiting for background and active states. This will solve an issue where some actions are not synchronized correctly.


LeoNatan:
@rotemmiz Let's merge and release?

PR #1037, opened 2018-11-11
Android: Increase RN-timers idling resources look-ahead to 1.5sec

d4vidi:
@rotemmiz just for review, don't merge

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/ReactNativeTimersIdlingResource.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/ReactNativeTimersIdlingResource.java
@@ -35,13 +35,15 @@ public class ReactNativeTimersIdlingResource implements IdlingResource, Choreogr
     private final static String METHOD_GET_NATIVE_MODULE = "getNativeModule";
     private final static String METHOD_HAS_NATIVE_MODULE = "hasNativeModule";
     private final static String FIELD_TIMERS = "mTimers";
-    private final static String FIELD_TARGET_TIME = "mTargetTime";
+    private final static String TIMER_FIELD_TARGET_TIME = "mTargetTime";
+    private final static String TIMER_FIELD_INTERVAL = "mInterval";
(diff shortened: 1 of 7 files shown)


detox/android/detox/src/main/java/com/wix/detox/espresso/ReactNativeTimersIdlingResource.java
```diff
@@ -144,7 +137,46 @@ public void doFrame(long frameTimeNanos) {
...
+    }
+
+    private boolean isTimerOutsideBusyWindow(Object nextTimer) {
+        final long currentTimeMS = System.nanoTime() / 1000000L;
+        final Reflect nextTimerReflected = Reflect.on(nextTimer);
+        final long targetTimeMS = nextTimerReflected.field(TIMER_FIELD_TARGET_TIME).get();
+        final int intervalMS = nextTimerReflected.field(TIMER_FIELD_INTERVAL).get();
+        final boolean isRepetitive = nextTimerReflected.field(TIMER_FIELD_REPETITIVE).get();
+
+//        Log.i(LOG_TAG, "Next timer has duration of: " + intervalMS
```
rotemmiz: Not sure those comments are necessary, unless there's a hack that needs an explanation.

d4vidi: Assuming you're referring to the comments and not the logs -
they're necessary because we didn't unit test, so there's nothing here to explain the idea behind the implementation. This will change soon, of course.

d4vidi:
@rotemmiz why not rebase over master, instead?

d4vidi:
Anyways this branch is becoming stale. If you're not comfortable with the code, let's try to think of smt else :)

rotemmiz:
We're going to squash it anyway, it doesn't matter. But build is still red.

rotemmiz:
Retriggerd the build, let's see

Issue #1034, opened 2018-11-09
iOS/android concurrency


jgreen210:
#### Description

I'm running detox concurrently with itself, since want parallelism across simulators/emulators/devices (and to manage their lifecycles myself). Each detox instance is given one android or iOS device/simulator. I don't run more than one iOS simulator at the same time, but do want to run one iOS simulator detox instance at the same time as multiple android detox instances.

Inspecting the detox source code, I notice that ~/Library/Detox/device.registry.state.lock is used to manage iOS concurrency within one detox invocation. It was added in: https://github.com/wix/Detox/pull/609

That's OK, since I only want to run at most one iOS detox instance at a time. I notice that the lock file is cleared even if start detox for just android though, so it will interfere with iOS detox invocations.

There is some locking around the lock file, but it's (correctly) only done after the lock file is cleared when start detox.

#### Steps to Reproduce

Run one iOS detox instance and many android detox instances concurrently with each other. Their use of ~/Library/Detox/device.registry.state.lock will collide.

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 9.0.6
* React Native: n/a
* Node: n/a
* Device: iOS and android
* Xcode: n/a
* macOS: n/a

jgreen210:
I think this should fix this:

https://github.com/jgreen210/Detox/commit/9078a7ab9307db5426ac76cc3a0e4bbe57834974
https://github.com/jgreen210/Detox/commits/registry_locking

I've not tried to run it locally or run this project's tests against this commit yet. I'd like a little feedback on the general approach before I invest time in that and create a PR.

jg210:
90778a7a isn't right - would need to clear registry somewhere else too. I'll fix it up on Monday.

jgreen210:
Another way to fix this would be to make the path to the lock file configurable via a command-line option. It would then be possible to use a different file for ios and android. This would also allow e.g. iOS-iOS concurrency to be configured in a variety of ways.

PR #1043, opened 2018-11-14
Add DETOX_DEVICE_LOCK_FILE environment variable.

jgreen210:
- [x] This is a small change 
- [ ] This change has been discussed in issue #1034 and the solution has been agreed upon with maintainers.

---

**Description:**

Add DETOX_DEVICE_LOCK_FILE environment variable.

### DIFF
--- a/docs/Guide.ParallelTestExecution.md
+++ b/docs/Guide.ParallelTestExecution.md
@@ -11,7 +11,7 @@ By default `detox test` will run the test runner with one worker (it will pass `
 #### Lock file
 Simulators/emulators run on a different process, outside of node, and require some sort of lock mechanism to make sure only one process controlls a simulator in a given time. Therefore, Detox 7.4.0 introduced `device.registry.state.lock`, a lock file controlled by Detox, that registers all in-use simulators.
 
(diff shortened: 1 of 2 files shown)


docs/Guide.ParallelTestExecution.md
```diff
@@ -22,6 +22,8 @@ The lock file location is determined by the OS, and [defined here](https://githu
 or<br>
 `$USERPROFILE/Application Data/Detox/device.registry.state.lock`
 
+The lock file location can be overridden using the DETOX_DEVICE_LOCK_FILE_PATH environment variable.
```
DanielMSchmidt:
```suggestion
The lock file location can be overridden using the `DETOX_DEVICE_LOCK_FILE_PATH` environment variable.
```

rotemmiz:
I went over this PR a few times, and each time I thought to myself that it requires much greater attention.
On one hand, adding another config env var requires us to map what we already have, and document it in a centralised place. It's a good idea, but so is another client argument (which is much better documented currently)
On the other hand, adding a cli arg would need to be synced with @DanielMSchmidt's work, as he is rewriting the detox CLI tool.
I don't have an operative remark here, but I want us to discuss this here and come out with the best possible solution.

Issue #1039, opened 2018-11-12
Support for pressing Delete key on Android


michaelgmcd:
Unlike iOS, there doesn't appear a way to press the "Delete" key on Android. For iOS, I'm am able to use the `typeText` method and the `"\b"` string. From what I can tell through looking at the code is that either the [UIDevice's pressDelete method](https://github.com/wix/Detox/blob/016adbb6ec37235b17c4d036fe221daa454d22d1/detox/src/android/espressoapi/UIDevice.js#L137-L143) or the [Android Espresso pressKey](https://developer.android.com/reference/android/support/test/espresso/action/ViewActions#presskey) method are the ways to solve this, however neither of these seem to be exposed to the JavaScript API.

michaelgmcd:
@LeoNatan, I can probably take care of the PR for this, we'll just need some discussion on the implementation. The way I see it, we can either:
* Manually handle the `\b` key in the typeText element method on Android:
```javascript
element.typeText('aa\bcc')
// on Android becomes 
element.typeText('aa')
element._pressKey(67) // internal
element.typeText('cc')
```
* Expose the `pressKey` command on Android
  * It doesn't seem like EarlGrey has a `pressKey` command, so this would be Android only
  * Issue with this is that users will have to manually implement what's in the first option for a cross-platform test
* Expose the `pressDelete` command on Android
  * Will also require users to manually implement the first option for cross-platform support
  * Doesn't support keys other than Delete

LeoNatan:
Sadly, `\n` and `\b` are not obvious at all on iOS. I think it's probably a good idea to have functions such as `pressReturnKey()` and `pressBackspaceKey()` on element (next to `typeText`), and then inside split the impl to typing `\n` on iOS and doing the necessary key press for Android. What do you think?

LeoNatan:
The `pressKey()` should remain internal, I think.

LeoNatan:
If you could submit a PR, that would be lovely. Thank you!

michaelgmcd:
I think the abstraction sounds nice. I'll take another look at this tonight. Should be able to open a PR.

PR #1044, opened 2018-11-14
New Actions: `tapBackspaceKey` and `tapReturnKey`

michaelgmcd:
- [X] This change has been discussed in issue #1039 and the solution has been agreed upon with maintainers.

---

This PR adds support for pressing the backspace and return keys on both iOS and Android. Although iOS support exists by using typeText and the `\b` and `\n` keys, this was unclear and not cross-platform. The pressBackspaceKey and pressReturnKey methods abstract these features for both platforms.

**Note:** I did try adding a `numberOfTimes` parameter as a shortcut for pressing the backspace or return key multiple times, however I wasn't able to figure out the simplest way for executing multiple actions:

For both Android and iOS, the addition `numberOfTimes` for pressReturnKey is pretty straightforward:
```javascript
async pressReturnKey(numberOfTimes = 1) {
  if (numberOfTimes && typeof numberOfTimes !== 'number') { throw new Error('...') }
  return await new ActionInteraction(this, new TypeTextAction('\n'.repeat(numberOfTimes))).execute();
}
```

For pressBackspaceKey on iOS, it can be implemented the same way as return, replacing `\n` with `\b`. For Android, however, I'm not sure how to invoke multiple `PressKeyActions`. Either way, I think this can be added in a follow up PR if it's determined to be useful.

cc @LeoNatan

### DIFF
--- a/detox/src/android/expect.js
+++ b/detox/src/android/expect.js
@@ -204,10 +204,10 @@ class Element {
   async multiTap(times) {
     return await new ActionInteraction(this, new MultiClickAction(times)).execute();
   }
-  async pressBackspaceKey() {
+  async tapBackspaceKey() {
     return await new ActionInteraction(this, new PressKeyAction(67)).execute();
   }
-  async pressReturnKey() {
+  async tapReturnKey() {
     return await new ActionInteraction(this, new TypeTextAction('\n')).execute();
   }
   async typeText(value) {
--- a/detox/src/android/expect.js
+++ b/detox/src/android/expect.js
@@ -54,6 +54,13 @@ class MultiClickAction extends Action {
   }
 }
 
+class PressKeyAction extends Action {
+  constructor(value) {
+    super();
+    this._call = invoke.callDirectly(ViewActionsApi.pressKey(value));
+  }
+}
+
 class TypeTextAction extends Action {
   constructor(value) {
     super();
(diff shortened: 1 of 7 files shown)


LeoNatan:
Fantastic, thank you!
Would I trouble you much if I ask to name them `tap...`? It was my fault with the `press` in the issue, but on iOS that has different meanings. Tapping is the agreed upon terminology for iOS and Android.

michaelgmcd:
Done @LeoNatan.

LeoNatan:
Looks good to me. Let's wait for @rotemmiz review as well.

michaelgmcd:
@rotemmiz Any idea when you'll be able to get to this? It's a blocker for my team moving to Detox (which we'd love to do asap)

rotemmiz:
Looks great! Let's just add a few lines of documentation in https://github.com/wix/Detox/blob/master/docs/APIRef.ActionsOnElement.md

rotemmiz:
Amaze!

michaelgmcd:
@rotemmiz @LeoNatan Are we good to go?

rotemmiz:
Thank you!

michaelgmcd:
Any timeframe on when this will be released?

rotemmiz:
After a few inconsistencies with lerna 3, and some faulty release, you can now find your changes in `
`detox@9.1.2`

PR #1048, opened 2018-11-19
Android Sync: Improve RN-timers idling resource such that it would skip over interval-timers

d4vidi:
**Description:**

Though in 5a179cac we've improved the RN-timers idling resource on Android - making it effective for the 1st time, we didn't take repeating timers (i.e. JS intervals) into account.

This branch introduces an iOS-equivalent solution that simply ignores intervals and takes the nearest non-repeating timer as the one to inspect.


d4vidi:
Seems that stress-timeouts are flaky on Android, I'll check.

PR #1053, opened 2018-11-25
Generation: Use git submodule instead of downloading files

DanielMSchmidt:
- [ ] This is a small change 
- [X] Fixed a TODO in the code

---

**Description:**

Previously we download uiautomator and espresso source files manually to generate code on their basis. This caused instability in the CI. 

With this change we use git submodules like for all our other not node dependencies with the expected result of an increased stability. 

As a result of commented out code there was a manual change merged in the generated code, I also provided are solution which generates the basis for it.


rotemmiz:
Whenever you update Espresso to AndroidX Espresso test, let’s add https://github.com/android/android-test/ as a submodule to generate the JS API (the ones added here are dead repos)

Issue #1059, opened 2018-11-30
resetContentAndSettings freezes simulator


andrejborstnik:
#### Description
The `resetContentAndSettings` does not work. It runs all the commands (shutdown, erase, boot), but they no longer work as intended. On `boot` the simulator phone opens, but is frozen. I need to restart the simulator app to get it to properly boot.

If the simulator app is closed and the simulator is opened with `open -a "Simulator" --args -CurrentDeviceUDID <udid>` the simulator resets properly.

Environment: `Detox 9.1.2, xCode/Simulator 10.1 (10B61), Mojave 10.14.1 (18B75)`

#### Steps to Reproduce
Just put `await device.resetContentAndSettings();` in the init of detox. Or run `xcrun simctl shutdown <udid>`, `xcrun simctl erase <udid>` and `xcrun simctl boot <udid>`.

![Hanging](https://giant.gfycat.com/JitteryObviousBluetickcoonhound.gif)

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 9.1.2
* React Native: 0.57.4
* Node: 9.11.2
* Device: iPhone SE
* Xcode: 10.1
* macOS: 10.14.1

#### Device and verbose Detox logs
Does not really need detox to run the commands. Probably the way erase/boot works changed and no longer works in detox.

LeoNatan:
This looks like an Xcode/Simulator app, not Detox. Open an issue with Applet at https://bugreport.apple.com

andrejborstnik:
@LeoNatan I agree it was probably caused by a change in the `simctl`. Whether this is a bug or an intended change I don't know - the documentation is sparse.

But I know two things:
 * The `resetContentAndSettings` does not work
 * If I use `Hardware > Erase All Contents and Settings` it works properly

This indicates erasing can be done correctly. Whether it is possible with `simctl` I don't know. I just wanted to bring attention to this issue, as it probably breaks some tests. 

The "solution" I presented was just to show it is possible to erase properly and is not a major Simulator bug.

PR #1060, opened 2018-11-30
Fixes the resetContentAndSettings command

andrejborstnik:
- [x] This is a small change #1059

---

**Description:**
Fixes `resetContentAndSettings` - no longer freezes the simulator.


rotemmiz:
The current implementation would break parallelism support.

Where do you see this issue now, on Xcode 10.1?

andrejborstnik:
@rotemmiz Yes, on Xcode 10.1. I described the problem in the issue #1059.

LeoNatan:
This looks like a bad solution. We do not control the Simulator app, the user does. Killing it because of an Apple bug does not sound to me like a good idea.

LeoNatan:
Closing this PR. Please open an issue with Apple at https://bugreport.apple.com

PR #1084, opened 2018-12-10
Add `visibility` to by()

EdwardDrapkin:
- [X] This is a small change 

I had a weird case where the easiest way for me to narrow my matchers was visibility.  I looked at the matchers code and noticed that the assertion was already written and was exposed as `waitFor(...).toBeVisible()` but not `element.by.visibility()`.  

I added a function with the following signature: `by.visibility(isVisible:boolean = true)`


rotemmiz:
How would this work? This matcher will find all the visible/invisible views, most of the times it will throw an error for multiple matched views found, right?
It can probably only work in conjunction with other matchers.

But even then, what's the usecase where you have more than one view per id?

rotemmiz:
Coverage on unit tests is missing, and also E2E to test the new API.

EdwardDrapkin:
I added the tests and pushed.

Basically what happened is we are using `react-native-popup-dialog` which renders a dialog off-screen with "Continue" and "Cancel" buttons, and then you use a ref to that element to animate it into display.  On one of the screens, That package does not forward testIDs into its components, so it was impossible to select the "Continue" button, because the only selector I can get it from is `by.label("Continue").withAncestor(...)`, but that matches multiple elements, so adding a `.and(by.visibility())` means that it'll find the only one that's visible.

It's working fine, I just have to do this nonsense for now:

```
            const { VisibleMatcher } = require(`detox/src/${device.getPlatform()}/matchers.js`)
```

EdwardDrapkin:
Yeah, we know, the issue is the app we're building is used to help finance retail purchases, so given the time of year that it is, we've let more than a few bad UI things creep into the app just to get features shipped out the door.   It's basically just a CRUD app that sets up some form submissions, so we're not too worried about performance or accessibility so much as getting people the ability to buy stuff.  The entire interface for that part of the app is scheduled for a rewrite because it's had so much stuff ad-hoc added to it over the last few months, but we can't do that until after Q4 ends.

LeoNatan:
I don't like this change. I always fear that exposing unnecessary API will encourage bad coding, and that is indeed the case here. If Detox gets confused about an element, the user can under special cases (such as when using Read Screen, using Voice Over for visually-impaired users or other accessibility features of the operating system), and I do not wish we encourage such bad practices by enabling them in the API. I understand that in this case, accessibility may not be of importance, but when exposing an API, it has to be considered what abuse it allows.

PR #1085, opened 2018-12-11
Introduce kotlin and reimplement RN-timers idling resource

d4vidi:
**Description:**

1. Introduce Kotlin into detox core (lib)
2. Reimplement the `ReactNativeTimersIdlingResource` in Kotlin, TDD.


d4vidi:
*NOTE* still need to check the effect kotiln would have on a potential detox `.aar`.
*NOTE 2* still need to verify versioning collision and the minimal kotlin version.

d4vidi:
Closing temporarily, waiting for https://jira.wixpress.com/browse/DBS-1103 to resolve.

d4vidi:
Follow-up on the notes:

Seems there are plenty of Kotlin based `.aar`/`.jar`'s [out there](https://github.com/mcxiaoke/awesome-kotlin#libraries). I've taken a deep look into some and it seems that while the delivered bytecode expects the presence of Kotlin elements (e.g. functions return `kotlin.String`), kotlin core is never bundled in. Rather, the hosting app should introduce Kotlin stdlib in its deps and - as long as the version is new enough (tested `1.1.0` in our case), it should all play nice together eventually.

As for versioning -- I worked out detox to default to `1.3.0` but to also expect the version to be defined by the hosting app by aligning a global called `detoxKotlinVersion` to their own kotlin version. We'll add that to the installation instructions and everything should be ok.

d4vidi:
Next up, we have the CI saying: `> Plugin with id 'kotlin-android' not found.` @yershalom and I trying to figure out why gradle doesn't download it automatically.

d4vidi:
@rotemmiz you oughta change the project's settings to deny merging as long as checks haven't succeeded + view approved...

d4vidi:
@yershalom holy c***, giving it a try...

rotemmiz:
@d4vidi what's the verdict? Were you able to solve it?

d4vidi:
@rotemmiz ios builds fail in very weird ways. I'm trying to figure where this is coming from, though in essence this PR has nothing to do with iOS at all.

d4vidi:
@yershalom connectivity issues? -
```
10:20:52 lerna info Installing external dependencies
... ( >4 minutes)
10:25:01 lerna ERR! execute callback with error
10:25:01 lerna ERR! Error: Command failed: npm install
```

d4vidi:
I know what this is about. I'm taking care of it.

d4vidi:
Fixes issue #1115

PR #1089, opened 2018-12-18
Set consistent semicolon style in firstTestContent

CodingItWrong:
- [X] This is a small change 

---

**Description:**

Previously all lines except the last ended with a semicolon, so the generated file did not satisfy linters that either required or disallowed semis

### DIFF
--- a/detox/local-cli/templates/firstTestContent.js
+++ b/detox/local-cli/templates/firstTestContent.js
@@ -16,6 +16,6 @@ const firstTestContent = `describe('Example', () => {
     await element(by.id('world_button')).tap();
     await expect(element(by.text('World!!!'))).toBeVisible();
   });
-})`;
+});`;
 
 module.exports = firstTestContent;


detox/local-cli/templates/firstTestContent.js
```diff
@@ -16,6 +16,6 @@ const firstTestContent = `describe('Example', () => {
     await element(by.id('world_button')).tap();
     await expect(element(by.text('World!!!'))).toBeVisible();
   });
-})`;
+});`;
```
LeoNatan: This change doesn't seem correct to me.

The content of this string is written to file [here](https://github.com/wix/Detox/blob/78575f07560b6f22c75cef28e551db6215ed2cb7/detox/local-cli/detox-init.js#L54), and it should contain that trailing semicolon at the end. No?

CodingItWrong: This is what is outputted currently:

```
describe('Example', () => {
  beforeEach(async () => {
  });

  it('should have welcome screen', async () => {
    await expect(element(by.id('welcome'))).toBeVisible();
  });

  it('should show hello screen after tap', async () => {
    await element(by.id('hello_button')).tap();
    await expect(element(by.text('Hello!!!'))).toBeVisible();
  });

  it('should show world screen after tap', async () => {
    await element(by.id('world_button')).tap();
    await expect(element(by.text('World!!!'))).toBeVisible();
  });
})
```

Note the final line doesn't have a semicolon. This is because the backtick-quoted string doesn't have a final semicolon inside it. After this change, the output is:

```
describe('Example', () => {
  beforeEach(async () => {
  });

  it('should have welcome screen', async () => {
    await expect(element(by.id('welcome'))).toBeVisible();
  });

  it('should show hello screen after tap', async () => {
    await element(by.id('hello_button')).tap();
    await expect(element(by.text('Hello!!!'))).toBeVisible();
  });

  it('should show world screen after tap', async () => {
    await element(by.id('world_button')).tap();
    await expect(element(by.text('World!!!'))).toBeVisible();
  });
});
```

Which correctly has the semicolon at the end.

LeoNatan: Oof I’m sorry, you added the semicolon.  Indeed. Thank you for your contribution!

CodingItWrong: Sweet, thanks for accepting!

PR #1104, opened 2018-12-30
iOS: Add pinch actions

sraikimaxime:
- [ ] This is a small change 
- [x] This change has been discussed in issue #<1105> and the solution has been agreed upon with maintainers. - This is wip, the solution has not been agreed on and I haven't followed the contribution guide yet, sorry. My work started after seeing issues #<589> and #<250>.

---

**Description:**
This is the first step of the implementation of tools allowing to manipulate react-native-maps MapView. I'm opening this PR to have the earliest challenge on how I'm going forward.

##iOS

I've been looking to how EarlGrey proposes to manipulate maps but as far as my understanding goes, it only allows to swipe on it and pinch on it to handle the zoom. Therefore I've been testing to connect expect with earl grey's actionForPinchFastInDirectionWithAngle and actionForPinchSlowInDirectionWithAngle And managed to manipulate the map like in the gif I add to this description.

So far I've been able to zoom in, zoom out and move around in the react native maps thanks to these actions.

##Android
Haven't started yet :)

pinch-and-swap


sraikimaxime:
Updating description as I implemented the functions inside the generation process.

Still have a few things to handle: 
- Fix the "double" type as it appears it should be NSNumber
- Reach 100% coverage to pass the CI

I also added a lot of type into "supported types" but I don't really understand what it is supposed to mean when we declare types are "supported" should I do something more inside other parts of the code ?

sraikimaxime:
I did the remapping from double to NSNumber and the generated GreyActions are working OK but I'm still not sure of how good a practice it is :)

sraikimaxime:
I realized I had generated a bunch of functions I didn't really know about by adding their arguments type to the supported types. Could you tell me more about what it does mean to add something in there, what kind of check am I supposed to do on variables before "supporting" them ? I need to support "double" but it also adds another action along with mines.

rotemmiz:
Seems like you managed with unit tests.

I also saw you removed many types from iOS generator adapter, any reason for that?

rotemmiz:
+ this PR is great! Its a very welcome addition. thanks!

rotemmiz:
Making it 100% safe to merge will require addition of new E2E tests that will actually test this new API.
This test needs to have no new dependencies on the test project, as we want to keep it clean from external noise.

sraikimaxime:
@rotemmiz 
"I also saw you removed many types from iOS generator adapter, any reason for that?" 
- Yes, it looked to me that they were duplicates in the list :)

"Making it 100% safe to merge will require addition of new E2E tests that will actually test this new API.
This test needs to have no new dependencies on the test project, as we want to keep it clean from external noise."
- I was thinking about it :) plan to work on a e2e test with a map this week end !

I still have a question: What is the aim of the the supported types array? Am I supposed to test some stuff before adding "double" in it ? I Also noted that the generator fails to generate a few functions because their type is not supported I'd love to remove these warnings but I don't want to manipulate these "supported types" before understanding what they mean :)

rotemmiz:
Let's add @DanielMSchmidt to the conversation

sraikimaxime:
Hello @DanielMSchmidt I'm planning on working on e2e tests this week end and would love to have your feedback before I start implementing a e2e project with a react native maps as e2e test, what do you think ?

Also I still don't really understand what is the aim of the "supportedTypes" array, if you know more about it i'd love to read from you :)

Thank you for your time and help anyway.

generation/core/generator.js
```diff
@@ -70,6 +70,9 @@ module.exports = function getGenerator({
       if (methodArg === null) {
         console.error(method);
       }
+      if (!supportedTypes.includes(methodArg.type)) {
```
DanielMSchmidt: I think we have a separate map for this, which gives you the methods that could not be generated for each class due to which types.

DanielMSchmidt:
I looked through it, seems good to me. Thank you for your patience 😇

generation/core/generator.js
sraikimaxime: Right ! I'll take it off

sraikimaxime:
@DanielMSchmidt Thank you very much for your review ! About the two questions I asked above:

- What is the use of supported type ? Should I add to supported types the ones that prevent the generator to create functions (several functions are blocked)
- Do you think it would be a good idea to create a blank project with a react-native-maps in order to do the e2e test for this action ?

generation/core/generator.js
DanielMSchmidt: 👍

DanielMSchmidt:
> What is the use of supported type?

We have a two-way contract: native side - client side. Both sides need to be able to handle the input of a type. We prevent generating methods for the client side that we can not use on the native side.

> Do you think it would be a good idea to create a blank project with a react-native-maps in order to do the e2e test for this action 

As @rotemmiz already said we would very much prefer a test case without an external dependency

sraikimaxime:
@rotemmiz  @DanielMSchmidt. So I did add the ios e2e tests, now looking for android but having an issue with enable aapt2, it looks like its not supported anymore this year :) Would love to fix this but I might need a little bit of help, any advice ?

sraikimaxime:
Hello @rotemmiz @DanielMSchmidt, I'm having trouble fixing the e2e tests on android as some of them look flaky. Here the stressTimeouts first test run three times in a row:
image

I don't really understand why it would break like this from time to time, do you have any advice ?

sraikimaxime:
I've been investigating and it seems these tests have been switched from :ios: to cross platform in this [PR](https://github.com/wix/Detox/commit/5a179cac4b0ac2195bc3949df1fbb226bbba8f0d) It looks the CI was red and became green on a rerun, I also noticed the artifacts were not updated at the time. I'm re switching them to :ios: only for now, what do you think ?

sraikimaxime:
@rotemmiz @DanielMSchmidt, Hello guys, I've been adding some docs :) Do you see something missing in order to merge this first step ?

In order to continue on the Android implementation, I have been looking at how it is done in Espresso and it looks like most people use UIAutomator in order to perform two swipes to realize the pinch gesture. If it looks ok to you guys I'll start implementing this ! :)

sraikimaxime:
@d4vidi Hello I'm adding you to the conversation as it looks you are pretty active these days :)

sraikimaxime:
up @LeoNatan @rotemmiz @DanielMSchmidt cf [My previous message](https://github.com/wix/Detox/pull/1104#issuecomment-453862099)

DanielMSchmidt:
Sorry for the delay @sraikimaxime, I'm going to take a look right now 💪

docs/Guide.Contributing.md
```diff
@@ -120,13 +120,15 @@ launchctl setenv PATH $PATH
...
-2. Update the snapshots with: `npm run verify-artifacts:ios -- -u`.
-3. Re-run Android tests with `npm run e2e:android`.
-4. Update the snapshots with: `npm run verify-artifacts:android -- -u`.
-5. Add the snapshots to your Git commit:
+of expected artifacts. It is usually done in seven steps:
+
+1. In `detox/test` project, build the ios project with `npm run build:ios`.
+2. Run all end-to-end tests on iOS with `npm run e2e:ios`.
+3. Update the snapshots with: `npm run verify-artifacts:ios -- -u`.
+4. In `detox/test` project, build the ios project with `npm run build:android`
```
DanielMSchmidt: I think this line was added, did you mean to build the android project with this line?
```suggestion
4. In `detox/test` project, build the android project with `npm run build:android`
```

detox/test/e2e/12.animations.test.js
```diff
@@ -51,7 +51,7 @@ describe('Animations', () => {
       await expect(element(by.id('UniqueId_AnimationsScreen_afterAnimationText'))).toNotExist();
     });
 
-    it(`should wait during delays shorter than 1.5s (driver: ${driver})`, async () => {
+    it(`:ios: should wait during delays shorter than 1.5s (driver: ${driver})`, async () => {
```
rotemmiz: Why did you disable tests for Android?

detox/test/e2e/09.stress-timeouts.test.js
```diff
@@ -4,7 +4,7 @@ describe('StressTimeouts', () => {
     await element(by.text('Timeouts')).tap();
   });
 
-  it('should handle a short timeout', async () => {
+  it(':ios: should handle a short timeout', async () => {
     await element(by.id('TimeoutShort')).tap();
```
rotemmiz: Why did you disable tests for Android?

detox/test/scripts/__snapshots__/verify_artifacts_are_not_missing.ios.test.js.snap
```diff
@@ -4,10 +4,10 @@ exports[` 1`] = `
 Array [
```
rotemmiz: We stopped testing artifacts file snapshots, no need to change it. PR might need a rebase/merge from master.

rotemmiz:
Thanks for this PR!
There are a few minor issues with it though, mentioned in comments.

detox/test/scripts/__snapshots__/verify_artifacts_are_not_missing.ios.test.js.snap
sraikimaxime: Ok removing it + unupdating the doc :)

detox/test/e2e/12.animations.test.js
sraikimaxime: as I said in [this comment](https://github.com/wix/Detox/pull/1104#issuecomment-452407831) these tests look flaky? cause looks like coming from [here](https://github.com/wix/Detox/pull/1104#issuecomment-452413650)

detox/test/e2e/09.stress-timeouts.test.js
sraikimaxime: I'll try to run them from updated master and let you know the result

detox/test/e2e/12.animations.test.js
sraikimaxime: I'll try to run them from updated master and let you know the result

detox/test/scripts/__snapshots__/verify_artifacts_are_not_missing.ios.test.js.snap
sraikimaxime: Still, shouldn't we remove this snapshot in this case ? if so I can do it in this PR :)

sraikimaxime:
@DanielMSchmidt Thank you ! I have added a small doc on APIRef.ActionsOnElement.md is there another place where you'd expect documentation ?

DanielMSchmidt:
seems fine to me

rotemmiz:
hey @sraikimaxime,

The PR looks good and I would like to merge it in, but still seems like artifact snapshot tests are deleted. We don't currently use them, but deleting them shouldn't be a part of this PR.

Please revert their deletion, and we'll merge it in.
Thanks for this PR!

sraikimaxime:
Thank you @rotemmiz  ! :) Should I open a new PR to remove the artifacts ?

PR #1083, opened 2018-12-09
 Allow Regular Expressions in Jest Runner

RCiesielczuk:
- [x] This change has been discussed in issue #1081  and the solution has been agreed upon with maintainers.

---

**Description:**

Jest allows you to [provide a regex pattern](https://jestjs.io/docs/en/cli#jest-regexfortestfiles) in order to run tests. Unfortunately I am unable to get the tests to run when the regex pattern is passed to Jest.

In the Jest documentation it's mentioned that:

>Depending on your terminal, you may need to quote this argument: jest "my.*(complex)?pattern"

The proposed solution wraps the file argument in quotation marks to in order to support regex patterns along side specifying a file path.

PR #1109, opened 2019-01-06
Switch CLI to yargs

### LINKED ISSUE #1132: pass in files to the cli like jest and mocha do by default
Currently to specify a specific file you have to do this

```bash
detox test --file 'some/file.js'
```

to test a folder you have to do this

```bash
detox test --specs 'some/folder'
```

and there's a PR open to allow for regex #1083 which you will be able to do like this

```bash
detox test --file 'some/**/*.js'
detox test --specs 'some/**/*.js'
```

Both Mocha and Jest have their setups to already support this by simply passing in all extra files at the end of the commands without having to specify `--file` or `--specs`.

```
jest 'something/**/*.js'
mocha 'something/**/*.js'
```

Detox can simplify it's arguments and make it more consistent with both of these testing libraries by following their pattern since commander already supports this simple feature.

This is pretty easy to implement all you would have to do is change this [line](https://github.com/wix/Detox/blob/master/detox/local-cli/detox.js#L7) to be

```bash
.command('test [testPattern...]]', 'Initiating your test suite')
```

and then deprecate the `--file` and `--specs` so that it shows they're deprecated but will still work.

DanielMSchmidt:
Switch commander interface out with [`yargs`](https://github.com/yargs/yargs/), which is a more reliable interface and the following awesome features:

- We can abstract common patterns (which config do I have, which platform am I supposed to run, etc) into nice middlewares, making the actual CLI code more human readable (To be done, if we agree that it makes sense)
- We can test each CLI command, which is really nice to be sure we don't break anything while improving the user experience
- It supports CLI plugins out of the box (which I think is cool because it allows third parties to make the CLI more powerful without writing their own wrappers with shell outs
- Yargs allows bash completion, I will investigate how this works, it was not my top prio
- [Extendable configs from files and package.json](https://github.com/yargs/yargs/blob/master/docs/api.md#extends-keyword) like with babel presets. 
- We can write a lot of validation inside the yargs real already, making this outer bound of our system as rigorous as possible, so we don't have to check everything 100 times within the main code base.

Resolves #1132.
Addresses #1238.


DanielMSchmidt:
@rotemmiz This PR is done now. It enables users to use the detox cli via environment variables, over the package.json and via configuration file. 

The CI seems to be flaky, so 🤷‍♂️

DanielMSchmidt:
Also to be done: https://github.com/wix/Detox/pull/1083#issuecomment-456713449

detox/local-cli/run-server.test.js
```diff
@@ -0,0 +1,33 @@
+describe('run-server', () => {
+  it('shows help text', async () => {
```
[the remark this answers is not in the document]
DanielMSchmidt: We could use [middlewares](https://github.com/yargs/yargs/blob/master/docs/advanced.md#middleware), but I would suggest moving this into a separate PR

detox/local-cli/test.js
```diff
@@ -0,0 +1,298 @@
...
+    describe: 'Override the device name specified in a configuration. Useful for running a single build configuration on multiple devices.'
+  }
+};
+
+module.exports.handler = function main(program) {
+  program.artifactsLocation = buildDefaultArtifactsRootDirpath(program.configuration, program.artifactsLocation);
+
+  clearDeviceRegistryLockFile();
+
+  if (!program.configuration) {
```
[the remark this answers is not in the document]
DanielMSchmidt: It does through [middlewares](https://github.com/yargs/yargs/blob/master/docs/advanced.md#middleware), but I would suggest to move it into a separate PR :)

detox/local-cli/test.js
```diff
@@ -0,0 +1,298 @@
...
+
+    return fallback;
+  }
+
+  function camelToKebabCase(string) {
+    return string.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
+  }
+
+  function runMocha() {
+    const loglevel = program.loglevel ? `--loglevel ${program.loglevel}` : '';
```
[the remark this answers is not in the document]
DanielMSchmidt: 🤷‍♂️ Not that I'm aware of, maybe we should have a map between our flags and mocha or jest flags.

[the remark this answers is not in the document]
rotemmiz: Interesting...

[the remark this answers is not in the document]
DanielMSchmidt: I still hate the fact that mocha has no mechanism like jest for having environment. This would be way more natural from a jest perspective than the wrapping that we currently do

DanielMSchmidt:
@rotemmiz This is ready for review now

detox/setupJest.js
```diff
@@ -0,0 +1,26 @@
+const yargs = require('yargs');
```
rotemmiz: Does this file contains only test helper functions? if it does, let's move it to test folder.

docs/APIRef.DetoxCLI.md
```diff
@@ -34,6 +34,53 @@ detox [options] [command]
 | --- | --- |
 | -h, --help | Output usage information |
 
+## Configuration Options
+
+Every flag can be passed in different ways, this section will show how `detox test --record-videos failing` can be expressed differently. Please note that you can use any combination of the usages below.
```
rotemmiz: I'd explain how the config is being prioritized (or did you already explain it and missed?)

detox/package.json
```diff
@@ -30,7 +30,7 @@
   "devDependencies": {
     "eslint": "^4.11.0",
     "eslint-plugin-node": "^6.0.1",
```
rotemmiz: remove "commander" from dependencies!

rotemmiz: 🎉

detox/cli.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
```
rotemmiz: Why did you prefer putting this here and not inside `/local-cli` ?

detox/local-cli/init.js
```diff
@@ -1,27 +1,43 @@
...
 
-if (program.runner) {
-  main(program);
-} else {
-  program.help();
+module.exports.handler = function main(argv) {
+  console.log("init handler", argv)
+  switch (argv.runner) {
+    case "mocha":
+    createMochaFolderE2E();
```
rotemmiz: please indent the code

detox/local-cli/test.js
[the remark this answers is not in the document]
rotemmiz: great!!

detox/setupJest.js
DanielMSchmidt: The `detox/test` path won't work as it's a separate project with a package.json and the module resolution won't get for example yargs as a package. I could move it in a `detox/__tests__` folder if you like, that should work

rotemmiz: yes, that's what I meant, `detox/test` is an e2e project

DanielMSchmidt: 👍

detox/cli.js
DanielMSchmidt: I can not put `.` as `commandDir` and I did not want to have this extra level of nesting 🤷‍♂️

DanielMSchmidt:
@rotemmiz Answered and fixed everything 💪

detox/cli.js
rotemmiz: Why not?
```diff
-  "detox": "local-cli/detox.js"
+  "detox": "cli.js"
```

This is a part of the changes you made in this PR, why can't it be reverted?

DanielMSchmidt: Because of [this part](https://github.com/wix/Detox/pull/1109/files#diff-aeb1fd007adaeb0ed56930b012da8ef1R11) in my changes. At least as far as I understand it. Feel free to try it, though.

rotemmiz:
@DanielMSchmidt I moved cli.js to local-cli.js.
I am still having a weird issue I can not trace:
running `detox test --configuration ios.sim.release` results in
```
node_modules/.bin/mocha e2e --opts e2e/mocha.opts --configuration ios.sim.release  --no-colors    --grep :android: --invert  --record-logs none --take-screenshots none --record-videos none --artifacts-location "artifacts/ios.sim.release.2019-02-07 20-17-13Z"  --testRunner mocha --configurations [object Object]
```

I couldn't find where `--configurations` is being added to `program`. The only place it exists in is:
```
const config = require(path.join(process.cwd(), 'package.json')).detox;
```
But I can't see how that is applied into `program`.
Your help is appreciated here :)

rotemmiz:
The issue seems bigger than the purposed fix in a0582e308cfea47b29f12f3cd77b68f7663b57c2.
Currently `--testRunner` is also leaked into `collectExtraArgs`.
Seems like these are all parameter of the detox config from package.json. I couldn't find where this is injected, but it probably should be filtered from `collectExtraArgs` as well.

DanielMSchmidt:
@rotemmiz Ah yeah, that explains why I did not catch it in the very first moment, when this feature is not enabled, interesting 🤔 I'll think a bit about how to fix this, I mean we basically would also need to be carful about all other input options, so just filtering them out might not be the best option

rotemmiz:
Both `confiurations` and `test-runner/testRunner` are not CLI args, they are configuration parameters, they are somehow leaked into `program`.

DanielMSchmidt:
@rotemmiz I think I got it, at least theoretically. We could create a new instance of a yargs program without any configuration within collectExtraArgs, then it should work just fine.

Issue #1125, opened 2019-01-22
Resuming the app with a url causes detox to hang and time-out


d4vidi:
#### Description
On Android, resuming a background app with a supported url (provided as an arg to `launchApp`) results in a timeout. This distinctively happens in case the app activity has been set `launchMode=singleInstance`.

```
     Error: Error: Could not launch intent Intent { act=android.intent.action.VIEW dat=mytestscheme://test-string flg=0x10000000 cmp=com.somevendor/com.example.MainActivity } within 45 seconds. Perhaps the main thread has not gone idle within a reasonable amount of time? There could be an animation or something constantly repainting the screen. Or the activity is doing network calls on creation? See the threaddump logs. For your reference the last time the event queue was idle before your activity launch request was 1548160484943 and now the last time the queue went idle was: 1548160534341. If these numbers are the same your activity might be hogging the event queue.
      at Client.execute (node_modules/detox/src/client/Client.js:80:28)
      at InvocationManager.execute (node_modules/detox/src/invoke.js:11:33)
      at EmulatorDriver.deliverPayload (node_modules/detox/src/devices/drivers/AndroidDriver.js:145:36)
      at Device.launchApp (node_modules/detox/src/devices/Device.js:87:33)
      at Context.it.only (e2e/15.urls.test.js:31:18)
      at <anonymous>
```

**The timeout takes place albeit the activity and expected content clearly residing in the foreground.**

#### Steps to Reproduce
The following combination of a test and a main-activity definition will result in the aforementioned timeout:

`AndroidManifest.xml`:
```xml
     <activity
        android:name=".MainActivity"
        android:label="@string/app_name"
        android:launchMode="singleInstance">
        <intent-filter>
            <action android:name="android.intent.action.MAIN" />
            <category android:name="android.intent.category.LAUNCHER" />
            <category android:name="android.intent.category.DEFAULT" />
        </intent-filter>
        <intent-filter>
            <action android:name="android.intent.action.VIEW" />
            <category android:name="android.intent.category.DEFAULT" />
            <data android:scheme="mytestscheme"/>
        </intent-filter>
      </activity>
```

Test (JS):
```javascript
const url = 'mytestscheme://test-string';
await device.launchApp({newInstance: true});
await device.sendToHome();
await device.launchApp({newInstance: false, url: url});
```

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 9.1.2
* React Native: 0.51
* Node: 9.11.2
[...]

PR #1128, opened 2019-01-23
Fix issues related to single-task activities

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1125.

**Description:**

Address the problem specified in #1125.
The root cause resides in what appears to be an Espresso bug (still need to validate this). This PR introduces a workaround, along with associated improvements around the current confusion the android driver currently has around activities launching. I suppose that can be further improved, but one step at a time.

### DIFF
--- a/detox/test/android/app/src/main/AndroidManifest.xml
+++ b/detox/test/android/app/src/main/AndroidManifest.xml
@@ -16,7 +16,8 @@
       android:allowBackup="true"
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
-      android:theme="@style/AppTheme">
+      android:theme="@style/AppTheme"
+    >
 
       <!--
        The main, default activity; Note: always keep launchMode 'standard'
@@ -53,7 +54,7 @@
         <intent-filter>
             <action android:name="android.intent.action.VIEW" />
             <category android:name="android.intent.category.DEFAULT" />
-            <data android:scheme="detoxtesturlscheme"/>
+            <data android:scheme="detoxtesturlscheme.singletask"/>
         </intent-filter>
        </activity>
 
--- a/detox/test/android/app/src/main/AndroidManifest.xml
+++ b/detox/test/android/app/src/main/AndroidManifest.xml
@@ -42,7 +42,7 @@
 
       <!--
        MainActivity flavor that only runs in a 'singleInstance' launch mode.
-       Executed instead of the main by the test app given the "forceSingleTask" arg to instrumentation.
+       Executed instead of the main by the test app given a "androidSingleTaskActivity" arg to instrumentation.
        -->
       <activity
         android:name=".SingleInstanceActivity"
--- a/detox/test/android/app/src/main/AndroidManifest.xml
+++ b/detox/test/android/app/src/main/AndroidManifest.xml
@@ -47,14 +47,14 @@
        -->
       <activity
         android:name=".SingleInstanceActivity"
-        android:label="@string/app_single_task_name"
+        android:label="@string/app_name_single_instance"
         android:launchMode="singleInstance"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
         android:windowSoftInputMode="adjustResize">
         <intent-filter>
             <action android:name="android.intent.action.VIEW" />
             <category android:name="android.intent.category.DEFAULT" />
-            <data android:scheme="detoxtesturlscheme.singletask"/>
+            <data android:scheme="detoxtesturlscheme.singleinstance"/>
         </intent-filter>
        </activity>
 
--- a/detox/test/android/app/src/main/AndroidManifest.xml
+++ b/detox/test/android/app/src/main/AndroidManifest.xml
@@ -17,9 +17,14 @@
       android:label="@string/app_name"
       android:icon="@mipmap/ic_launcher"
       android:theme="@style/AppTheme">
+
+      <!--
+       The main, default activity; Note: always keep launchMode 'standard'
+       -->
       <activity
         android:name=".MainActivity"
         android:label="@string/app_name"
+        android:launchMode="standard"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
         android:windowSoftInputMode="adjustResize">
         <intent-filter>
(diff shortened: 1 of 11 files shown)


detox/test/e2e/15.urls.test.js
```diff
@@ -1,23 +1,29 @@
...
     await device.launchApp({newInstance: true});
     await device.sendToHome();
-    await device.launchApp({newInstance: false, url: url});
-    await expect(element(by.text(url))).toBeVisible();
+    await device.launchApp({newInstance: false, url: testUrl});
+    await expect(element(by.text(testUrl))).toBeVisible();
+  });
+
+  it(':android: device.launchApp() with a URL should trigger url handling in activities running in single-task mode', async () => {
+    await device.launchApp({newInstance: true, androidSingleTask: true});
```
rotemmiz: Can't we detect task mode through the intent instead of asking the user to pass this metadata?

d4vidi: The user would never do this. This is internal so that detox could rerun this specific test with its custom single-task activity.

d4vidi: Come to think of it, perhaps I should do this duplication with the other url tests...

d4vidi: err there are some typos there, I'll fix and force-push

d4vidi: ok done with the dups.

d4vidi: You're right, but I couldn't think of anything either...

d4vidi: The `launchMode` cannot be set programatically

LeoNatan: Why not a unit test?

d4vidi:
On 2nd thought, I could have an idea on how to get rid of `forceSingleTask`. I'll give it a go and we'll see.

d4vidi:
@rotemmiz @LeoNatan CI looks ok (red due to ios flakiness). Waiting for your feedback.

detox/test/android/app/src/main/AndroidManifest.xml
```diff
@@ -16,25 +16,50 @@
...
+      <activity
+        android:name=".SingleInstanceActivity"
+        android:label="@string/app_single_task_name"
+        android:launchMode="singleInstance"
+        android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+        android:windowSoftInputMode="adjustResize">
+        <intent-filter>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.DEFAULT" />
+            <data android:scheme="detoxtesturlscheme.singletask"/>
```
rotemmiz: `launchMode` is `singleInstance` but the intent filter name is `singletask` it may be a bit confusing.

detox/test/android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,4 @@
 <resources>
     <string name="app_name">Detox Test</string>
+    <string name="app_single_task_name">Detox Test (SingleTask)</string>
```
rotemmiz: singleInstance

detox/test/e2e/15.urls.test.js
```diff
@@ -1,23 +1,46 @@
...
   });
 
-  it('device.openURL({url: url}) should trigger open url handling in app when app is in foreground', async () => {
-    const url = 'detoxtesturlscheme://such-string';
-    await device.launchApp({newInstance: true});
-    await device.openURL({url: url});
-    await expect(element(by.text(url))).toBeVisible();
+  const withAndroidSingleTaskActivityArgs = () => ({
+    url: 'detoxtesturlscheme.singletask://such-string',
+    launchArgs: Object.freeze({ androidSingleTaskActivity: true }),
```
rotemmiz: singleInstance

What is the reason for freezeing the object?

detox/test/e2e/15.urls.test.js
```diff
@@ -1,23 +1,46 @@
...
-    await device.launchApp({newInstance: false, url: url});
-    await expect(element(by.text(url))).toBeVisible();
+  [
+    {
+      platform: '',
+      ...withDefaultArgs(),
+    },
+    {
+      platform: 'android',
+      ...withAndroidSingleTaskActivityArgs(),
```
rotemmiz: singleInstance

detox/test/e2e/15.urls.test.js
rotemmiz: why do you need `androidSingleTaskActivity` as a param here? isn't the custom url enough?

detox/test/e2e/15.urls.test.js
```diff
@@ -1,23 +1,46 @@
...
+    }
+  ].forEach(({platform, url, launchArgs}) => {
```
rotemmiz: I'm inclined to say that having explicit tests without looping, param preparation, and conditions is much more readable.
+ adding `android` to the platform name -> test name is a bit confusing in this way.

I suggest you duplicate the tests, and on Android only tests, such as signleInstance tests, add `:android:` to the test name. these tests will be ignored on iOS runs.

docs/Introduction.Android.md
```diff
@@ -107,7 +107,7 @@ buildscript {
 
 ### 4. Create Android Test class
 
-You need to add the file `android/app/src/androidTest/java/com/[your.package]/DetoxTest.java` and fill it like [this](../detox/test/android/app/src/androidTest/java/com/example/DetoxTest.java), except that you need to change the package to your projects name.
```
rotemmiz: I've been planning on changing this text for too long now... Thank you! <3

detox/test/android/app/src/main/AndroidManifest.xml
d4vidi: I hate `singleInstance`. It isn't very descriptive. But I agree that it could be confusing so I'll change it nonetheless.

detox/test/e2e/15.urls.test.js
d4vidi: I'm afraid the url is insufficient since we need to launch the app in single-instance mode if we really want to test this behaviour out.
As for `Object.freeze()`-ing, that is due to another bug I found; I've introduced a new unit test instead of `Object.freeze()`-ing here, check it out.

detox/test/e2e/15.urls.test.js
d4vidi: I have to disagree in this case...

docs/Introduction.Android.md
d4vidi: `¯\_(ツ)_/¯`

rotemmiz:
♨️

PR #1133, opened 2019-01-25
deprecated `--file` and `--specs`

tjbenton:
- [x] This is a small change 
- [x] This change has been discussed in issue #1132

---

**Description:**

This PR deprecates the `--file` and `--specs`. The reason for this change is that `jest` and `mocha` already have their own implementation of how their default test pattern is set in their configs, and how you can override them in the cli. So instead of changing that to be different this is just using what those libraries already offer. Which in turn makes it more consistent for the end users.

```bash
detox test --spec 'test-ui/**/*.test.jsx'
detox test --file 'test-ui/**/*.test.jsx'
```

now becomes

```bash
detox test 'test-ui/**/*.test.jsx'
```

Also the default `detox test` now defaults to what ever is set in the jest/mocha configurations as the default.

Another advantage to this change is that when `--spec` and `--file` are fully removed from the cli it will allow mocha users to use the `--file` option that already exists in the mocha framework.

```bash
--file <file>                           include a file to be ran during the suite [file]
```

I have also accounted for for the quoting of the glob patterns like was done in #1083 so that functionality isn't removed.

### DIFF
--- a/detox/local-cli/detox-test.js
+++ b/detox/local-cli/detox-test.js
@@ -11,11 +11,12 @@ const DetoxConfigError = require('../src/errors/DetoxConfigError');
 const config = require(path.join(process.cwd(), 'package.json')).detox;
 
 program
+  .usage('[testPatterns ...]')
   .allowUnknownOption()
   .option('-o, --runner-config [config]',
     `Test runner config file, defaults to e2e/mocha.opts for mocha and e2e/config.json' for jest`)
   .option('-s, --specs [relativePath]',
-    `Root of test folder`)
+    `[DEPRECATED] Root of test folder`)
   .option('-l, --loglevel [value]',
     'Log level: fatal, error, warn, info, verbose, trace')
   .option('--no-color',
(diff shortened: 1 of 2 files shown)


rotemmiz:
Hey @tjbenton,
I like the idea, and the fact that this brings back a bit of order into the cli and that fact that `file` and `specs` are kinda overlapping.

Regarding the implementation, I think it would be wrong to merge it before #1109 by @DanielMSchmidt  lands, as it is a total rewrite of Detox cli. 
Do you think you can rework this on top of the new implementation?

tjbenton:
I can rework it on top of the new implementation, I just need to know what branch to base it off of and submit a pr to

rotemmiz:
#1109 , that's `danielmschmidt/move-to-yargs`

tjbenton:
do you want the `--file` and `--spec` to be completely removed? or do you want me to add in the deprecated verbiage in there like I had in this one?

noomorph:
The functionality has entered the #1109. Hence, closing the issue.
@tjbenton, thanks much for the idea! Hope to merge `danielmschmidt/move-to-yargs` within a day.

Issue #1119, opened 2019-01-14
`invocationManager` is cached by nodejs' require cache and prevent multiple client


manuquentin:
#### Description
I'm facing an issue where I can't have multiple detox clients that handle multiple simulators, like :
```js
const Detox = require('detox/src/Detox');
const firstDetox = new Detox({ deviceConfig });
const otherDetox = new Detox({ deviceConfig: { ...deviceConfig, name: 'iPhone 7' } });
```
But in [the driver](https://github.com/wix/Detox/blob/c76dca9113b8e5a97a3f6f54d93e530159cd85f4/detox/src/devices/drivers/IosDriver.js#L15), the `expect.js` file is required directly and so the [`invocationManager`](https://github.com/wix/Detox/blob/master/detox/src/ios/expect.js#L19) is cached for all the subsequent calls and it will be overridden. This is due to the nodejs [require caching mechanism](https://nodejs.org/api/modules.html#modules_caching).

So if you have 2 clients, the `invocationManager` will be the same for both and be set from the last one and so will only run in the second device.

#### Proposition
Encapsulate the `expect` module in a function lilke :
```js
module.exports = function() {
  let invocationManager;
  function setInvocationManager(im) {
    invocationManager = im;
  }
  // ....
  return {
    setInvocationManager,
    //... 
  };
};
```
And in the driver :
```diff
+ const configureExpect = require('../../ios/expect');
// ...

- this.expect = require('../../ios/expect');
+ this.expect = configureExpect();
```

This way the `invocationManager` will not be cached and each driver will have it own `invocationManager`.

Regards,
Emmanuel

rotemmiz:
We never though Detox will be used that way, but it actually makes sense I guess, if you want to create a test that interacts with two or more devices (testing chat between two clients for example), this is required.

 If you are up to the task, you are more than welcome to submit a PR, including a proposed fix, and some documentation explaining what a world of possibilities this opens.

noomorph:
Resolved by #1378 .

PR #1138, opened 2019-02-03
Avoid to cache invocationManager with nodejs require caching

manuquentin:
- [x] This is a small change 
- [x] This change has been discussed in issue #1119 and the solution has been agreed upon with maintainers.

---

**Description:**
I've wrapped `[ios|android]/expect.js` into a function so the nodejs cache is not used each time.


manuquentin:
Tests can't pass this way... I'll figure out another way to avoid `InvocationManager` caching

Issue #1051, opened 2018-11-23
UIPickerView does not contain desired value


petrmrakava:
Sorry guys, but all about iOS Picker was closed without any conclusion. I have exactly the same issue as https://github.com/wix/Detox/issues/798. Also, the  question on stackoverflow has no answer https://stackoverflow.com/questions/50082463/testing-picker-with-detox-in-react-native 

so

if I use "testID":
`await element(by.id('testIdPicker')).setColumnToValue(0, 'en');`
I get
`"Description" : "Interaction cannot continue because the desired element was not found."`
If I use "type":
`await element(by.type('UIPickerView')).setColumnToValue(0, 'en');`
I get
`"Description" : "UIPickerView does not contain desired value!"`

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 9.0.4
* React Native: 0.57.4
* Node: 11.1.0
* Device:*
* Xcode: 10.0
* macOS: 10.13.6

![detox](https://user-images.githubusercontent.com/43537625/48945878-2d281080-ef2c-11e8-8c88-2c90736ed305.png)

LeoNatan:
Can you please create a demo project that reproduces this?

petrmrakava:
Sorry I can not share our project with you, but I created an example based on https://github.com/wix/Detox/issues/798. There are the same issues as I reported.
You can find it here: https://github.com/petrmrakava/Detox/tree/petrmrakava/iosPicker

LeoNatan:
Great, thank you!

fedeerbes:
Hi @LeoNatan, any progress in this? i'm facing the same issue.

LeoNatan:
Sorry, I am busy with other projects at the moment.

LeoNatan:
See this:
https://github.com/facebook/react-native/pull/22537

LeoNatan:
It seems like an RN bug.

fedeerbes:
thanks @LeoNatan i'll be waiting for the next version with that fix.

LeoNatan:
@fedeerbes Can you please test to see if the issue still reproduces? Thanks

revh:
Hi, first of all thanks for the great work.

It seems that the issue still reproduces also with RN 0.59.9

`await element(by.type("UIPickerView")).setColumnToValue(0, "value");`

results in 

`"Description" : "UIPickerView does not contain desired value!" `

noomorph:
Just to make sure, I wonder how this test passes then?

[...]

Issue #778, opened 2018-06-07
Unable to select the date from 'UIPickerView'


zetendra:
#### Description
Unable to select the date from the date picker. I am able to select the time and AM/PM

#### Steps to Reproduce
Its not working for the below case
element(by.type('UIPickerView')).setColumnToValue(0, Fri Jun 8');

Its working for the below
element(by.type('UIPickerView')).setColumnToValue(1, '11');
element(by.type('UIPickerView')).setColumnToValue(2, '22');
element(by.type('UIPickerView')).setColumnToValue(3, 'AM');

Can some one help me on this?

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 7.3.6
* React Native: 0.53.0
* Node: 8.11.2
* Device: iPhone 7
* Xcode: 9.1
* macOS: 
<img width="238" alt="datepicker" src="https://user-images.githubusercontent.com/1446116/41111511-c79ffc44-6a99-11e8-996b-54ae50a04755.png">

rotemmiz:
Try to use the layout inspector and see the actual string rendered in the UI, is it exactly `Fri Jun 8` ?

zetendra:
@rotemmiz  Thank you for responding to my query.

I used iOS inspector to find the exact format of the value, i see it is "Thu, Jun 14"
I tried the same in my code, i didnt work. I am attaching the screen shots and pasting the error

`await datePicker().setColumnToValue(0, `${day}, ${month} ${date}`);`

`     Error: Error: An action failed. Please refer to the error trace below.
Exception with Action: {
  "Action Name" : "Set picker column 0 to value 'Thu, Jun 14'",
  "Element Matcher" : "((kindOfClass('UIPickerView') && !(kindOfClass('RCTScrollView'))) || (kindOfClass('UIScrollView') && ((kindOfClass('UIView') || respondsToSelector(accessibilityContainer)) && ancestorThatMatches((kindOfClass('UIPickerView') && kindOfClass('RCTScrollView'))))))"
}

Error Trace: [
  {
    "Description" : "UIPickerView does not contain desired value!",
    "Error Domain" : "com.google.earlgrey.ElementInteractionErrorDomain",
    "Error Code" : "2",
    "File Name" : "GREYPickerAction.m",
    "Function Name" : "-[GREYPickerAction perform:error:]",
    "Line" : "120"
  }
]`

[...]

PR #1148, opened 2019-02-08
Select date on IOS UIDatePicker with actionForSetDate

matthewrfindley:
- [x] This is a small change 

---

**Description:**

Addresses the following issues:
https://github.com/wix/Detox/issues/778
https://github.com/wix/Detox/issues/1051

## The problem:
Using `setColumnToValue` to select the correct date on a DatePickerIOS does not work as expected.  This [issue](https://github.com/wix/Detox/issues/1051) defines a solution to select a valid value from the `0` column by using the value of the label.  The solution breaks down because the date picker contains the same label for every year in the list.  Attempting to select `Feb 8` for 2019 would result in selecting `Feb 8` for 2006, which is the minimum date in the list.

Example usage of setColumnToValue on DatePickerIOS:
```
await element(by.type('UIPickerView')).setColumnToValue(0,"Feb 6");
```

react-native DatePickerIOS renders the date in the UIDatePicker as 2 labels.  The first being the 3 letter month abbreviation and day, the second being the 3 letter day abbreviation.  Using `setColumnToValue` breaks down when trying to select the correct date since the implementation of Earl Grey's `GREYPickerAction` selects only the first label see [here](https://github.com/google/EarlGrey/blob/master/EarlGrey/Action/GREYPickerAction.m#L97).  Even if Earl Grey were able to compare using all the labels with a list item, a UIDatePicker would be limited to only 7 years worth of dates.  This would not be maintainable since we are not in control of the default range for UIDatePicker, nor is it acceptable to force users to limit the list in their application.

## Solution:
Expose an existing Earl Grey action called [actionForSetDate](https://github.com/google/EarlGrey/blob/master/EarlGrey/Action/GREYActions.m#L272) within Detox.


LeoNatan:
Thanks for submitting this PR. Will review shortly.

detox/ios/Detox/GREYActions+Detox.h
```diff
@@ -0,0 +1,16 @@
...
+//  Created by Matt Findley on 2/7/19.
+//  Copyright © 2019 Wix. All rights reserved.
+//
+
+@import Foundation;
+#import <EarlGrey/EarlGrey.h>
+
+@interface GREYActions (Detox)
+
++ (id<GREYAction>)detoxSetDatePickerDateIOSOnly:(NSString *)dateString withFormat:(NSString *)dateFormat;
```
LeoNatan: Let's not call this "IOSOnly". Should be `detoxSetDatePickerDate:withFormat:`.

detox/ios/Detox/GREYActions+Detox.m
```diff
@@ -0,0 +1,23 @@
...
+//  Copyright © 2019 Wix. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "GREYActions+Detox.h"
+#import <EarlGrey/GREYActions.h>
+
+@implementation GREYActions (Detox)
+
++ (id<GREYAction>)detoxSetDatePickerDateIOSOnly:(NSString *)dateString withFormat:(NSString *)dateFormat
```
LeoNatan: Likewise, `detoxSetDatePickerDate:withFormat:`.

detox/test/e2e/17.datePicker.test.js
```diff
@@ -3,12 +3,17 @@ describe(':ios: DatePicker', () => {
...
     it('datePicker should trigger change handler correctly', async () => {
       await element(by.type('UIPickerView')).setColumnToValue(1,"6");
       await element(by.type('UIPickerView')).setColumnToValue(2,"34");
       await expect(element(by.id("timeLabel"))).toHaveText('choosenTime is 6:34');
     });
 
-  });
-  
+    it('can select dates on a UIDatePicker', async () => {
+      await element(by.type('UIDatePicker')).setDatePickerDateIOSOnly('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ");
```
LeoNatan: This should be `setDatePickerDate` and documentation should be updated to state it is for iOS only, as are the current picker view APIs.

detox/src/ios/expect.js
```diff
@@ -185,6 +186,14 @@ class ScrollColumnToValue extends Action {
   constructor(column,value) {
     super();
     this._call = invoke.callDirectly(GreyActions.actionForSetPickerColumnToValue(column,value))
+    this._call = invoke.callDirectly(GreyActions.actionForSetPickerColumnToValue(column, value));
```
LeoNatan: I think this line was duplicated by mistake?

matthewrfindley: 📗 ` [fe3a615](/wix/Detox/pull/1148/commits/fe3a6152e98f1d8fe352c78fde9ded4fe7d2168e)`

detox/test/e2e/17.datePicker.test.js
matthewrfindley: 📗 9470d65

detox/ios/Detox/GREYActions+Detox.m
matthewrfindley: 📗 1d6913c

detox/ios/Detox/GREYActions+Detox.h
matthewrfindley: 📗 1d6913c

matthewrfindley:
@LeoNatan https://jenkins-oss.wixpress.com/job/multi-detox-pr/920/ - Failure seems unrelated to any of the recent changes. 🤷‍♂️

LeoNatan:
Looks good. CI is flaky.

LeoNatan:
@rotemmiz Let's merge?

rotemmiz:
Android e2e hangs, wonder why

detox/test/src/Screens/DatePickerScreen.js
```diff
@@ -1,47 +1,49 @@
 import React, { Component } from 'react';
-import {
-  Text,
-  View,
-  StyleSheet,
-  DatePickerIOS
-} from 'react-native';
+import { Text, View, StyleSheet, DatePickerIOS } from 'react-native';
```
rotemmiz: You can't add `DatePickerIOS` on a shared component, it needs to run on both iOS and Android.
You might need to split this screen to .ios.js and .android.js.
This is probably the reason why the e2e hangs, the app crashes.

rotemmiz:
Please see my comment, I think I found the reason for Android E2E's hangs

detox/test/src/Screens/DatePickerScreen.js
LeoNatan: But that’s not a new change. How did it work until now?

rotemmiz: ho, you're right, I haven't noticed that

Issue #1149, opened 2019-02-10
device.launchApp() ends too soon on Android


d4vidi:
#### Description
On Android, it happens at times that `device.launchApp()` returns (promise resolves) before the app is indeed fully launched on the device. This is applicable for apps that only take some time to load, unlike the detox test-app itself.

#### Steps to Reproduce
The behavior is flaky, so only running a reproducing use case as a stress (i.e. in a loop) can reproduce. In essence, this is the conceptual code:

```js
await device.relaunchApp({ delete: true });
while (true) {
   await expect(element(by.id(someMainScreenTestID))).toBeVisible();
   await device.launchApp({newInstance: true});
}
```

Sooner than later, the `someMainScreenTestID` element will not be visible when reaching the `expect` line and the test will fail.

#### Detox, Node, Device, Xcode and macOS Versions
* Detox: `10.0.5`
* React Native: `0.57`
* Node: `9.11.2`
* Device: Pixel 2 API 28 emulator
* IDE: Android Studio 3.3
* macOS: 10.13.6 (HS)

#### Device and verbose Detox logs
N/A

PR #1150, opened 2019-02-10
Idle wait before ready

d4vidi:
- [x] This is a small change 
- [x] This change has been discussed in issue #1149 and the solution has been agreed upon with maintainers.

---

**Description:**
A fix for issue #1149. Putting the testable refactoring aside, the suggested solution is to add an `Espresso.onIdle()` before sending `ready` messages over the web socket.


detox/android/detox/src/main/java/com/wix/detox/DetoxActionsFacade.kt
```diff
@@ -0,0 +1,11 @@
+package com.wix.detox
+
+import android.content.Context
+import android.support.test.espresso.Espresso
+import com.wix.detox.espresso.UiAutomatorHelper
+
+class ActionsFacade {
```
rotemmiz: What did we do wrong to have a `Facade` as a class name? 😣

detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
```diff
@@ -102,6 +101,14 @@ public void onAction(final String type, final String params, final long messageI
         handler.post(new Runnable() {
             @Override
             public void run() {
+
+                final ExternalAction externalAction = externalActions.get(type);
```
rotemmiz: What does External mean?

detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
```diff
@@ -213,14 +212,34 @@ public void run() {
...
     }
 
     @Override
     public void onClosed() {
         stop();
     }
 
+    private void initReactNativeIfNeeded() {
+        if (ReactNativeSupport.isReactNativeApp()) {
+            ReactNativeCompat.waitForReactNativeLoad(reactNativeHostHolder);
```
rotemmiz: We need to think of a way to support native apps with no RN as dependency. Currently native apps crash in classNotFound when running a detox test .
But this is probably a different issue

detox/android/detox/src/main/java/com/wix/detox/DetoxActionsFacade.kt
d4vidi: Lol, I don't care much about names. It's something that simplifies and aggregates other things have spread around in other places (mainly helpers), instead of providing its own direct implementation (which is a facade basically 🤔). In our case, it also makes actions unit-testable. wdyt?

detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
d4vidi: actions that are externally invoked (by the test runner).

rotemmiz: Do you mean `device` API?

detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
d4vidi: It's different but sounds important nonetheless. Do we have an actual issue? I don't understand why this would happen, though, because we only check for the `ReactApplication` class' existence

detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
d4vidi: hmm not necessarily, i think

PR #1165, opened 2019-02-21
Detox Instruments Integration (Phase 1)

LeoNatan:
This PR aims to add initial integration between Detox and Detox Instruments.

Detox will load or use the already loaded Profiler framework and call its APIs to record performance metrics for each test. Each recording will be collected by the artifacts manager.

Discussion/Tasks

- [x] Determine optimal way of collecting performance metrics (one recording for entire suite vs one recording per describe vs one recording per test)
- [x] Artifacts manager should collect the recordings
- [x] Finalize the actual performance collection config
- [x] Prerequisite: complete and merge yargs PR #1109 
- [x] Add Detox argument to enable recording
- [x] Handle app termination from Detox API—need to stop recording before using `xcrun simctl terminate`.
- [x] Handle crashing or otherwise app exist—need to stop recording
- [x] JS coverage
- [x] Change Detox server to notify testee in case tester is disconnected (to support ⌃+C)
- [x] Handle cases when Detox Instruments is not installed on user's system


LeoNatan:
@noomorph Could you please take the "Artifacts manager should collect the recordings" and "Change Detox server to notify testee in case tester is disconnected" tasks? Thanks

LeoNatan:
For now, every test has its own recording file. I first sanitize the fullName string for file name: `:/\?%*|"<>` characters turn into `_`. To that the `.dtxrec` extension is added, and these files are saved under the app's Documents folder.

detox/ios/DetoxUserNotificationTests/TestableAppDelegate.swift
```diff
@@ -33,13 +33,13 @@ class TestableAppDelegate: NSObject, UIApplicationDelegate {
 	var userNotificationTriggerType = TestableAppDelegateNotifcationTriggerType.unknown
 	var userNotificationTitle : String?
 	
-	func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
+	func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey : Any]? = nil) -> Bool {
```
LeoNatan: Why was this reverted?

PR #1169, opened 2019-02-25
Add LongPressAtPoint action for iOS

AngelVenchev:
This is a small change.

** Description **

This is adding the LongPressAtPoint action for iOS that is already available in EarlGrey.


AngelVenchev:
The build failed. I need to add tests. I will get to that.

noomorph:
@AngelVenchev, I think at the very least you need to supplement your changes with an extra test in [03.actions.test.js](https://github.com/wix/Detox/blob/master/detox/test/e2e/03.actions.test.js).

PR #1177, opened 2019-02-26
PR (Issue #1176) - Stop Forcing -gpu host on android emulators

### LINKED ISSUE #1176: Stop forcing -gpu host when launching android emulator
#### Description
Launching an android emulator via the `detox test` command always uses the cli parameters `-gpu host` which will fail in some scenarios where host gpu emulation is not available. This flag should not be included allowing the emulator to launch with whatever gpu mode is specified in the config.ini. 

I believe the default behavior when an explicit gpu mode is not specified in the config.ini is to use the host gpu, anyway.

#### Steps to Reproduce
Run the `detox test -c <config>` command. Use a configuration of type `android.emulator` and select an avd name with a `config.ini` file that has an explicit hw.gpu.mode set.

e.g.
```
hw.gpu.mode=swiftshader_indirect
```

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: `10.0.7`
* React Native: `0.57.8`
* Node: `11.6.0`
* Device: `Android Emulator`
* Xcode: N/A 
* macOS: N/A
* Windows: `Windows 10 Pro v1703`

#### Device and verbose Detox logs
For the sake of brevity, only including the relevant error message portion of the log:
```
detox[1400] ERROR: [Emulator.js/SPAWN_FAIL] path\to\emulator -verbose -gpu host -no-audio @DEFAULT_EMULATOR failed with code 3221225477

emulator: ERROR: OpenGLES emulation failed to initialize. Please consider the following troubleshooting steps:

1. Make sure your GPU drivers are up to date.

2. Try software rendering: Go to Extended Controls > Settings > Advanced tab and change "OpenGL ES renderer (requires restart)" to "Swiftshader".

Or, run emulator from command line with "-gpu swiftshader_indirect". 3. Please file an issue to https://issuetracker.google.com/issues?q=componentid:192727 and provide your complete CPU/GPU info plus OS and display setup.
```

RyanThomas73:
Resolves #1176 

Only adds the `-gpu [gpu mode]` flags to the android `emulator` command if the `--headless` parameter is specified or if an explicit `--gpu [gpu mode]` cli argument is provided.

### DIFF
--- a/detox/src/devices/android/Emulator.js
+++ b/detox/src/devices/android/Emulator.js
@@ -29,11 +29,15 @@ class Emulator {
   async boot(emulatorName) {
     const emulatorArgs = _.compact([
       '-verbose',
-      '-gpu', this.gpuMethod(),
       '-no-audio',
       argparse.getArgValue('headless') ? '-no-window' : '',
       `@${emulatorName}`
     ]);
+
+    const gpuMethod = this.gpuMethod();
+    if(gpuMethod) {
+      emulatorArgs.push('-gpu', gpuMethod);
+    }
 
     let childProcessOutput;
     const tempLog = `./${emulatorName}.log`;
@@ -93,9 +97,9 @@ class Emulator {
         default:
           return 'auto';
       }
-    } else {
-      return 'host';
     }
+
+    return undefined;
   }
 }
 
--- a/detox/src/devices/android/Emulator.js
+++ b/detox/src/devices/android/Emulator.js
@@ -86,6 +86,11 @@ class Emulator {
   }
 
   gpuMethod() {
+    const gpuArgument = argparse.getArgValue('gpu');
+    if(gpuArgument) {
+      return gpuArgument;
+    }
+
     if (argparse.getArgValue('headless')) {
       switch (os.platform()) {
         case 'darwin':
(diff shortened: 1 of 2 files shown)


noomorph:
@rotemmiz, looks good to me. Shall we merge?

detox/src/devices/android/Emulator.js
```diff
@@ -93,9 +102,9 @@ class Emulator {
         default:
           return 'auto';
       }
-    } else {
-      return 'host';
     }
+
+    return undefined;
```
d4vidi: @RyanThomas73 for backwards compatibility, please return `host` as the default.

RyanThomas73: @d4vidi Returning host as the default reintroduces the problem where it will always include `-gpu host` with the emulator command preventing the emulator command from honoring an explicit gpu mode specific in the emulator config via `hw.gpu.mode=`

RyanThomas73: We could choose a specific value to map to `undefined` so it uses the emulator config default (e.g. `detox test .... --gpu default`) but I feel its better the way it is in the PR now, the user shouldn't have to explicitly specify default to get the default behavior from the emulator.

d4vidi: Yes, you are right - I was not aware of that.
Nevertheless the issue of back-compatibility still _is_ concerning (keeping the community happy). That being said, I take it that the crucial use cases are of the CI, where typically emulators are run `headless`, which isn't affected in this case. So, bottom line here - I'll approve and merge.
Thank you :)

PR #1179, opened 2019-02-27
Xcode10

yershalom:
- [X] This is a small change 

---

**Description:**
Add `-UseModernBuildSystem=NO` for unit test iOS.
mac_slave_4 is mojave with xcode10.1.

If you'll approve this I'll upgrade the rest of the agents.

https://jenkins-oss.wixpress.com/job/detox-ios-xcode10.1/4/

### DIFF
--- a/scripts/unit.ios.sh
+++ b/scripts/unit.ios.sh
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-set -o pipefail && xcodebuild -project detox/ios/Detox.xcodeproj -scheme Detox -configuration Debug -sdk iphonesimulator build-for-testing -UseModernBuildSystem=NO | xcpretty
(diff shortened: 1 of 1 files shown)


rotemmiz:
Is the build pending on an xcode10 agent?

LeoNatan:
I'm not sure this is needed. Does Detox itself (there is RN in Detox) not support new build system? I don't see why not.

fbartho:
I'm not sure we should disable the new build system globally. The new build system works for us with React-Native 57 (mostly), 58, 59

LeoNatan:
@fbartho Detox itself is a lean framework and never had an issue with the new build system (there is no RN dependency in Detox). 

I’ll close this as I don’t believe it’s needed.

PR #1189, opened 2019-03-01
Resolves #1100

### LINKED ISSUE #1100: Issue with _queryPID function
#### Description
I was faced with the problem with `async _queryPID` function in AdroidDriver.js file.
For example, the `pidof` function returns a value only from the 3rd time, then the previous 2 write errors to the console

`detox[14376] ERROR: [exec.js/EXEC_FAIL, #15] "C:\Users\User_Name\AppData\Local\Android\Sdk\platform-tools\adb -s emulator-5554 shell "ps | grep \"com\.example\.nameOfProject$\""" failed with code = 1, stdout and stderr:
`
Although this function works, and the application starts, the tests go. Is it possible to solve this problem?
At the moment I had to change the line in AdroidDriver.js file:
`if (waitAtStart) {
      await sleep(500);
    }`
to
`if (waitAtStart) {
      await sleep(3000);
    }`

* Detox: "9.1.2"
* React Native: "0.57.4"
* Node: "10.14.2"
* Device: "Nexus_5X_API_26"

noomorph:
Resolves #1100

### DIFF
--- a/detox/src/devices/android/ADB.js
+++ b/detox/src/devices/android/ADB.js
@@ -131,7 +131,7 @@ class ADB {
   async pidof(deviceId, bundleId) {
     const bundleIdRegex = escape.inQuotedRegexp(bundleId) + '$';
 
-    const processes = await this.shell(deviceId, `ps | grep "${bundleIdRegex}"`).catch(() => '');
+    const processes = await this.shell(deviceId, `ps | grep "${bundleIdRegex}"`, {silent: true}).catch(() => '');
     if (!processes) {
       return NaN;
     }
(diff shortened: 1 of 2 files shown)


rotemmiz:
Is there a way to add a test to make sure this is really resolved in CI?

noomorph:
@rotemmiz , compare the logs.

Master:
screen shot 2019-03-03 at 12 15 52 pm

PR:
screen shot 2019-03-03 at 12 15 47 pm

The only weird thing is why iOS is so flaky in CI... 😕

PR #1190, opened 2019-03-03
Fix main.jsbundle does not exist

msand:
- [x] This is a small change


LeoNatan:
I'm not sure I understand this change. Please explain.

msand:
On travis this is required to alias node to the correct node.js version, otherwise the packager fails to create the bundle, as it's calling the previously existing node instead.

LeoNatan:
Merged, thanks!

Issue #1173, opened 2019-02-26
Migrate to Android jetpack (androidx)


d4vidi:
#### Description

Switch to importing `Espresso` and the test framework from Jetpack (i.e. import from `androidx.*`).

d4vidi:
WIP in (stale?) branch `androidx-upgrade`

d4vidi:
Also upgrade the engine: https://github.com/wix-private/wix-one-app-engine/pull/787

PR #1193, opened 2019-03-05
(Android) Migrate instrumentation support lib to androidx libs

d4vidi:
- [x] This is a breaking change, requiring a major version upgrade
- [x] This change has been discussed in issue #1173 and the solution has been agreed upon with maintainers.

---

**Description:**

With respect to #1173 --
This is phase 1 of a 2-phase procedure towards fully migrating:
1. Switch to `androidx` and provide a migration guide, bundled with the regeneration of the JS Espresso-contract classes based on this androidx test [repo](https://github.com/android/android-test/tree/androidx-test-1.1.0) and tag
2. Introduce the androidx test repo as a submodule and take the contracts from there, similar to what we already do with EarlGrey.

Reasons not to do step 2 right now are that this change is already pretty big as-is, and that step-2 will not be breaking in the users' perspective.


d4vidi:
@LeoNatan @rotemmiz I need you guys to have a look at this - go over the docs, in the very least.

generation/utils/downloadEspresso.js
```diff
@@ -3,7 +3,5 @@ const downloadFile = require('./downloadFile');
 
 module.exports = function downloadEspresso(fullyQualifiedClass) {
   const path = fullyQualifiedClass.replace(/\./g, '/');
-  return downloadFile(
-    `http://android.googlesource.com/platform/frameworks/testing/+/android-support-test/espresso/core/src/main/java/${path}.java?format=TEXT`
-  );
+  return downloadFile(`https://raw.githubusercontent.com/android/android-test/androidx-test-1.1.0/espresso/core/java/${path}.java`);
```
lucasnlm: Is there any new Google link to be used here?

d4vidi: Hey @lucasnlm, not sure I understand the question. This is a link to raw content of this repo: https://github.com/android/android-test

d4vidi:
Update: I've decided to skip part 2 of this upgrade, as I don't see too much point in adding the submodule unless we effectively compile it along with Detox.
Currently the status is that we grab the code from google's android-test repo on github from a label that is perfectly aligned to the version we use for the testing artifacts we defined as dependencies in our `build.gradle`.

PR #1210, opened 2019-03-17
Restore UIAutomator as an external file in Android generation

d4vidi:
_Starting as a PR to see how the CI behaves_


d4vidi:
This fails at times due to a CI error where the agent's date is out of sync (and is effectively pretty old). In turns, this breaks `https` requests as the SSL certificate is considered too old by the other end. Thus the generator cannot fetch the raw source file.

Issue #1230, opened 2019-03-26
iOS keyWindow.rootViewController would be null when app is driven by Detox.


Ponyets:
**Describe the bug**
In some legacy iOS code, some developers use `[UIApplicaiton sharedApplication].keyWindow.rootViewController` as a convenient way to get a VC instance.
When the app is driven by Detox, this call would return null.

It's because Detox creates a DTXTouchVisualizerWindow without any ViewController, then set it's windowLevel to a very high value and make it the keyWindow.

Surely, `[UIApplicaiton sharedApplication].keyWindow.rootViewController` isn't good way to access a VC instance, but it's widely used.

**To Reproduce**
1. Write a code block in an iOS app, like the callback of a button click, which tries to get a VC instance by calling `[UIApplicaiton sharedApplication].keyWindow.rootViewController`  and use the VC instance to present another ViewController.
2. Build and run the app in simulator, then manually test the functionality of the above code block. The new ViewController would be presented. Everything is fine.
3. Write a test case for this code block, and run it with Detox.
4. The code block doesn't work properly, the new ViewController wouldn't be presented.

**Expected behavior**
The functionality keeps exactly the same as it's in manual test.

**Environment (please complete the following information):**
 - Detox: 11.0.2
 - React Native: 0.55.4
 - Node: 11.9.0
 - Device: iOS Simulator & iPhone Xs Max
 - Xcode: 10.1
 - iOS: 12.1
 - macOS: 10.14.3

LeoNatan:
Hello,

I am looking at the code, and touch indicator window is not made key at any point. 
Perhaps the legacy applications forgot to call `makeKeyAndVisible` or `makeKeyWindow`?
From what you describe, the issue can occur if any system window (of which there can be any number of them—such as the dictionary when you tap "Look Up").

I am not sure about your PR either, because it can be often incorrect when several `UIWindow` instances exist (more complex apps). I don't think it's the right move to break Detox for complex apps in favor of apps that cannot be bothered to use iOS APIs correctly.

LeoNatan:
I am playing around, trying to reproduce a case where the touch indicator window is key, and cannot. In some cases, it seems `[[UIApplication sharedApplication] keyWindow]` returns `nil`, which makes sense.

Ponyets:
Hello Leo,

Thanks.

I just double checked our code, and confirmed that we did call `makeKeyAndVisible` of the AppDelegate's window in the `didFinishLaunchingWithOptions` method. 

[...]

PR #1231, opened 2019-03-26
Fix keyWindow.rootViewController nullability on iOS

Ponyets:
- [x] This is a small change 

---

**Description:**

***Issue: https://github.com/wix/Detox/issues/1230***

Some iOS developer use `[UIApplicaiton sharedApplication].keyWindow.rootViewController` as a convenient way to get a VC instance. Surely, this isn't suggested, but it's widely used anyway.
When the app is driven by Detox, this call would return null. It's because DTXTouchVisualizerWindow has a very high windowLevel and doesn't has any ViewController.

This PR fixes this issue by avoid making DTXTouchVisualizerWindow the keyWindow.

### DIFF
--- a/detox/ios/Detox/DetoxAppDelegateProxy.m
+++ b/detox/ios/Detox/DetoxAppDelegateProxy.m
@@ -23,6 +23,11 @@
 	return self;
 }
 
+- (BOOL)_canBecomeKeyWindow
+{
+	return NO;
+}
+
 @end
 
 @class DetoxAppDelegateProxy;


LeoNatan:
I don't think this is the right solution. It can be often incorrect when several `UIWindow` instances exist (more complex apps). The PR assumes that there is only one user window, but that could be a very wrong assumption. 

I don't think it's the right move to break Detox for complex apps in favor of apps that cannot be bothered to use iOS APIs correctly.

If I could suggest something else, but I don't have time to check. If you could test with that instead, I will be happy to include it. `UIWindow` has a private method called `- (BOOL)_canBecomeKeyWindow;`

So if we define this method for the touch indicator window, like so:
```objc
- (BOOL)_canBecomeKeyWindow
{
	return NO;
}
```

It could solve the problem people are seeing?

detox/ios/Detox/DetoxAppDelegateProxy.m
```diff
@@ -23,6 +23,11 @@ - (UIWindow *)overlayWindow
 	return self;
 }
 
+- (UIViewController *) rootViewController
```
LeoNatan: Let's try with
```objc
- (BOOL)_canBecomeKeyWindow
{
	return NO;
}
```
instead and see if that helps.

Ponyets: It works. Thanks for you suggestion. I will update this PR soon.

LeoNatan:
Thanks for checking! I will accept it.

LeoNatan:
Thanks for checking! I will accept it.

PR #1237, opened 2019-03-27
Update guide for Detox config for React-native v0.59

huyb1991:
React-native `v0.59` change build structure, so, the Detox configurations `binaryPath` should be update too


noomorph:
It is an interesting observation, but I think that this notice is in the wrong place.
1. Theoretically, it should belong to the Guide.Migration.md
2. Besides, you should check for all the other places with an outdated path in an example and add a notice there too.

LeoNatan:
The package.json in that doc file is an example. People should fill in the correct path for themselves. In this case, it doesn't seem like a serious change that should warrant such a notice.

@noomorph The migration guide is use to explain how to upgrade from one version of Detox to another, not how to adapt to changes in React Native itself.

PR #1244, opened 2019-03-30
ci: attempt to fix SSL issue with another curl version

noomorph:


noomorph:
The issue indeed was caused by broken time sync.

Issue #1247, opened 2019-04-01
`detox test` fails without status code


tancredi:
#### Description
After [introducing `yargs`](https://github.com/wix/Detox/pull/1109) in v12.0.0 it seems the detox cli exits normally (no error code) even when tests are failing

I must be missing something because I found no way at all to run the tests using the local cli in a way that allows me to detect failures and block the execution of further commands - which seems pretty fundamental for a tool that has to run on CI

Please let me know whether I'm missing something here - will be happy to submit a PR otherwise

#### Steps to Reproduce
In a codebase containing failing tests
```
detox test && echo "Success"
```
Always outputs `Success` after execution

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 12.1.1
* React Native: 0.59.2
* Node: 11.10.1
* Device: -
* Xcode: 10.1
* macOS: 10.14

noomorph:
Thanks for pointing that out. I'm checking.

LeoNatan:
The PR has landed, so I’m closing the issue. Thanks

noomorph:
@tancredi, fixed in detox@12.1.2 .

tancredi:
Cheers

PR #1248, opened 2019-04-01
Exit process with error status code when a cli command fails

tancredi:
Will wait for #1247 to be discussed

- [x] This is a small change 
- [x] This change has been discussed in issue #1247 and the solution has been agreed upon with maintainers.

---

**Description:**
Exits with status code `1` when an actual error (not a `yargs` validation error) is passed to the `.fail` callback

### DIFF
--- a/detox/local-cli/cli.js
+++ b/detox/local-cli/cli.js
@@ -26,6 +26,7 @@ yargs
         logger.error(line);
       }
       console.error('');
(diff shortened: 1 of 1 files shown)


LeoNatan:
@noomorph

noomorph:
Thank you, @tancredi ! Your prompt feedback is very welcome.

PR #1269, opened 2019-04-10
feat(log): dump pending network requests on test timeout

noomorph:
**Description:**

Currently, it is not easy to track what requests to `testee` have never been resolved, leading to timeout errors in tests.

To make it Detox debug-friendly in those cases, I suggest dumping pending network requests if:

1. a test has failed with a timeout (mocha offers this as an API, in Jest I scan for timeouts in failed expectations)
2. in `detox.cleanup()`, before we clean up the client.

Although it does not cover all the cases (e.g. cannot catch infinite `detox._client.cleanup()` but looks like a good start IMO.

### DIFF
--- a/detox/src/client/AsyncWebSocket.js
+++ b/detox/src/client/AsyncWebSocket.js
@@ -59,7 +59,7 @@ class AsyncWebSocket {
 
     return new Promise(async(resolve, reject) => {
       message.messageId = messageId || this.messageIdCounter++;
-      this.inFlightPromises[message.messageId] = {resolve, reject};
+      this.inFlightPromises[message.messageId] = {message, resolve, reject};
       const messageAsString = JSON.stringify(message);
       this.log.trace({ event: 'WEBSOCKET_SEND' }, `${messageAsString}`);
       this.ws.send(messageAsString);
@@ -94,6 +94,12 @@ class AsyncWebSocket {
       return false;
     }
     return this.ws.readyState === WebSocket.OPEN;
+  }
+
+  resetInFlightPromises() {
+    _.forEach(this.inFlightPromises, (_, messageId) => {
+      delete this.inFlightPromises[messageId];
+    });
   }
 
   rejectAll(error) {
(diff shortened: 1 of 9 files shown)


LeoNatan:
@d4vidi

LeoNatan:
I think this is a bad approach. Sometimes, as we have discovered, things take more than 60 seconds. For example, first launch of an app that has to go through the packager and that takes a long time in case of a large app. Detox already has a configurable timeout setting, and that should be respected, rather than a hardcoded number.

detox/src/client/AsyncWebSocket.js
```diff
@@ -12,6 +12,7 @@ class AsyncWebSocket {
     this.inFlightPromises = {};
     this.eventCallbacks = {};
     this.messageIdCounter = 0;
+    this.timeout = 60000;
```
LeoNatan: Pls use the Detox configurable timeout, not magic numbers.

noomorph: Okay, I hope I'll find a way to do it.

rotemmiz: We discussed a way to use the test runner timeout parameter, it can be added to the test runner adapter for easier extraction

noomorph:
Oh, I have another idea on how to rewrite this logic while keeping the feature. We might want to run a self-check on `detox.cleanup()` which will print out the stuck requests.

noomorph:
I found how to do it right. :smiley: 
I'll implement soon.

noomorph:
@LeoNatan, I've completely changed the approach.

Now we just dump pending network requests if any, in 2 cases:
1) a test has failed with a timeout (mocha offers this as an API, in Jest I scan for timeouts in failed expectations)
2) in `detox.cleanup()`, before we clean up the client.

It does not cover all the cases (e.g. cannot catch infinite `detox._client.cleanup()` but looks like a good start to me.

PR #1276, opened 2019-04-13
fix(android): install APK with a file path that needs escaping

mrgoodrich:
- [x] This is a small change 

---

**Description:**

Windows users with spaces in their APK path face an issue where the path is broken by spaces. I faced this issue and tried escaping the space, but it turned out the solution was to remove the backslash escapes I added and use quotation marks, like "path". Apostrophes like 'path' don't work.

For reference,
https://android.stackexchange.com/questions/185951/script-to-pull-from-directory-containing-spaces

This fixed the problem for me and I'm now able to run `detox build`.

I considered other places in the code where the path could be wrapped in quotes, but this seemed like the simplest and only place necessary.

### DIFF
--- a/detox/src/devices/android/AAPT.js
+++ b/detox/src/devices/android/AAPT.js
@@ -2,6 +2,7 @@ const _ = require('lodash');
 const Environment = require('../../utils/environment');
 const path = require('path');
 const exec = require('../../utils/exec').execWithRetriesAndLogs;
+const escape = require('../../utils/pipeCommands').escape.inQuotedString;
 const egrep = require('../../utils/pipeCommands').search.fragment;
 const fsext = require('../../utils/fsext');
 
@@ -23,7 +24,7 @@ class AAPT {
   async getPackageName(apkPath) {
     await this._prepare();
     const process = await exec(
-      `${this.aaptBin} dump badging "${apkPath}" | ${egrep("package: name=")}`,
+      `${this.aaptBin} dump badging "${escape(apkPath)}" | ${egrep("package: name=")}`,
       undefined, undefined, 1
     );
 
--- a/detox/src/devices/android/ADB.js
+++ b/detox/src/devices/android/ADB.js
@@ -102,7 +102,7 @@ class ADB {
   }
 
   async install(deviceId, apkPath) {
-    apkPath = `"${apkPath}"`;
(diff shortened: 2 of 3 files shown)


noomorph:
Can you elaborate about the backslash escapes? Do you mean Detox tries somewhere to escape spaces with backslashes on Windows?

And, in any case, please use `escape` utility methods to work around the edge cases with quotes inside the directory names. You will see the reference in the very top of the file you edited.

mrgoodrich:
Hey Yaroslav. Sorry about not explaining the issue very well.

My (simplified) APK path is `C:\Users\me\Google Drive\myproject\android\app\build\outputs\apk\debug\app-debug.apk`. I think it depends on the shell, but with PowerShell on Windows in my case, the command currently attempted as a part of `detox test` is `C:\Users\me\AppData\Local\Android\sdk\platform-tools\adb -s mydevice install -r -g -t C:\Users\me\Google Drive\myproject\android\app\build\outputs\apk\debug\app-debug.apk`. This results in the message `adb: failed to stat Drive\myproject\android\app\build\outputs\apk\debug\app-debug.apk: No such file or directory` failing `detox test`. Notice how it splits the path into two arguments at "Google Drive" due to the space.

In my case with PowerShell, the only solution is to wrap the path to the APK in quotation marks. Neither escaping the space nor wrapping with apostrophes fix the parsing of the path.

I tried both of the functions in escape from pipeCommands.js, but they didn't fix the issue.
* `escape.inQuotedString(apkPath)` returns `C:\Users\me\Google Drive\myproject\android\app\build\outputs\apk\debug\app-debug.apk`
* `escape.inQuotedRegexp(apkPath)` returns `C:\\Users\\me\\Google Drive\\myproject\\android\\app\\build\\outputs\\apk\\debug\\app-debug\.apk`

noomorph:
Well, I understood the issue immediately, I was just wondering if you had this or similar issue in any other places of Detox, with escaping whitespace-containing paths on Windows. That's it.

As for `escape.inQuotedString`, it works but in the current implementation you just need to add quotes manually: `"${escape.inQuotedString(apkPath)}"`. Maybe that was the thing that confused you.

Try to test manually the case when your folder name contains not only spaces but double quotes.

PR #1290, opened 2019-04-16
feat: undeprecate "specs" in package.json

noomorph:
**Description:**

In `detox@12.1.0` there was a premature deprecation of `"specs"` property in detox section of `package.json`. It appeared inconvenient for anyone who prefers terse commands like `detox test` over verbose `detox test e2e`. From now on, "specs" property will act as a fallback if a specific test folder has not been specified. For example (1):

```sh
detox test   # translates to: mocha <...args> e2e
detox test e2e/01.sanity.test.js  # translates to: mocha <...args> e2e/01.sanity.test.js
```

For the reference, previously it used to work like (2):

```sh
detox test   # translates to: mocha <...args> e2e
detox test e2e/01.sanity.test.js  # translates to: mocha <...args> e2e e2e/01.sanity.test.js
```

The problem of the latter was that you could not focus (override defaults) on a specific test file without having to use `specs` in package.json and deprecated `--specs` in the command line (3):

```sh
detox test --specs e2e/01.sanity.test.js  # overrides specs in package.json
# mocha <...args> e2e/01.sanity.test.js
```

Since (2) has been introduced very recently and did not function well, I suggest to consider this rather a minor version than a breaking change.

Your considerations are welcome.

P. S. As for `file` deprecation warning, it has been reworked and looks like on the screenshot:

Screenshot from 2019-04-16 12-20-49


LeoNatan:
Any luck finding why it showed an unrelated error when specs was not provided?

noomorph:
@LeoNatan, yes, it was a bug in `mocha@^5.0.0`. After upgrading to `mocha@^6.0.0`, that weird error has disappeared, now it is comprehensible:

Screenshot from 2019-04-17 08-44-07

docs/Guide.Migration.md
```diff
@@ -5,24 +5,61 @@ title: Migration Guide
...
+
 ## Migrating from Detox 12.0.x to 12.1.x
 
 This is not a breaking change yet, but starting from `detox@12.1.0` you'll start seeing warnings like:
 
 ```
-WARN:  [test.js] Deprecation warning: "file" and "specs" support will be dropped in the next Detox version.
-WARN:  [test.js] Please edit your package.json according to the migration guide: https://wix.to/I0DOAK0 
+detox[21201] WARN:  [deprecation.js] Beware: -f, --file will be removed in the next version of Detox.
+detox[21201] WARN:  [deprecation.js] See the migration guide: https://wix.to/I0DOAK0
```
LeoNatan: Why is this shortened and pointing the same document? Shortened URLs should be used in Detox docs, especially not if tracking is used.

detox/test/package.json
```diff
@@ -26,7 +26,6 @@
     "mocha": "^5.0.0"
```
LeoNatan: Let's update to 6 if 5 has issues.

docs/Guide.Migration.md
noomorph: Makes sense. Done.

detox/test/package.json
noomorph: Looks like a glitch. Check master, if we already have `^6.1.3` there in `detox/test/package.json`.
Anyways, thanks, I saw some of the example projects had `4.0.0`, so your comment is relevant!

noomorph: Resolved in #1308 and #1298.

PR #1295, opened 2019-04-16
docs(APIRef.DetoxCLI) fix: missing - for options

guhungry:
Add missing - for --configuration and --gpu

- [x] This is a small change 

---

**Description:**
Long option should be -- instead of - or else it will be error on parameter when run command

Screen Shot 2019-04-17 at 00 25 11
Reference: https://travis-ci.org/guhungry/react-native-photo-manipulator/jobs/520852289


LeoNatan:
Thanks!

Issue #1279, opened 2019-04-15
App uninstalls and reinstalls for every test file when using Jest


hammadj:
**Describe the bug**
After refactoring our tests to use Jest instead of Mocha, Detox started uninstalling / reinstalling the app for every test file. This greatly slows down our tests and is not ideal. 

**To Reproduce**

- [x] I have tested this issue on the latest Detox release and it still reproduces

Any Detox setup with Jest will exhibit this behaviour while Mocha setups will not. 

**Expected behavior**
The app should not be reinstalled for every test file unless the user tells detox to do so. 

**Environment (please complete the following information):**
 - Detox: 12.2.1
 - Jest: 24.4.0
 - React Native: 0.58.6
 - Node: 8.11.2
 - Device: iPhone X
 - Xcode: 10
 - iOS: 12.1
 - macOS: Mojave

**Device and Verbose Detox Logs**
Example of it reinstalling between tests

```
detox[14058] DEBUG: [exec.js/EXEC_CMD, #11] /usr/bin/xcrun simctl shutdown FCC3C45C-8F27-44F6-8E0B-AB5A271D1532
detox[14058] DEBUG: [exec.js/EXEC_TRY, #11] Shutting down FCC3C45C-8F27-44F6-8E0B-AB5A271D1532...
detox[14058] DEBUG: [DetoxServer.js/DISCONNECT] role=tester, sessionId=386a6d9f-0821-fab3-433c-e7e7d1428729
detox[14058] DEBUG: [DetoxServer.js/DISCONNECT] role=testee, sessionId=386a6d9f-0821-fab3-433c-e7e7d1428729
PASS e2e/tests/Contacts.spec.ts (380.165s)
  Checkins
    ✓ User adds a new contact to the space via checkin screen|user can add contact note on new checkin (53889ms)
    ✓ User adds a new contact via the select contact screen (49995ms)
    ✓ user performs checkin with existing contact|user adds inspection note on checkin|user edits contact while checking it in (84214ms)
  Contacts
    ✓ user can search for existing contact by name and mobile (26458ms)
    ✓ Create new contact and check contact is created (65661ms)
detox[14058] DEBUG: [exec.js/EXEC_SUCCESS, #11] FCC3C45C-8F27-44F6-8E0B-AB5A271D1532 shut down
detox[14058] INFO:  [DetoxServer.js] server listening on localhost:50929...
[...]

PR #1328, opened 2019-04-23
docs: add an example of using --reuse within Jest test runner

noomorph:
- [x] This is a small change 

---

**Description:**

Inspired by people asking questions, like in https://github.com/wix/Detox/issues/1279 .


LeoNatan:
I think this needs to be updated with the new combined demo projects as there were changes to this area. Is this PR still needed?

noomorph:
Tough question. On one hand, I would not be very happy to add this to e2e scripts on Jenkins CI. On the other hand, this covers a scenario people were asking how to do. This solution has a caveat, and that makes me hesitant to merging it to official docs/codebase. There is another issue that also has a more interesting solution. https://github.com/wix/Detox/issues/1331

So let's close.

PR #1333, opened 2019-04-25
feat: record network in Detox Instruments artifacts

noomorph:
- [x] This is a small change 

---

**Description:**

Adds network requests to `*.dtxrec` recordings produced by Detox artifacts subsystem.

### DIFF
--- a/detox/ios/Detox/DetoxInstrumentsManager.m
+++ b/detox/ios/Detox/DetoxInstrumentsManager.m
@@ -17,6 +17,7 @@ DTX_CREATE_LOG_PREFIX(DetoxInstrumentsManager, @"🥶")
 @property (class, nonatomic, strong, readonly) id defaultProfilingConfiguration;
 @property (nonatomic, readwrite) NSTimeInterval samplingInterval;
 @property (nonatomic, readwrite) BOOL recordNetwork;
+@property (nonatomic, readwrite) BOOL recordInternalReactNativeEvents;
 @property (nonatomic, readwrite) BOOL recordThreadInformation;
 @property (nonatomic, readwrite) BOOL collectStackTraces;
 @property (nonatomic, readwrite) BOOL symbolicateStackTraces;
--- a/detox/ios/Detox/DetoxInstrumentsManager.m
+++ b/detox/ios/Detox/DetoxInstrumentsManager.m
@@ -146,7 +146,6 @@ static void (*__DTXProfilerMarkEvent)(NSString* category, NSString* name, __DTXE
 	[config setRecordingFileURL:URL];
 	
 	//TODO: Finalize the actual config for Detox perf recording.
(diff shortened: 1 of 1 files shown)


detox/ios/Detox/DetoxInstrumentsManager.m
```diff
@@ -146,7 +146,6 @@ - (id)_configForDetoxRecordingWithURL:(NSURL*)URL
 	[config setRecordingFileURL:URL];
 	
 	//TODO: Finalize the actual config for Detox perf recording.
-	[config setRecordNetwork:NO];
```
LeoNatan: Set it explicitly to yes.

LeoNatan: It's better to be explicit, not rely on defaults.

LeoNatan:
You wanted to add the JS events too no?

LeoNatan:
`@property(nonatomic, readonly) BOOL recordInternalReactNativeEvents;`

LeoNatan:
Add that property and set it to `YES`.

LeoNatan:
@noomorph  ^

noomorph:
@LeoNatan , thanks for the reminder! 👍

detox/ios/Detox/DetoxInstrumentsManager.m
```diff
@@ -147,6 +147,7 @@ - (id)_configForDetoxRecordingWithURL:(NSURL*)URL
 	
 	//TODO: Finalize the actual config for Detox perf recording.
 	[config setRecordNetwork:YES];
+	[config setRecordInternalReactNativeEvents:YES];
```
LeoNatan: This will not compile. You need to expose the property.

LeoNatan: Also, it's probably best to wrap this with `respondsToSelector:`, because this is a very recent feature and maybe One App has an old profiler.

detox/ios/Detox/DetoxInstrumentsManager.m
```diff
@@ -146,7 +149,14 @@ - (id)_configForDetoxRecordingWithURL:(NSURL*)URL
 	[config setRecordingFileURL:URL];
 	
 	//TODO: Finalize the actual config for Detox perf recording.
-	[config setRecordNetwork:NO];
+	[config setRecordEvents:YES];
+	if ([config respondsToSelector:@selector(setRecordInternalReactNativeEvents:)])
```
rotemmiz: In what cases setRecordInternalReactNativeEvents doesn't exist?

noomorph: When users have older version of Detox Instruments. The supported version was released only yesterday.

PR #1347, opened 2019-04-30
feat: allow disabling of post install with env var

### LINKED ISSUE #1338: Allow disabling of post install with env var
In certain situations, like our quick internal deployment CI, we want builds to happen as fast as possible and since no tests are run we don't want to build the detox framework. Or situations where we only build for release which requires some dev dependencies but not detox.

Building the framework for Detox takes considerable time when it isn't actually used.

Allow and env var to be set and checked in postinstall scripts to disable building the framework is present.

removing the dependency from the project is not really an option as it would impact other automated and manual workflows.

I understand there is something we could maybe do with yarn workspaces but yarn is not a tool we use internally nor on our automated systems.

**Additional context**

noomorph:
Resolves #1338

@rotemmiz, what's your opinion on that issue? I understand what problem they might have and indeed there's no other solution so far as I am concerned - that's because `npm install` is not really flexible, and there is a goal to speed up some builds on @haswalt's CI, we should support that from our codebase.

- [x] This is a small change

### DIFF
--- a/detox/scripts/postinstall.js
+++ b/detox/scripts/postinstall.js
@@ -1,4 +1,4 @@
-if (process.platform === "darwin") {
+if (process.platform === "darwin" && !process.env.DETOX_DISABLE_POSTINSTALL) {
 	require("child_process").execFileSync(`${__dirname}/build_framework.ios.sh`, {
 		stdio: "inherit"
 	});


haswalt:
@rotemmiz any thoughts on this? Looks like a simple change that would greatly improve portability of Detox to CI environments. Especially those that don't support iOS / run release builds.

LeoNatan:
@noomorph Please merge master and once the build passes I will accept it.

noomorph:
@LeoNatan, done. Let's wait for the build.

PR #1348, opened 2019-04-30
Introduce basic Detox Instruments demo app 

rotemmiz:
- [ ] This is a small change 

---

**Description:**
Introduce a demo app that includes an integration of detox-instruments for demo and testing purposes inside detox repo.


examples/demo-react-native-detox-instruments/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "test:ios": "detox test --configuration ios.sim.debug --record-performance all",
+    "e2e:ios": "npm run build:ios && npm run test:ios"
+  },
+  "dependencies": {
+    "react": "16.8.4",
+    "react-native": "0.59.x"
+  },
+  "devDependencies": {
+    "detox": "^12.4.1",
+    "detox-instruments": "1.x.x",
```
LeoNatan: This is unsupported. Please do not include it on any external project.

LeoNatan: The only supported method is `brew cask`.

rotemmiz: I can remove it from the examples folder, but this as a test project

LeoNatan: It doesn’t matter. Detox Instruments should not be installed from npm. It should be installed as described in the developer guide.

LeoNatan:
As discussed, I will make a change to install Detox Instruments from `brew cask`. In order to test custom path, the script will provide the default path as a custom path.

PR #1385, opened 2019-05-15
Launch args

d4vidi:
- [x] This is a small change 
- [x] This change has been discussed in issue #981 and the solution has been agreed upon with maintainers.

---

**Description:**

Fixes issue #981. Based on the original PR by @elyalvarado (set as co-author) - #982, which has been left hanging.

Notes:
- I've introduced support for _all_ launch argument types (i.e. more than just plain strings). This means that I've left iOS behind in that sense. Should be fairly easy to implement once viable.
- Nevertheless I've implemented proper e2e tests for _both_ platforms (iOS - where applicable), in the form of: `launchApp(<some test args>)` > tap the `Launch App` button > `expect()` the launch args' visibility on a dedicated screen (key & value).

Notes re implementation:
- I had to break the technique in `Device.js` - of adding prefixes to the args, implemented by calling a `deviceDriver.defaultLaunchArgsPrefix()` method. Instead, the `device` now passes the args directly through to `launchApp()` of the delegated implementation (`IOSDriver` or `AndroidDriver`), putting this concern in their scope (which I find, btw, a better design practice to begin with).
For that, I added test coverage of the delegated drivers, TDD.

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -38,6 +38,7 @@
     "wtfnode": "^0.8.0"
   },
   "dependencies": {
+    "nodejs-base64": "^1.0.3",
     "bunyan": "^1.8.12",
     "bunyan-debug-stream": "^1.1.0",
     "chalk": "^2.4.2",
--- a/detox/package.json
+++ b/detox/package.json
@@ -38,7 +38,6 @@
     "wtfnode": "^0.8.0"
   },
   "dependencies": {
-    "nodejs-base64": "^1.0.3",
     "bunyan": "^1.8.12",
     "bunyan-debug-stream": "^1.1.0",
     "chalk": "^2.4.2",
--- a/detox/src/devices/drivers/SimulatorDriver.js
+++ b/detox/src/devices/drivers/SimulatorDriver.js
@@ -69,7 +69,9 @@ class SimulatorDriver extends IosDriver {
     await this.applesimutils.uninstall(deviceId, bundleId);
   }
 
-  async launchApp(deviceId, bundleId, launchArgs, languageAndLocale) {
+  async launchApp(deviceId, bundleId, _launchArgs, languageAndLocale) {
+    const launchArgs = this._transformLaunchArgs(_launchArgs);
+
     await this.emitter.emit('beforeLaunchApp', {bundleId, deviceId, launchArgs});
     const pid = await this.applesimutils.launch(deviceId, bundleId, launchArgs, languageAndLocale);
     await this.emitter.emit('launchApp', {bundleId, deviceId, launchArgs, pid});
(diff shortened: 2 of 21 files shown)


d4vidi:
@noomorph wow, I've been meaning to talk to you about this and forgot. Great this we have tests to bite us in the ass (jenkins) is red). Thanks!

detox/test/ios/example/NativeModule.m
```diff
@@ -54,30 +54,63 @@ @implementation NativeModule
...
     UITabBarController* tbc = [UITabBarController new];
     tbc.viewControllers = @[newRoot, newRoot2, newRoot3, newRoot4];
-    
+
     [[delegate window]setRootViewController:tbc];
     [[delegate window] makeKeyAndVisible];
   });
 }
 
+RCT_EXPORT_METHOD(getLaunchArguments:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject)
```
LeoNatan: This should not be here. This logic should not be in a demo project.

detox/test/src/app.js
```diff
@@ -98,6 +100,7 @@ class example extends Component {
             throw new Error('Simulated Crash')
           })}
           {this.renderScreenButton('Shake', Screens.ShakeScreen)}
+          {this.renderScreenButton('Launch Args', Screens.LaunchArgsScreen)}
```
LeoNatan: This should only be rendered on Android or it will crash on iOS.

d4vidi: I changed the test itself to e2e-only, but now also made this render only on Android 👍

LeoNatan: 👍 Thanks

LeoNatan:
Amit, why are you closing and reopening PRs?

detox/package.json
```diff
@@ -38,6 +38,7 @@
     "wtfnode": "^0.8.0"
   },
   "dependencies": {
+    "nodejs-base64": "^1.0.3",
```
rotemmiz: I would like to discourage the usage of an external library for base64 encoding/decoding. It is available in vanilla node:
`return Buffer.from(str, 'base64').toString('utf8');`

Moreover, this specific library is only 34 lines of code, which basically just runs the line I pasted above.

detox/src/devices/drivers/SimulatorDriver.js
```diff
@@ -127,6 +129,13 @@ class SimulatorDriver extends IosDriver {
   async waitForBackground() {
     return await this.client.waitForBackground();
   }
+
+  _transformLaunchArgs(launchArgs) {
```
rotemmiz: What is the reason for switching to `_transformLaunchArgs` instead of using `defaultLaunchArgsPrefix` per platform?
How does it work with Android Instrumentation which needs to have `-e` before an argument?

detox/package.json
d4vidi: awesome, switching-to!

d4vidi: pushed a new commit

d4vidi:
Merging this -- only failure was the crash handling e2e on Android, which I've fixed on a parallel track.

PR #1390, opened 2019-05-16
Add rebuild-framework-cache command to local CLI

LeoNatan:
In most cases, `clean-framework-cache` and `build-framework-cache` are called together. This PR adds `rebuild-framework-cache` to local CLI so that these two steps are easier to accomplish.

Right now, it doesn't replace the existing ones. I don't think `clean-framework-cache` should be removed, but `build-framework-cache` may be unnecessary.

One caveat here is that `rebuild-framework-cache` cleans all cache but only builds for the current Detox. Not sure how bad this is, considering the use-cases.

### DIFF
--- a/detox/local-cli/rebuild-framework-cache.js
+++ b/detox/local-cli/rebuild-framework-cache.js
@@ -1,13 +1,17 @@
+const fs = require('fs-extra');
 const path = require('path');
 const cp = require('child_process');
 const os = require('os');
 const log = require('../src/utils/logger').child({ __filename });
 
-module.exports.command = 'build-framework-cache';
-module.exports.desc = 'MacOS only. Build Detox.framework to ~/Library/Detox. The framework cache is specific for each combination of Xcode and Detox versions';
+module.exports.command = 'rebuild-framework-cache';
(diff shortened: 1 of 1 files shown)


noomorph:
So, let's extend a bit this command, so it cleans only the required cache.
If cleaning all is required, let's add an extra CLI flag

noomorph:
By the way, what if we reintroduce this command in a bit different fashion. 

```
detox framework-cache clean
detox framework-cache clean --all
detox framework-cache build
detox framework-cache build -f
```

This way, it's easier to work with it in terminal. You just erase the ending and complete with another subcommand.

And all the framework cache functionality will live in the same file. It's quite small, in any case.

noomorph:
Anyway, guys, as you wish. That were just my two cents.

LeoNatan:
Suggestions make sense, but I think out of scope for this PR. It's just a quick-fix for an annoyance.

Also, I didn't want to make changes to existing commands for BC reasons.

LeoNatan:
Merged as is. @noomorph Let's discuss improvements when you come back from vacation.

Issue #1369, opened 2019-05-07
Integrate parallel testing using Jest in Detox CI (iOS)


d4vidi:
Introduce parallel test-execution in CI, for iOS, using the latest Jest tweaks made in `12.7.0`.

LeoNatan:
Please wait with this. We had many issues and moved back to mocha. If debugability is still bad, we should remain with mocha.

rawqing:
I had some problems with Mocha : https://github.com/mochajs/mocha/issues/94 ; https://github.com/ampproject/amphtml/issues/15658 , so I had to use jest

PR #1394, opened 2019-05-21
Introduce parallel e2e tests execution using Jest in Detox demo app (iOS only)

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1369

---

### Description:

The content of the PR is twofold:
1. Resolving #1369: Switching to Jest as the test runner. Slightly extra complex here due to coverage reconfiguration -- Jest has the Istanbul reporter built into it, rendering `nyc` useless, thus requiring it's configuration to be replaced with Jest's config-schema.

2. Addressing an interesting find revealed thanks to coverage:  In Jest, rather than in Mocha, the environment's tear-down has been done ungracefully. Namely, the testing used to end before the Detox server had finished closing its connections to the tester and testee.
I'm not 100% sure why this doesn't happen with Mocha, but possibly its because mocha's own tear-down takes longer than Jest's and thus disconnection manages to complete in time (to make a stronger case - essentially adding a `sleep()` in our `afterAll()` in Jest yields the same results).
Nevertheless read below  for info regarding the fix.

### Jest tear down fix:

Important preliminary facts:
- Cleanup sequence is as follows: client (i.e. tester), device (i.e. testee) and then the server.
- On Android, by the time cleanup gets to the server, it has no effective ws connections. Hence, client and device cleanups on Android are effective and fully synchronous.
- On iOS, in server cleanup the server  typically has 1 active connection -- to the device. Hence client clean up works, but - as it turns out, device cleanup doesn't really trigger a socket disconnection. Rather, it just cleans up some internals; Effectively, disconnection relies on the server-side cleanup that follows.
@LeoNatan do you think you could improve the iOS part of the sequence in that sense?

As for the actual fix: I've simply added logic to make `DetoxServer`'s cleanup fully synchronous, forcing it to wait (with a timeout!) for all of its running ws connections to go down before resolving the returning promise. Please review.


detox/src/client/AsyncWebSocket.js
```diff
@@ -71,7 +71,7 @@ class AsyncWebSocket {
   }
 
   async close() {
-    return new Promise(async (resolve, reject) => {
```
d4vidi: this was never needed here, a bit out of scope.

detox/src/server/DetoxServer.js
```diff
@@ -71,8 +77,52 @@ class DetoxServer {
     }
   }
 
-  close() {
+  async close() {
+    if (this._hasConnections()) {
```
d4vidi: The heart of the tear-down matter.
Note that the server web-socket's `close()` allows for giving it a callback that is essentially called when closing end. I've tried it - it just doesn't work (called prematurely).

LeoNatan:
As discussed with @rotemmiz yesterday, let’s wait for @noomorph to return before pushing this. Cleanup is of utmost importance when recording with Detox Instruments. In general, lifecycle is very important. We need to make sure the lifecycle and state machine are correctly managed in Jest, and in Jest when multiple instances exist.

detox/src/server/DetoxServer.js
```diff
@@ -71,8 +77,52 @@ class DetoxServer {
     }
   }
 
-  close() {
+  async close() {
```
d4vidi: The heart of the tear-down matter.
Note that the server web-socket's close() allows for giving it a callback that is essentially called when closing ends. I've tried it - it just doesn't work (called prematurely).

detox/src/server/DetoxServer.js
```diff
@@ -1,6 +1,11 @@
 const _ = require('lodash');
 const WebSocket = require('ws');
 const WebSocketServer = WebSocket.Server;
+const delegatedPromise = require('../utils/delegatedPromise');
+
+const CLOSE_TIMEOUT = 10000;
+const ROLE_TESTER = 'tester';
+const ROLE_TESTEE = 'testee';
```
d4vidi: boyscouting

detox/src/server/DetoxServer.js
```diff
@@ -71,8 +77,52 @@ class DetoxServer {
...
+  _isClosedownPending() {
+    return !!this.pendingClosedown;
+  }
+
+  _resolvePendingClosedown() {
+    this.pendingClosedown.resolve();
+    this.pendingClosedown = undefined;
+  }
+
+  _hasConnections() {
```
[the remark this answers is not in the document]
noomorph: Also, I would suggest renaming `_hasConnections` to `_hasActiveSessions` to make it fit with the terminology in the file.

detox/src/server/DetoxServer.js
noomorph: Let's nevertheless try to update to `ws@^7.0.0`.
We have a too outdated `ws@^1.1.1`, and maybe they have a fix already there.
If so, we won't have to write so much workaround code.

detox/src/utils/delegatedPromise.test.js
```diff
@@ -0,0 +1,48 @@
+describe('Delegated-promise util', () => {
```
noomorph: Might be not needed if we reuse some npm package.

d4vidi:
@noomorph I'm happy to say `ws` version `2.3.1` (latest v2) solves the premature `close()` callback invocation! So I removed most of the logic from the server's impl. Please re-review :-)

d4vidi:
Needless to say, this does not waiver us from further upgrading to newer version. But I must consider this as out of scope, because we are 5 breaking versions away from the latest and need to pick our battles.

Issue #853, opened 2018-07-24
[SimulatorDriver] semantics of device.name has been unexpectedly changed since 7.4.0


bartolkaruza:
#### Description
After upgrading to Detox 8.0.0 from 7.3.3, detox is not able to find the same iOS simulator using the name of the simulator. After downgrading, it works again, with no changes to the simulator config in between.

```
Error: a BeforeAll hook errored, process exiting: e2e/support/index.js:48: Can't find a simulator to match with "DETOX", run 'xcrun simctl list' to list your supported devices.
      It is advised to only state a device type, and not to state iOS version, e.g. "iPhone 7"
```

 Output of `xcrun simctl list`
```
== Devices ==
-- iOS 11.2 --
    DETOX (4E4CECC2-AA47-437F-B9B6-EC5917751329) (Booted)
-- iOS 11.3 --
    iPhone 7 (621A1BEF-EEF1-4017-BE73-A46DDAC3A0DF) (Shutdown)
-- tvOS 11.3 --
-- watchOS 4.3 --
== Device Pairs ==

``` Detox config in package.json
"detox": {
    "test-runner": "cucumber",
    "configurations": {
      "iphone": {
        "binaryPath": "ios/build/Build/Products/Debug-iphonesimulator/xx.app",
        "build": "yarn ios-detox-build",
        "type": "ios.simulator",
        "name": "DETOX"
      },
      "android": {
        "binaryPath": "./android/app/build/outputs/apk/debug/app-debug.apk",
        "build": "yarn android-detox-build",
        "type": "android.emulator",
        "name": "Nexus_5X_API_27"
      }
    }
  }
```
#### Steps to Reproduce
1. Create a working detox configuration with version `7.3.3` 
2. Upgrade to 8.0.0
3. Run tests

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: 8.0.0
* React Native: 0.55
* Node: v8.11.3
* Device: iOS 11.2 simulator 
* Xcode: 9.3
* macOS: 10.13.4

#### Device and verbose Detox logs
```
$ yarn e2e-test e2e/features/
$ ENVFILE=.env.e2e export $(cat .env.e2e | xargs -I{} echo 'DETOX_{}') && ./node_modules/.bin/cucumber-js --require-module babel-core/register --require-module babel-preset-stage-2 -f node_modules/cucumber-pretty --artifacts-location ./artifacts --configuration iphone e2e/features/
Started mock server on PORT 9000
[...]

PR #905, opened 2018-08-29
[WIP] Returns back the previous device selection logic on iOS simulator

noomorph:
Resolves #853 .

TODO list:

- [ ] fix and resolve merge conflicts
- [ ] make the code terser, maybe split into two PRs
- [ ] pass code review

PR #1404, opened 2019-05-27
Allow specifying the simulator name or UDID as a device name for iOS simulators

jeromecornet:
- [x] This is a small change 

**Description:**

This PR allows specifying a UDID as the --device-name option, which allows targeting a specific simulator instance rather than just one that matches the device type/OS.

This is very useful when the test environment contains multiple of the same device types that are pre-configured for different locales for example.

### DIFF
--- a/detox/src/devices/ios/AppleSimUtils.js
+++ b/detox/src/devices/ios/AppleSimUtils.js
@@ -30,7 +30,10 @@ class AppleSimUtils {
 
     let type;
     let os;
-    if (_.includes(query, ',')) {
+    if (query.match(/[0-9A-F]{8}-[0-9A-F]{4}-[4][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}/i)) {
+      return [query]
+    }
+    else if (_.includes(query, ',')) {
       const parts = _.split(query, ',');
       type = parts[0].trim();
       os = parts[1].trim();
--- a/detox/src/devices/ios/AppleSimUtils.js
+++ b/detox/src/devices/ios/AppleSimUtils.js
@@ -30,10 +30,7 @@ class AppleSimUtils {
 
     let type;
     let os;
-    if (query.match(/[0-9A-F]{8}-[0-9A-F]{4}-[4][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}/i)) {
-      return [query]
-    }
-    else if (_.includes(query, ',')) {
+    if (_.includes(query, ',')) {
       const parts = _.split(query, ',');
       type = parts[0].trim();
       os = parts[1].trim();
@@ -45,10 +42,14 @@ class AppleSimUtils {
 
     const response = await this._execAppleSimUtils({ args: `--list --byType "${type}" --byOS "${os}"`}, statusLogs, 1);
     const parsed = this._parseResponseFromAppleSimUtils(response);
-    const udids = _.map(parsed, 'udid');
+    let udids = _.map(parsed, 'udid');
     if (!udids || !udids.length || !udids[0]) {
-      throw new Error(`Can't find a simulator to match with "${query}", run 'xcrun simctl list' to list your supported devices.
-      It is advised to only state a device type, and not to state iOS version, e.g. "iPhone 7"`);
+      const devices = await this.devicesMatching(query)
+      udids = _.map(devices, 'udid')
+      if (!udids || !udids.length || !udids[0]) {
+        throw new Error(`Can't find a simulator to match with "${query}", run 'xcrun simctl list' to list your supported devices.
+        It is advised to only state a device type or name, and not to state iOS version, e.g. "iPhone 7"`);
+      }
     }
     return udids;
   }
@@ -85,6 +86,14 @@ class AppleSimUtils {
   async isBooted(udid) {
     const device = await this.findDeviceByUDID(udid);
     return (_.isEqual(device.state, 'Booted') || _.isEqual(device.state, 'Booting'));
+  }
+
+  async devicesMatching(name) {
+    const result = await this._execSimctl({ cmd: `list -j` });
+    const stdout = _.get(result, 'stdout');
+    const output = JSON.parse(stdout);
+    const devicesWithRunTimes = _.flatMap(output.devices, (runtime) => _.map(runtime, (device) => ({ runtime, ...device })) );    
+    return _.filter(devicesWithRunTimes, (device) => device.name == name || device.udid == name)
   }
 
   async deviceTypeAndNewestRuntimeFor(name) {


noomorph:
@LeoNatan, what's your opinion on this? IMO, looks like a valid idea, why not?

LeoNatan:
But why not a new param —device-udid?

noomorph:
@LeoNatan, well, that's possible too, but how do you define it in `package.json` (in `detox.configurations['ios.something'].name`, I mean)? Do you suggest to add there `udid` field too?

LeoNatan:
Yep, sounds reasonable. Name is not UDID, doesn’t sound like a good idea to mix.

jeromecornet:
Thanks for the comment guys.

Sorry this comment is going to be long, but I think I should explain some of my thoughts behind the PR

### Device name isn't name
The reason I went and added to the --device-name logic is that this param already doesn't really represent the device name. It can be either 
- a device type 
- or a combination of device type, OS
(see https://github.com/wix/Detox/blob/be6ea6709c3a7f2785e94f2fbba3484bdc60cdcb/detox/src/devices/ios/AppleSimUtils.js#L33-L43 )

So the way I see it is that --device-name is really a matcher for something that uniquely identifies the device, rather than the actual name (which BTW isn't currently matched by the --device-name parameter / "name" configuration)

### Compatibility

If backwards compatibility was out the window, I would personally rename the "name" configuration and --device-name parameter, and simply use a "device" configuration and --device parameter, which would match the device with any possible way (name, device type, UDID).

But I'm not sure that it's worth breaking backwards compatibility for this. That said, I'm open to make the change if you feel it's necessary.

### Adding udid interacts oddly with name 

I thought of adding  add a "udid" field to package.json and adding a --device-udid runtime option that would behave the same as --device-name (i.e. override the device in the default configuration) but then the interaction with the actual --device-name would be weird.

For example, if you have one configuration that is : 
```json
"ios.sim": {
      "binaryPath": "bin/Exponent.app",
      "type": "ios.simulator",
      "name": "iPhone 7"
    }
```
And 2 simulators: UDID-1 which is an iPhone 7 and UDID-2 which is an iPhone 8.

When running `detox test`, it runs on UDID-1, when running `detox test --device-name "iPhone 8"` it runs on UDID-2
If you run `detox test --device-udid "UDID-1"` it should probably run on UDID-1, but if you run 
`detox test --device-udid "UDID-2"` does it still make sense to run on UDID-2 even though the configuration Name says it should be an iPhone 7 ?

Also if you had a configuration with 
```json
"ios.sim": {
      "binaryPath": "bin/Exponent.app",
      "type": "ios.simulator",
      "name": "iPhone 7"
      "udid": "UDID-2"
    }
```

Should the device be not found because there is no simulator that matches both the type (in the name configuration) and the udid. Or should udid override name or the other way around ?

If you still would prefer to add a separate parameter called --device-udid (and configuration "UDID"), can you please give me some guidance on what you would expect the interaction with --device-name and "name" would be ?

### Code stuff

For now, I noticed the merge conflict so:
- I rebased on master to remove the test files that were removed in https://github.com/wix/Detox/pull/1405
- as well as improved the UDID regexp to match the UUIDv4 spec (my previous regexp was too lose)

noomorph:
1. @LeoNatan, I side with @jeromecornet in a sense that `"name"` acts more than just a name, it is a matcher currently (a broken one, I must admit, till https://github.com/wix/Detox/pull/905 lands, hopefully in June :crossed_fingers: ).

    I don't see clear benefits in adding yet another CLI/JSON param such as `--device-udid` which from its birth is doomed to have a conflict resolution with `--device-name` in current codebase, and also be specific to Apple terminology - for instance, Android devices do not have a mirror concept to [UDID](https://en.wikipedia.org/wiki/UDID) if we talk about interacting with them via ADB.

2. @jeromecornet, on the other hand, I am not sure, how is this going to play together with `--max-workers <n>` feature in `Detox + Jest` chain? Have you tested your change in that scenario on iOS? How will the new devices be created and named in this case?

3. @jeromecornet, [how are you sure that UDID complies to UUIDv4 spec if the former is proprietary and not standardized?](https://stackoverflow.com/questions/4547369/how-can-i-recognize-whether-a-string-is-a-udid-for-an-ios-device) I refer to what you wrote above:
> as well as improved the UDID regexp to match the UUIDv4 spec (my previous regexp was too lose)

----

To conclude, I am okay with extending the semantics of `--device-name` even further. We might want to redesign lots of CLI/JSON parameters in the next major version when we have a clearer understanding of what separation of device querying concerns we want to achieve. To me, this pull request looks harmless, for the feature what it adds.

LeoNatan:
We shouldn't abuse existing bugs to make a bigger mess. "Name" should be a name, while others such as "type" and "udid" should be separate ones. Device type is not a "name", and neither is "udid". If each OS has its own terminology, we should support it as expected. "Name" isn't it.

UDID is indeed not a UUID always. Simulators use UUID as their UDID, but hardware device, which we may want to support soon once Facebook's idb is more stable, use the hardware UDID.

LeoNatan:
The UUID parsing logic is a perfect example why this should be split. Since there is no documentation that states that a UDID is a UUID, having such logic is wrong. If this was a separate field, we would just pass the string as-is and let the system take care of itself.

jeromecornet:
@LeoNatan :
I agree with you, the name doesn't match the feature, and my code only compounds the existing problem. 
Even without adding any UDID-related functionality, what would you like to see ?
- make `--device-name` select a device by name (not type or type/os as it does currently)
- rename `--device-name` to something more descriptive, like `--device-matcher` or `--device` 

@noomorph :
- I have not tested with --max-workers although I assume this won't play nicely as it assume a single device. I will try this and get back to you.
- my original regexp was way too loose, and I think @LeoNatan has a point which is that the code now needs to know too much about how a udid is created. 


Maybe a way forward is to change the matcher code to look for the parameter passed as device name, and match on device name OR type OR udid ? And probably rename that parameter to just be 'device' (although that can be done independently). Thoughts ?

jeromecornet:
OK, so I reworked a bit the approach (@noomorph this will likely conflict with https://github.com/wix/Detox/pull/905)
Instead of hardcoding the regexp for UDID, instead I am:
- using the current detection logic (i.e. match a device type or device type/os)
- but if no devices match that, the code falls match to try and match either simulator name or udid.

This way, if you pass a custom simulator name or udid it will work as expected. If the simulator name is a device type (like the default simulators), it will pick the device type instead.

I've updated the PR title to reflect that

@LeoNatan do you think that's an improvement ? It doesn't really change the fact that `--device-name` is more than the name, but at least it _can_ be the name 😄 unlike the current state where it's (a least) the device type

d4vidi:
Sorry for the CI issues - 'been doing some migrations. The builds should be green in a few mins.

jeromecornet:
@noomorph @LeoNatan thoughts on this approach ? I'd love to see this functionality added, and I think this current version is an improvement over the status quo

noomorph:
@jeromecornet, what is below that is my personal opinion. So, I think it is acceptable to stick to simplicity of your approach, provided that:
1. Configuration undergoes some renaming: 
    1. `--device-name` becomes `--device`. The former parameter becomes deprecated (see example in `local-cli/test.js`).
    2. `name`(as  in `package.json -> detox -> configurations -> [conf] -> name`) becomes `device`. The same thing with the deprecation.
2. Priority of device query will be something like:
    1. Device ID (as they are GUID-like, should not be a breaking change)
    2. Device name (that will resolve #853)
    3. Device type, OS version (backward compatibility)
    4. Device type [latest available version]
    5. We might agree on something different than 1-4.
 3. You don't duplicate code like you did here:
    ```js
    async devicesMatching(name) {
        const result = await this._execSimctl({ cmd: `list -j` });
    // ...     
    }

    async deviceTypeAndNewestRuntimeFor(name) {
        const result = await this._execSimctl({ cmd: `list -j` });
    // ...     
    }
    ```
    and also try to avoid so heavy and verbose calls such as `list -j` until it is unavoidable. AppleSimUtils has a lot of query operators, so you might want to rework your queries based on its capabilities.

@LeoNatan, please give me your $0.02 on my comment above. If we agree, then @jeromecornet will know what to do. If not — then probably he will blocked until we come to some other solution.

LeoNatan:
Closing in favor of https://github.com/wix/Detox/issues/853
@noomorph Please update #853 with the latest status.

@jeromecornet If you'd like to see udid support, please submit a PR taking into account what is discussed in #853. The solution needs to take into account all problems. I don't want to complicate the existing code which is obviously not good (`name` is mapped to `type` 🤦‍♂️ in https://github.com/wix/Detox/blob/bb0730e43bdb984fad2bf9ccd02dfae752f0f3e4/detox/src/devices/ios/AppleSimUtils.js#L35). If we are going to touch this place, let's pls do it right.

PR #1407, opened 2019-05-28
Make crash e2e more stable

d4vidi:
**Description:**

Currently, crash e2e's start by reloading RN instead of restarting from scratch. This means the test doesn't really begin fresh, from scratch - as some native state properties can leak in. Not just theoretical -- on Android, this made the app load inside a different activity that's related to url launch checking (previous test was urls related).

### DIFF
--- a/detox/test/e2e/19.crash-handling.test.js
+++ b/detox/test/e2e/19.crash-handling.test.js
@@ -1,7 +1,7 @@
 describe('Crash Handling', () => {
-
   it('Should throw error upon app crash', async () => {
-    await device.reloadReactNative();
+    await device.launchApp({newInstance: true});
+
     let failed = false;
 
     try {


LeoNatan:
How is leaking related to the crash test? If the app crashes, it should be caught in whatever view or activity the crash occurred. The crash is unrelated to UI anyway?

d4vidi:
well yes, it _is_ related since the app has been launched last with a url. in that case, it loads up a screen that shows the url rather than the usual main menu.
Perhaps that was also an issue on iOS, except that on iOS the suite has other tests in between.

LeoNatan:
But why does it matter one what screen the app crashes?

LeoNatan:
Or is the "crash" button on the main menu?
In that case, why isn't reloadReactNative reseting correctly to the main screen?

d4vidi:
You crash the app by tapping a `crash` button from the main menu

PR #1408, opened 2019-05-28
 Use EarlGrey from fork, rather than official

LeoNatan:
The fork includes a further fix to network swizzling, which fixes a crash in New Relic due to a bug in their own swizzles. A comedy of errors.


noomorph:
@LeoNatan, could you attach the link to the original issue and PR in EarlGrey they have not accepted/fixed? Just for history and easier tracking. Thanks in advance!

LeoNatan:
https://github.com/google/EarlGrey/pull/840

PR #1412, opened 2019-05-29
fix: include launch args that come from artifact plugin

noomorph:
- [x] This is a small change 

---

**Description:**

This is a hotfix for https://github.com/wix/Detox/pull/1385.

There is a not very obvious logic in `SimulatorInstrumentsPlugin` that mutates `launchArgs` in order to be able to launch the app with Detox Instruments recording enabled from the very beginning. Also, it can specify `detoxInstrumentsPath` via that specific mutation.

This particular oversight had two implications:

1. You can’t override Detox Instruments Path when launching an app.
2. And you can’t collect profiling if you launch an app inside `it()`.

This fix moves all the command-line argument manipulation logic of launchApp to `AppleSimUtils` class.

### DIFF
--- a/detox/src/devices/drivers/SimulatorDriver.js
+++ b/detox/src/devices/drivers/SimulatorDriver.js
@@ -70,10 +70,8 @@ class SimulatorDriver extends IosDriver {
   }
 
   async launchApp(deviceId, bundleId, launchArgs, languageAndLocale) {
-    const _launchArgs = this._transformLaunchArgs(launchArgs);
-
     await this.emitter.emit('beforeLaunchApp', {bundleId, deviceId, launchArgs});
-    const pid = await this.applesimutils.launch(deviceId, bundleId, _launchArgs, languageAndLocale);
+    const pid = await this.applesimutils.launch(deviceId, bundleId, launchArgs, languageAndLocale);
     await this.emitter.emit('launchApp', {bundleId, deviceId, launchArgs, pid});
 
     return pid;
@@ -128,13 +126,6 @@ class SimulatorDriver extends IosDriver {
 
   async waitForBackground() {
     return await this.client.waitForBackground();
-  }
-
-  _transformLaunchArgs(launchArgs) {
-    return _.reduce(launchArgs, (result, value, key) => {
-      result[`-${key}`] = value;
-      return result;
-    }, {});
   }
 }
 
--- a/detox/src/devices/drivers/SimulatorDriver.test.js
+++ b/detox/src/devices/drivers/SimulatorDriver.test.js
@@ -30,13 +30,13 @@ describe('IOS simulator driver', () => {
 
     it('should be passed to AppleSimUtils even if some of them were received from `beforeLaunchApp` phase', async () => {
       uut.emitter.on('beforeLaunchApp', ({ launchArgs }) => {
-        launchArgs.dog3 = 'from-plugin';
+        launchArgs.dog3 = 'Chika, from plugin';
       });
 
       await uut.launchApp(deviceId, bundleId, launchArgs, languageAndLocale);
       expect(uut.applesimutils.launch).toHaveBeenCalledWith(deviceId, bundleId, {
         ...launchArgs,
-        dog3: 'from-plugin',
+        dog3: 'Chika, from plugin',
       }, '');
     });
   });
--- a/detox/src/devices/drivers/SimulatorDriver.test.js
+++ b/detox/src/devices/drivers/SimulatorDriver.test.js
@@ -1,34 +1,43 @@
 describe('IOS simulator driver', () => {
+  let uut;
 
   const deviceId = 'device-id-mock';
   const bundleId = 'bundle-id-mock';
 
+  beforeEach(() => {
+    jest.mock('../ios/AppleSimUtils', () => mockAppleSimUtils);
+  });
+
   describe('launch args', () => {
+    let launchArgs, languageAndLocale;
 
-    let uut;
     beforeEach(() => {
-      jest.mock('../ios/AppleSimUtils', () => mockAppleSimUtils);
-
-      const SimulatorDriver = require('./SimulatorDriver');
-      uut = new SimulatorDriver({
-        client: {},
-      });
-    });
-
-    it('should inject a prefix to arg keys', async () => {
-      const launchArgs = {
+      launchArgs = {
         'dog1': 'dharma',
         'dog2': 'karma',
       };
 
-      await uut.launchApp(deviceId, bundleId, launchArgs, '');
+      languageAndLocale = '';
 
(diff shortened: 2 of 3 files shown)


detox/src/devices/drivers/SimulatorDriver.js
```diff
@@ -70,10 +70,8 @@ class SimulatorDriver extends IosDriver {
   }
 
   async launchApp(deviceId, bundleId, launchArgs, languageAndLocale) {
-    const _launchArgs = this._transformLaunchArgs(launchArgs);
-
     await this.emitter.emit('beforeLaunchApp', {bundleId, deviceId, launchArgs});
-    const pid = await this.applesimutils.launch(deviceId, bundleId, _launchArgs, languageAndLocale);
```
d4vidi: @noomorph this definitely breaks unit tests I've created.

detox/src/devices/drivers/SimulatorDriver.js
```diff
@@ -70,10 +70,8 @@ class SimulatorDriver extends IosDriver {
...
+    const pid = await this.applesimutils.launch(deviceId, bundleId, launchArgs, languageAndLocale);
```
d4vidi: @noomorph this definitely breaks unit tests I've created.

detox/src/devices/drivers/SimulatorDriver.js
```diff
@@ -70,10 +70,8 @@ class SimulatorDriver extends IosDriver {
   }
 
   async launchApp(deviceId, bundleId, launchArgs, languageAndLocale) {
-    const _launchArgs = this._transformLaunchArgs(launchArgs);
-
     await this.emitter.emit('beforeLaunchApp', {bundleId, deviceId, launchArgs});
```
d4vidi: @noomorph shouldn't you be changing plugins code so they'd adhere to this new way launch args are passed in?

LeoNatan:
The battle of the map/reduce 😂

detox/src/devices/drivers/SimulatorDriver.test.js
```diff
@@ -1,34 +1,43 @@
...
+    it('should be passed to AppleSimUtils even if some of them were received from `beforeLaunchApp` phase', async () => {
+      uut.emitter.on('beforeLaunchApp', ({ launchArgs }) => {
+        launchArgs.dog3 = 'from-plugin';
       });
+
+      await uut.launchApp(deviceId, bundleId, launchArgs, languageAndLocale);
+      expect(uut.applesimutils.launch).toHaveBeenCalledWith(deviceId, bundleId, {
+        ...launchArgs,
+        dog3: 'from-plugin',
+      }, '');
```
d4vidi: kudos

detox/src/devices/drivers/SimulatorDriver.test.js
```diff
@@ -1,34 +1,43 @@
...
+
+    it('should be passed to AppleSimUtils even if some of them were received from `beforeLaunchApp` phase', async () => {
+      uut.emitter.on('beforeLaunchApp', ({ launchArgs }) => {
+        launchArgs.dog3 = 'from-plugin';
       });
+
+      await uut.launchApp(deviceId, bundleId, launchArgs, languageAndLocale);
+      expect(uut.applesimutils.launch).toHaveBeenCalledWith(deviceId, bundleId, {
+        ...launchArgs,
+        dog3: 'from-plugin',
```
LeoNatan: Should be "chika"!

noomorph: My bad 😊 I'll fix that, of course.

detox/src/devices/ios/AppleSimUtils.js
```diff
@@ -248,7 +248,7 @@ class AppleSimUtils {
   }
 
   _joinLaunchArgs(launchArgs) {
-    return _.map(launchArgs, (v, k) => `${k} "${v}"`).join(' ').trim();
+    return _.map(launchArgs, (v, k) => `-${k} "${v}"`).join(' ').trim();
```
LeoNatan: Since we are making a change here, let's consider the use case where iOS supports a flat list of arguments, as opposed to key-value pairs.

Spreading an array into an object creates the following object:
`{'0':'arg1', '1':'arg2'}`

Perhaps we can catch this inside this map, and instead of producing `-0 arg1 -1 arg2`, just produce `arg1 arg2`.

LeoNatan: Right now we spread the user provided object (array or object), into the lauchArgs object. This would create a problem if the user wants to provide an array of arguments, rather than a key value object.

detox/src/devices/drivers/SimulatorDriver.test.js
```diff
@@ -30,13 +30,13 @@ describe('IOS simulator driver', () => {
 
     it('should be passed to AppleSimUtils even if some of them were received from `beforeLaunchApp` phase', async () => {
       uut.emitter.on('beforeLaunchApp', ({ launchArgs }) => {
-        launchArgs.dog3 = 'from-plugin';
+        launchArgs.dog3 = 'Chika, from plugin';
```
d4vidi: 👍

detox/src/devices/ios/AppleSimUtils.js
d4vidi: @LeoNatan this was broken either way (regardless of where the dashes logic is...), are you sure this is the right scope for this?

d4vidi: I think that in order to _start_ fixing that we first need to replace the initial args spread with a lodash clone() or smt

LeoNatan: Not sure if this is the scope. I was just thinking out loud. It depends on how difficult it is to support. If it's a single `if()`, I don't think it's bad to have here. If it requires a full retooling of how we pass arguments top-down, then not the scope.

noomorph: @LeoNatan, I would enhance that in another pull request. Judging from other places in the code, `launchArgs` is supposed to be a key-value map anyways. Making it work as an array will take more effort than changing one line of code.

LeoNatan: 👍

d4vidi: BTW I think it would only make sense to allow that if we also support it on Android, otherwise we'd make the API platform specific. This isn't super trivial since on Android we're force onto a key-value scheme. I suppose what makes sense here is to map the array's values onto $key=true pairs...

LeoNatan: I don’t see a problem with supporting it only on iOS, since Android enforces key/value.

Issue #1181, opened 2019-02-27
Upgrade min supported RN version to 59


d4vidi:
### Description

Android should support .59, even if with an exception remarks due to the [RN clipping issue](https://github.com/facebook/react-native/issues/23870) on Android.

Also, the demo apps, which are currently 51 and 56, need latest (i.e. `.59`) on iOS, android and maybe `.57` (RN clipping issue, again - i.e. the full detox test suite, and a partial suite).

d4vidi:
Android blocker: https://github.com/facebook/react-native/issues/23870

fungilation:
Quoting @yungsters at https://github.com/facebook/react-native/issues/23870#issuecomment-497204209

> I do not have any good ideas for how to resolve this. I understand the dilemma that this places on Detox, but I do not know how else we could possibly have implemented support for `overflow: visible` without defaulting `setClipChildren(false)`. Due to the viral nature of `overflow`, a default of `overflow: hidden` (and `setClipChildren(true)`) prevents any meaningful support for `overflow: visible` which, if you might recall, was one of the top requested features for React Native.

LeoNatan:
Clowns 🤦‍♂️

LeoNatan:
We've introduced a workaround for this and support 0.59.x.

PR #1416, opened 2019-05-30
Introduce restricted support in RN .58+ for Android

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1181 and the solution has been agreed upon with maintainers.

---

**Description:**

Introduce required changes in order to have our *demo* app test suite successfully run on Android (with the necessary restrictions introduced by the [RN issue](https://github.com/facebook/react-native/issues/23870)).
* Documentation will follow.
* We nevertheless have to upgrade our visibility matcher in order to work around the RN issue, even though it's far from trivial.

### DIFF
--- a/detox/test/e2e/05.waitfor.test.js
+++ b/detox/test/e2e/05.waitfor.test.js
@@ -1,3 +1,5 @@
+const custom = require('./utils/custom-it');
+
 describe('WaitFor', () => {
   beforeEach(async() => {
     await device.reloadReactNative();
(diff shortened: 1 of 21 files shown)


detox/test/e2e/05.waitfor.test.js
```diff
@@ -11,18 +13,17 @@ describe('WaitFor', () => {
...
-  it('should wait until an element is removed', async() => {
+  it('should wait until an element is removed', async () => {
     await expect(element(by.id('deletedFromHierarchyText'))).toBeVisible();
     await element(by.id('GoButton')).tap();
     await waitFor(element(by.id('deletedFromHierarchyText'))).toBeNotVisible().withTimeout(20000);
     await expect(element(by.id('deletedFromHierarchyText'))).toBeNotVisible();
   });
 
-  it('should find element by scrolling until it is visible', async() => {
+  custom.it.withFailureIf.android.rn58OrNewer('should find element by scrolling until it is visible', async () => {
```
LeoNatan: What is the point of this if we know it will fail anyway? We will be running the RN59 suite from now on, on both iOS and Android. Why not just close it off for Android until we have a proper fix?

d4vidi: This is so we'd know when (if) they decide to get it fixed.

PR #1400, opened 2019-05-23
Removed triggering of instruments recording  (--record-performance) for apps launched in before/beforeApp scopes

rotemmiz:
**Description:**

Removed triggering of instruments recording  (--record-performance) for apps launched in before/beforeApp scopes

```js
  before(async () => {
    await device.launchApp({newInstance: true, permissions: {notifications: 'YES'}});
  });
```

In that case, the beforeEach scope handles starting of a recording on an already launched app, and instruments will crash on an invalid state.

```
Suppressed error inside function call: SimulatorInstrumentsPlugin.onBeforeEach({ title: 'Launch logged in app from dead state',
  fullName:
   'Open installed app - logged in Launch logged in app from dead state',
  status: 'running' })
  err: A recording is already in progress
  (
      0   CoreFoundation                      0x00000001101421bb __exceptionPreprocess + 331
      1   libobjc.A.dylib                     0x000000010e076735 objc_exception_throw + 48
      2   CoreFoundation                      0x0000000110142015 +[NSException raise:format:] + 197
      3   DTXProfiler                         0x000000010afeef72 -[DTXProfiler _startProfilingWithConfiguration:deleteExisting:] + 875
      4   Detox                               0x000000010af5dca1 -[DetoxInstrumentsManager startRecordingAtURL:] + 149
      5   Detox                               0x000000010af5c567 -[DetoxManager _handlePerformanceRecording:isFromLaunch:completionHandler:] + 413
      6   Detox                               0x000000010af5ad28 -[DetoxManager websocket:didReceiveAction:withParams:withMessageId:] + 266
      7   Detox                               0x000000010af4fa97 -[WebSocket receiveAction:] + 457
      8   SocketRocket                        0x000000012f522508 __43-[SRWebSocket _handleFrameWithData:opCode:]_block_invoke.295 + 214
      9   EarlGrey                            0x000000012f375b32 __60-[GREYDispatchQueueTracker grey_dispatchAsyncCallWithBlock:]_block_invoke + 50
      10  libdispatch.dylib                   0x00000001153224e1 _dispatch_call_block_and_release + 12
      11  libdispatch.dylib                   0x000000011532354b _dispatch_client_callout + 8
      12  libdispatch.dylib                   0x000000011532f380 _dispatch_main_queue_callback_4CF + 1290
      13  CoreFoundation                      0x00000001100a73e9 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
[...]

PR #1418, opened 2019-05-30
Revert #1400

noomorph:
I suggest reverting the revert https://github.com/wix/Detox/commit/1c76b5ac7c8cd91ff0240bb6d3b3597656090e8e from PR https://github.com/wix/Detox/pull/1400.

My pull request reproduces the situation in the description and it does not result into any crashes.

Let's think what was wrong in the original situation, and fix the original issue without throwing away essential parts of `SimulatorInstrumentsPlugin`. As the saying goes, the baby should not be thrown out with the bath water.


LeoNatan:
What was wrong was that code. `this.testRecording` was being set incorrectly, causing a crash due to incorrect state management.

noomorph:
After all the checks, I confirm you were absolutely right. That fragment of code was erroneous and simply redundant. Removing it makes the lifecycle of SimulatorInstrumentsPlugin work much more correctly.

PR #1430, opened 2019-06-04
Merge demo binaries

LeoNatan:
Merge all RN projects to use the same binary so that build time is reduced significantly.


LeoNatan:
We'll merge this after #1348

PR #1436, opened 2019-06-06
Cucumber Example IOS Native

Srutherford2407:
- [x] This is a small change 

---

**Description:**

Added Cucumber Example
Updated ReadMe
3 Scenario Examples to ensure relauchAPP works
Broke out one Scenario into 2 Steps to ensure but steps ran before relaunchAPP was called
Updated package.json

### DIFF
--- a/examples/demo-native-ios-cucumber/.gitignore
+++ b/examples/demo-native-ios-cucumber/.gitignore
@@ -1,0 +1,31 @@
+####
+# Webstrom
+
+.idea
+node_modules/
+npm-debug.log
+.log
+
+# OSX
+#
+.DS_Store
+
+# Xcode
+#
+build/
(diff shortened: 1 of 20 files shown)


LeoNatan:
Hello,

Thank you for this PR!
I am not sure this is something we should accept. I think it's better to have this as a different repo, which we can link in the documentation. We don't have the capacity to maintain such a project (we're spread thin as it is).

Srutherford2407:
@LeoNatan 

That works for me! My company will continue to test Detox with cucumber and hopefully be able to address an issues we find between the two.

LeoNatan:
Great. So let's close this PR, and once you have your repo, you are welcome to submit a new PR in documentation, pointing to your repo. Thanks

PR #1437, opened 2019-06-07
Update APIRef.ActionsOnElement.md

leotm:
- [x] This is a small change 

---

**Description:**

This appears to be an important pre-requisite to successfully calling `typeText`, because it uses the builtin keyboard in all text fields.

To use the builtin keyboard, it must be toggled on first in the simulator in any text field, otherwise this error will be thrown in the terminal:

```sh
Detox _ Failed to type string because keyboard was not shown on screen
```

One can achieve this by tapping on any app text field, then selecting *Hardware -> Keyboard -> Toggle Software Keyboard (⌘K)*.

This will now automatically toggle the builtin keyboard on during tests.

### DIFF
--- a/docs/APIRef.ActionsOnElement.md
+++ b/docs/APIRef.ActionsOnElement.md
@@ -78,6 +78,10 @@ Use the builtin keyboard to type text into a text field.
 await element(by.id('textField')).typeText('passcode');
 ```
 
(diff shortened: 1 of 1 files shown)


LeoNatan:
Thanks

PR #1438, opened 2019-06-11
Update GettingStarted.md with yarn alternatives

leotm:
- [x] This is a small change 

---

Add `yarn` scripts alongside `npm` scripts.


rotemmiz:
I don't see the need to add yarn commands on top of the current ones, there is plenty documentation on migrating from npm to yarn, or using it in general.

leotm:
Just seeing it in more and more commonplace documentation nowadays.

E.g. https://www.learnstorybook.com/react/en/get-started/

But of course, not needed.

LeoNatan:
WE have considered this and have decided not to accept it. Thank you for your PR.

PR #1443, opened 2019-06-12
Update Introduction.GettingStarted.md

lyqht:
changed command given for brew install applesimutils. directly running just brew applesimutils does not find the package.

- [x] This is a small change 

---

**Description:**


LeoNatan:
Did you miss the `brew tap wix/brew` command one line before?

PR #1452, opened 2019-06-14
correcting to install wix/brew/applesimutils, instead of applesimutil…

jeffrey-l-turner:
The documentation and scripts are incorrect as we are getting errors now trying to `brew install applesimutils` instead of `brew install wix/brew/applesimutils`

- [x] This is a small change 

---

**Description:**
Changed documentation, and scripts in addition to a minor shellcheck error.

### DIFF
--- a/detox/src/devices/ios/AppleSimUtils.js
+++ b/detox/src/devices/ios/AppleSimUtils.js
@@ -237,7 +237,7 @@ class AppleSimUtils {
 
     } catch (ex) {
       throw new Error(`Could not parse response from applesimutils, please update applesimutils and try again.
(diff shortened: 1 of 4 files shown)


LeoNatan:
Documentation says to `brew tap wix/brew` before installing. This PR is not needed.

LeoNatan:
Out of curiosity, @jeffrey-l-turner what errors were you seeing before this change? If you tapped the `wix/brew` repo, it should have been visible. If you seen some `<<<<` errors, the solution is to `brew untap wix/brew && brew tap wix/brew`.

Issue #1427, opened 2019-06-04
Occasional ECONNREFUSED connecting to emulator via telnet


rotemmiz:
**Describe the bug**
Every now and then we get a `ECONNREFUSED 127.0.0.1:23`, trying to connect to the android emulator in the internal CI's.

In the standard case, we expect for the device list to be as follows:
```
List of devices attached
emulator-5554	device
```
...in which case we should connect to the device via the `5554` port rather than via `23`. Yet something happens and we fall back to the `23` default.

As the emulator is listening on `5554` rather than `23`, it is obvious that the telnet connection attempt eventually times out. Hence the problem lies in port lookup code.

**Environment (please complete the following information):**
 - Detox: 12.10.x
 - React Native: *
 - Node:*
 - Device:Android emulator
 - OS: *

**Device and Verbose Detox Logs**
```
detox[17696] ERROR: [EmulatorTelnet.js/TELNET_ERROR] connect ECONNREFUSED 127.0.0.1:23
err: Error: connect ECONNREFUSED 127.0.0.1:23
       at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1117:14)
```

noomorph:
@rotemmiz, in fact, those lines go to stderr:
```
* daemon not running; starting now at tcp:5037
* daemon started successfully
```
So, technically, I see no difference (stdout-wise) between your first and second example.

rotemmiz:
Then maybe its due to the fact that the output is not buffered, and so `stdout` has only one line?
`List of devices attached`

noomorph:
@rotemmiz, not likely. Exec waits till the app termination. Why are you sure it has to do with "when daemon starts"? I could not reproduce this issue by just doing what you described in the steps to reproduce.

d4vidi:
@noomorph I've verified what you're saying, indeed you are right. Which means we don't know the cause for telnet issues but we do know they seem to pop up every now and then.
Please review my PR which aims at collecting more info rather than fixing this.

d4vidi:
@noomorph @rotemmiz I've retrofitted the description to fit better to what we know now.

noomorph:
Please open then :)

d4vidi:
@rotemmiz @noomorph I haven't seen instances of this error for quite a while now. Am I the only one?

mglaze:
[...]

PR #1453, opened 2019-06-16
Increase log-level for result of 'adb devices' command

d4vidi:
- [x] This is a small change 
- [x] This change has been discussed in issue #1427 and the solution has been agreed upon with maintainers.

---

**Description:**

Since we don't know what the root cause is for the sporadic failures of #1427, for now we'll have to settle on getting more info. With this PR, we'll `debug`-log the result of each `adb devices` instead of just `trace`-ing it. This will give us high visibility the next time this happens.

### DIFF
--- a/detox/src/devices/android/ADB.js
+++ b/detox/src/devices/android/ADB.js
@@ -13,7 +13,7 @@ class ADB {
   }
 
   async devices() {
-    const output = (await this.adbCmd('', 'devices')).stdout;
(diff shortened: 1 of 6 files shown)


d4vidi:
@rotemmiz @noomorph following our talk from this morning, I now have 2nd thoughts regarding introducing a log wrapper. Feels like over-engineering for just 2 `if`'s. Plus we need to start getting answers and we're losing time... So, merging in.

noomorph:
@d4vidi, good luck!

PR #1464, opened 2019-06-19
(Android) Introduce exception-based, edge-monitoring in scrolling actions of all flavours (equiv to iOS)

### LINKED ISSUE #1409: (Android) Endless scrolling attempts if item not found using `waitFor`
**Describe the bug**
With `waitFor()` + `whileElement()` over a scroll-view - namely, in a search for a specific item with repeated scrolling, the scrolling never ends if the list's end is reached without the item being found.
Eventually, the tests fails due to the global timeout.

**To Reproduce**

Happens on one of the e2e tests of the Wix app. Reproduceable by @viktorijasujetaite.

**Expected behavior**
As with iOS -- should throw an exception when scroll'ed all the way to the end.

**Screenshots**
N/A

 - Detox: `12.9.0`
 - React Native: `0.59.x`
 - Node: -
 - Device: `Pixel_API_26`
 - OS: -

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1409 and the solution has been agreed upon with maintainers.

---

**Description:**

Fixes #1409 by introducing a custom exception thrown whenever a scrollable view is scrolled (in a single direction) all the way through to the end by detox, thus making the api equivalent to iOS'.

The change here can, in a way, be broken down into 2 aspects:
1. Make scrolling throw an exception when the 'edge' is reached. This adheres to basic expectations from the tester's side - of android behaving like iOS in that sense.
2. Integrate into the `waitFor()` + `whileElement()` matcher +`scroll()` action combo, such that when an element is visible (or hidden) the scrolling stops gracefully or due to the above error, respectively. Once again -- same as iOS.

fyi @viktorijasujetaite


d4vidi:
Note: build is failing due to technical android artifacts publishing issues. I'll take of it but in essence e2e's are green.

d4vidi:
@viktorijasujetaite 'fix is in now with Detox `12.11.2`, could you please try it out with your tests? 🙏

viktorijasujetaite:
@d4vidi - verified, it works! Thank you so much 👏

Issue #1460, opened 2019-06-18
bug: timeout error on detox.cleanup() if an app had been terminated earlier in a test


noomorph:
**Describe the bug**

Detox hangs up (both for Mocha and Jest, iOS and Android) if an app is terminated inside a test.
That leads to a timeout error on `await detox.cleanup()`. 

**To Reproduce**

- [x] I have tested this issue on the latest Detox release and it still reproduces

Check out branch: https://github.com/wix/Detox/tree/noomorph/fix/cleanup-terminated-app
Run: `npx detox test -c android.emu.release --loglevel trace e2e/23.flows.test.js`

**Expected behavior**

Detox ends execution successfully, within a relatively short period of time.

**Environment (please complete the following information):**
 - Detox: master
 - React Native: 0.59.x
 - Node: 8.11.3
 - Device: both Android emulator and iOS simulator are prone to this.
 - Xcode: 10.1 

**Device and Verbose Detox Logs**

```
detox[10017] TRACE: [ArtifactsManager.js/LIFECYCLE] artifactsManager.onBeforeTerminateApp({ deviceId: 'emulator-5554', bundleId: 'com.wix.detox.test' })
detox[10017] DEBUG: [exec.js/KILL] sending SIGINT to [pid = 10075]: /home/x/Android/Sdk/platform-tools/adb -s emulator-5554 shell am instrument -w -r -e detoxServer ws://localhost:46175 -e detoxSessionId ac2f7792-aa72-b01e-4546-dee0312e7f92 -e debug ZmFsc2U= com.wix.detox.test.test/androidx.test.runner.AndroidJUnitRunner
detox[10017] DEBUG: [exec.js/KILL] sending SIGINT to [pid = 10075]: /home/x/Android/Sdk/platform-tools/adb -s emulator-5554 shell am instrument -w -r -e detoxServer ws://localhost:46175 -e detoxSessionId ac2f7792-aa72-b01e-4546-dee0312e7f92 -e debug ZmFsc2U= com.wix.detox.test.test/androidx.test.runner.AndroidJUnitRunner
[...]

PR #1465, opened 2019-06-20
fix: conduct cleanup even if app has been already terminated

noomorph:
- [x] This change has been discussed in issue #1460 and the solution has been agreed upon with maintainers.


rotemmiz:
Looks good IMO.
Did the new test fail before this change?

noomorph:
Yes, it fails. At least, I've verified on Android.

PR #1470, opened 2019-06-22
Update Guide.Expo.md

yaron1m:
- [X] This is a small change 

---

Added the know issue of expo not working with detox version > 12.3.0.
Also - the sample app is deprecated. After contacting the author I created a new updated example

### DIFF
--- a/docs/Guide.Expo.md
+++ b/docs/Guide.Expo.md
@@ -39,5 +39,8 @@ beforeEach(async () => {
 
 - Usage with Android is currently TBD.
 
+## Known issues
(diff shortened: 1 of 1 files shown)


LeoNatan:
Please note that Detox older than 12.11.1 will not work with recent versions of applesimutils, so some portions of Detox will be broke.

yaron1m:
I guess broken portions are better than tests not working at all :)

LeoNatan:
I fear the results will be similar for some tests. OK, let’s push this. Any insight on what broke?

PR #1471, opened 2019-06-23
(Android) Fix a unidentified non-visibility matching when using waitFor()

### LINKED ISSUE #1463: (Android) optimize waitFor() + toBeNotVisible() + withTimeout() combo
**Describe the bug**

On android, waiting for an element to be _not_ visible, as performed by [this e2e](https://github.com/wix/Detox/blob/c5c8a519d36356bacdaafdc2aca3cfcb50e4d85c/detox/test/e2e/05.waitfor.test.js#L17), isn't optimal: it firsts waits for the timeout to expire, then performs the matching validations.

**To Reproduce**

- [x] I have tested this issue on the latest Detox release and it still reproduces

Run detox' own existing demo app's `waitFor` test suite, and see that the aforementioned test takes 20s to complete (i.e. as long as the time period given to `withTimeout()`).

**Expected behavior**
Once the associated element is _not_ visible, the final promise from the `waitFor()` chain should resolve immediately (or close to that).

**Screenshots**
N/A

 - Detox: 12.10.3
 - React Native: 59.8
 - Node: 10
 - Device: Pixel_API_26
 - OS: MacOS 10.14.3

**Device and Verbose Detox Logs**

Test suite summary (jest):

```
 PASS  e2e/05.waitfor.test.js (42.115s)
  WaitFor
    ✓ should wait until an element is created and exists in layout (3563ms)
    ✓ should wait until an element is removed (21830ms)
    ✓ should find element by scrolling until it is visible (1393ms)
    ✓ should abort scrolling if element was not found (9821ms)
```
> i.e. 2nd test in the list took >20s to run.

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1463  and the solution has been agreed upon with maintainers.

---

# Description

Bottom line is - this fixes #1463. Effectively, however, in order to do so, it introduces an essential (mini) overhaul in the way view matching works on Android.

### Overhaul details

A typical (native) Espresso matching test-code looks like this:

```java
onView(withId(R.id.viewId)).check(matches(isDisplayed()))
```
Or to be on the accurate side, since in Detox we don't use native ID's - but testID's, the code should actually like this:
```java
onView(matcherForTestId('someJSElement')).check(matches(isDisplayed())
```
But when trying to compose an equivalent assertion to represent the `waitFor` API (e.g. `waitFor(element(...)).toBeVisible().withTimeout()`), what we accidentally had created, was this:

```java
onView(allOf(matcherForTestId('someJSElement'), isDisplayed())).check(matches(isDisplayed())
```

This is bad, because:
a. `isDisplayed()` is duplicated, unnecessarily.
b. According to [Espresso's basic guide](https://developer.android.com/training/testing/espresso/basics#checking-view-assertions), this is considered bad practice.
c. Most importantly, in the [e2e case](https://github.com/wix/Detox/blob/c5c8a519d36356bacdaafdc2aca3cfcb50e4d85c/detox/test/e2e/05.waitfor.test.js#L17) described by the associated issue (i.e. #1463), this implementation eventually prevents us from breaking the `waitFor()` loop (on the native side) when the view is indeed gone, as we should. Why? because even when the view _is_ gone, the preliminary matcher (to `onView()`) doesn't find an element, nonetheless (!!!), and thus errors keep on being repeatedly thrown (in the [assertion loop](https://github.com/wix/Detox/blob/3ffaaae84e1b6e26c05ebd24934e100ff913841d/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java#L65)) - just like _before_ the view was gone. In other words, the lookup-matcher and assertion-matcher cannot be distinguished, and so the two states (before and after the view is gone) cannot be differentiated.

### Fix

The general idea is to adhere to Espresso's recommendation and separate lookup matchers (e.g. `matcherForTestId()`) from assertion matchers (e.g. visibility, `withText`).

This falls into a few major change blocks:
a. Change on JS-side (`expect.js`):  omitted the _assertion_ matcher from the final one passed in to `onView()`.
b. Introduced an alternative implementation of Espresso's `matches()` -- that _allows_ for nulls to pass on through to matchers (i.e. invokes the matcher even when the view doesn't exist).
c. Introduced an alternative implementation of the not-visible matcher, which short-circuits to true when the view is null.

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
@@ -2,20 +2,20 @@ package com.wix.detox.espresso;
 
 import android.view.View;
 
+import com.wix.detox.espresso.DetoxErrors.DetoxRuntimeException;
 import com.wix.detox.espresso.DetoxErrors.StaleActionException;
 
 import junit.framework.AssertionFailedError;
 
 import org.hamcrest.Matcher;
 
-import androidx.test.espresso.EspressoException;
 import androidx.test.espresso.ViewAction;
 import androidx.test.espresso.ViewInteraction;
 
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.assertion.ViewAssertions.doesNotExist;
-import static androidx.test.espresso.assertion.ViewAssertions.matches;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static com.wix.detox.espresso.assertion.ViewAssertions.matches;
 import static org.hamcrest.Matchers.not;
 
 /**
@@ -48,25 +48,21 @@ public class DetoxAssertion {
     }
 
     public static void waitForAssertMatcher(final ViewInteraction i, final Matcher<View> m, double timeoutSeconds) {
-        long originTime = System.nanoTime();
-        long currentTime;
+        final long originTime = System.nanoTime();
 
         while (true) {
-            currentTime = System.nanoTime();
+            long currentTime = System.nanoTime();
             long elapsed = currentTime - originTime;
-            double seconds = (double)elapsed / 1000000000.0;
+            double seconds = (double) elapsed / 1000000000.0;
             if (seconds >= timeoutSeconds) {
-                return;
+                throw new DetoxRuntimeException("" + timeoutSeconds + "sec timeout expired without matching of given matcher: " + m);
             }
+
             try {
                 i.check(matches(m));
                 break;
-            } catch (Exception e) {
-                if (e instanceof EspressoException) {
-                    UiAutomatorHelper.espressoSync();
-                } else {
-                    throw e;
-                }
+            } catch (AssertionFailedError err) {
+                UiAutomatorHelper.espressoSync(20);
             }
         }
     }
@@ -81,16 +77,12 @@ public class DetoxAssertion {
             try {
                 assertMatcher(i, vm);
                 break;
-            } catch (Exception e) {
-                if (e instanceof EspressoException) {
-                    try {
-                        onView(searchMatcher).perform(searchAction);
-                    } catch (StaleActionException exStaleAction) {
-                        assertMatcher(i, vm);
-                        break;
-                    }
-                } else {
-                    throw e;
+            } catch (AssertionFailedError err) {
+                try {
+                    onView(searchMatcher).perform(searchAction);
+                } catch (StaleActionException exStaleAction) {
+                    assertMatcher(i, vm);
+                    break;
                 }
             }
         }
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
@@ -2,11 +2,8 @@ package com.wix.detox.espresso;
 
 import android.view.View;
 
-import org.hamcrest.BaseMatcher;
-import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 
-import androidx.test.espresso.matcher.ViewMatchers;
 import androidx.test.espresso.matcher.ViewMatchers.Visibility;
 
 import static androidx.test.espresso.matcher.ViewMatchers.hasDescendant;
@@ -15,8 +12,11 @@ import static androidx.test.espresso.matcher.ViewMatchers.isDescendantOfA;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayingAtLeast;
 import static androidx.test.espresso.matcher.ViewMatchers.withContentDescription;
+import static androidx.test.espresso.matcher.ViewMatchers.withEffectiveVisibility;
 import static androidx.test.espresso.matcher.ViewMatchers.withTagValue;
 import static androidx.test.espresso.matcher.ViewMatchers.withText;
+import static com.wix.detox.espresso.matcher.ViewMatchersKt.isMatchingAtIndex;
+import static com.wix.detox.espresso.matcher.ViewMatchersKt.isOfClassName;
 import static org.hamcrest.Matchers.allOf;
 import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.is;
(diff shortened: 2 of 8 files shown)


detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -67,24 +67,8 @@ private DetoxMatcher() {
...
-                return false;
-            }
-
-            @Override
-            public void describeTo(Description description) {
-                description.appendText("Class " + className + " not found on classpath. Are you using full class name?");
-            }
-        };
+    public static Matcher<View> matcherForClass(String className) {
+        return isOfClassName(className);
```
d4vidi: boyscouting, going once

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -94,7 +78,7 @@ public void describeTo(Description description) {
     // Special ViewAssertion is needed for GONE views
     @Deprecated
     public static Matcher<View> matcherForNotVisible() {
-        return not(isDisplayed());
+        return isNotVisible();
```
d4vidi: core change

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -106,27 +90,7 @@ public void describeTo(Description description) {
...
-                ++count;
-                return false;
-            }
-
-            @Override
-            public void describeTo(Description description) {
-                description.appendText("matches " + index + "th view.");
-            }
-        };
+        return isMatchingAtIndex(index, innerMatcher);
```
d4vidi: boyscouting, going twice

detox/android/detox/src/main/java/com/wix/detox/espresso/assertion/ViewAssertions.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static androidx.test.espresso.core.internal.deps.guava.base.Preconditions.checkNotNull;
+import static androidx.test.espresso.matcher.ViewMatchers.assertThat;
+
+/**
+ * A custom extension of {@link androidx.test.espresso.assertion.ViewAssertions}.
+ *
+ * <p/>Perhaps in the future we could extend Espresso's using Kotlin's extension functions.
+ */
+public class ViewAssertions {
```
d4vidi: core stuff

detox/src/android/expect.js
```diff
@@ -149,9 +149,8 @@ class WaitForInteraction extends Interaction {
   constructor(invocationManager, element, matcher) {
     super(invocationManager);
     this._element = element;
+    this._element._selectElementWithMatcher(this._element._originalMatcher);
```
d4vidi: a very really important fix...

detox/test/e2e/05.waitfor.test.js
```diff
@@ -15,9 +15,19 @@ describe('WaitFor', () => {
...
     await element(by.id('GoButton')).tap();
-    await waitFor(element(by.id('deletedFromHierarchyText'))).toBeNotVisible().withTimeout(20000);
+
+    const startTime = new Date().getTime();
+    await waitFor(element(by.id('deletedFromHierarchyText'))).toBeNotVisible().withTimeout(timeout);
+    const endTime = new Date().getTime();
+
+    if (endTime - startTime > timeout) {
+      throw new Error(`Action not expired even after a timeout`);
+    }
```
d4vidi: Can't be more confident than that

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
```diff
@@ -48,25 +48,21 @@ public static ViewInteraction assertNotExists(ViewInteraction i) {
...
 
         while (true) {
-            currentTime = System.nanoTime();
+            long currentTime = System.nanoTime();
             long elapsed = currentTime - originTime;
-            double seconds = (double)elapsed / 1000000000.0;
+            double seconds = (double) elapsed / 1000000000.0;
             if (seconds >= timeoutSeconds) {
-                return;
+                throw new DetoxRuntimeException("" + timeoutSeconds + "sec timeout expired without matching of given matcher: " + m);
```
d4vidi: @LeoNatan Need to also do this for iOS. Ack?

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
```diff
@@ -2,20 +2,20 @@
...
 
-import androidx.test.espresso.EspressoException;
 import androidx.test.espresso.ViewAction;
 import androidx.test.espresso.ViewInteraction;
 
 import static androidx.test.espresso.Espresso.onView;
 import static androidx.test.espresso.assertion.ViewAssertions.doesNotExist;
-import static androidx.test.espresso.assertion.ViewAssertions.matches;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static com.wix.detox.espresso.assertion.ViewAssertions.matches;
```
d4vidi: a core fix

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -67,34 +67,16 @@ private DetoxMatcher() {
...
 
     public static Matcher<View> matcherForSufficientlyVisible() {
         return isDisplayingAtLeast(75);
     }
 
-    // Special ViewAssertion is needed for GONE views
-    @Deprecated
     public static Matcher<View> matcherForNotVisible() {
-        return not(isDisplayed());
+        return anyOf(nullValue(), not(isDisplayed()));
```
d4vidi: a core fix

detox/android/detox/src/main/java/com/wix/detox/espresso/matcher/NullableTypeSafeMatcher.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.internal.ReflectiveTypeFinder;
+
+/**
+ * Identical to Espresso's {@link org.hamcrest.TypeSafeMatcher}, but allows nulls (in which case
+ * the "item" is passed-in into the sub-class nonetheless, as a null).
+ */
+public abstract class NullableTypeSafeMatcher<T> extends BaseMatcher<T> {
```
d4vidi: a core fix

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAssertion.java
LeoNatan: Fixed. #1473

PR #1472, opened 2019-06-24
Upgrade to RN 0.59.9 to support Xcode 11

LeoNatan:
.

### DIFF
--- a/detox/test/index.ios.js
+++ b/detox/test/index.ios.js
@@ -22,4 +22,5 @@ class exampleIos extends example {
   }
 }
 
+console.disableYellowBox = true;
 AppRegistry.registerComponent('example', () => exampleIos);
--- a/detox/test/package.json
+++ b/detox/test/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "moment": "^2.24.0",
     "react": "16.8.3",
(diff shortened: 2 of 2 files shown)


LeoNatan:
Test failure seems unrelated to RN upgrade in test suite.

detox/test/index.ios.js
```diff
@@ -22,4 +22,5 @@ class exampleIos extends example {
   }
 }
 
+console.disableYellowBox = true;
```
d4vidi: @LeoNatan why is this needed? why do we all of a sudden have yellow boxes?

d4vidi:
@LeoNatan did you also update the builds on jenkins (master and pr)?

detox/test/index.ios.js
d4vidi: @LeoNatan I strongly advise against this

LeoNatan: This is only on iOS on debug. I don't need them.

LeoNatan: RN by default has yellowboxes on iOS 12 and 13.

LeoNatan:
No, it's just in the project to have support for Xcode 11. CI is unchanged.

PR #1474, opened 2019-06-24
(Android) Improve reporting of 'adb devices' for #1427

d4vidi:
Improve reporting of 'adb devices' for #1427, so that users could report important stuff back when (if) this happens.

### DIFF
--- a/detox/src/devices/android/ADB.js
+++ b/detox/src/devices/android/ADB.js
@@ -288,16 +288,19 @@ class ADB {
 }
 
 function _reportTelnetPortResolutionError(input, devicesList, deviceAdbName, port) {
-  const log = require('../../utils/logger').child({ __filename });
   const {encodeBase64} = require('../../utils/encoding');
-  log.error({event: 'DEVICE_NAME_ERROR'}, `Failed to determine port for emulator device '${deviceAdbName}!!!`);
-  log.error({event: 'DEVICE_NAME_ERROR'}, `Please help us out by reporting all DEVICE_NAME_ERROR logs in: https://github.com/wix/Detox/issues/1427`);
-  log.error({event: 'DEVICE_NAME_ERROR'}, `State dump ==>`);
-  log.error({event: 'DEVICE_NAME_ERROR'}, `adb devices: ${input}`);
-  log.error({event: 'DEVICE_NAME_ERROR'}, `adb devices (base64): '${encodeBase64(input)}'`);
-  log.error({event: 'DEVICE_NAME_ERROR'}, `devicesList: ${devicesList}`);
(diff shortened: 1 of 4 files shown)


detox/src/devices/android/ADB.js
```diff
@@ -283,4 +287,17 @@ class ADB {
...
+  const log = require('../../utils/logger').child({ __filename });
+  const {encodeBase64} = require('../../utils/encoding');
+  log.error({event: 'DEVICE_NAME_ERROR'}, `Failed to determine port for emulator device '${deviceAdbName}!!!`);
+  log.error({event: 'DEVICE_NAME_ERROR'}, `Please help us out by reporting all DEVICE_NAME_ERROR logs in: https://github.com/wix/Detox/issues/1427`);
+  log.error({event: 'DEVICE_NAME_ERROR'}, `State dump ==>`);
+  log.error({event: 'DEVICE_NAME_ERROR'}, `adb devices: ${input}`);
+  log.error({event: 'DEVICE_NAME_ERROR'}, `adb devices (base64): '${encodeBase64(input)}'`);
+  log.error({event: 'DEVICE_NAME_ERROR'}, `devicesList: ${devicesList}`);
+  log.error({event: 'DEVICE_NAME_ERROR'}, `port: ${port}`);
+  throw new Error(`Unable to determine port for emulator device '${deviceAdbName}'!`);
```
noomorph: I suggest extracting `throw new Error` out of `_report...` method, to the `if` branch above.

detox/src/utils/exec.js
```diff
@@ -43,7 +43,7 @@ async function execWithRetriesAndLogs(bin, options, statusLogs, retries = 10, in
     throw new DetoxRuntimeError(`command ${cmd} returned undefined`);
   }
 
-  _logExecOutput(log, result, verbosity === 'high' ? 'debug' : 'trace');
+  _logExecOutput(log, result, verbosity === 'high' ? 'info' : 'trace');
```
noomorph: Are you sure we need to print adb devices output in absolutely all cases, @d4vidi? We all know that there is no other way to get into Telnet error other than entering that unfortunate `if` branch. So why pollute logs for other people?

d4vidi:
@noomorph well I was contemplating between logs and and error message;  I suppose coming from Android where the log size is limited, I got used to crunching it. Anyways though I don't believe it matters much I _will_ change it - seems like a more proper way to do it. tnx

detox/src/utils/exec.js
d4vidi: I can't see how 2-3 lines can pollute anything, but nevertheless after adding detailed logging this is indeed pointless...

PR #1478, opened 2019-06-25
Update APIRef.ActionsOnElement example

leotm:
- [x] This is a small change 

---

**Description:**

The documentation is good, but the example is deceptive.
This makes it super clear, especially to those skim-reading documentation (around 60% of people).


LeoNatan:
I don't think this is necessary.
The reason the examples are column 1 and 2 is because in a 12-hour locale, the first column is AM/PM.

LeoNatan:
I added a note about setting explicit locale to prevent issues like this:

https://github.com/wix/Detox/blob/master/docs/APIRef.ActionsOnElement.md#setcolumntovaluecolumn-value--ios-only

PR #1496, opened 2019-07-02
OR implementation main docs

Usamaliaquat123:
Review detox docs but i confused a little but then i see this needs a liitle bit change

- [x ] This is a small change 
---


LeoNatan:
This looks unnecessary. The wording is fine in that section. Thank you for your contribution.

Issue #1157, opened 2019-02-14
Wont Launch Valid Android Emulator AVD


RyanThomas73:
#### Description
The `detox test` command fails with the error message "Error: Emulator with name <emulator name> has a corrupt config.ini file (/path/to/config.ini), try fixing it by recreating an emulator.

This happens for a perfectly valid emulator avd that starts up without problem. In my particular case the avd was created from the command line using the `avdmanager.bat create avd` command.

Looking through the source code, this error is thrown from `EmulatorDriver.js _fixEmulatorConfigIniSkinName(..)` when the config.ini file for the emulator does not have a 'skin.name' value and does not have `hw.lc.width` and `hw.lcd.height` values.

It is unclear why the emulator driver is trying to fix the config.ini in this way. Ideally the command should work for such an emulator avd since the avdmanager does not create it with those values and the avd works without them.

If there is a specific scenario where this behavior is required or it is a requirement specific to detox then, at a minimum, the error message should be made more descriptive to provide a clear solution. Such as instructing the user that skin and/or height + width specification is required in the config.ini.

#### Steps to Reproduce
Run the `detox test -c <config>` command. Use a configuration of type `android.emulator` and select an avd name with a `config.ini` file that has neither the `skin.name` property nor the `hw.lcd.width` / `hw.lcd.height` properties.

#### Detox, Node, Device, Xcode and macOS Versions

* Detox: `10.0.7`
* React Native: `0.57.8`
* Node: `11.6.0`
* Device: `Android Emulator`
* Xcode: N/A 
* macOS: N/A
* Windows: `Windows 10 Pro v1703`

#### Device and verbose Detox logs
```
detox[716] INFO:  [DetoxServer.js] server listening on localhost:50369...
detox[716] DEBUG: [AsyncWebSocket.js/WEBSOCKET_OPEN] opened web socket to: ws://localhost:50369
detox[716] TRACE: [AsyncWebSocket.js/WEBSOCKET_SEND] {"type":"login","params":{"sessionId":"3a17d4e8-c480-0900-5910-4068d33e6274","role":"tester"},"messageId":0}
detox[716] DEBUG: [DetoxServer.js/LOGIN] role=tester, sessionId=3a17d4e8-c480-0900-5910-4068d33e6274
detox[716] DEBUG: [DetoxServer.js/LOGIN_SUCCESS] role=tester, sessionId=3a17d4e8-c480-0900-5910-4068d33e6274
[...]

PR #1504, opened 2019-07-03
Allow valid Android AVDs to run (#1157)

laurentS:
- [x] This is a small change 

---

**Description:**
This is a proposal to fix https://github.com/wix/Detox/issues/1157 (I sent a PR instead of commenting on the issue as it is locked).

- `git blame` shows that the problematic code dates back to https://github.com/wix/Detox/commit/419d356506f0c1751072b8c44bfc663bece4a856
- `skin.name` (or even `skin`) does not appear anywhere else in the `detox` code base (as of tag `13.0.2`)
- the added `skin.name` in `config.ini` does not seem required, as emulators created and launched by the latest android SDK function well without this added line.
- detox works well on my app with this patch applied, although I've not been able to run the e2e part of the test suite (JS tests were fine)

I hope it's ok to propose this PR before having discussed it in the issues. If not, please reopen #1157 :)

### DIFF
--- a/detox/src/devices/drivers/EmulatorDriver.js
+++ b/detox/src/devices/drivers/EmulatorDriver.js
@@ -18,24 +18,6 @@ class EmulatorDriver extends AndroidDriver {
     this.emulator = new Emulator();
   }
 
-  async _fixEmulatorConfigIniSkinName(name) {
-    const configFile = `${os.homedir()}/.android/avd/${name}.avd/config.ini`;
-    const config = ini.parse(fs.readFileSync(configFile, 'utf-8'));
-
-    if (!config['skin.name']) {
-      const width = config['hw.lcd.width'];
-      const height = config['hw.lcd.height'];
-
-      if (width === undefined || height === undefined) {
-        throw new Error(`Emulator with name ${name} has a corrupt config.ini file (${configFile}), try fixing it by recreating an emulator.`);
-      }
-
-      config['skin.name'] = `${width}x${height}`;
-      fs.writeFileSync(configFile, ini.stringify(config));
-    }
-    return config;
-  }
-
   async boot(avdName) {
     let adbName = await this._findADBNameByAVDName(avdName, { strict: false });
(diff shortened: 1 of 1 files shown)


rotemmiz:
The original issue was with emulators created with `avdmanager create avd` and launched through CLI, not Android Studio (issue described here https://stackoverflow.com/a/47265664/930391).

I am not sure Google has solved the issue, last time I tried this, about 6 months ago, the issue was still present on macOS. Either if its solved or not in newer version, I'd rather keep this piece of code to support older SDK versions. Let's find a better solution together, but I need to understand:
1. How your emulators were created (cli, Android Studio?)
2. How the config.ini looks like (paste it if you can please)
3. How are you launching the emulator? (With Detox? from Android Studio?)

I'll copy this comment to #1157 and let's continue discussing there.

laurentS:
Hi @rotemmiz thanks for looking into it. I'm replying here as #1157 is locked.

- I created my emulator with the command line `tools/bin/avdmanager create avd --force -n ci_avd -k 'system-images;android-27;google_apis;x86' --abi x86 -c 100M`
- Here is the `config.ini` that was created:
```
PlayStore.enabled=false
abi.type=x86
avd.ini.encoding=UTF-8
hw.cpu.arch=x86
hw.sdCard=yes
image.sysdir.1=system-images/android-27/google_apis/x86/
sdcard.size=100M
tag.display=Google APIs
tag.id=google_apis
```
- I have launched the emulator with the command line `emulator/emulator @ci_avd -no-audio -verbose`

Here's the version of emulator:
```bash
$ emulator/emulator -version
Android emulator version 27.3.10.0 (build_id 4969155) (CL:751f3c993c814a52adf144c159615ce457c2d282)
```

For now I've been manually adding either a `skin.name=...` line or `hw.lcd.height=...` and `hw.lcd.width=...` to the end of the file, which allows detox to run (it doesn't without)

rotemmiz:
Unlocked #1157

PR #1510, opened 2019-07-07
Add option to keep simulator lock file to cli

swabbass:
- [x] This is a small change 

---

**Description:**
when running detox test, the devices lock file is removed, detox test -keepsimlock will not delete the lock file.
it will be manually managed.

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -134,7 +134,7 @@ module.exports.builder = {
   },
   keepLockFile:{
     group: 'Configuration:',
-    describe:'Keep Simulators lock file when running detox test'
+    describe:'Keep the device lock file when running Detox tests'
   },
   n: {
     alias: 'device-name',
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -126,11 +126,11 @@ module.exports.builder = {
   H: {
     alias: 'headless',
     group: 'Execution:',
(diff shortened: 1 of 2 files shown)


LeoNatan:
Please explain how this will work.
The lock file is an internal implementation detail of Detox.

swabbass:
The way this flag will be used is when running detox test from two separate processes it will prevent removing the lock file for devices once the detox test command starts working.

detox/local-cli/test.js
```diff
@@ -132,6 +132,10 @@ module.exports.builder = {
     group: 'Execution:',
     describe: '[Android Only] Launch Emulator with the specific -gpu [gpu mode] parameter.'
   },
+  keepsimlock:{
```
LeoNatan: Please name this `keepLockFile`

detox/local-cli/test.js
```diff
@@ -132,6 +132,10 @@ module.exports.builder = {
...
+    group: 'Configuration:',
+    describe:'keep simulator lock file when running detox test'
```
LeoNatan: Please use proper capitalization. Thanks

detox/local-cli/test.js
swabbass: done

LeoNatan:
Thanks!

PR #1515, opened 2019-07-09
Add explicit note about React Native 0.60 and workspace

CodingItWrong:
- [x] This is a small change 

---

As mentioned in [the React Native 0.60 announcement blog post](https://facebook.github.io/react-native/blog/2019/07/03/version-60#cocoapods-by-default), React Native now uses Cocoapods, and therefore `.xcworkspace` files, by default.

To make the Getting Started instructions easier for users who aren't familiar with Cocoapods, this PR explicitly says that the `-workspace` option is needed for React Native 0.60 and above.

I've confirmed in [a tutorial project](https://github.com/learn-tdd-in/react-native) that in RN 0.60 Detox does not work with the `-project` flag but does work with this change applied.


LeoNatan:
Yuck, I hope that is not mandatory.

LeoNatan:
Thanks!

Issue #1513, opened 2019-07-08
Disable RN Timers and Animation synchronization to prevent tests from stalling


leanmazzu:
**Is your feature request related to a problem? Please describe.**
It's related to a problem we've been facing at Skyscanner. Some of our tests were stalling in Android because of `ReactNativeTimersIdlingResource.isIdleNow()` and  `AnimatedModuleIdlingResource.isIdleNow()` never returning true in some cases. I couldn't identify exactly _when_ this happens since some of the times this happened we didn't have any infinite animations running in that particular screen. 

**Describe the solution you'd like**
We already have the `disableSynchronization()` function but that's only disabling the network synchronization. So my proposal is to add similar methods to disable RN timers synchronization and animation synchronization. You can see an already implemented version of this here: https://github.com/leanmazzu/Detox/commit/efb0dcfd799ffea3d91bd1c8e46ee84df3950dd6. This is what we currently have in place in our Skyscanner fork and it's working just fine for us.

**Describe alternatives you've considered**
N/A

**Additional context**
This would be a **non-breaking** change as it's not modifying the current `disableSynchronization()` method or anything else.

LeoNatan:
Hmm `disableSynchronization()` should disable all synchronization. @rotemmiz Why is this not the case in Android?

@leanmazzu Feel free to submit a PR that implements this.

leanmazzu:
Cool will do, just one question... the solution I proposed above is only adding the functionality to disable `ReactNativeTimersIdlingResource` and `AnimatedModuleIdlingResource` that was enough for us but, do you think it's worth adding the functionality to disable `ReactNativeUIModuleIdlingResource` as well? Looks like it could help solve this issue https://github.com/wix/Detox/issues/1435

LeoNatan:
I think a `disableSynchronization()` option should disable everything, like it does on iOS. Waiting for @rotemmiz

rotemmiz:
It should, and if it doesn't we need to fix it.
Also, we should probably add per IR disable mechanism (as much as we can)

LeoNatan:
We don't have that mechanism on iOS either for now. But at least the main `disable` should disable 😄 

@leanmazzu We'd be happy to receive a pull request. Thank you!

leanmazzu:
Ok cool, I'll do this then:

- `disableSynchronization()` will now disable **all** the IR.
- There will be a `disable` method for every IR so that they can be disabled/enabled separately.

[...]

PR #1521, opened 2019-07-11
Add disable/enable synchronisation for each individual IR. Refactor c…

leanmazzu:
…urrent disableSynchronization method to disable ALL sync.

- [ ] This is a small change 
- [X] This change has been discussed in issue <https://github.com/wix/Detox/issues/1513> and the solution has been agreed upon with maintainers.

---

**Description:**

This PR does two things:

- Refactor `disableSynchronization()` which will now disable all the IRs. As discussed in issue <https://github.com/wix/Detox/issues/1513> `disableSynchronization()` should disable everything.
- Add methods to allow enable/disable synchronization of each IR individually.

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/ReactNativeSupport.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/ReactNativeSupport.java
@@ -184,18 +184,24 @@ public class ReactNativeSupport {
 
         setupReactNativeQueueInterrogators(reactContext);
 
-        rnBridgeIdlingResource = new ReactBridgeIdlingResource(reactContext);
-        rnTimerIdlingResource = new ReactNativeTimersIdlingResource(reactContext);
-        rnUIModuleIdlingResource = new ReactNativeUIModuleIdlingResource(reactContext);
-        animIdlingResource = new AnimatedModuleIdlingResource(reactContext);
-
-        IdlingRegistry.getInstance().register(rnTimerIdlingResource);
-        IdlingRegistry.getInstance().register(rnBridgeIdlingResource);
-        IdlingRegistry.getInstance().register(rnUIModuleIdlingResource);
-        IdlingRegistry.getInstance().register(animIdlingResource);
-
         if (networkSyncEnabled) {
             setupNetworkIdlingResource();
+        }
+
+        if (animationSyncEnabled) {
+            setupAnimationIdlingResource();
+        }
+
+        if (rnTimersSyncEnabled) {
+            setupRNTimersIdlingResource();
+        }
+
+        if (rnUIModuleIdlingResourceEnabled) {
+            setupRNUIModuleIdlingResource();
+        }
+
+        if (rnBridgeIdlingResourceEnabled) {
+            setupRNBridgeIdlingResource();
         }
     }
 
@@ -204,7 +210,7 @@ public class ReactNativeSupport {
         Looper JSMessageQueue = getLooperFromQueue(reactContext, FIELD_JS_MSG_QUEUE);
         Looper JMativeModulesMessageQueue = getLooperFromQueue(reactContext, FIELD_NATIVE_MODULES_MSG_QUEUE);
 
-//        IdlingRegistry.getInstance().registerLooperAsIdlingResource(UIBackgroundMessageQueue);
+        //        IdlingRegistry.getInstance().registerLooperAsIdlingResource(UIBackgroundMessageQueue);
         IdlingRegistry.getInstance().registerLooperAsIdlingResource(JSMessageQueue);
         IdlingRegistry.getInstance().registerLooperAsIdlingResource(JMativeModulesMessageQueue);
 
@@ -245,6 +251,14 @@ public class ReactNativeSupport {
         reactContext.getCatalystInstance().removeBridgeIdleDebugListener(rnBridgeIdlingResource);
     }
 
+    public static void enableSynchronization(boolean enable) {
+        enableRNTimersSynchronization(enable);
+        enableAnimationSynchronization(enable);
+        enableNetworkSynchronization(enable);
+        enableRNBridgeSynchronization(enable);
+        enableRNUIModuleSynchronization(enable);
+    }
+
     private static boolean networkSyncEnabled = true;
     public static void enableNetworkSynchronization(boolean enable) {
         if (!isReactNativeApp()) return;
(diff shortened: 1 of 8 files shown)


LeoNatan:
👑

rotemmiz:
Thanks for this precious PR.
1. API looks great
2. I see a lot of code repetition in Java, I believe line count can be decreased if you add one generic function, WDYT?
3. This new API must be tested in E2E (same as we do here
https://github.com/wix/Detox/blob/master/detox/test/e2e/14.network.test.js#L30), maybe a new e2e test file, something like x.synchronization.test.js

rotemmiz:
@LeoNatan is the API sufficient for iOS, when we'll implement it?

LeoNatan:
@rotemmiz Seeing that on iOS, disabling synchronization fully works, I don't think it's a priority to implement the granularity. We'll get to it eventually.

leanmazzu:
Hey sorry for the late response guys:

> I see a lot of code repetition in Java, I believe line count can be decreased if you add one generic function, WDYT?

If you're talking about `ReactNativeSupport.java` I agree. I'll add generic `setupIdlingResource(IdlingResource ir)` and `removeIdlingResource(IdlingResource ir)` methods that contain the repeated logic and that can be called from the specific setup and remove methods.

> This new API must be tested in E2E (same as we do here
https://github.com/wix/Detox/blob/master/detox/test/e2e/14.network.test.js#L30), maybe a new e2e test file, something like x.synchronization.test.js

Cool, this one might take me a little while and I might need some help from your side, but I'll try my best to implement it and will come back to you with questions if it gets too beyond my knowledge.

I'll work on the rest of the requested changes as well, thanks!

rotemmiz:
Glad to help, ping us whenever you need.

LeoNatan:
Something we discussed internally but I think we should discuss here is potential issues that such granularity might reveal. For instance, disabling RN timers (`setTimeout()`) might appear logical when some endless JS animation is preventing tests, but counterintuitively, it might have an effect that other RN synchronization becomes broken because RN uses timers for many things internally. I cannot make an educated guess how much of an issue this is. Most likely, if we see an issue, it would be during initial RN load (or during reload), where RN uses timers, and if timer idle resources are disabled, we might not wait for RN to load correctly. I wonder how we can test this in the testing suite to make sure we minimize these side effects.

leanmazzu:
@LeoNatan I was wondering exactly the same thing while thinking of how to implement e2e testing for some of these enable/disable IR methods.

For some of them, like `AnimatedModuleIdlingResource` or `ReactNativeUIModuleIdlingResource` I think we could assert they are actually working by just creating some view with an endless animation and asserting that the synchronization is actually disabled and using manual synchronization (`waitFor`).

But for `ReactNativeTimersIdlingResource` or `ReactNativeBridgeIdlingResource` I don't really know how to proceed. 

I get your point on that RN timers are used for more than animations. One thing that I noticed when our tests where stalling is that the `timerReflected.interval` was always `1000` and, therefore, the `isTimerBusyWindow()` function was returning always `true`. See here: https://github.com/wix/Detox/blob/263ae4fc7770d2f9b3d2caee06e83591da16fe5d/detox/android/detox/src/main/java/com/wix/detox/espresso/ReactNativeTimersIdlingResource.kt#L99

I don't know the specifics of why the `BUSY_WINDOW_THRESHOLD` constant was set to `1500` nor why `timerReflected.interval` is returning that specific value in our case though.

Maybe we should stick the granularity in this PR only to `AnimatedModuleIdlingResource` and `ReactNativeUIModuleIdlingResource` and, if it's needed, think of fixing `ReactNativeTimersIdlingResource` in a different PR, WDYT?

LeoNatan:
It was set to 1500 as a compromise to ignore very long timers, while catching most common timers and animations.

mnzaki:
Any updates on this? I would like to help if possible and it would be nice to start from somewhere if there are already some tests implemented. Can you push any updates?

Also I think that changing the behavior of `disableSynchronization` to disable everything *may* be a breaking change, if some users accidentally depend on the behavior that it only disables network synchronization.

> For instance, disabling RN timers (setTimeout()) might appear logical when some endless JS animation is preventing tests, but counterintuitively, it might have an effect that other RN synchronization becomes broken because RN uses timers for many things internally.

Can someone please elaborate on this? As I understand it when we say "disable RN timers synchronization" we mean that the `IdlingResource` is unregistered, so that resources is not waited for any more when detox needs to determine if the app is busy. But how would that affect RN's internal use of timers? Would it affect how other `*IdlingResource` classes determine if their respective resource is currently busy?

leanmazzu:
Sorry for not being really active on this! It's been super busy days. Feel free to pick up on this! I think what's left is mostly tests coverage.

> Also I think that changing the behavior of `disableSynchronization` to disable everything _may_ be a breaking change, if some users accidentally depend on the behavior that it only disables network synchronization.
> 
> >

This is a fair point, but at the same time if we're planning to add specific disable methods then `disableSynchronization()` is not doing exactly what it says if it only disables the network synchronization.

Maybe we could mark this as a breaking change and remove/deprecate `disableSynchronization()` instead of changing its implementation so that existing users of this function don't get unexpected results. So my idea would be:

- remove/deprecate `disableSynchronization()`
- add `disableAllSynchronization()` this does what it says, just disable all the idling resources.
- add `disableNetworkSynchronization()` 
- add `disableRNTimersSynchronization()`
- add `disableUIModuleSynchronization()`
- add `disableAnimationSynchronization()`
- add `disableRNBridgeSynchronization()`

What do you all think?

noomorph:
I was just passing by, but IMO the latter syntax you suggested looks better than the other options:

```js
setSynchronization({
  timers: false,
  animations: false
});
```

Less boilerplate, more action.

chevonc:
Will this be merged soon? Would love to get android test suite running alongside iOS, but currently blocked on synchronization issues. Can we have some movement on this?

rotemmiz:
We can continue further discussing the API ( I personally like the API suggested by @mnzaki), but the most important part here is comment 3 (https://github.com/wix/Detox/pull/1521#issuecomment-510591964), this PR must include E2E tests, or else it will be super fragile.
Comment 2 also has some validity, as it will be hard to maintain the code with this much duplication.
As I said before, we are here to help if you need it, this PR is not complete yet, it needs some more attention.

leanmazzu:
Hi folks! I was trying to implement the suggested API:
```
device.setSynchronization(true);
device.setSynchronization(false);  
device.setSynchronization({
  timers: true,
  animations: false
});
```

So to support that, on the Java side, I would need to do something like this:
```
public static void setSynchronization(HashMap<String, Boolean> options) {
        if (options.containsKey(OPTIONS_ANIMATIONS)) {
            ReactNativeSupport.enableAnimationSynchronization(options.get(OPTIONS_ANIMATIONS));
        }

        if (options.containsKey(OPTIONS_BRIDGES)) {
            ReactNativeSupport.enableRNBridgeSynchronization(options.get(OPTIONS_BRIDGES));
        }

        if (options.containsKey(OPTIONS_NETWORK)) {
            ReactNativeSupport.enableNetworkSynchronization(options.get(OPTIONS_NETWORK));
        }

        if (options.containsKey(OPTIONS_TIMERS)) {
            ReactNativeSupport.enableRNTimersSynchronization(options.get(OPTIONS_TIMERS));
        }

        if (options.containsKey(OPTIONS_UIMODULE)) {
            ReactNativeSupport.enableRNUIModuleSynchronization(options.get(OPTIONS_UIMODULE));
        }
    }
```

But there's a problem with this solution. We are using code generation to create `EspressoDetox.js`and this code generation is done with [`java-method-parser`](https://github.com/DanielMSchmidt/java-method-parser). 

This library is using the comma `,` as argument separator, so when it finds `HashMap<String,Boolean>` it splits this argument and takes it as two different arguments like `HashMap<String` and `Boolean>`. 

So then the code generator fails as it can't these to anything in the supported types list (I added `HashMap<String,Boolean>` to the list of supported types in `generation/adapters/android.js`).

noomorph:
@leanmazzu, I'll see if I can help with fixing `java-method-parser`, or maybe someone else volunteers. In any case, let's strive to introduce a good API, rather than an API which is a victim of the situation.

rotemmiz:
We would need to replace it with an AST java parser, this is something we need to put in our backlog.

rotemmiz:
Anyway, for any suggested fix to java-method-parser we'll need @DanielMSchmidt's help merging and releasing a new version, or we can fork or merge java-method-parser into this repo if that doesn't work.

DanielMSchmidt:
If you send in a PR I can review and merge it, no problem :)

noomorph:
@DanielMSchmidt , sure, please see https://github.com/DanielMSchmidt/java-method-parser/pull/152

noomorph:
@leanmazzu, you should already be unblocked. Ping me if you bump into other similar troubles.

chevonc:
Any eta on when this will get merged? Awaiting these updates so we can run our android tests :)

rotemmiz:
Ping @leanmazzu

leanmazzu:
Sorry for taking so long to getting back on this!

I had a discussion about this solution with Omri Bruchim from Wix while he was visiting us in London and the conclusion was that disabling the synchronization kind of defeats the purpose of Detox, which is that synchronization should be handled automatically. So the solution should be around trying to fix the actual IdlingResources to improve the synchronization. With that in mind, do we still think this is a good approach? 

If we think this PR is still a valid approach, I could try and commit the changes to the API. But I'd need for some volunteers to take care of covering this with E2E testing since I honestly lack the knowledge on how to do that or how to actually assert this is working as expected.

PR #1530, opened 2019-07-12
Change Expo guide to recommend `.bin` folder prefixed by a dot

CodingItWrong:
- [x] This is a small change 

---

When following the previous instructions of naming the folder `bin`, Expo sees assets within `bin/Exponent.app` as assets to be included in the published bundle. It gives the following warning upon launch:

```
Warning: This project contains unoptimized assets. Please run `expo optimize` in your project directory.
```

The app still runs locally and Detox succeeds, but when attempting to publish, a long sequence of `bin/Exponent.app` assets are optimized:

```
Optimizing /Users/josh/apps/ribbons/native/bin/Exponent.app/AppIcon20x20@2x.png
Compressed version of /Users/josh/apps/ribbons/native/bin/Exponent.app/AppIcon20x20@2x.png same size as original. Using original instead.
Optimizing /Users/josh/apps/ribbons/native/bin/Exponent.app/AppIcon20x20@3x.png
Compressed version of /Users/josh/apps/ribbons/native/bin/Exponent.app/AppIcon20x20@3x.png same size as original. Using original instead.
```

Eventually for me the optimization process hangs. Preventing it from hanging isn't the fix, though--we don't want these assets included in the published build, whether optimized or unoptimized.

I tried a few options to configure Expo to exclude `bin` from the build without success. But simply naming the folder to `.bin` seems to work as well--we do not get the "unoptimized assets" warning upon launch, and the publish operation succeeds without attempting to include `bin/Exponent.app` assets.

### DIFF
--- a/docs/Guide.Expo.md
+++ b/docs/Guide.Expo.md
@@ -9,7 +9,7 @@
 "detox": {
   "configurations": {
     "ios.sim": {
-      "binaryPath": "bin/Exponent.app",
+      "binaryPath": ".bin/Exponent.app",
       "type": "ios.simulator",
       "name": "iPhone 7"
     }
@@ -19,7 +19,7 @@
 
 - Download the Expo Client iOS App from [Expo.io/tools](https://expo.io/tools#client).
 - Unzip the iOS IPA and **rename the folder** to `Exponent.app`. It'll have a file icon but will still be a folder.
-- Create `bin` folder and put `Exponent.app` inside so it matches the binaryPath set above.
+- Create `.bin` folder and put `Exponent.app` inside so it matches the binaryPath set above.
 - Create an `e2e` and copy over the settings from [the example app](https://github.com/expo/with-detox-tests/tree/master/e2e)
 
 The most important piece of this the `reloadApp` from `detox-expo-helpers`. Don't forget this.


LeoNatan:
I think under default rules, Finder doesn't show hidden files and folders (name has a "." prefix). That would be problematic.

CodingItWrong:
That makes sense that it's not ideal, but I think the current error behavior is more problematic—you can't publish to Expo.

I'm up for suggestions of other fixes; maybe you know more about the React Native Build process to have a way to filter out the bin folder.

LeoNatan:
I really don't. 😂 I am the last person you want to ask RN questions.
I stay away from that as much as humanly possible.

Can't the folder be simply deleted or moved when publishing? Isn't there some kind of ignore file that can be used?

CodingItWrong:
I tried a few things with ignore files; I'll add details here in case someone with more info than us can chime in.

Let me know if you'd prefer for me to close this PR and open an Issue with the same info; that way the discussion can proceed apart from a specific solution approach.

LeoNatan:
We can discuss here or in issue, not really a problem for now. But I wonder how no one has complained about this so far.

CodingItWrong:
Closing this as we don't want to move forward with it.

I will say that it's a bit discouraging that the docs describe Expo support, I'm raising an issue with Expo support, and it's not really being looked into. If you can't reproduce the problem I'm reporting then it's not a valid issue. But if no one can take the time to try to reproduce it, maybe the Expo guide should be prefaced with a note that it's an unsupported configuration.

LeoNatan:
Nobody in Wix uses Expo. Expo support has been, from the beginning, a community effort. You've mentioned an issue here that nobody else has complained about, and offered a solution which, as discussed before, has drawbacks. The PR was left open to allow other members of the community to weight in, but nobody did. Perhaps this issue is better discussed in Expo community?

CodingItWrong:
I decided to just drop the issue as it wasn't urgent for me.

Apart from that, I feel like the official Expo page in the docs communicates that Expo support is an official Detox effort, and I've made a suggestion of how you might communicate that it's not. You can feel free to take that input if you like.

LeoNatan:
I will add a few words there.

PR #1531, opened 2019-07-14
(Android) Convert unit-tests to specs using Spek [fixed]

d4vidi:
**Description:**

Reimplement Android unit tests as Specs using the [Spek](https://spekframework.org) framework (and JUnit 5 under the hood).


rotemmiz:
Showoff!

PR #1551, opened 2019-07-24
(Android) Fix undesired turning of typeText() taps into long taps over InputText components

### LINKED ISSUE #1462: typeText() performs a long tap
On Android when I use typeText('text') it performs maybe a long tap because it shows 'paste' before type 'text'. 
I assume that you have some tap interaction behind the typeText that is causing this behavior.

Screenshot 2019-06-19 at 10 03 36

d4vidi:
- [x] This is a small change 
- [x] This change has been discussed in issue #1462 and the solution has been agreed upon with maintainers.

---

**Description:**

Fix #1462 by replacing Espresso's implementation of `typeText()` with a custom one that manages the pre-tap in a way that is safe for RN apps (just like detox does for standard clicks).

As a side note, the challenge here was not in fixing the issue (except for researching it), but in having associated use cases tested. This isn't trivial since `TextInput` components cannot be set a long-press callback from the JS side.

### DIFF
--- a/detox/test/e2e/03.actions.test.js
+++ b/detox/test/e2e/03.actions.test.js
@@ -32,34 +32,46 @@ describe('Actions', () => {
   });
 
   it('should type in an element', async () => {
-    let typedText = device.getPlatform() === 'ios' ? 'Type Working 123 אֱבּג абв!!!' : "Type Working!!!";
-    await element(by.id('UniqueId937')).tap();
+    const typedText = device.getPlatform() === 'ios' ? 'Type Working 123 אֱבּג абв!!!' : "Type Working!!!";
+
+    if (device.getPlatform() === 'ios') {
+      await element(by.id('UniqueId937')).tap();
+    }
     await element(by.id('UniqueId937')).typeText(typedText);
     await expect(element(by.text(typedText))).toBeVisible();
   });
 
   it('should press the backspace key on an element', async () => {
-    let typedText = 'test';
-    await element(by.id('UniqueId937')).tap();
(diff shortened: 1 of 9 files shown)


detox/test/e2e/03.actions.test.js
```diff
@@ -32,34 +32,46 @@ describe('Actions', () => {
   });
 
   it('should type in an element', async () => {
-    let typedText = device.getPlatform() === 'ios' ? 'Type Working 123 אֱבּג абв!!!' : "Type Working!!!";
-    await element(by.id('UniqueId937')).tap();
+    const typedText = device.getPlatform() === 'ios' ? 'Type Working 123 אֱבּג абв!!!' : "Type Working!!!";
+
+    if (device.getPlatform() === 'ios') {
```
d4vidi: @LeoNatan could you get around to fixing this, soon?

LeoNatan: Is there a specific problem here or what we discussed in Slack?

d4vidi: Yes, as I explained on Slack

LeoNatan: Yes, yes, I thought the comment was about the line above, but it's about the .tap(). Looking into it.

d4vidi: Cool, let me know what you come up with - this is interesting :)

LeoNatan: Found another issue with `clearText()`. Fixing.

Issue #1255, opened 2019-04-04
Emulator Config file path is not built correctly


geraniumdt:
**Describe the bug**
I'm running on a Windows machine and I have changed the ANDROID_SDK_HOME location to point another drive. However, the configFile path from EmulatorDriver (_fixEmulatorConfigIniSkinName function) is using the homeDir as a base for the path instead of the environment variable of the sdk, resulting in an invalid path.

**To Reproduce**

1. Change the ANDROID_SDK_HOME to point to another location
2. Open an emulator
3. Run detox test

**Expected behavior**
The emulator config file should be correctly identified at the new location

**Actual behavior**

The following error is thrown:
```
 { Error: ENOENT: no such file or directory, open 'C:\Users\<placeholder_for_user>\.android\avd\Nexus_5X_API_27.avd\config.ini'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.readFileSync (fs.js:551:33)
    at EmulatorDriver._fixEmulatorConfigIniSkinName (D:\myApp\node_modules\detox\src\devices\drivers\EmulatorDriver.js:23:33)
    at EmulatorDriver.acquireFreeDevice (D:\myApp\node_modules\detox\src\devices\drivers\EmulatorDriver.js:91:16)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\Users\\<placeholder_for_user>\\.android\\avd\\Nexus_5X_API_27.avd\\config.ini' }
```

Let me know if more details are needed.

Thanks

noomorph:
@geraniumdt, I might be a bit busy with the other issues, so if you feel the strength to make a pull request addressing this shortcoming, I'd be very grateful! If not, well, I think I'll find a spare hour somewhere in April-May for that.

Grohden:
@noomorph i think that this issue is on the EmulatorDriver.js:

https://github.com/wix/Detox/blob/61ad39b7a84a6e73c2c7ecc0199ba3a13c391772/detox/src/devices/drivers/EmulatorDriver.js#L21-L25

It uses `os.homedir()`, we probably could solve this issue using these lines:
```javascript
 const sdkHome = process.env.ANDROID_SDK_HOME || os.homedir()
 const configFile = path.normalize(`${sdkHome}/.android/avd/${name}.avd/config.ini`);
```

[...]

PR #1573, opened 2019-07-31
Look for emulators in ANDROID_SDK_HOME or os.homedir()

pettersamuelsen:
- [x] This is a small change 
---

Fixes the issue https://github.com/wix/Detox/issues/1255#issuecomment-510211996 where the emulator path is incorrect in certain environments.

### DIFF
--- a/detox/src/devices/drivers/EmulatorDriver.js
+++ b/detox/src/devices/drivers/EmulatorDriver.js
@@ -19,7 +19,8 @@ class EmulatorDriver extends AndroidDriver {
   }
 
   async _fixEmulatorConfigIniSkinName(name) {
-    const configFile = `${os.homedir()}/.android/avd/${name}.avd/config.ini`;
+    const sdkHome = process.env.ANDROID_SDK_HOME || os.homedir();
+    const configFile = path.normalize(`${sdkHome}/.android/avd/${name}.avd/config.ini`);
     const config = ini.parse(fs.readFileSync(configFile, 'utf-8'));
 
     if (!config['skin.name']) {


detox/src/devices/drivers/EmulatorDriver.js
```diff
@@ -19,7 +19,8 @@ class EmulatorDriver extends AndroidDriver {
   }
 
   async _fixEmulatorConfigIniSkinName(name) {
-    const configFile = `${os.homedir()}/.android/avd/${name}.avd/config.ini`;
+    const sdkHome = process.env.ANDROID_SDK_HOME || os.homedir();
+    const configFile = path.normalize(`${sdkHome}/.android/avd/${name}.avd/config.ini`);
```
noomorph: I am not entirely sure, but isn't `sdkHome` supposed to include that `.android` as well?
I made a quick look up in: https://stackoverflow.com/questions/23042638/how-do-i-set-android-sdk-home-environment-variable
And among examples of how can `ANDROID_SDK_HOME` look like, I see strings like that:

```
C:\Users\<USERNAME>\.android\
```

Hence, it is meant to include `.android` dir path within. So why do you append `.android` again to it?

noomorph: @pettersamuelsen, ping.

noomorph: I have checked that stuff on SO once again, and it seems that the OP made a typo in the answer there.

Nevertheless, I dug into the docs:  https://developer.android.com/studio/command-line/variables

As far, as I understand, what really matters, is this environment variable: `ANDROID_EMULATOR_HOME`. Experimentally verified:

```bash
> emulator -list-avds
# Nexus_4_4
# Nexus_4_API_25
# Nexus_5X_API_23
# Nexus_5X_API_24
# Nexus_5X_API_25
# Nexus_5X_API_26
# Nexus_5X_API_28
# Nexus_5_0
```

and with it:

``` bash
> ANDROID_EMULATOR_HOME=/tmp emulator -list-avds
#
```

The same goes for `ANDROID_AVD_HOME`. It affects `emulator` executable.

But as for `ANDROID_SDK_HOME` - alas, I have not found any effect on the emulator boot process. Looks like it is ignored in this regard.

So, maybe you should use `ANDROID_AVD_HOME` + `ANDROID_EMULATOR_HOME` in your logic there?

benoitdion: @pettersamuelsen thoughts on next steps?

Issue #1575, opened 2019-08-01
Refactor apk build path to not include architecture in its flavors


ayelenmarie:
Hi guys! I'm opening this enhancement request that I think will be very useful for Android developers and QAs.

**Is your feature request related to a problem? Please describe.**
Upon the new Play Store requirements for uploading different .apk for every architecture available in our app, we've encountered an issue while trying to test a project that deals with this. 
After running `detox build -c android.emu.debug` successfully, our `detox test -c android.emu.debug` fails and leaves us this error:

```  
Error: '.../mobile/android/app/build/outputs/apk/androidTest/developmentX86/debug/app-development-x86-debug-androidTest.apk' could not be found, did you run './gradlew assembleAndroidTest' ?
```
You can see that Detox is trying to access an APK for x86 in a folder that does not exist (developmentX86). This is because it's treating the architecture as a flavor. 

**Describe the solution you'd like**
Since builds for Android testing don't need to specify its architecture version, it would be great if Detox could just not include them when trying to make flavors for its `androidTest.apk`, as `APKPath.js` does when slices all different parts of the `originalApkPath`.

**Describe alternatives you've considered**
In `APKPath.js`, if we refactor line 7 in `getTestApkPath(originalApkPath)` to replace all possible architecture combinations into `""`, we could make flavors that don't have those, eliminating the wrong path that Detox tries to find when running tests:

```
    const originalApkPathObj = path.parse(originalApkPath.replace(/x86\w*-/g, ''));
```
This should solve this problem as I've already tested it as well and it works with the two possible combinations of architecture: x86 and x86_64.

**Additional context**
Our Android configuration for Detox in the project in question:

```
      "android.emu.debug": {
        "binaryPath": "android/app/build/outputs/apk/development/debug/app-development-x86-debug.apk",
        "build": "cd android && ./gradlew assembleDevelopmentDebug assembleAndroidTest -DtestBuildType=debug && cd ..",
        "type": "android.emulator",
        "name": "Pixel_2_API_28"
      },
```

Detox version used: ` "detox": "13.3.2" `

I hope I explained myself well enough. I'll send a PR with the refactored line and the corresponding unit test as well. Just let me know if you need anything else.

Thanks!

rotemmiz:
Hey @ayelenmarie, 

I think that there are a few things that need to be cleared, since I'm not entirely sure how Detox fails to resolve the testApk path.
[...]

PR #1576, opened 2019-08-02
Refactor apk build path to not include architecture in its flavors #1575

ayelenmarie:
- [x] This is a small change 

---

**Description:**
Regarding my enhancement request #1575 :
This change allows `getTestApkPath()` in `APKPath.js` to return a correct path when it receives an APK with a specific architecture in it. I also included the corresponding unit test to this PR. You can check all the details in the issue.

Note: I think ESLint formatted some lines apart from line 7 and the added test and that's why they appear as changes too. Feel free to make it as it was before! 

Thanks!

### DIFF
--- a/detox/src/devices/android/APKPath.js
+++ b/detox/src/devices/android/APKPath.js
@@ -3,9 +3,8 @@ const _ = require('lodash');
 const string = require('../../utils/string');
 
 class APKPath {
-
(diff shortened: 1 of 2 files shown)


rotemmiz:
Added my notes in https://github.com/wix/Detox/issues/1575#issuecomment-517653167

PR #1554, opened 2019-07-25
Bugfix on Biometrics

abunsen:
**Description:**

This is a small change - I missed the "s" in `includes` on 3 methods and had an extra "m" in the anchor link on the docs.

PR #1547, opened 2019-07-22
Add Support for Biometrics on SimulatorDriver

abunsen:
**Description:**

This is a small change (which was requested in #958).

It adds support for Biometrics by adding the following methods to DeviceDriver:

- `enrollBiometrics` and `disenrollBiometrics` - for setting up biometrics on a Simulator
- `matchFace` and `unmatchFace` - for matching a face on a Simulator
- `matchFinger` and `unmatchFinger` - for matching a finger on a Simulator

It uses an underlying call in `AppleSimUtils.js` to three new methods:

- `setBiometricEnrollment` - which accepts a UDID and either `"YES"` or `"NO"`
- `matchBiometric` - which accepts a UDID and either `"Face"` or `"Finger"`
- `unmatchBiometric` - which accepts a UDID and either `"Face"` or `"Finger"`

I believe I added the appropriate tests so I think this should be good to go - however, I'm happy to change if the core maintainers are not content with my proposed changes. Please provide some suggestions on how it can be improved. Thank you!

PR #1587, opened 2019-08-06
Fix matchBiometric and unmatchBiometric commands

avanderberg:
- The matchBiometric and unmatchBiometric commands failed because of an incorrect parenthesis.
- The correct commands are --matchFace or --matchFinger, see https://github.com/wix/AppleSimulatorUtils

The biometrics functionality was introduced with https://github.com/wix/Detox/pull/1554 and https://github.com/wix/Detox/pull/1547

- [x] This is a small change 

---

**Description:**

Using the biometrics commands caused errors on my computer. I fixed them with this PR.

Fyi, using `await device.matchFinger();` still fails on my computer with the timeout below. I could not yet find the reason why. There may be additional errors that need fixing.
```
    1) with his matching finger
detox[71526] WARN:  [Client.js/PENDING_REQUESTS] App has not responded to the network requests below:
  (id = 12) waitForBackground: {}

That might be the reason why the test "Login with the correct fingerprint with his matching finger" has timed out.

  1 passing (6m)
  1 failing

  1) Login with the correct fingerprint
       with his matching finger:
     Error: Timeout of 280000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (e2e/tests/login.test.js)
```

### DIFF
--- a/detox/src/devices/ios/AppleSimUtils.js
+++ b/detox/src/devices/ios/AppleSimUtils.js
@@ -146,7 +146,7 @@ class AppleSimUtils {
       successful: `Matched ${matchType}!`
     };
 
-    await this._execAppleSimUtils({ args: `--byId ${udid} --${matchType}}` }, statusLogs, 1);
+    await this._execAppleSimUtils({ args: `--byId ${udid} --match${matchType}` }, statusLogs, 1);
   }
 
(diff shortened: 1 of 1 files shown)


LeoNatan:
Thanks!

PR #1588, opened 2019-08-06
use adb, aapt, and emulator on PATH if not found in ANDROID_SDK_ROOT

### LINKED ISSUE #1579: [Android] manually define adb path and skip check for local sdk installation

We are running a setup where we lease a remote android emulator and control it via adb (I think this is similar to how Genymotion works). In this setup, the local machine has adb but does not have a a full sdk installed. I am using the `android.attached` type since the emulator is already attached to adb.

When I try to use detox in this setup the error `$ANDROID_SDK_ROOT is not defined...` is thrown. If I go in and explicitly set the adb path in [detox's  source](https://github.com/wix/Detox/blob/dd34555bc2c23b24c16ad8e0b407c92115b0de8e/detox/src/devices/android/ADB.js#L12), I'm able to successfully run a Detox test since `getAndroidSDKPath()` isn't called. 

(I also have to run `adb reverse tcp:1234 tcp:1234` for the app to login to the detox server, but that works fine.)

A flag or argument to define the path to `adb` instead of requiring the entire sdk to be installed.(https://github.com/wix/Detox/blob/dd34555bc2c23b24c16ad8e0b407c92115b0de8e/detox/src/devices/android/ADB.js#L12)

cc @zsol

cs01:
- [x] This is a small change 
- [ ] This change has been discussed in https://github.com/wix/Detox/issues/1579 and the solution has been agreed upon with maintainers.

---

**Description:**
Allows the arguments `adb` and `aapt` to be passed when running a test. Fixes #1579.


rotemmiz:
Why not set ANDROID_SDK_ROOT env var instead?

cs01:
> Why not set ANDROID_SDK_ROOT env var instead?

In the setup we are using, the emulator exists remotely and is accessible via adb and port forwarding. The local machine does not have the android sdk, so trying to run detox with the `android.attached` type fails because detox still tries to locate the local SDK even though it's not needed.

More details are in https://github.com/wix/Detox/issues/1579.

rotemmiz:
And what about setting into an empty string? Still fail?

cs01:
If I run detox with
```
ANDROID_SDK_ROOT= 
```
I get 
```
detox[3775873] ERROR: [DetoxExportWrapper.js/DETOX_INIT_ERROR]
 Error: $ANDROID_SDK_ROOT is not defined, set the path to the SDK installation directory into $ANDROID_SDK_ROOT,
    Go to https://developer.android.com/studio/command-line/variables.html for more details
    at getAndroidSDKPath (/data/users/cssmith/fbsource/xplat/js/gauntlet/node_modules/detox/src/utils/environment.js:13:11)
    at Object.getAndroidEmulatorPath (/data/users/cssmith/fbsource/xplat/js/gauntlet/node_modules/detox/src/utils/environment.js:20:19)
...
```

But if I run with 
```
ANDROID_SDK_ROOT=/dev/null 
```
and set adb and aapt via the new arguments added in this PR, it does work.

noomorph:
I'm against this implementation. At the moment I see no need to create new `--adb`, `--aapt` CLI arguments.

If I grasp your situation correctly, you have `adb`, `aapt` & Co. resolved via `$PATH` variable.

If so, the point is just to teach Detox code to use non-absolute paths to `adb`, `aapt`, `emulator` (ideally, that is enabled when you have no `$ANDROID_HOME` and `$ANDROID_SDK_ROOT` variables set, but the said binaries still can be found somewhere in `$PATH`).

So, if you even would want to override them, then you should have configured your `$PATH` like `PATH=/more/prioritised/android-tools-path:$PATH`.

rotemmiz:
I also think its not a good API to add paths as a cli override command, especially if they are individual and don't act the same for all cli tools used by Detox.
What I think can be a good compromise is try to use the current implementation (the one that is in master), and if the tool is not found, try to find if the tool exist in path (either with `which`/`where` and their Windows counterpart), and then just use it.

cs01:
Thanks for the feedback. Agreed that searching the PATH is cleaner. I removed all arguments and made the following changes instead:

* add `which` utility functions
* search for `adb` on PATH
* search for `aapt` on PATH
* search for `emulator` on PATH
* remove unused Emulator in `AttachedAndroidDriver` class

cs01:
I just realized you requested the PATH be searched after ANDROID_SDK_ROOT was already searched. Right now it's searching PATH before ANDROID_SDK_ROOT. I'll fix it in the next commit (probably some time next week).

noomorph:
@cs01, sorry, I am not sure that your 'which' util really belongs to Detox project. Can we use third party there, e.g.: https://github.com/mathisonian/command-exists? Actually I am not insisting on specifically that module, but honestly I'd like to see that we import some module which is properly tested cross-platform.

As for SDK root precendce, yeah, that makes sense. Better to rewrite, indeed. That way the results are more predictable in a sense that the tools will be taken always from the same folder without a chance that one specific version from elsewhere will override something just because of PATH.

cs01:
No worries, I thought there would be hesitation to add a new dependency but if that's not the case I can use https://www.npmjs.com/package/npm-which. I'll push an update next week.

cs01:
This is ready for another review. 

Summary of changes:

* Remove unused `Emulator` in `AttachedAndroidDriver.js `
* If `adb`, `aapt`, or `emulator` are not found in `ANDROID_SDK_ROOT`, look for them on the PATH using the `which` package. If found, use it. If not found, re-throw the original error (`$ANDROID_SDK_ROOT is not defined, set the path to the SDK installation directory into $ANDROID_SDK_ROOT...`)

detox/src/utils/environment.js
```diff
@@ -17,12 +18,19 @@ function getAndroidSDKPath() {
...
 
 function getAndroidEmulatorPath() {
-  const sdkPath = getAndroidSDKPath();
-  const extension = os.platform() === 'win32' ? '.exe' : '';
-  const newEmulatorPath = path.join(sdkPath, 'emulator', `emulator${extension}`);
-  const oldEmulatorPath = path.join(sdkPath, 'tools', `emulator${extension}`);
-
-  return fs.existsSync(newEmulatorPath) ? newEmulatorPath : oldEmulatorPath;
+  try{
+    const extension = os.platform() === 'win32' ? '.exe' : '';
```
noomorph: I have an assumption, that you can do here much better, given that now you are leveraging `which` logic.

```js
return
  which.sync('emulator', { path: preferedPath, nothrow: true }) ||
  which.sync('emulator', { path: alternativePath, nothrow: true }) ||
  which.sync('emulator'); /* the latter will throw */
```

And no dances with `if fs.existsSync()` and `os.platform()`.

What do you think?

detox/src/devices/android/ADB.js
```diff
@@ -9,7 +10,16 @@ const Environment = require('../../utils/environment');
 class ADB {
   constructor() {
     this._cachedApiLevels = new Map();
-    this.adbBin = path.join(Environment.getAndroidSDKPath(), 'platform-tools', 'adb');
+    try{
+      this.adbBin = path.join(Environment.getAndroidSDKPath(), 'platform-tools', 'adb');
```
noomorph: Doesn't it make sense to convert this code into something like below?

```js
this.adbBin = which.sync('adb', { path: platformToolsPath });
```

detox/src/devices/android/ADB.js
```diff
@@ -9,7 +10,16 @@ const Environment = require('../../utils/environment');
...
+    } catch (err){
+      const adbOnPath =  which.sync('adb', {nothrow: true});
+      if (adbOnPath) {
+        this.adbBin = adbOnPath;
+        return;
+      }
+      throw new Error(err);
```
noomorph: So, if you care that much about rethrowing the original error, can you show please how does this error look like? Is it comprehensible enough for a human?

detox/src/devices/android/ADB.js
```diff
@@ -9,7 +10,16 @@ const Environment = require('../../utils/environment');
 class ADB {
   constructor() {
     this._cachedApiLevels = new Map();
-    this.adbBin = path.join(Environment.getAndroidSDKPath(), 'platform-tools', 'adb');
+    try{
+      this.adbBin = path.join(Environment.getAndroidSDKPath(), 'platform-tools', 'adb');
+    } catch (err){
+      const adbOnPath =  which.sync('adb', {nothrow: true});
+      if (adbOnPath) {
+        this.adbBin = adbOnPath;
```
noomorph: Why so many dances? Why not just `this.adbBin = which.sync('adb');`?

cs01: The reason I went through all this was to prefer the original logic and try that first, and if something wasn't found to throw the same error as detox has been throwing (see below).  I can simplify the code in a new commit.

```
function getAndroidSDKPath() {
  let sdkPath = process.env.ANDROID_SDK_ROOT || process.env.ANDROID_HOME;
  if (!sdkPath) {
    throw new Error(`$ANDROID_SDK_ROOT is not defined, set the path to the SDK installation directory into $ANDROID_SDK_ROOT,
    Go to https://developer.android.com/studio/command-line/variables.html for more details`);
  }
  return sdkPath;
}
```

cs01:
@noomorph thanks for the review and suggestions. Let me know what you think of my latest changes. 

The one tricky thing I ran into was in AAPT.js where the latest build tools version is found based on the directory contents. Since the environment variable may not exist and the error is no longer thrown at that point, the directory may not exist. So I added a check to confirm the directory exists before searching it.

detox/src/devices/android/AAPT.js
```diff
@@ -1,23 +1,44 @@
...
+      latestBuildToolsVersion = _.last(buildToolsDirs);
+    }
+
+    const buildToolsDirLatestVersion = path.join(
+      sdkPath,
+      'build-tools',
+      latestBuildToolsVersion,
+    );
+
+    this.aaptBin =
```
noomorph: One more issue emerged.
I see the same repetitive pattern across those multiple "Android tools"-related wrapper classes.
You get some path, you perform some `path.join`s, you perform some `exist`ence checks, you call `which.sync` multiple times, rinse and repeat.

I won't be repeating this message for all those three files you have changed, just examine them and get the idea.

I suggest extracting this logic to a separate class (e.g. `AndroidSDK`), thereby eliminating a single concern spread across multiple classes. You might then export an instance of it in `environment.js`.

On the low-level scale, your flow now looks better than previously.

detox/src/devices/android/ADB.js
```diff
@@ -9,7 +10,14 @@ const Environment = require('../../utils/environment');
...
   constructor() {
     this._cachedApiLevels = new Map();
-    this.adbBin = path.join(Environment.getAndroidSDKPath(), 'platform-tools', 'adb');
+    const platformToolsDir = path.join(Environment.getAndroidSDKPath(), 'platform-tools')
+    this.adbBin =
+      which.sync('adb', {path: platformToolsDir, nothrow: true}) ||
+      which.sync('adb', {nothrow: true});
+
+    if (this.adbBin == null){
+      throw new Error(Environment.MISSING_SDK_ERROR)
```
noomorph: I came to the conclusion that you wrote the code you haven't tested manually or via unit tests.
Not good, not good.

You see, `MISSING_SDK_ERROR` is not exported from `./environment.js`, and had you been striving to check all cases manually at least, you would have noticed this issue earlier than me. CI is red now for your pull request, pls fix.

cs01: I actually did test this code manually, but maybe didn't cover all the code paths.

cs01:
I am not able to address all your comments today, but wanted to suggest that since you have so many thoughts on how to implement this, that it would be fine with me if you wanted to take over this Pull Request and add the functionality. If not, I am fine continuing work on it but won't be able to get to it until next week.

detox/src/devices/android/ADB.js
noomorph: Okay, I see... maybe you haven't added `export ...` to staged changes in Git by an accident and committed without it. Anyways, no offence, man, I'm just concerned about testing edge cases in every piece of functionality that *cannot get a proper unit or e2e test coverage* for some historical or other reasons. As a result, sometimes I don't trust people who claim they have tested their PR thoroughly and if I have spare time I manually check their code on a few different filesystems and OSes. And sometimes those checks turn out to be not so useless. =)

noomorph:
@cs01, probably I'll be fine with your timeline but might help if I have an opportunity.

You see, here we've got a situation with two factors: an already existing aspiration to improve the longlasting situation around the functionality of Environment that is smeared among more files than IMO it really should be, and a boy scout rule, because the functionality in those places with this pull requests tends to grow a bit.

I would be really happy if this story gets a happy ending, and Detox gets the means to run those Android SDK tools using `$PATH` variable as a fallback.

cs01:
I see the CI tests are failing and can see why you suggested the class-based solution. I tried to run the tests locally but they failed, so I didn't attempt to make any more changes.
```
> scripts/ci.android.sh
Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlSchema
	at com.android.repository.api.SchemaModule$SchemaModuleVersion.<init>(SchemaModule.java:156)
	at com.android.repository.api.SchemaModule.<init>(SchemaModule.java:75)
	at com.android.sdklib.repository.AndroidSdkHandler.<clinit>(AndroidSdkHandler.java:81)
	at com.android.sdklib.tool.sdkmanager.SdkManagerCli.main(SdkManagerCli.java:73)
	at com.android.sdklib.tool.sdkmanager.SdkManagerCli.main(SdkManagerCli.java:48)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.annotation.XmlSchema
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	... 5 more
```

noomorph:
@cs01, as far as I can see, the errors on CI (at least) are in unit test layer, not sure why we are talking about Java errors now, though it's frustrating, of course, to have `scripts/ci.android.sh` not passing. I might look into it sometime later, but anyway, the problem is on `npm run unit` level in `Detox/detox` project.

noomorph:
@cs01, is there anything pending or shall I review again?

noomorph:
@cs01, sorry for keeping you waiting, I had many distractions last week (other tasks and three days off in a row), so I'll get to the review apparently only this week, hopefully on Monday or Tuesday.

cs01:
Merged changes from master and resolved conflicts. The unit tests are passing, but one e2e test is failing (`test/e2e/06.device.test.js`). Not sure if this is related to this PR or not.

noomorph:
@cs01, waiting for the CI to pass.

The summary of the change is: if `$ANDROID_SDK_ROOT` environment variable is not set (and `$ANDROID_HOME` as well), the following functions will try to search for the executable in `$PATH`:
* getAaptPath()
* getAdbPath()
* getAndroidEmulatorPath()

If they don't find the executable in $PATH, they will throw the same exception as usual.

As a side effect, the edge case situation where SDK root points to some invalid directory, is also handled via exception of format:

    There was no "adb" executable file in directory: /opt/sdk/platform-tools
    Check integrity of your Android SDK.

Chad, any suggestions from your side before we merge?

cs01:
I looked over the code and I see it searches the `PATH` for the various executables, so it accomplishes what was set out in the PR. I have no suggestions.

Issue #1572, opened 2019-07-31
Parallel testing on Android using Jest, phase #1


d4vidi:
Much like on iOS, we need support for running multiple test workers simultaneously on Android using Jest, each worker running on a preassigned, unique emulator instance.

Current challenge is in allowing for multiple emulator instances - all of the same type (e.g. `Pixel_2_API_28`). In essence, the SDK tools allow for a single emulator (instance) per AVD.

PR #1595, opened 2019-08-08
Android parallelism (WIP)

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1572 and the solution has been agreed upon with maintainers.

---

**Description:**

The first step towards officially introducing parallel testing on Android (multiple jest workers).
Reasons for this being beta:
- Uses a feature of the android emulator's binary which is still in beta (`-read-only`). It was introduced [~1 year ago](https://androidstudio.googleblog.com/2018/07/emulator-2800-canary.html), and looks stable - but many versions have been released since then and I couldn't test all of them.
- Some error-logs show when using 3 workers or more (though everything works fine).
- The whole feature hasn't been documented yet.
- Worker-device assignment reporter doesn't seem completely stable yet, and it hasn't been documented.

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -240,9 +240,13 @@ module.exports.handler = async function test(program) {
   }
 
   function runJest() {
-    if (platform === 'android' && program.workers !== 1) {
-      log.warn('Can not use -w, --workers. Parallel test execution is only supported on iOS currently');
-      program.w = program.workers = 1;
+    if (platform === 'android') {
+      program.readOnlyEmu = false;
+      if ((program.workers || 0) > 1) {
+        program.readOnlyEmu = true;
+        log.warn('Multiple workers is an experimental feature on Android and requires an emulator binary of version 28.0.1 or higher. ' +
+          'Check your version by running: $ANDROID_HOME/tools/bin/sdkmanager --list');
+      }
     }
 
     const jestReportSpecsArg = program['jest-report-specs'];
@@ -276,6 +280,7 @@ module.exports.handler = async function test(program) {
       'recordPerformance',
       'deviceName',
       'reportSpecs',
+      'readOnlyEmu',
     ]);
 
     log.info(printEnvironmentVariables(detoxEnvironmentVariables) + command);
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -147,7 +147,7 @@ const collectExtraArgs = require('./utils/collectExtraArgs')(module.exports.buil
 
 module.exports.handler = async function test(program) {
   program.artifactsLocation = buildDefaultArtifactsRootDirpath(program.configuration, program.artifactsLocation);
-  
+
   if(!program.keepLockFile){
     clearDeviceRegistryLockFile();
   }
@@ -320,9 +320,13 @@ module.exports.handler = async function test(program) {
   }
 
   function clearDeviceRegistryLockFile() {
-    const lockFilePath = environment.getDeviceLockFilePath();
-    fs.ensureFileSync(lockFilePath);
-    fs.writeFileSync(lockFilePath, '[]');
+    const lockFilePathIOS = environment.getDeviceLockFilePathIOS();
+    fs.ensureFileSync(lockFilePathIOS);
+    fs.writeFileSync(lockFilePathIOS, '[]');
+
+    const lockFilePathAndroid = environment.getDeviceLockFilePathAndroid();
+    fs.ensureFileSync(lockFilePathAndroid);
+    fs.writeFileSync(lockFilePathAndroid, '[]');
   }
 
   run();
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -240,20 +240,22 @@ module.exports.handler = async function test(program) {
   }
 
   function runJest() {
+    const hasMultipleWorkers = ((program.workers || 0) > 1);
     if (platform === 'android') {
       program.readOnlyEmu = false;
-      if ((program.workers || 0) > 1) {
+      if (hasMultipleWorkers) {
         program.readOnlyEmu = true;
         log.warn('Multiple workers is an experimental feature on Android and requires an emulator binary of version 28.0.1 or higher. ' +
           'Check your version by running: $ANDROID_HOME/tools/bin/sdkmanager --list');
       }
     }
+    program.reportWorkerAssign = hasMultipleWorkers;
 
     const jestReportSpecsArg = program['jest-report-specs'];
     if (!_.isUndefined(jestReportSpecsArg)) {
       program.reportSpecs = (jestReportSpecsArg.toString() === 'true');
     } else {
-      program.reportSpecs = (program.workers === 1);
+      program.reportSpecs = !hasMultipleWorkers;
     }
 
     const command = _.compact([
@@ -280,6 +282,7 @@ module.exports.handler = async function test(program) {
       'recordPerformance',
       'deviceName',
       'reportSpecs',
+      'reportWorkerAssign',
       'readOnlyEmu',
     ]);
 
--- a/detox/src/Detox.js
+++ b/detox/src/Detox.js
@@ -127,6 +127,10 @@ class Detox {
     });
   }
 
+  deviceName() {
+    return this.device.name();
+  }
+
   _logTestRunCheckpoint(event, { status, fullName }) {
     log.trace({ event, status }, `${status} test: ${JSON.stringify(fullName)}`);
   }
--- a/detox/src/DetoxExportWrapper.js
+++ b/detox/src/DetoxExportWrapper.js
@@ -40,6 +40,13 @@ class DetoxExportWrapper {
     }
   }
 
+  deviceName() {
+    if (this[_detox]) {
+      return this[_detox].deviceName();
+    }
+    return '';
+  }
+
   _definePassthroughMethod(name) {
     this[name] = (...args) => {
       if (this[_detox]) {
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -96,6 +96,10 @@ class Device {
     }
   }
 
+  name() {
+    return this.deviceDriver.name();
+  }
+
   async takeScreenshot(name) {
     if (!name) {
       throw new Error('Cannot take a screenshot with an empty name.');
--- a/detox/src/devices/DeviceRegistry.js
+++ b/detox/src/devices/DeviceRegistry.js
@@ -2,16 +2,15 @@ const _ = require('lodash');
 const fs = require('fs-extra');
 const plockfile = require('proper-lockfile');
 const retry = require('../utils/retry');
-const environment = require('../utils/environment');
-const DEVICE_LOCK_FILE_PATH = environment.getDeviceLockFilePath();
 
 const LOCK_RETRY_OPTIONS = {retries: 10000, interval: 5};
 
 class DeviceRegistry {
 
-  constructor({getDeviceIdsByType, createDevice}) {
+  constructor({getDeviceIdsByType, createDevice, lockfile}) {
     this.getDeviceIdsByType = getDeviceIdsByType;
     this.createDevice = createDevice;
+    this.lockfile = lockfile;
     this._createEmptyLockFileIfNeeded();
   }
 
@@ -39,11 +38,11 @@ class DeviceRegistry {
   }
 
   async _lock() {
-    await retry(LOCK_RETRY_OPTIONS, () => plockfile.lockSync(DEVICE_LOCK_FILE_PATH));
+    await retry(LOCK_RETRY_OPTIONS, () => plockfile.lockSync(this.lockfile));
   }
 
   async _unlock() {
-    await plockfile.unlockSync(DEVICE_LOCK_FILE_PATH);
+    await plockfile.unlockSync(this.lockfile);
   }
 
   clear() {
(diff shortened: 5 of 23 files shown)


detox/local-cli/test.js
```diff
@@ -240,16 +240,22 @@ module.exports.handler = async function test(program) {
...
+    const hasMultipleWorkers = ((program.workers || 0) > 1);
+    if (platform === 'android') {
+      program.readOnlyEmu = false;
+      if (hasMultipleWorkers) {
+        program.readOnlyEmu = true;
+        log.warn('Multiple workers is an experimental feature on Android and requires an emulator binary of version 28.0.1 or higher. ' +
+          'Check your version by running: $ANDROID_HOME/tools/bin/sdkmanager --list');
+      }
     }
+    program.reportWorkerAssign = hasMultipleWorkers;
```
noomorph: There is my gut feeling that something is wrong with making semi-anonymous `program` parameters which get handled elsewhere by very optional Jest-specific reporters. Looks implicit, distributed and not really unmanageable to me. 😕 

Are you sure this is the only way?

detox/local-cli/test.js
```diff
@@ -320,9 +328,13 @@ module.exports.handler = async function test(program) {
...
 
   function clearDeviceRegistryLockFile() {
-    const lockFilePath = environment.getDeviceLockFilePath();
-    fs.ensureFileSync(lockFilePath);
-    fs.writeFileSync(lockFilePath, '[]');
+    const lockFilePathIOS = environment.getDeviceLockFilePathIOS();
+    fs.ensureFileSync(lockFilePathIOS);
+    fs.writeFileSync(lockFilePathIOS, '[]');
+
+    const lockFilePathAndroid = environment.getDeviceLockFilePathAndroid();
```
noomorph: What was the motivation to create two separate lock files for each platform? Theoretically, we could use only one, so why did you discard that option — what are advantages and disadvantages?

detox/local-cli/test.js
```diff
@@ -240,16 +240,22 @@ module.exports.handler = async function test(program) {
   }
 
   function runJest() {
-    if (platform === 'android' && program.workers !== 1) {
-      log.warn('Can not use -w, --workers. Parallel test execution is only supported on iOS currently');
-      program.w = program.workers = 1;
+    const hasMultipleWorkers = ((program.workers || 0) > 1);
```
noomorph: Why `|| 0`? Was there something wrong with `type: number, default: 1` handling in yargs?

detox/src/Detox.js
```diff
@@ -127,6 +127,10 @@ class Detox {
     });
   }
 
+  deviceName() {
```
noomorph: Was there a specific urge to add a getter for the device name specifically to Detox object instance?

Usually we `export { detox, device, expect, ... }`, so what is the issue with simply accessing `device` and getting its name?

detox/src/DetoxExportWrapper.js
```diff
@@ -40,6 +40,13 @@ class DetoxExportWrapper {
     }
   }
 
+  deviceName() {
```
noomorph: See my comment above about `Detox.prototype.deviceName`.

detox/src/devices/Device.js
```diff
@@ -96,6 +96,10 @@ class Device {
     }
   }
 
+  name() {
```
noomorph: Why not `get name() {`?

detox/src/devices/DeviceRegistry.js
```diff
@@ -2,16 +2,15 @@ const _ = require('lodash');
...
 const retry = require('../utils/retry');
-const environment = require('../utils/environment');
-const DEVICE_LOCK_FILE_PATH = environment.getDeviceLockFilePath();
 
 const LOCK_RETRY_OPTIONS = {retries: 10000, interval: 5};
 
 class DeviceRegistry {
 
-  constructor({getDeviceIdsByType, createDevice}) {
+  constructor({getDeviceIdsByType, createDevice, lockfile}) {
```
noomorph: Everything here is fine per se, but see my comment at the beginning about what is *that good reason* to split lock files into 2, by platform?

detox/src/devices/android/ADB.js
```diff
@@ -13,8 +13,22 @@ class ADB {
   }
 
   async devices() {
-    const output = (await this.adbCmd('', 'devices', { verbosity: 'high' })).stdout;
-    return this.parseAdbDevicesConsoleOutput(output);
+    const parser = await this._adbDevices();
+    const devices = [];
+    for (let device = await parser.parseNext(); device; device = await parser.parseNext()) {
```
noomorph: Not sure, but isn't there something wrong with:
```
for (let device = await parser.parseNext(); device; device = await parser.parseNext())
```
?

You have two `parseNext()` methods in one row, so you will skip every 2nd device, no? read device, do something, read device, read device, do something -> 1, 3, 5, 7, ... ?

detox/src/devices/android/ADB.js
```diff
@@ -13,8 +13,22 @@ class ADB {
...
+      devices.push(device);
+    }
+    return devices;
+  }
+
+  async findDevice(matcher) {
+    const parser = await this._adbDevices();
+    for (let device = await parser.parseNext(); device; device = await parser.parseNext()) {
```
noomorph: The same here.

detox/src/devices/android/ADB.js
```diff
@@ -287,6 +269,59 @@ class ADB {
   }
 }
 
+class AdbDevicesParser {
```
noomorph: Why not adhere to the built-in `AsyncIterator` and [for await..of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)? On one hand, `for await of` is supported only by Node 10.x, but if we implement AdbParser as AsyncIterable right now, then we can easily rewrite your findDevice() and devices() the moment we jump onto node>=10.x

detox/src/devices/android/ADB.js
```diff
@@ -287,6 +269,59 @@ class ADB {
...
+    } else if (isGenymotion(adbName)) {
+      device = {type: 'genymotion', name: adbName, adbName};
+    } else {
+      device = {type: 'device', name: adbName, adbName};
+    }
+    return device;
+  }
+}
+
+function isEmulator(deviceAdbName) {
```
noomorph: Maybe turn them into static methods?

detox/src/devices/android/ADB.js
```diff
@@ -287,6 +269,59 @@ class ADB {
...
+  return _.includes(deviceAdbName, 'emulator-');
+}
+
+function isGenymotion(adbName) {
+  return (/^((1?\d?\d|25[0-5]|2[0-4]\d)(\.|:)){4}[0-9]{4}/.test(adbName));
+}
+
 function _reportTelnetPortResolutionError(input, devicesList, deviceAdbName, port) {
```
noomorph: Maybe make `_reportTelnetPortResolutionError` a member of `AdbDevicesParser`?

detox/src/devices/android/Emulator.js
```diff
@@ -38,7 +41,7 @@ class Emulator {
     }
 
     let childProcessOutput;
-    const tempLog = `./${emulatorName}.log`;
+    const tempLog = `./${emulatorName}-${options.port}.log`;
```
noomorph: If options.port is undefined, will it produce `./emulator-5555-undefined.log`?

detox/src/devices/drivers/AttachedAndroidDriver.js
```diff
@@ -7,12 +7,18 @@ class AttachedAndroidDriver extends AndroidDriver {
     super(config);
 
     this.emulator = new Emulator();
+    this._name = 'Unspecified Device';
+  }
+
+  name() {
```
noomorph: See my comment above about `get name()`.

detox/src/devices/drivers/EmulatorDriver.js
```diff
@@ -117,6 +102,55 @@ class EmulatorDriver extends AndroidDriver {
     await telnet.kill();
     await this.emitter.emit('shutdownDevice', { deviceId });
   }
+
+  async _fixEmulatorConfigIniSkinNameIfNeeded(avdName) {
+    const configFile = `${os.homedir()}/.android/avd/${avdName}.avd/config.ini`;
```
noomorph: Please extract getting `configFile` to an outer method, we have a pending pull request on this, I believe. That's where you can configure from where to take config.ini.

detox/src/devices/drivers/EmulatorDriver.js
```diff
@@ -117,6 +102,55 @@ class EmulatorDriver extends AndroidDriver {
...
+    const device = await this.adb.findDevice(async (candidate) => {
+      const isEmulator = candidate.type === 'emulator';
+      const isMatchingName = candidate.name === name;
+      if (!(isEmulator && isMatchingName)) {
+        return false;
+      }
+
+      // Note: though not entirely pure, this is an important optimization so as to avoid preparsing of all
+      // potentially fit emulators, which can be time consuming - mostly because of the telnet.
+      const isBusy = await this.deviceRegistry._getBusyDevices().includes(candidate.adbName);
```
noomorph: Why use a private method of deviceRegistry? Why not extend and change something else?

detox/src/devices/drivers/EmulatorDriver.js
```diff
@@ -117,6 +102,55 @@ class EmulatorDriver extends AndroidDriver {
...
+    if (device) {
+      devices.push(device.adbName);
+    }
+    return devices;
+  }
+
+  async _createDevice() {
+    const {min, max} = DetoxEmulatorsPortRange;
+    let port = Math.random() * (max - min) + min;
+    port = port & (~0 - 1);
```
noomorph: Maybe write your bit mask explicitly `0xXXXXX` ?

detox/src/utils/environment.js
```diff
@@ -53,6 +58,7 @@ module.exports = {
   getAndroidSDKPath,
```
noomorph: You could add getting config.ini folder somewhere here by the way..

detox/test/e2e/init.js
```diff
@@ -2,13 +2,18 @@ const detox = require('detox');
...
 
 jasmine.getEnv().addReporter(adapter);
 
 // This takes care of generating status logs on a per-spec basis. By default, jest only reports at file-level.
 // This is strictly optional.
 jasmine.getEnv().addReporter(specReporter);
 
+// This will post which device has assigned to run a suite, in a multiple-worker test session.
+// This is strictly optional.
+jasmine.getEnv().addReporter(assignReporter);
```
noomorph: Is it justified to add and add and add new and new and new microreporters? What do you think?

detox/src/devices/drivers/EmulatorDriver.js
```diff
@@ -1,104 +1,89 @@
...
-  async acquireFreeDevice(avdName) {
+  async _validateAvd(avdName) {
     const avds = await this.emulator.listAvds();
-
     if (!avds) {
-      const avdmanagerPath = path.join(Environment.getAndroidSDKPath(), 'tools', 'bin', 'avdmanager');
+      const avdmanagerPath = path.join(environment.getAndroidSDKPath(), 'tools', 'bin', 'avdmanager');
 
       throw new Error(`Could not find any configured Android Emulator. 
-      Try creating a device first, example: ${avdmanagerPath} create avd --force --name Nexus_5X_API_24 --abi x86 --package 'system-images;android-24;google_apis_playstore;x86' --device "Nexus 5X"
```
d4vidi: no `Pixel 2` target anymore, only `pixel`

detox/local-cli/test.js
d4vidi: Well the alternative is to just propagate a `multipleWorkers=true/false` an push the operative logic down to actual consumers (reporters in this case). I believe the logic should be separated. Nevertheless I agree we're on the verge of config-vars boom. Perhaps in this case it's safe to always report assignment (unconditionally), in case the reporter was registered.

In any case I admit contemplating the two approaches beforehand, couldn't conclude an absolute case to go with either.

detox/local-cli/test.js
d4vidi: I'm picturing an exotic use case where on a CI two detox instances run multiple testing for both iOS and Android separately... a single file would have them mutually exclude access to that file, though really they are not fighting over the same resources.

The overhead of having another file is scarce... why not, then? Only caveat is that if we ever allow for multi-platform runs using a single detox instance, it wouldn't work. But... yeah.

detox/src/Detox.js
d4vidi: 👍

detox/local-cli/test.js
noomorph: 1. Mmmmm, okay, but it would exclude access for less than a second, I believe — just at the moment, when some of them stringifies JSON array and writes it back to the file.
2. Okay, if you want to keep them separate, why do you clean both lock files regardless of what platform you do run on?? That confuses me, to be honest.

detox/local-cli/test.js
noomorph: > Perhaps in this case it's safe to always report assignment (unconditionally), in case the reporter was registered, since the output is minimal.

Agree.

detox/src/Detox.js
noomorph: Ahem, it seems that the code was left intact while you had resolved the conversation. Did you forget to push, maybe?

detox/src/devices/drivers/EmulatorDriver.js
noomorph: Not sure I have understood — you have changed the code to use `Pixel 2` target, no?

d4vidi:
Need to stabilize this on CI. Closing for the time being so that PR builds won't be triggered until then.

d4vidi:
I've changed the PR's title because it better conveys how I've since learned there's more serious work to be done in order to make this feature solid and well polished.
In a nutshell, once this lands in the status will be:
1. Parallelism will be largely possible, though still experimental.
2. It will not always be useable (e.g. with 3 or more devices), mostly because - as it turns out, the ADB local server appears to be a bottleneck, causing serious time penalties during emulators boot time. Most notable issue is that with a device that can properly *quick* boots, apk's installations takes forever to complete in some of the devices since they happen concurrently, and thus often one device is held back by installation on others. Typically this results in a timeout.
3. Last issue thus remains to be fixed using multiple ADB servers running concurrently, which is nontrivial but possible.

detox/test/e2e/init.js
d4vidi: Should we unify to a 'detox' reporter for simplicity?

PR #1597, opened 2019-08-08
feat(android): add platform-specific API for clearing package data

lokendrarawat97:
- [x] This is a small change 

---

**Description:**

added clearStorage option to clear app data for android

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -235,7 +235,7 @@ class Device {
   }
 
   async resetContentAndSettings() {
-    await this.deviceDriver.resetContentAndSettings(this._deviceId);
+    await this.deviceDriver.resetContentAndSettings(this._deviceId, this._bundleId);
   }
 
   getPlatform() {
--- a/detox/src/devices/android/ADB.js
+++ b/detox/src/devices/android/ADB.js
@@ -130,6 +130,10 @@ class ADB {
     await this.adbCmd(deviceId, `uninstall ${appId}`);
   }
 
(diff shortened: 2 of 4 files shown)


lokendrarawat97:
@noomorph ok will do it today.

detox/src/devices/Device.js
```diff
@@ -235,7 +235,7 @@ class Device {
   }
 
   async resetContentAndSettings() {
-    await this.deviceDriver.resetContentAndSettings(this._deviceId);
+    await this.deviceDriver.resetContentAndSettings(this._deviceId, this._bundleId);
```
noomorph: @LeoNatan, I now wonder whether `resetContentAndSettings` process can be made app-specific, not global for the whole device (simulator)?

LeoNatan: Not possible. On iOS there is no such functionality, and it is the same with simulators. We can delete the app, but that's not the same thing and doesn't reset the settings related to the app.

heroic:
@noomorph 
`adb shell recovery --wipe_data` is not recommended or even useful, as it reboots the emulator. In essence `shell pm clear ${packageId}` resets the app, not the phone. There isn't a way to do a light factory reset on android without a reboot.

So it would probably be better to use it as a new method.

noomorph:
@heroic, if you check the behavior on iOS, that's exactly what resetContentAndSettings does - wipes and reboots a device. I see no contradictions there.

As for clean app data, I agree and that is what I also said earlier, that it has to be added as a New method

PR #1598, opened 2019-08-08
Text input improvements

LeoNatan:
Added explicit errors on failure which are sent to Earl Grey for clarity on which stage of text input failed.

### DIFF
--- a/detox/ios/Detox/GREYActions+Detox.m
+++ b/detox/ios/Detox/GREYActions+Detox.m
@@ -78,71 +78,104 @@ static void _DTXTypeText(NSString* text)
 	}));
 }
 
-static BOOL _ensureFirstResponderIfNeeded(id expectedFirstResponderView, NSError* __strong * errorOrNil)
+static UIView* _isViewOrDescendantFirstResponder(UIView* view)
 {
-	BOOL isFirstResponder = [expectedFirstResponderView isFirstResponder];
+	UIView* currentFirstResponder = view.window.firstResponder;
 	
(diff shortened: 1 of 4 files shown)


LeoNatan:
@d4vidi It appears two tests failed here. The first is one that I added and is the scenario we discussed internally with @viktorijasujetaite 
```
<View testID='wrapper'><TextInput /></View>
...
await element(by.id('wrapper')).typeText("hello");
```

This now works on iOS, but it seems to fail on Android. I can make the test iOS only, but I wonder if we shouldn't have the same functionality on both platforms.

d4vidi:
I'm not entirely convinced that this is a use case we should support in
Detox. Before typing in text, we typically make sure that the view is
visible and that it is text-editable. If the test id is on a view then the
user's test doesn't make sense... Also, fixing it on the user's end is a
pretty easy thing to do.

LeoNatan:
The way I fixed it was by tapping on the user view. If the first responder after tap is a text editable, I take that rather than the view. It makes sense on iOS to have this for other scenarios as well, so I think I want to merge this anyway. The question is, would Android also benefit from such logic.

LeoNatan:
I want to land this because people are depending on this. I will make the failing test iOS only for now.

LeoNatan:
@d4vidi So Android is still broken. I added one element to that screen and somehow it broke the scroll?

Please take a look.

https://github.com/wix/Detox/pull/1598/files#diff-bed70c67a493bc438deb1689cadd32eeR68

Thanks

Issue #1200, opened 2019-03-11
Using in a physical android device 


jramalho:
**Describe the bug**
When i try to use it on a physical device, it boot the app but never automate a test.
This using the examples files.

**To Reproduce**

1. Download base project
2. detox build the base react-native project or react-native-jest projet
3. detox test

the issue happen on both react-native or react-native-jest demo project on the examples folder

**Expected behavior**
It should automate tests

**Screenshots**

**Environment (please complete the following information):**
 - Detox: 10.0.12
 - React Native: 0.51.1
 - Device: Moto e 2
 - OS: Linux Ubuntu

**Device and Verbose Detox Logs**
1) "before all" hook

  0 passing (2m)
  1 failing

  1) "before all" hook:
     Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  

  0 passing (2m)
  1 failing

  1) "before all" hook:
     Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

JsonDo:
Hello @jramalho ,
For me to be able to test on physical Android devices, I have to change some adb settings first.

With only one device connected through USB, I usually do the following:

1. `adb -d reverse tcp:8099 tcp:8099`
1. `detox run-server`
1. `detox test -c yourConfigHere`

In your configuration file, instead of giving detox the emulator name, give it its adb serial number, obtained by running `adb -d devices`.

Issue #1174, opened 2019-02-26
Detox error running tests on real Android device


quirimmo:
#### Description

I am trying to run some simple test on my real Android device, attached with USB to my machine. 

If I run the same tests through the emulator everything works fine, but with the real android device I get the following errors:

>  Example › should show the welcome screen
> 
>     Timeout - Async callback was not invoked within the 120000ms timeout specified by jest.setTimeout.
> 
>       at mapper (../node_modules/jest-jasmine2/build/queue_runner.js:41:52)
> 
>   ● Example › should show the welcome screen
> 
>     ReferenceError: device is not defined
> 
>        5 | describe('Example', () => {
>        6 |   beforeAll(async () => {
>     >  7 |     await device.launchApp({ delete: true, newInstance: true });
>          |     ^
>        8 |   });
>        9 |
>       10 |   it('should show the welcome screen', async () => {
> 
>       at Object.beforeAll (firstTest.spec.js:7:5)

This is my detox config:

```
"android.emu.debug-original-device-win": {
        "binaryPath": "android/app/build/outputs/apk/dev/debug/app-dev-debug.apk",
        "build": "pushd android && gradlew assembleDebug assembleAndroidTest -DtestBuildType=debug && popd",
        "type": "android.attached",
        "name": "73QDU17803001744"
      },
```

The app runs correctly on my device, but the tests do not start at all.

I've tried to run the tests using the `RCT_METRO_PORT=8099` env variable and doing the `adb reverse tcp:8099 tcp:8099`, but no luck. 

What is this issue related to? And how can I get it through? 

Thanks a lot

#### Steps to Reproduce
The official page related to Android works totally fine with emulator. 
When I try to run tests on my real Android device, I get the issue above. 

#### Detox, Node, Device, Xcode and macOS Versions
* Detox: 10.0.9
* React Native: 0.57.7
* Node: 8.11.1
* Device: Honor 8

#### Device and verbose Detox logs

> $ set RCT_METRO_PORT=8099&& detox test --configuration android.emu.debug-original-device-win -l verbose
> configuration="android.emu.debug-original-device-win" loglevel="verbose" artifactsLocation="artifacts\\android.emu.debug-original-device-win.2019-02-26 14-24-33Z" node_modules\.bin\jest "e2e" --config=e2e/config.json --maxWorkers=1 "--testNamePattern=^((?!:ios:).)*$"
> detox[13064] INFO:  [DetoxServer.js] server listening on localhost:61909...
[...]

Issue #968, opened 2018-10-03
Running Detox on a real Android device - supported? 


EdwardDrapkin:
I'm trying to run my tests on a real device or in a Genymotion cloud AWS instance.  They complete successfully on a local emulator.

I've tried with a local Android emulator as device type `android.emulator` and a local Genymotion emulator as `android.attached`; both work correctly.  When I try to run `android.attached` with a real device ID, things start to go wrong.  The application builds and installs successfully, and it even launches successfully as well.   But then nothing ever happens until the Jest tests timeout with errors about `device` and `element` not being defined.

Looking at the log files, it seems like the test can't connect to the Detox server but I'm not sure if I'm reading things correctly.

* Detox: 9.0.4
* React Native: 0.56
* Node: 10.5
* Device: Google Pixel / ZTE Z983 / Genymotion AWS 8.0 image
* Xcode: n/a
* macOS: High Sierra

Here's my configuration...

```json
        "android.wallet.device": {
                "binaryPath": "android/app/build/outputs/apk/flexwallet/release/app-flexwallet-release.apk",
                "build": "cd android && ./gradlew assembleFlexwalletRelease assembleAndroidTest -DtestBuildType=release && cd ..",
                "type": "android.attached",
                "name": "9802d885"
            },
```

...and I launch it like so....

```
detox test --configuration=android.wallet.device --runner-config=test/detox-jest.json --specs=e2e/ --artifacts-location=./.e2e/flexmobile --loglevel=trace --record-logs=all
```

... and I get these logs: [detox_pid_87613.log](https://github.com/wix/Detox/files/2443477/detox_pid_87613.log) and [test.log](https://github.com/wix/Detox/files/2443485/test.log)

EdwardDrapkin:
I read through the source code and it looks like running `detox run-server` and adding:

```
"session": {
            "sessionId": "manualsession",
            "server": "ws://192.168.41.137:8099" #output from detox run-server
        },```

to my Detox config that gets passed to init() was what I needed to do.  

Is this documented somewhere I missed?

frankgerhardt:
I'm interested in this as well. I'm surprised that this issue is new. For iOS #95 is the second oldest open issue.

EdwardDrapkin:
Upon further research, it seems that running `adb reverse tcp:8099 tcp:8099` and running the detox server on 8099 every time is the best solution.

PR #1601, opened 2019-08-12
fix(attached-android): reverse port forwarding, and screen unlocking

mnzaki:
- [x] This is a small change 

---

**Description:**
This PR addresses 2 things for testing on android devices:
- Automatic reverse port forwarding for `DetoxServer` to connect back to the host machine. Relevant issues: #1174 #1200 #968 
- Screen unlocking from `'Dozing'` state

### DIFF
--- a/detox/src/devices/drivers/AndroidDriver.js
+++ b/detox/src/devices/drivers/AndroidDriver.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const URL = require('url').URL;
 const _ = require('lodash');
 const { encodeBase64 } = require('../../utils/encoding');
 const log = require('../../utils/logger').child({ __filename });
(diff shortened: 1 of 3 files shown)


mnzaki:
The port forwarding thing was quite a bugger to figure out since the error is happening inside the detox app on the device. Perhaps it would be useful to have some sort of graphical feedback (notification, error message, etc) when the app faces an issue.

And for the `'Dozing'` state, I'm note entirely sure if this makes sense cross-device, I only tested on a Nokia 6.1 with Android 9.0, for which this was necessary.

detox/src/devices/drivers/AndroidDriver.js
```diff
@@ -199,11 +200,16 @@ class AndroidDriver extends DeviceDriverBase {
...
     const additionalLaunchArgs = this._prepareLaunchArgs({debug: false});
+    const serverPort = new URL(this.client.configuration.server).port;
+    this.adb.reverse(deviceId, serverPort)
     const testRunner = await this.adb.getInstrumentationRunner(deviceId, bundleId);
     const spawnFlags = [`-s`, `${deviceId}`, `shell`, `am`, `instrument`, `-w`, `-r`, ...launchArgs, ...additionalLaunchArgs, testRunner];
 
     this.instrumentationProcess = spawnAndLog(this.adb.adbBin, spawnFlags, { detached: false });
-    this.instrumentationProcess.childProcess.on('close', () => this._terminateInstrumentation());
+    this.instrumentationProcess.childProcess.on('close', () => {
+      this._terminateInstrumentation()
```
noomorph: Could you rewrite to that to `async`-`await` style, please?
async-await-style

noomorph:
Thank you for the pull request - it looks quite useful. I have one stylistic comment about async-await, and let's recheck what's up with CI build.

mnzaki:
I rewrote with async/await.

The test failure is still mysterious to me though, it fails to install the app for one of the tests :/

mnzaki:
Oh it actually passed. I think it was just probably an unrelated failure

noomorph:
@mnzaki, thank you.

PR #1605, opened 2019-08-13
Fix biometric matching timeout

brandons:
- [x ] This is a small change 

Attempting to match biometrics always times out. These functions call 'waitForBackground' which is unnecessary as far as I can tell. The app does not background after matching a face or finger and this causes the test to always time out.

I believe there is still an issue with biometrics where `waitForActive` will always time out when a Face ID or Touch ID dialog is opened upon app launch. However that may be a broader issue.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -142,22 +142,18 @@ class Device {
 
   async matchFace() {
     await this.deviceDriver.matchFace(this._deviceId);
-    await this.deviceDriver.waitForBackground();
   }
 
   async unmatchFace() {
     await this.deviceDriver.unmatchFace(this._deviceId);
-    await this.deviceDriver.waitForBackground();
   }
 
   async matchFinger() {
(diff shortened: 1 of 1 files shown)


LeoNatan:
Why was that even added in there?
(We haven't worked in this, this is all a community effort. I only added support for it in AppleSimUtils)

The app does not go into background mode, and it could be that it doesn't even go into inactive.

brandons:
The app goes inactive when the Face ID prompt appears, not after. Maybe the original intent was to waitForActive and not background?

LeoNatan:
Right, that makes sense.

LeoNatan:
Could you please test to see if it works with or without waitForActive?

brandons:
@LeoNatan Sorry for the delay. It is working with `waitForActive`. I have updated the PR.

brandons:
Anything else required on my part to get this merged?

LeoNatan:
Thanks!

PR #1612, opened 2019-08-19
Log reflection errors from looper introspection

rodrigos-facebook:
- [ x] This is a small change 

**Description:**
If you are trying to set-up Detox in your app and there is a proguard misconfiguration then this usage of reflection may fail and it will be silent. If this log were here, it would have been a huge clue that would have saved me from having to manually debug Detox to find this thrown exception.


noomorph:
@rodrigos-facebook , why not `Log.e` ?

rodrigos-facebook:
Makes sense, updated

LeoNatan:
@rotemmiz @d4vidi @noomorph  Looks like a small change. Can you pls review?

PR #1614, opened 2019-08-20
chore(npmignore): exclude extra files in npm package

Diokuz:
Just small extension of npmignore, to save npm traffic and npminstall time.


noomorph:
Thanks!

PR #1629, opened 2019-08-28
Make Expo setup guide a bit clearer

MarkPollmann:
- [x] This is a small change 

---

**Description:**
Make Detox setup with Expo a bit clearer.

### DIFF
--- a/docs/Guide.Expo.md
+++ b/docs/Guide.Expo.md
@@ -21,10 +21,10 @@
 
 - Download the Expo Client iOS App from [Expo.io/tools](https://expo.io/tools#client).
 - Unzip the iOS IPA and **rename the folder** to `Exponent.app`. It'll have a file icon but will still be a folder.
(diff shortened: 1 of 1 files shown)


rotemmiz:
I would argue that we should remove the expo page completely, as we do not maintain the expo support. Support can only come from Expo themselves. Having this page in Detox's docs may confuse people that support can even be handled by the Detox team, it can not (technically it has to be managed by the native app code owners)

LeoNatan:
@rotemmiz I've been thinking the same. Almost every issue opened here has been because people read the guide and think we provide support for Expo, when in reality, the bugs are usually stemming from changes in Expo itself doing stuff improperly, people misunderstanding how Expo works or Expo not supporting Detox in their tooling. I too think we should remove the page, and perhaps add a line in the top README that support for Expo falls squarely on Expo maintainers or the community. What do you think?

rotemmiz:
Let's do it, for the sake of clarity.

PR #1650, opened 2019-09-10
Corrects React Native version specifier in Mocking

alexanderwallin:
- [x] This is a small change 

---

**Description:**
The two scenarios for the different React Native version ranges misses 0.59. This corrects the latter range.


LeoNatan:
Thanks!

PR #1656, opened 2019-09-15
Upgrade Gradle to 5.1.1

d4vidi:
- [x] This is a small change 
- [ ] This change has been discussed in issue #0 and the solution has been agreed upon with maintainers.

---

**Description:**

Upgrade Gradle to `5.1.1` and the Android gradle-plugin to equivalent version `3.4.1` in all demo apps. Biggest motivation for doing this is the [HTTP retries](https://docs.gradle.org/5.0/release-notes.html#http-retries-during-dependency-resolution) feature.

Disclaimer: the feature may not be fully functional as described [in this report to gradle](https://github.com/gradle/gradle/issues/8264). I hope it'd be ok for us nevertheless.

### DIFF
--- a/detox/android/build.gradle
+++ b/detox/android/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext.isOfficialDetoxLib = true
-    ext.kotlinVersion = '1.3.0'
+    ext.kotlinVersion = '1.3.41'
     ext.detoxKotlinVerion = ext.kotlinVersion
     ext.dokkaVersion = '0.9.18'
     ext.buildToolsVersion = '28.0.3'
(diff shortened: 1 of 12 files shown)


d4vidi:
Build fails on CI because RN < 0.59.9 has [issues with new versions of gradle](https://github.com/facebook/react-native/releases/tag/v0.59.9). I've thus upgraded RN from `59.8` to `59.9` - things seem to be okay.

d4vidi:
With respect to these limitations, however, I've also wanted to make sure that the detox .aar built with gradle version >=5 will be properly supported by apps with older gradle.

I've tested it so in these two custom builds (branch), where I've upgraded Gradle only for the lib, and not the demo apps themselves:
* https://jenkins-oss.wixpress.com/view/Tree/job/detox-demo-projects-android-59-8-master/196/
* https://jenkins-oss.wixpress.com/view/Tree/job/detox-android-59-8-master/460/

(the 1st of the two is more significant seeing that it prebuilts the .aar and uses it as its detox dep artifact).

Issue #1659, opened 2019-09-18
[ios] performance issue: unnecessary slower wait time for all taps


doronpr:
**Describe the bug**
Detox uses touch indicator to visualize taps when the test runs.
The visualization creates an animation therefore can slow down the test and therefore there is an option to disable the indicator (`disableTouchIndicator`).
When the indicator disabled, the "ripple effect" animation is indeed not showing. However,  COSTouchVisualizer.removeInactiveFingerTips code is still running (https://github.com/conopsys/COSTouchVisualizer/blob/b00850925cd312430c0f716cde53cd6053f214d4/Classes/COSTouchVisualizerWindow.m#L194) this causes an unnecessary  wait time of **~100ms** on **EVERY** tap (measured using `debugSynchronization`). This can easily be avoided to speed up tests (see PR)

**To Reproduce**
- [x ] I have tested this issue on the latest Detox release and it still reproduces
1. Set disableTouchIndicator to true and enable debugSynchronization.
2. Run a simple ios test that uses tap.
3. Measure the time it takes for the `.tap()` promise to resolve.

**Expected behavior**
Logs should show no trace of touch indicator.

**Actual behaviour**
Interaction is waiting for `removeInactiveFingerTips`. This slows down the `.tap` promise resolve time.

**Environment (please complete the following information):**
 - Detox: 14.3.2
 - React Native: 0.59.10 
 - Node: 10.15.0
 - Device: Simulator
 - Xcode: 10.1
 - iOS:
 - macOS: 19.14.6

**Device and Verbose Detox Logs**
**BEFORE the fix:**
```
detox[86565] INFO:  [actions.js] Sync App State: undefined
detox[86565] INFO:  [actions.js] Sync Dispatch Queue: com.apple.main-thread
detox[86565] INFO:  [actions.js] Sync App State: undefined
detox[86565] INFO:  [actions.js] Sync Dispatch Queue: com.apple.main-thread
detox[86565] INFO:  [actions.js] Sync Timed: performSelector @selector(removeInactiveFingerTips) on DTXTouchVisualizerWindow
detox[86565] INFO:  [actions.js] Sync Dispatch Queue: com.apple.main-thread
detox[86565] INFO:  [actions.js] Sync Timed: performSelector @selector(removeInactiveFingerTips) on DTXTouchVisualizerWindow
detox[86565] INFO:  [actions.js] Sync Dispatch Queue: com.apple.main-thread
[...]

PR #1660, opened 2019-09-18
ios+performance: speed up tap wait time when disableTouchIndicator is on

doronpr:
refs https://github.com/wix/Detox/issues/1659

- [x] This is a small change 

---

**Description:**


d4vidi:
@LeoNatan failed on one flaky test, you can go ahead and merge.

LeoNatan:
> @LeoNatan failed on one flaky test. Putting the fix of it aside, you can go ahead and merge.

It seems to happen a lot. If it's always the same test, please disable it. If not, then we have problem with the Android stability.

@doronpr Thanks!

PR #1671, opened 2019-09-24
Update expect.js to include swipe() action on whileElement() method

AylwSt01:
Allow the `.swipe()` action to be used when using the `whileElement()` method on the `waitFor()` manual synchronization.

- [ ] This is a small change 

---

**Description:**

### DIFF
--- a/detox/src/android/expect.js
+++ b/detox/src/android/expect.js
@@ -195,6 +195,11 @@ class WaitForActionInteraction extends WaitForActionInteractionBase {
     this._prepare(new ScrollAmountStopAtEdgeAction(direction, amount));
     await this.execute();
   }
+  
+  async swipe(direction, speed = 'fast') {
+    this._prepare(new SwipeAction(direction, speed));
+    await this.execute();
+  }
 }
 
 class Element {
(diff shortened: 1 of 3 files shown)


LeoNatan:
Can you please elaborate why this is needed? I don't see a point of this for normal scroll views.

AylwSt01:
> Can you please elaborate why this is needed? I don't see a point of this for normal scroll views.

The `.swipe()` action happens faster than the scroll does. I don't know why `.whileElement()` is limited to just `.scroll()`. There could be many use cases for other actions to be used with `.whileElement()`.

For me, I'm using the .swipe() action to scroll to the bottom of a long Terms and Conditions screen, where an agreement button becomes enabled. So I can just swipe until that button is enabled. I tried the `.scrollTo(bottom)` however this method seemed far more efficient because it's so much quicker to repeat a "fast" swipe action that it is to leave the framework to scroll down to the bottom edge.

I saw another "issue" reported on this exact request so I thought I'd patch locally and submit a pull request for it. I'm aware it's incomplete thus far without tests etc, I'm finding time to sort that out.

LeoNatan:
Looks like `WaitForActionInteraction` is perhaps better suited to be an `Element`, so it can inherit all the normal actions. Out of scope for this PR. It looks good.

LeoNatan:
Can you please add a unit test for the new functionality you've added so that the CI passes? Thanks

AylwSt01:
will do, just finding time between day job priorities :)

PR #1674, opened 2019-09-24
chore: remove snapshot testing from the project

### LINKED ISSUE #1310: Remove unneeded tests
As discussed offline with @noomorph, let's remove useless tests, as well as the 100% coverage requirement. Better approach is needed, rather than silly tests that only test trivial things but give no confidence that the process they test is working or not.

noomorph:
Resolves #1310. 

- [x] This is a small change 
- [x] This change has been discussed in issue #1310 and the solution has been agreed upon with maintainers.

---

**Description:**

Removes snapshot testing paradigm from the project.


LeoNatan:
I see that everything failed. Is this with the latest master fixes?

PR #1676, opened 2019-09-25
feat: --device-launch-args passthrough parameter

### LINKED ISSUE #1669: Support emulator custom-args passthrough
Can't pass in custom arguments to emulators dispatched by Detox. For example, I would like to specify in command line a specific proxy url or a DNS server.

Pass such arguments in via CLI, or specify them in detox' config in `package.json`.

None

**Additional context**
N/A

---

@swabbass  @silyevsk

d4vidi:
Resolves #1669.

- [x] This is a small change 
- [x] This change has been discussed in issue #1669 and the solution has been agreed upon with maintainers.

---

**Description:**

Allows users to provide custom args in CLI that would be passed-through onto the device (Android emulator / iOS simulator) when (if) launched by Detox.

@swabbass @silyevsk

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -243,6 +243,7 @@ module.exports.handler = async function test(program) {
     const detoxEnvironmentVariables = _.pick(program, [
       'deviceLaunchArgs',
     ]);
+
     launchTestRunner(command, detoxEnvironmentVariables);
   }
 
@@ -291,6 +292,7 @@ module.exports.handler = async function test(program) {
       'readOnlyEmu',
       'deviceLaunchArgs',
     ]);
+
     launchTestRunner(command, detoxEnvironmentVariables);
   }
 
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -140,6 +140,10 @@ module.exports.builder = {
     alias: 'device-name',
     group: 'Configuration:',
     describe: 'Override the device name specified in a configuration. Useful for running a single build configuration on multiple devices.'
+  },
+  'device-launch-args': {
+    group: 'Execution:',
+    describe: 'Custom arguments to pass (through) onto the device (emulator/simulator) binary when launched.'
   }
 };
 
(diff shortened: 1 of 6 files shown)


detox/local-cli/test.js
```diff
@@ -235,8 +239,19 @@ module.exports.handler = async function test(program) {
       ...getPassthroughArguments(),
     ]).join(' ');
 
-    log.info(command);
-    cp.execSync(command, { stdio: 'inherit' });
+
```
noomorph: You could extract the repetitive code to `exec(command, detoxEnvironmentVariables)`.

```
log.info(printEnvironmentVariables(detoxEnvironmentVariables) + command);
cp.execSync(command, {
      stdio: 'inherit',
         env: {
           ...process.env,
           ...detoxEnvironmentVariables
         }
       });
```

detox/src/devices/ios/AppleSimUtils.js
```diff
@@ -70,8 +71,9 @@ class AppleSimUtils {
     const isBooted = await this.isBooted(udid);
 
     if (!isBooted) {
+      const userLaunchArgs = (argparse.getArgValue('deviceLaunchArgs') || '');
```
noomorph: Why not call the variable `deviceLaunchArgs`, as is?

d4vidi: well then I suppose the outcome would be `boot UDID undefined`, wouldn't it?

noomorph: but I meant rename in two places, no?

detox/src/devices/ios/AppleSimUtils.js
```diff
@@ -70,8 +71,9 @@ class AppleSimUtils {
     const isBooted = await this.isBooted(udid);
```
noomorph: @d4vidi , you should probably change the signature of `boot(udid)`, and add to it `deviceLaunchArgs = ''` as a second argument.

AppleSimUtils.js is meant to be stateless, not depending on environment variables.

detox/src/devices/android/Emulator.js
```diff
@@ -24,6 +24,7 @@ class Emulator {
   }
 
   async boot(emulatorName, options = {port: undefined}) {
+    const userLaunchArgs = (argparse.getArgValue('deviceLaunchArgs') || '').split(/\s+/);
```
noomorph: Most likely, you should extend `options` with `userLaunchArgs` property too, similar to AppleSimUtils.js. And then you would get args from the command line/env var on the level of DeviceDriver.

noomorph: It is a suggestion though, not a mandatory thing to do.

Issue #1103, opened 2018-12-30
Incorrect OS version of specified iPhone simulator name


Wo1v3r:
https://github.com/wix/Detox/blob/45c895aa6abb02823f325f0b274ac63e12b05760/detox/src/devices/ios/AppleSimUtils.js#L38

E.G;

- delete iPhone 8 iOS 12.1 simulators
- pass only device name; `iPhone 8`

When other devices have 12.1 simulators installed for them, 12.1 will be picked even though device won't match;

```
 Error: Can't find a simulator to match with "iPhone 8", run 'xcrun simctl list' to list your supported devices.
      It is advised to only state a device type, and not to state iOS version, e.g. "iPhone 7"
    at AppleSimUtils.findDevicesUDID (/Users/jonathanl/Desktop/dev/crm/invoices-mobile/node_modules/detox/src/devices/ios/AppleSimUtils.js:47:13)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

LeoNatan:
The logic should be fixed to search the latest available OS for the specified simulator name, not latest available OS version in general.

youfoundron:
Affecting me as well.

noomorph:
Related to https://github.com/wix/Detox/pull/905 .

noomorph:
Indirectly related to https://github.com/wix/Detox/issues/853 .

youfoundron:
> Affecting me as well. @Wo1v3r what was your workaround?

Having other issues now, but updating to latest version of `detox@^12.1.1` fixed this.

matt-oakes:
For anyone coming through this through searching for the error message, you might want to upgrade you `applesimutils` to at least version 0.7.0 using:

`brew upgrade applesimutils`

My issue was that I was on version 0.6.3, which was not able to return a list of any simulators. You can test this by running the command:

`applesimutils --list`

If you get an empty array, you might fix it by upgrading :+1:

mattijsf:
I have the same problem. I also tried upgrading `applesimutils` but the problem remains.


```js
"configurations": {
  "ios.sim.debug": {
  	...
    "type": "ios.simulator",
    "name": "iPhone SE"
  },
  ...
 }
```

```bash
$ brew upgrade applesimutils                                                                                                                                                                                                                     
Error: wix/brew/applesimutils 0.7.0 already installed

$ detox test -c ios.sim.debug
[...]

PR #1694, opened 2019-10-08
refactor: device selection revisited

### LINKED ISSUE #1565: TypeError: Cannot read property 'identifier' of undefined
**Description**
I have a problem from this lib during running detox tests, simulators for react native works okay. 

**Screenshots**
detox[17229] ERROR: [DetoxExportWrapper.js/DETOX_INIT_ERROR] 
 TypeError: Cannot read property 'identifier' of undefined
    at AppleSimUtils.create (/.../node_modules/detox/src/devices/ios/AppleSimUtils.js: 100:100)

**Environment**
 - macOS version: 10.14.5
 - Xcode version: 10.3

Before Xcode update I didn't have this issue

noomorph:
- [x] This change has been discussed in issues #853, #1103,  #1565 and pull requests #905, #1404 .

This pull request prepares the ground to finally deal with #853 and also should potentially resolve #1565. 

The consequent PR will add `device` property to a configuration, which can be configured in a more powerful way.

**Changes:**

1. **AppleSimUtils**: make it thinner and more explicit, therefore the search method is called now `list` (similar to the CLI). The query handler logic (`iPhone X, iOS 12.0`) is moved 1 layer upwards, to SimulatorDriver.

2. **SimulatorDriver**: logic of search remains intact - device type (+ device os, if specified via comma).

3. **AppleSimUtils**: remove the dependency on a heavy and non-maintainable direct call `xcrun simctl list -j`, by allowing `new AppleSimUtils().create(device)` to accept a "prototype" for a device you want to create. I pick it from the result the `list()` method returns as-is, and pass it to `create()`.

4. **SimulatorDriver**: remove redundant calls to applesimutils, querying the same things twice when creating a device. **Important:** *now, when creating a device, it takes an OS version of the first matching but a busy device.*

5. **AndroidDriver**: remove redundant Telnet queries to busy devices. That can happen, because getting a name now done explicitly, via `queryName()` method, so we can postpone it till the very last moment.

6. **ADB**: recompose an async-iterator-like ADB output parser to lazy DeviceHandle classes, to allow bullet (5) to happen.

7. **DeviceRegistry**: refactor to accommodate changes to `SimulatorDriver` and `AndroidDriver`. API was simplified to just `allocateDevice` and `disposeDevice`. This way, we remove 2-way connecting between *Driver -> DeviceRegistry -> injected methods of *Driver, therefore we have better maintainability.

TODO(@noomorph): conduct some more manual QA.

### DIFF
--- a/detox/src/devices/drivers/SimulatorDriver.js
+++ b/detox/src/devices/drivers/SimulatorDriver.js
@@ -1,11 +1,12 @@
-const exec = require('child-process-promise').exec;
-const path = require('path');
-const fs = require('fs');
 const _ = require('lodash');
+const fs = require('fs');
+const path = require('path');
+const exec = require('child-process-promise').exec;
+const DeviceRegistry = require('../DeviceRegistry');
 const IosDriver = require('./IosDriver');
 const configuration = require('../../configuration');
+const DetoxRuntimeError = require('../../errors/DetoxRuntimeError');
 const environment = require('../../utils/environment');
-const DeviceRegistry = require('../DeviceRegistry');
 const argparse = require('../../utils/argparse');
 
 class SimulatorDriver extends IosDriver {
@@ -14,9 +15,7 @@ class SimulatorDriver extends IosDriver {
     super(config);
 
     this.deviceRegistry = new DeviceRegistry({
-      getDeviceIdsByType: async type => await this.applesimutils.findDevicesUDID(type),
-      createDevice: type => this.applesimutils.create(type),
-      lockfile: environment.getDeviceLockFilePathIOS(),
+      lockfilePath: environment.getDeviceLockFilePathIOS(),
     });
 
     this._name = 'Unspecified Simulator';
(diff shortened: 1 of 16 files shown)


LeoNatan:
> If it is a complex query like iPhone X, iOS 12.0 (contains comma-separated values), it immediately searches by device type and OS.

I don't think this is correct. There can be simulators of the same name for different OS versions. We need to be able to specify a more complex query, rather than just a string with commas.

LeoNatan:
We discussed in the past the ability to provide an object as device selector.
```
{
  name: "iPhone X",
  os: "iOS 13.0"
}
```
vs
```
{
  type: "iPhone X",
  os: "iOS 13.0"
}
```

detox/src/devices/drivers/SimulatorDriver.js
```diff
@@ -157,6 +155,73 @@ class SimulatorDriver extends IosDriver {
...
+    }
+
+    if (!udid) {
+      throw new Error(`Failed to find device matching "${deviceQuery}"`);
+    }
+
+    return udid;
+  }
+
+  async _groupDevicesByStatus(query) {
```
LeoNatan: Query should be an object rather than string. If we are making such important changes, let's do it fully. Otherwise, there is ambiguity that cannot be resolved with this query. Devices with the same name are allowed to exist, If I have two devices named "MySim", one for iOS 13 and one for iOS 12, the current proposed solution will fail for a `"MySim, iOS 13.0"` query.

LeoNatan: Perhaps we can have a fallback like this for legacy strings, but we should add the full solution. I don't think it's a hard solution, as the object can travel, just like the string, from the package.json to here.

noomorph: I decided to split the refactor and the change to the feature, on Friday it will appear in a consequent pull request.

Issue #1371, opened 2019-05-07
Implement facebook/idb tool (fbsimctl replacement)


ayelenmarie:
Talking with LeoNatan in #629 I mentioned Facebook released a new tool that replaces\includes `fbsimctl` that is called `idb`. They will start monitoring and updating that repo instead of the other one and they are suggesting everyone to move to this new tool instead. So it will be great if Detox could migrate as well!

Currently it's **under construction** but as it supposedly fixes some issues that `fbsimctl` has, I might as well suggest it here.

Here is the repo for `idb` : [https://github.com/facebook/idb/issues/512](url)

Thanks!

lawrencelomax:
Hello, one of the authors of the `idb` project here!

More than happy to give pointers/advice if you need them.

As far as the maturity of the project goes, we've been using this in production for nearly 2 years now and it's only just now that we've open sourced it.

rotemmiz:
Hey @lawrencelomax !!

Good to have you here, actually I do have a few questions:
1. One of the reason that made us ditch fbsimctl (aside from the multi simulator support introduced in Xcode9) was the fact that there were no stable releases, and the only way to install it was through brew at the repo's HEAD. I see that idb is being released regularly, and brew uses the latest version, which is great!
2. I did not dive into the code, but couldn't find anything in docs regarding ability to pass launch args and environment variables to app on device, Is that possible?
3. We never really considered running Detox on iOS devices for more than a POC, but if `idb` can get us closer, this may change they way we approach it. here's the latest status written by @LeoNatan https://github.com/wix/Detox/issues/95#issuecomment-483348724.

lawrencelomax:
> One of the reason that made us ditch fbsimctl (aside from the multi simulator support introduced in Xcode9) was the fact that there were no stable releases, and the only way to install it was through brew at the repo's HEAD. I see that idb is being released regularly, and brew uses the latest version, which is great!

I hear you on this as know that this has been a problem in the past. The biggest reason for this is that `fbsimctl` was very much in maintenance mode whilst we were working on `idb` itself. We're more than happy to keep changelogs and make regular releases from master.

> I did not dive into the code, but couldn't find anything in docs regarding ability to pass launch args and environment variables to app on device, Is that possible?

[...]

PR #1697, opened 2019-10-14
feat(ios): add method device.clearKeychain()

santiagofm:
- [x] This is a small change 
- [x] This change has been discussed in issue https://github.com/wix/Detox/issues/1371

---

**Description:**

Adds the ability to delete the iOS Keychain by adding `clearKeychain` to `Device`.

I opted to not add a new e2e test as a dependency would need to be added, but will if you guys think it's needed.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -210,6 +210,10 @@ class Device {
     await this.deviceDriver.setLocation(this._deviceId, lat, lon);
   }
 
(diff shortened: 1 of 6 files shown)


LeoNatan:
Thanks
I will not ask for an e2e test here because of the terrible Keychain API.
But please make sure the unit tests pass (there is a problem with coverage right now).

Thanks

PR #1705, opened 2019-10-19
More helpful error message if Detox.init() fails

noomorph:
I am very tired of user reports that tell `device (or element) is not defined` and so on.
Let's solve it.

The goal of this pull request to transform those errors:

Old errors

To these, telling that there was an error in detox.init():

New errors

### DIFF
--- a/detox/runners/jest/WorkerAssignReporterImpl.js
+++ b/detox/runners/jest/WorkerAssignReporterImpl.js
@@ -1,3 +1,4 @@
+const _ = require('lodash');
 const chalk = require('chalk').default;
 const ReporterBase = require('./ReporterBase');
 const log = require('../../src/utils/logger').child();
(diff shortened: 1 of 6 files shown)


detox/runners/jest/WorkerAssignReporterImpl.js
```diff
@@ -8,8 +9,10 @@ class WorkerAssignReporterImpl extends ReporterBase {
   }
 
   report(workerName) {
-    this._traceln(`${chalk.whiteBright(workerName)} assigned to ${chalk.blueBright(this.device.name)}`);
-    this._traceln('');
+    const deviceName = _.attempt(() => this.device.name);
+    if (!_.isError(deviceName)) {
+      this._traceln(`${chalk.whiteBright(workerName)} assigned to ${chalk.blueBright(this.device.name)}\n`);
+    }
```
d4vidi: @noomorph I'd actually want to keep this log even if the device name isn't available because it helps with bug hunting in this area

d4vidi:
Love it!

rotemmiz:
I was looking for this one, remembered you showed it to me but couldn't find it. It's amazing!

Issue #1718, opened 2019-10-29
Example not working. Just with adjustment


Nestoro:
In the example you write

```
await device.reloadReactNative();
await expect(element(by.id('email'))).toBeVisible();
```

but that doesn't seem to work 100% of the cases. i didn't inspect your structure further but it seems that every element() call should be awaited.

so i would like to change that to:

```
await device.reloadReactNative();
await expect(await element(by.id('email'))).toBeVisible();
```

or:

```
await device.reloadReactNative();
var element = await element(by.id('email'));
await expect(element).toBeVisible();
```

This could reduce problems new users encounter since this is a pretty easy to oversee thing if you try to get detox to run for the first time.

LeoNatan:
Pull requests are welcome.

LeoNatan:
The await is not necessary for the internal element. The entire call is serialized to the native iOS/Android test runner.

PR #1719, opened 2019-10-30
Fix examples. Issue (#1718)

Nestoro:
- [x] This is a small change 
- [~] This change has been discussed in issue #1718 and the solution has been agreed upon with maintainers.

---

In this request i added await to all `expect(element.by` calls. I did this since it shouldn't hurt in any case and is necessary in some.

In addition you could remove line 105 in docs/Troubleshooting.RunningTests.md because the case addressed in that line should not be possible with the updated syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,14 +22,14 @@ describe('Login flow', () => {
     
   it('should login successfully', async () => {
     await device.reloadReactNative();
-    await expect(element(by.id('email'))).toBeVisible();
+    await expect(await element(by.id('email'))).toBeVisible();
       
     await element(by.id('email')).typeText('john@example.com');
     await element(by.id('password')).typeText('123456');
     await element(by.text('Login')).tap();
       
-    await expect(element(by.text('Welcome'))).toBeVisible();
-    await expect(element(by.id('email'))).toNotExist();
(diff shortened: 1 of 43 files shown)


LeoNatan:
Hello,
These changes are invalid and are not necessary for the function of the system. You are welcome to debug the system and see for yourself, that the entire call is serialized to the native sides of Detox, thus making the await calls redundant.

Thank you for your contribution, but we will not accept this PR.

Nestoro:
I can't agree
the only problem my changes kept encountering where one test-case that literally threw this stacktrace:

`13:03:45         "Description":  "Cannot scroll, the scrollview is already at the edge.",
`
what indicates a broken test case in general not broken syntax....

i am not disagreeing with the fact that the entire call should be serialized. i am disagreeing with the assumption that this is always the case. this 'redundancy' is a crucial fix for an implementation of detox i had to work with because the expect function kept receiving a unresolved Promise (element(by.id call).
this await doesn't hurt the existing code in anyway and is just an addition for the case where it hurts to not have it included. but if you continue to disagree just keep this closed.

Best Regards!

LeoNatan:
But it literally has no effect. I'm not sure how that would solve your issue.

Issue #1524, opened 2019-07-11
Change visibility percentage in toBeVisible()


sergiomap:
I would like to be able to change the visibility percentage of an element.

I know that toBeVisible() will return true if the visibility is 75% or greater but in some cases I need that the element be 100% visible, otherwise I cannot tap the element.

rotemmiz:
You can potentially expose https://github.com/wix/Detox/blob/master/detox/src/ios/earlgreyapi/GREYMatchers.js#L277 as a new API call.

rotemmiz:
And on Android you'll need to add it as a param (with 75 as default)
https://github.com/wix/Detox/blob/f63424ccbcb0224f51674c5c0e3c7bc73122d721/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java#L75

LeoNatan:
In a future release to come soon, visibility will no longer be calculated by percentage.

PR #1720, opened 2019-10-31
Allow expecting visibility with specified percentage

maxkomarychev:
- [x] This is a small change 
- [x] This change has been discussed in issue #1524 and the solution has been agreed upon with maintainers.

---

**Description:**

This PR changes signature of `toBeVisible` allowing to match visibility by any arbitrary %

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
@@ -2,6 +2,8 @@ package com.wix.detox.espresso;
 
 import android.view.View;
 
+import java.lang.Math;
+
 import org.hamcrest.Matcher;
 
 import androidx.test.espresso.matcher.ViewMatchers.Visibility;
@@ -73,6 +75,10 @@ public class DetoxMatcher {
 
     public static Matcher<View> matcherForSufficientlyVisible() {
         return isDisplayingAtLeast(75);
+    }
+
+    public static Matcher<View> matcherForMinimumVisiblePercent(double percentage) {
+        return isDisplayingAtLeast((int)percentage * 100);
     }
 
     public static Matcher<View> matcherForNotVisible() {
--- a/detox/src/ios/expect.js
+++ b/detox/src/ios/expect.js
@@ -359,8 +359,8 @@ class ExpectElement extends Expect {
     super(invocationManager);
     this._element = element;
   }
-  async toBeVisible() {
-    return await new MatcherAssertionInteraction(this._invocationManager, this._element, new VisibleMatcher()).execute();
+  async toBeVisible(percentage) {
+    return await new MatcherAssertionInteraction(this._invocationManager, this._element, new VisibleMatcher(percentage)).execute();
   }
   async toBeNotVisible() {
     return await new MatcherAssertionInteraction(this._invocationManager, this._element, new NotVisibleMatcher()).execute();
@@ -397,8 +397,8 @@ class WaitForElement extends WaitFor {
     //if ((!element instanceof Element)) throw new Error(`WaitForElement ctor argument must be a valid Element, got ${typeof element}`);
     this._element = element;
   }
-  toBeVisible() {
-    return new WaitForInteraction(this._invocationManager, this._element, new VisibleMatcher());
+  toBeVisible(percentage) {
+    return new WaitForInteraction(this._invocationManager, this._element, new VisibleMatcher(percentage));
   }
   toBeNotVisible() {
     return new WaitForInteraction(this._invocationManager, this._element, new VisibleMatcher())._not();
--- a/detox/src/android/espressoapi/DetoxMatcher.js
+++ b/detox/src/android/espressoapi/DetoxMatcher.js
@@ -155,6 +155,21 @@ class DetoxMatcher {
     };
   }
 
+  static matcherForMinimumVisiblePercent(percentage) {
+    if (typeof percentage !== "number") throw new Error("percentage should be a number, but got " + (percentage + (" (" + (typeof percentage + ")"))));
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.espresso.DetoxMatcher"
+      },
+      method: "matcherForMinimumVisiblePercent",
+      args: [{
+        type: "Double",
+        value: percentage
+      }]
+    };
+  }
+
   static matcherForNotVisible() {
     return {
       target: {
--- a/detox/src/android/expect.js
+++ b/detox/src/android/expect.js
@@ -271,8 +271,8 @@ class ExpectElement extends Expect {
     super(invocationManager);
     this._element = element;
   }
-  async toBeVisible() {
-    return await new MatcherAssertionInteraction(this._invocationManager, this._element, new VisibleMatcher()).execute();
+  async toBeVisible(percentage) {
+    return await new MatcherAssertionInteraction(this._invocationManager, this._element, new VisibleMatcher(percentage)).execute();
   }
   async toBeNotVisible() {
     return await this._invocationManager.execute(DetoxAssertionApi.assertNotVisible(call(this._element._call)));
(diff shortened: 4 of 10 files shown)


maxkomarychev:
hi, I am slightly confused with failed tests, can anyone explain what happened? did I edit files which are supposed to be generated?

noomorph:
@d4vidi, could you take a look at this PR?

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -75,6 +77,10 @@ private DetoxMatcher() {
         return isDisplayingAtLeast(75);
     }
 
+    public static Matcher<View> matcherForMinimumVisiblePercent(double percentage) {
+        return isDisplayingAtLeast((int)Math.floor(percentage * 100));
```
d4vidi: I believe `Math.floor` isn't really needed here as the int-cast has the same effect.

d4vidi:
'restarted the CI job as we're now past at least some of the recent flakiness (I hope). Waiting for its  feedback in order to decide how to proceed.

detox/src/ios/expect.js
```diff
@@ -397,8 +397,8 @@ class WaitForElement extends WaitFor {
     //if ((!element instanceof Element)) throw new Error(`WaitForElement ctor argument must be a valid Element, got ${typeof element}`);
     this._element = element;
   }
-  toBeVisible() {
-    return new WaitForInteraction(this._invocationManager, this._element, new VisibleMatcher());
+  toBeVisible(percentage) {
```
d4vidi: I think the default (i.e. `.75`) should be set here rather than in the platform-specific impl., such that it would by-definition be identical across all platforms. @noomorph wdyt?

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -75,6 +77,10 @@ private DetoxMatcher() {
         return isDisplayingAtLeast(75);
     }
 
+    public static Matcher<View> matcherForMinimumVisiblePercent(double percentage) {
```
d4vidi: I really think we should stick with `matcherForSufficientlyVisible()` in this context, so as to stay coherent with iOS (EarlGrey). In any case, no need for both methods...

d4vidi:
@maxkomarychev could you please rebase over the Detox/master so we could rerun with RN .60 support?

detox/src/ios/expect.js
noomorph: Hmm, I'm for adding a new matcher, first of all, `toBeVisibleBy` or something like that.

maxkomarychev:
what about introducing new matcher `toBeVisibleBy` mentioned in the other comment? is that still relevant? @d4vidi @noomorph

maxkomarychev:
I rebased. Please let me know final decision of how you want this to be shaped (e.g. new matcher, where to really apply default 0.75). Thanks.

noomorph:
@d4vidi , please voice your opinion. I am not sure I am right by demanding a new matcher `toBeVisibleBy`, but, on the other hand, `toBeVisibleBy(0.75)` feels more intelligible than `toBeVisible(0.75)`.

d4vidi:
There are 2 things that I find unsettling here:
1. @noomorph if I try to think about whether the API is going to be future-proof, neither a param nor a separate method really solve it well. How would you find a more extensible `toBeVisible({threshold: 0.75})` as a solution?
2. Is such a generic solution needed here, really? I think we could slightly be over designing here, since - as @maxkomarychev has explained, he/she either needs an element to be _visible_ or _completely_ visible. Possibly, we could settle for a `toBeFullVisible()` method for the task at hand.

maxkomarychev:
my 2 cents on both items:

- majority of Detox's methods take positional params (`scroll`, `swipe` etc) and not dictionaries (except for init-like methods and some tap-related functionality)
- some methods provide sane defaults (`scroll`)
- I can't really imagine more complex scenarios for visibility matcher and even then there is a room for extension either by adding more positional parameters or switching to dictionary parameter if it grows really big.

That said and also considering there is no unified way of handling parameters in Detox it feels like an overengineering at this stage. I don't mind refactoring this to separate method and/or moving a place where default value 0.75 is set along the stack though.

> Is such a generic solution needed here, really? I think we could slightly be over designing here, since - as @maxkomarychev has explained, he/she either needs an element to be visible or completely visible. Possibly, we could settle for a toBeFullVisible() method for the task at hand.

I never said I need to assert full visibility, I'm not an author of https://github.com/wix/Detox/issues/1524 ;).

Quite contrary my original case demands <75% visibility which is enough for my tests.

In addition it feels weird that today `toBeVisible` is `>=75%` while `toBeNotVisible` is actually `==0%`. With this PR Detox will allow matching full scope of visibility.

Thanks.

In fact I find it rather weird allowing assertions for 75% without ability to change it to one needs.
In addition

d4vidi:
@maxkomarychev duly noted, thanks! @noomorph could we compromise on something like the `toBeVisibleBy()` as you suggested only that it would accept an enumeration of say `INVISIBLE` = 0, `PARTLY_VISIBLE`=0.5, `VISIBLE` = 0.75 and `COMPLETELY_VISIBLE` = 1 (give or take, you get the idea). Perhaps `toBeVisibleAs()` would be more appropriate?

noomorph:
@d4vidi , I like the latter option, `toHaveVisibility(...)`. As for enumerations, well, maybe, but it creates another confusion: does it require it to be visible at least by %%, or it should be visible by %% approximately. When we assert that an element has to be PARTLY_VISIBILE, does it have to be PARTLY_VISIBLE indeed? Does the full visibility actually break the assertion? That's the question. 🤔

viktorijasujetaite:
@d4vidi on *Android*, I am experiencing interactions with elements that are visible less than 75%. Detox scrolls to element. Element is being covered huge part by floating action button, but it's found as visible. Detox tries to tap element expected to be visible, but it interacts with FAB instead. 
Behavior expected here is scrolling screen more, until element is not covered anymore.

maxkomarychev:
why do we need any enums for an expectation against a continuous value? imo EarlGrey's default value of `0.75` visibility is strange to begin with, why not 60, 80 or 90? at the same time why `PARTLY` is `0.5` and not `0.25` or `0.123141`?

would it help if we introduce new method named `toBeVisibleAtLeast` which should clear ambiguity - the element should be visible at least X percent.

maxkomarychev:
Added complementary pr to for of early grey to include X in assertion for min visibility on ios https://github.com/wix-playground/EarlGrey/pull/2/files

d4vidi:
@viktorijasujetaite if the problem is the tap action being performed at the wrong place (i.e. in coordinates where a view is party hidden), [`tapAtPoint()`](https://github.com/wix/Detox/blob/master/docs/APIRef.ActionsOnElement.md#tapatpoint) can be used.

d4vidi:
@maxkomarychev @noomorph the enumeration both makes the API more self-explanatory and would keep users from making mistakes. I am, however, not strongly opinionated on that.

maxkomarychev:
@d4vidi I am not against constants with sane defaults in general. But my primary goal for doing this PR is to unleash api that allow expectations with arbitrary values.

one thing about your suggestion:

> `INVISIBLE = 0`

this is not going to work as expected since EarlGrey checks if view is visible `> X` and not `= X` so expecting a view with 0 percent visibility will effectively evaluate to true if `> 0` of a view is visible. (https://github.com/wix-playground/EarlGrey/blob/master/EarlGrey/Matcher/GREYMatchers.m#L223-L234)
And earlgrey exposes separate matcher for

so how about we do this:

```js
// predefined values for day to day usage
export const FULL = 1
export const SUFFICIENT = 0.75
export const HALF = 0.5
export const ANY = 0 // (0.01 ?)


function toBeVisible(percent = SUFFICIENT) {} // default for non-breaking change
```

so all existing code will work the same way
`toBeVisible()` will expect `.75`
and then people can call `toBeVisible(FULL)` or `toBeVisible(HALF)` or `toBeVisible(0.1)` per their scenarios.

Thoughts?

d4vidi:
@maxkomarychev point taken, thanks! But anyways, I'm not sure 0 would be useful at all tbh :)

PR #1741, opened 2019-11-07
Set debug-synchronization by default to 10000, with an option to set to off

rotemmiz:
**Description:**
Set debug-synchronization by default to 10000, with an option to set to off

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -63,20 +63,18 @@ module.exports.builder = {
   d: {
     alias: 'debug-synchronization',
     group: 'Debugging:',
+    default: '10000',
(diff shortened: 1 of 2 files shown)


noomorph:
Interesting... :thinking:  I was musing here in a branch about a non-CLI solution, similar to what I was doing with artifacts: https://github.com/wix/Detox/compare/feat/debugSyncAlternative  . The dark side is that that solution requires fighting with quite a lot of old unit tests.

If you need this feature urgently, at least I would suggest printing deprecation warning about (`--debug-synchronization` and `--debug-synchronization true`, it used to be converted to 3000, no it is no longer so)

Issue #1748, opened 2019-11-10
Allow to override status bar on iOS simulator


sidferreira:
**Is your feature request related to a problem? Please describe.**
Screenshots comparison may fail because of different times the tests run.
This feature allows overriding status bar settings ensuring it will always have the same information.

**Describe the solution you'd like**
https://github.com/sidferreira/Detox/tree/featureStatusBar

**Describe alternatives you've considered**
None

**Additional context**
Focused on iOS for now

LeoNatan:
Please submit as a pull request.

PR #1751, opened 2019-11-11
Adds device.setStatusBar and device.resetStatusBar

sidferreira:
- [x] This is a small change 
- [x] This change has been discussed in issue #1748 
  - [ ] and the solution has been agreed upon with maintainers.

---

**Description:**

This change adds `device.setStatusBar` and `device.resetStatusBar` allowing override status bar information, specially for screenshot tests

### DIFF
--- a/detox/src/devices/drivers/AndroidDriver.js
+++ b/detox/src/devices/drivers/AndroidDriver.js
@@ -252,12 +252,6 @@ class AndroidDriver extends DeviceDriverBase {
       return result;
     }, []);
   }
-
-  async setStatusBar(deviceId, flags) {
-  }
-
-  async resetStatusBar(deviceId) {
-  }
 }
 
 module.exports = AndroidDriver;
--- a/detox/src/devices/drivers/AndroidDriver.js
+++ b/detox/src/devices/drivers/AndroidDriver.js
@@ -252,6 +252,12 @@ class AndroidDriver extends DeviceDriverBase {
       return result;
     }, []);
   }
+
+  async setStatusBar(deviceId, flags) {
+  }
+
+  async resetStatusBar(deviceId) {
+  }
 }
 
 module.exports = AndroidDriver;
--- a/detox/test/e2e/25.statusbar.test.js
+++ b/detox/test/e2e/25.statusbar.test.js
@@ -1,170 +1,47 @@
 
-function sleep(ms) {
-  return new Promise(resolve => setTimeout(resolve, ms));
-}
 
 describe('StatusBar Override', () => {
-  it('Time', async () => {
-    await device.setStatusBar({
-      time: '11:11',
-    });
-    await sleep(500);
-    await device.setStatusBar({
-      time: '22:22',
-    });
-    await sleep(500);
-    await device.setStatusBar({
-      time: '33:33',
-    });
-    await sleep(500);
+  afterEach(async () => {
     await device.resetStatusBar();
-    await sleep(500);
   });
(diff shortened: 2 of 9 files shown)


sidferreira:
Hey @LeoNatan !
LMK if there's anything missing for this merge!

detox/src/devices/drivers/AndroidDriver.js
```diff
@@ -252,6 +252,12 @@ class AndroidDriver extends DeviceDriverBase {
       return result;
     }, []);
   }
+
+  async setStatusBar(deviceId, flags) {
```
noomorph: According to OOP common sense, you should have defined them in the base driver class.

detox/test/e2e/25.statusbar.test.js
```diff
@@ -0,0 +1,170 @@
+
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+describe('StatusBar Override', () => {
+  it('Time', async () => {
```
noomorph: Could you use [it.each](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout) instead of these numerous repetitions?

noomorph: Why there is no platform-specific marker for `:ios:` in these e2e tests?

detox/test/e2e/25.statusbar.test.js
```diff
@@ -0,0 +1,170 @@
...
+    await sleep(500);
+    await device.setStatusBar({
+      time: '22:22',
+    });
+    await sleep(500);
+    await device.setStatusBar({
+      time: '33:33',
+    });
+    await sleep(500);
+    await device.resetStatusBar();
```
noomorph: If there is no way to assert that the commands have an effect on the simulator status bar, could you at least ensure that each test runs a single command? So that `beforeEach` and `afterEach` screenshots will be demonstrating the difference.

noomorph: Move `await device.resetStatusBar();` in `beforeEach()` scope.

noomorph: Most likely, you'll have to use `await device.takeScreenshot()` API to yield two different screenshots in this case.

detox/test/e2e/25.statusbar.test.js
sidferreira: @noomorph should I use one test for each so we can compare the screenshots?

detox/src/devices/drivers/AndroidDriver.js
sidferreira: Fixed

detox/test/e2e/25.statusbar.test.js
sidferreira: Fixed

sidferreira: Fixed

detox/test/e2e/25.statusbar.test.js
sidferreira: Fixed

noomorph: Is there a reason why you have removed `sleep(500)` now in e2e tests?

sidferreira: The `sleep(500)` was to visually follow the test.
As we have the screenshots now, there's no reason to sleep as we can check the screenshots.

But I can place it back if you want.

noomorph:
Thanks, @sidferreira !

PR #1778, opened 2019-11-22
Respect ANDROID_AVD_HOME environment variable

lvanengelen:
- [x] This is a small change 
---

**Description:**

Android SDK tools respect the `ANDROID_AVD_HOME` environment variable to locate installed emulator instances. This change adds support for that setting when fixing up the `config.ini` file.


noomorph:
Could you check this? https://github.com/wix/Detox/pull/1573#discussion_r314241163

lvanengelen:
Looked at #1573 and I investigated a bit. I ran `emulator -list-avds` in different configurations and got the following to list the avds correctly:

- With avds inside `/tmp/foo/.android/avd`: `ANDROID_SDK_HOME=/tmp/foo emulator -list-avds`
- With avds inside `/home/len/android/avd`: `ANDROID_EMULATOR_HOME=/home/len/android emulator -list-avds`
- With avds inside `/home/len/android/avd`: `ANDROID_AVD_HOME=/home/len/android/avd`

I will update the branch to reflect this.

Reading #1573 I am not sure why `ANDROID_SDK_HOME` did not work.

lvanengelen:
Hmm, I have a hard time deciphering the Jenkins logs to figure out what went wrong during the checks of the second commit. I also fail to understand why it would break while the first commit succeeded. Any pointers are welcome.

lvanengelen:
Because the original code also contained `/`, but I can fix that. Also gives me a way of retriggering the checks :) Or is there some easier way than pushing a new commit?

noomorph:
If nobody looks earlier, I can look on Monday.

lvanengelen:
That would be nice. As an experiment I force-pushed my initial commit (which completed fine earlier) and the checks also failed, so something seems to have changed in the test setup after my initial commit?

For now I re-pushed my other commits.

noomorph:
@lvanengelen, thank you for your contribution!

lvanengelen:
Just out of curiosity, could you explain why CI kept failing after my earlier pushes?

noomorph:
I am not sure about everything, but, in particular, the Jenkins OSS CI had undergone some configuration changes and only the recent commits to master have fixed android-rn-59 problems and maybe something else. I have noticed that your branch has been not in sync with master for enough time to start getting some of those issues.

Issue #1798, opened 2019-12-13
Update getDirectories to follow symbolic links


dicearr:
#### Problem description

In my current configuration `$ANDROID_HOME/build-tools` contains symbolic links to the build tools. When Detox looks for the Aapt binary it tries to get the latest available build tools installed by listing all the folders in `$ANDROID_HOME/build-tools`. The function used for that is `getDirectories` which is not following symlinks. All this results in Detox not being able to find the Aapt binary and throwing.

```console
$ ls -l $ANDROID_HOME/build-tools
total 4
lrwxrwxrwx 1 nobody nogroup 101 Jan  1  1970 28.0.3 -> /somepath/build-tools/28.0.3
$ detox test --config android.emu.debug --cleanup -l verbose
...
detox[10764] ERROR: [DetoxExportWrapper.js/DETOX_INIT_ERROR] 
 Error: There was no "aapt" executable file in directory: /path/to/sdk/android-sdk
...
```

#### Proposed solution

Modify the `getDirectories` function so, in case a symbolic link is found inside the root folder,  `fs.realpath` is used to determine the destination path and then check if that path is a folder or not. `fs.realpath` already deals with chains of symbolic links.

LeoNatan:
As I mentioned in the PR review, I think the simplest solution here is to just use `stat`, rather than `lstat`.

LeoNatan:
Merged

PR #1799, opened 2019-12-13
Update getDirectories to include symbolic links to directories

dicearr:
- [x] This is a small change 
- [ ] This change has been discussed in issue #1798  and the solution has been agreed upon with maintainers.

---

**Description:**

Update `getDirectories` to follow symbolic links and return those ones whose destinations are folders.

### DIFF
--- a/detox/src/utils/fsext.js
+++ b/detox/src/utils/fsext.js
@@ -6,7 +6,7 @@ async function getDirectories (rootPath) {
   let dirs = [];
   for (let file of files) {
     let pathString = path.resolve(rootPath, file);
-    if ((await fs.lstat(pathString)).isDirectory()) {
+    if ((await fs.stat(pathString)).isDirectory()) {
       dirs.push(file);
     }
   }


detox/src/utils/fsext.js
```diff
@@ -6,8 +6,14 @@ async function getDirectories (rootPath) {
   let dirs = [];
   for (let file of files) {
     let pathString = path.resolve(rootPath, file);
-    if ((await fs.lstat(pathString)).isDirectory()) {
+    const fileStats = await fs.lstat(pathString);
```
LeoNatan: Any reason not to change the original line to
```js
if ((await fs.stat(pathString)).isDirectory()) {
```
?

As I understand it, `stat` follows symbolic links, while `lstat` returns information about the link itself.

From the documentation:
https://nodejs.org/api/fs.html#fs_fs_lstat_path_options_callback

> `lstat()` is identical to `stat()`, except that if path is a symbolic link, then the link itself is stat-ed, not the file that it refers to.

dicearr: No reason, I just didn't see it in the docs. I will change it to `fs.stat`.

LeoNatan:
Thanks

PR #1808, opened 2019-12-18
Implement DTXProfiler subset for Android

valentynberehovyi:
Added basic integration with DetoxInstruments for Android.


detox/src/artifacts/instruments/recording/AndroidRecordingPathCreator.js
```diff
@@ -0,0 +1,14 @@
+const RecordingPathBuilder = require('./RecordingPathCreator');
+
+class AndroidRecordingPathCreator extends RecordingPathBuilder {
+  constructor(devicePathBuilder) {
+    super();
+    this.devicePathBuilder = devicePathBuilder;
+  }
+
+  createRecordingPath() {
+    return this.devicePathBuilder.buildTemporaryArtifactPath('.dtxplain');
```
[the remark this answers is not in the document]
valentynberehovyi: RecordingPathCreator is abstraction which may have different implementations as example for android and iOS when DevicePathBuilder is just an android util.

detox/test/android/app/src/main/java/com/example/NativeModule.java
```diff
@@ -111,42 +110,40 @@ public void getLaunchArguments(Promise promise) {
...
-     * called concurrently in the ActionsScreen of this demo app; could it be that it is not always called
-     * from the main thread?).
+     * If real life (CI), we've genuinely seen that it throws ConcurrentModificationException exceptions,
+     * on occasions (and why wouldn't it? - our demo app's ActionsScreen has multiple views subscribing and
+     * called concurrently; could it be that not always everything is run in the main thread?).
      * Therefore, we use here {@link ReactFindViewUtil#addViewsListener(OnMultipleViewsFoundListener, Set)},
      * which is too generic but nevertheless allows us to better control when we are to be removed.
      */
-    private static class LongTapsSpy implements OnMultipleViewsFoundListener {
+    private static class LongTapCrasher implements OnMultipleViewsFoundListener {
```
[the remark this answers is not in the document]
d4vidi: It does not. It's here because the branch was merged-to from master instead of rebasing it @valentynberehovyi ❤️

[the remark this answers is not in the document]
d4vidi: @valentynberehovyi you can still solve this by rebasing the branch. In the future, let's keep this in mind as the reason behind why we always rebase rather than merge 😄 
@rotemmiz #not_nitpickying_anymore

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -110,3 +110,61 @@ class QueryStatusActionHandler(
...
         }
 }
+
+class RecordingStateActionHandler(
+        private val instrumentsManager: DetoxInstrumentsManager,
+        private val wsClient: WebSocketClient
+) : DetoxActionHandler {
+
+    override fun handle(params: String, messageId: Long) {
+        val recordingPath = JSONObject(params).optString("recordingPath", null)
```
d4vidi: While this works, the contract by `optString` says the fallback value shouldn't be a `null`. Could you consider using `has()` and `getString()` instead?

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -110,3 +110,61 @@ class QueryStatusActionHandler(
...
+        if (recordingPath != null) {
+            instrumentsManager.startRecordingAtLocalPath(recordingPath)
+        } else {
+            instrumentsManager.stopRecording()
+        }
+
+        wsClient.sendAction("setRecordingStateDone", mapOf<String, Any>(), messageId)
```
d4vidi: Can use `emptyMap` instead of `mapOf` here

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
```diff
@@ -142,5 +142,82 @@ object DetoxActionHandlersSpec: Spek({
                         eq(messageId))
             }
         }
+
+        describe("Recording state actions") {
```
d4vidi: This is good, but need to also add a use case checking that done is sent back when path isn't available.

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -110,3 +110,61 @@ class QueryStatusActionHandler(
...
+        private val wsClient: WebSocketClient
+) : DetoxActionHandler {
+
+    override fun handle(params: String, messageId: Long) {
+        val recordingPath = JSONObject(params).optString("recordingPath", null)
+        if (recordingPath != null) {
+            instrumentsManager.startRecordingAtLocalPath(recordingPath)
+        } else {
+            instrumentsManager.stopRecording()
+        }
```
d4vidi: While I'm ok with the code, overall, I don't entirely understand the contract here. The name `recordingPath` lacks the meaning that the recording starts; The omission of it lacks the meaning that the recording stops.
I think you guys need to rethink this protocol -- namely, make it more explicit (e.g. record="start"/"stop" alongside recordPath as a parameter when record="start")

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
d4vidi: Also, what happens if params is null or malformed (i.e. invalid json)? Better cover that as well (expect as exception)

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -110,3 +110,61 @@ class QueryStatusActionHandler(
             })
         }
 }
+
+class RecordingStateActionHandler(
```
d4vidi: Please add an `Instrum` prefix to the calls name so we could categorize it better :)

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -110,3 +110,61 @@ class QueryStatusActionHandler(
...
+            instrumentsManager.startRecordingAtLocalPath(recordingPath)
+        } else {
+            instrumentsManager.stopRecording()
+        }
+
+        wsClient.sendAction("setRecordingStateDone", mapOf<String, Any>(), messageId)
+    }
+}
+
+class EventsActionsHandler(
```
d4vidi: `Instrum` prefix plz

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
valentynberehovyi: Agree

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
valentynberehovyi: has will check for key with any type of value instead of opting string type exactly.

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
valentynberehovyi: @d4vidi Protocol is the same as for iOS as well.

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
```diff
@@ -142,5 +142,82 @@ object DetoxActionHandlersSpec: Spek({
...
+                verify(instrumentsManager).stopRecording()
+            }
+
+            it("should reply with a 'done' ACK on set state finish") {
+                uut().handle(params, messageId)
+                verify(wsClient).sendAction(eq("setRecordingStateDone"), any(), eq(messageId))
+            }
+        }
+
+        describe("Events actions") {
```
d4vidi: not bad!

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
valentynberehovyi: Agree

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
```diff
@@ -142,5 +142,82 @@ object DetoxActionHandlersSpec: Spek({
...
+            lateinit var instrumentsManager: DetoxInstrumentsManager
+
+            fun uut() = EventsActionsHandler(instrumentsManager, wsClient)
+
+            beforeEachTest {
+                instrumentsManager = mock()
+            }
+
+            describe("begin interval") {
+                val json = "{\"category\":\"MockCategory\",\"name\":\"MockName\",\"id\":\"MockId\",\"additionalInfo\":\"MockAdditionalInfo\",\"action\":\"begin\"}"
```
d4vidi: this is impossible to read :) let's figure out a friendlier solution here. For example, a parser of a multiline JSON string. Or maybe a JSON builder, finally passing to the uut it's parsed (string) form...

d4vidi: It's actually gets worse when this content is used in the call to `verify`. Need to come up with a way to predefine and reuse these strings...

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
```diff
@@ -142,5 +142,82 @@ object DetoxActionHandlersSpec: Spek({
...
+                    verify(instrumentsManager).eventMark(eq("MockCategory"), eq("MockName"), eq("MockId"), eq("MockStatus"), eq("MockAdditionalInfo"))
+                }
+
+                it("should reply with a 'done' ACK") {
+                    uut().handle(json, messageId)
+                    verify(wsClient).sendAction(eq("addActionDone"), any(), eq(messageId))
+                }
+            }
+        }
     }
```
d4vidi: Need to also test the 'else'. Please reimplement using a custom, descriptive Detox exception (rather than a RuntimeException); you can add it to the `DetoxErrors` class...

detox/android/detox/src/main/java/com/wix/detox/DetoxInstrumentsManager.java
```diff
@@ -0,0 +1,141 @@
...
+            methodEventEndInterval = null;
+            methodEventMark = null;
+        }
+        hasProfiler = methodGetInstanceOfProfiler != null;
+    }
+
+    private final Context context;
+    private Object dtxProfilerInstance;
+
+    DetoxInstrumentsManager(Context context) {
```
d4vidi: Please specify which context type exactly you're expecting here (we have too many in instrumentation testing...)

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
valentynberehovyi: Good catch! Agree.

detox/android/detox/src/main/java/com/wix/detox/DetoxManager.java
```diff
@@ -145,5 +145,9 @@ public Unit invoke() {
                 return null;
             }
         }));
+
+        final DetoxInstrumentsManager instrumentsManager = new DetoxInstrumentsManager(reactNativeHostHolder);
+        actionHandlers.put("setRecordingState", new RecordingStateActionHandler(instrumentsManager, wsClient));
+        actionHandlers.put("addEvent", new EventsActionsHandler(instrumentsManager, wsClient));
```
d4vidi: Could we change this so these will only be added conditionally (i.e. if indeed instrumentation exists)?

detox/android/detox/src/main/java/com/wix/detox/DetoxInstrumentsManager.java
```diff
@@ -0,0 +1,141 @@
...
+                    boolean.class//recordReactNativeTimersAsEvents
+            );
+            methodGetInstanceOfProfiler = profilerClass.getDeclaredMethod("getInstance", Context.class);
+            methodStartRecording = profilerClass.getDeclaredMethod("startProfiling", Context.class, configurationClass);
+            methodStopRecording = profilerClass.getDeclaredMethod("stopProfiling");
+            methodEventBeginInterval = profilerClass.getDeclaredMethod("eventBeginInterval", String.class, String.class, String.class, String.class, String.class);
+            methodEventEndInterval = profilerClass.getDeclaredMethod("eventEndInterval", String.class, String.class, String.class);
+            methodEventMark = profilerClass.getDeclaredMethod("eventMark", String.class, String.class, String.class, String.class, String.class);
+        } catch (Exception e) {
+            Log.i("DetoxInstrumentsManager", "Instruments not found", e);
```
d4vidi: I'm not happy with detox starting up in standard mode with an exception on the device. Let's not `static` this and run this conditionally, and only when the proper time comes during detox init....

detox/android/detox/src/main/java/com/wix/detox/DetoxInstrumentsManager.java
```diff
@@ -0,0 +1,141 @@
...
+                    long.class,//samplingIntervalMillis
+                    File.class,//recordingFile
+                    boolean.class//recordReactNativeTimersAsEvents
+            );
+            methodGetInstanceOfProfiler = profilerClass.getDeclaredMethod("getInstance", Context.class);
+            methodStartRecording = profilerClass.getDeclaredMethod("startProfiling", Context.class, configurationClass);
+            methodStopRecording = profilerClass.getDeclaredMethod("stopProfiling");
+            methodEventBeginInterval = profilerClass.getDeclaredMethod("eventBeginInterval", String.class, String.class, String.class, String.class, String.class);
+            methodEventEndInterval = profilerClass.getDeclaredMethod("eventEndInterval", String.class, String.class, String.class);
+            methodEventMark = profilerClass.getDeclaredMethod("eventMark", String.class, String.class, String.class, String.class, String.class);
```
d4vidi: I can't accept this... could you please abstract this away in a delegate (for example, a `DetoxProfilerReflected` class?)

detox/android/detox/src/main/java/com/wix/detox/DetoxInstrumentsManager.java
```diff
@@ -0,0 +1,141 @@
+package com.wix.detox;
+
+import android.content.Context;
+import android.util.Log;
+
+import java.io.File;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+
+public class DetoxInstrumentsManager {
```
d4vidi: What about `DetoxProfilingProxy`? 🤷‍♂

detox/android/detox/src/main/java/com/wix/detox/DetoxInstrumentsManager.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private Object obtainProfiler() {
+        try {
+            return methodGetInstanceOfProfiler.invoke(null, context);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+}
```
d4vidi: As in some of my previous comments, let's move some of this to a delegate class (e.g. `DetoxProfilerReflected`) and un-static stuff and then add some unit tests, plzzzz

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
noomorph: @d4vidi, Valentyn is right. That is what was implemented for iOS.

detox/src/artifacts/instruments/InstrumentsArtifactRecording.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  async doStop() {
+    if (this._isClientConnected()) {
+      await this._client.stopInstrumentsRecording();
+    }
+  }
+
+  _isClientConnected() {
+    return this._client.isConnected && !this._client.pandingAppCrash;
```
noomorph: @valentynberehovyi, we'll need to merge the latest changes from master to here.

detox/src/artifacts/instruments/android/AndroidInstrumentsRecording.js
```diff
@@ -0,0 +1,22 @@
...
+class AndroidInstrumentsRecording extends InstrumentsArtifactRecording {
+  constructor({ adb, client, deviceId, temporaryRecordingPath }) {
+    super({ client, temporaryRecordingPath });
+    this.adb = adb;
+    this.deviceId = deviceId;
+  }
+
+  async doSave(artifactPath) {
+    await super.doSave(artifactPath);
+    await this.adb.pull(this.deviceId, this.temporaryRecordingPath, artifactPath);
```
noomorph: Just musing, what will we do if we don't find a file there... ? If actual profiler code won't be there, I mean. @valentynberehovyi

noomorph: Should we hint as well like below... ?
```
SimulatorInstrumentsRecording.hintAboutDetoxInstruments = _.once(() => {
   log.warn(`Make sure either:
 1. You have installed Detox Instruments:
    https://github.com/wix/DetoxInstruments#installation 
 2. You have integrated Detox Instruments in your app:
    https://github.com/wix/DetoxInstruments/blob/master/Documentation/XcodeIntegrationGuide.md 
 3. You have set the environment variable with your custom Detox Instruments location:
    export DETOX_INSTRUMENTS_PATH="/path/to/Detox Instruments.app"`);
 });
```

valentynberehovyi: We'll add this later. For now recordings will not start when instruments is not installed.

detox/android/detox/build.gradle
```diff
@@ -93,6 +93,16 @@ dependencies {
...
+    buildscript {
+        repositories {
+            jcenter()
+        }
+
+        dependencies {
+            classpath 'de.mannodermaus.gradle.plugins:android-junit5:1.5.2.0'
+        }
+    }
+
```
d4vidi: @valentynberehovyi why is this needed? it's already in the root build.gradle, were you trying to move it?

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
d4vidi: Would you consider changing this?

detox/android/detox/build.gradle
valentynberehovyi: In case of using detox as android project dependency classpath from root project is not available.

detox/android/detox/src/main/java/com/wix/detox/DetoxInstrumentsManager.java
d4vidi: This is still applicable

detox/android/detox/build.gradle
d4vidi: Could you elaborate? what wasn't working specifically?

detox/android/detox/src/main/java/com/wix/detox/instruments/DetoxInstrumentsManager.java
```diff
@@ -0,0 +1,79 @@
...
+            String eventStatus,
+            String additionalInfo
+    ) {
+        if (recording != null) {
+            recording.eventMark(category, name, id, eventStatus, additionalInfo);
+        }
+    }
+
+    public static boolean supports() {
+        return getDefaultInstruments().installed();
```
d4vidi: All initializations made in order to answer this question are done static-ly, so is it really necessary to have an instance here? It seems a bit redundant and also I found it a bit confusing tbh 🤷‍♂

detox/android/detox/src/main/java/com/wix/detox/instruments/reflected/InstrumentsRecordingReflected.java
```diff
@@ -0,0 +1,109 @@
+package com.wix.detox.instruments.reflected;
+
+import com.wix.detox.instruments.DetoxInstrumentsException;
+import com.wix.detox.instruments.InstrumentsRecording;
+
+import java.lang.reflect.Method;
+
+public class InstrumentsRecordingReflected implements InstrumentsRecording {
```
d4vidi: Love this, so clean...

detox/android/detox/src/test/java/com/wix/detox/instruments/DetoxInstrumentsManagerSpec.kt
```diff
@@ -0,0 +1,130 @@
...
+import com.wix.invoke.MethodInvocation
+import org.json.JSONObject
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+import java.io.File
+import java.lang.reflect.InvocationTargetException
+import java.util.*
+import java.util.concurrent.Executors
+
+object DetoxInstrumentsManagerSpec : Spek({
```
d4vidi: boom

detox/android/detox/build.gradle
valentynberehovyi: Spek tests were not working in case of using detox as android project dependency.

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
valentynberehovyi: Not for this iteration. As then we'll need to change the iOS part as well.

detox/android/detox/src/main/java/com/wix/detox/instruments/DetoxInstrumentsManager.java
valentynberehovyi: Static-ly initialised only *Reflected part, which is only one of possible instruments implementations.

detox/android/detox/build.gradle
rotemmiz: @valentynberehovyi can you explain exactly what failed (what gradle task/command) for you that required you to add this?

valentynberehovyi: Mine test app uses detox as project dependency. Also mine app set ext.isOfficialDetoxApp=true to develop test spek as well.
Without classpath in buildscript dependency gradle not able to resolve plugin **'de.mannodermaus.android-junit5'** . I'll move it to mine root project also for now.

detox/src/artifacts/instruments/InstrumentsArtifactRecording.js
noomorph: @valentynberehovyi , what is the status here?

PR #1816, opened 2019-12-26
Extra step to find config file in case emulator has been renamed

gustavos60:
- [X] This is a small change 

---

When renaming an emulator on Android Studio, inside home/.android/avd/ only the .ini file gets renamed, the .avd folder stays with the old name. Passing an emulator that has been renamed on detox configuration causes detox to crash as it tries to find the config file inside a folder that doesn't exist. This PR means to fix the crash by adding an extra step, getting path to the .avd folder from the .ini file.


noomorph:
@gustavos60, thank you for your contribution!

PR #1819, opened 2019-12-30
Update docs

pacozaa:
For MacOS Mojave 10.14.6 you need to install first

`xcode-select install` before Step number 1.3

```sh
brew tap wix/brew
brew install applesimutils
```

- [ ] This is a small change 

---

**Description:**

### DIFF
--- a/docs/Introduction.GettingStarted.md
+++ b/docs/Introduction.GettingStarted.md
@@ -39,7 +39,10 @@ Node is the JavaScript runtime Detox will run on. **Install Node 8.3.0 or above*
 
 A collection of utils for Apple simulators, Detox uses it to communicate with the simulator.
 
+If you are using `MacOS Mojave` run `xcode-select install` first
+
 ```sh
+xcode-select install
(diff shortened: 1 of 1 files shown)


LeoNatan:
I don’t think this is necessary if you have Xcode installed and have run it at least once before installing.

pacozaa:
I have been using Xcode for years and updated it to latest version and yet still got this problem. Yes it is necessary.

docs/Introduction.GettingStarted.md
```diff
@@ -39,7 +39,10 @@ Node is the JavaScript runtime Detox will run on. **Install Node 8.3.0 or above*
 
 A collection of utils for Apple simulators, Detox uses it to communicate with the simulator.
 
+If you are using `MacOS Mojave` run `xcode-select install` first
```
LeoNatan: Let’s just put the line in the instructions. In April, Apple will require Xcode 11, which is Mojave and above only.

PR #1856, opened 2020-01-14
Add clearer logging when detox init fails

yedidyak:
- [x] This is a small change 

---

**Description:**
When detox init fails for any reason, the only indication is that the global injection does not happen and so globals like device and element are undefined.

This PR adds an explicit log that in it failed, without changing any other behaviour.

### DIFF
--- a/detox/src/index.js
+++ b/detox/src/index.js
@@ -4,7 +4,12 @@ let detox;
 
 async function init(config, params) {
   detox = new Detox(config);
-  await detox.init(params);
+  try {
+    await detox.init(params);
+  } catch (e) {
+    console.error(`Error during detox initialization: ${e.message}`);
+    throw e;
+  }
 }
 
 async function cleanup() {
--- a/detox/src/index.test.js
+++ b/detox/src/index.test.js
@@ -1,11 +1,14 @@
 const schemes = require('./configurations.mock');
 describe('index', () => {
-  let detox;
+  let detox, mockDetox;
   beforeEach(() => {
     jest.mock('detox-server');
     jest.mock('./devices/Device');
     jest.mock('./client/Client');
+    jest.mock('./Detox');
+    mockDetox = require('./Detox');
     detox = require('./index');
+    global.console.error = jest.fn();
   });
 
   it(`Basic usage`, async() => {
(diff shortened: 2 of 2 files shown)


detox/src/index.test.js
```diff
@@ -17,6 +20,18 @@ describe('index', () => {
     await detox.cleanup();
   });
 
+  it('Basic usage, if init throws, log a clear explanation', async () => {
+    mockDetox.mockImplementation(() => ({
+      init: () => Promise.reject(new Error('Some tragic event occurred'))
```
LeoNatan: 😂

rotemmiz:
@noomorph tried to tackle this issue before, not sure what the verdict was. @noomorph  can you elaborate please?

detox/src/index.test.js
```diff
@@ -17,6 +20,18 @@ describe('index', () => {
...
+    }));
+    try {
+      await detox.init(schemes.validOneDeviceNoSession);
+    } catch (e) {
+      expect(e).toBeDefined();
```
d4vidi: can this ever run without `e` being defined?

detox/src/index.test.js
```diff
@@ -17,6 +20,18 @@ describe('index', () => {
     await detox.cleanup();
   });
 
+  it('Basic usage, if init throws, log a clear explanation', async () => {
+    mockDetox.mockImplementation(() => ({
+      init: () => Promise.reject(new Error('Some tragic event occurred'))
+    }));
+    try {
+      await detox.init(schemes.validOneDeviceNoSession);
```
d4vidi: better add a `fail()` after this...?

yedidyak:
This checks that the error thrown by detox internally is logged and then
rethrown

detox/src/index.js
```diff
@@ -4,7 +4,12 @@ let detox;
 
 async function init(config, params) {
   detox = new Detox(config);
-  await detox.init(params);
+  try {
+    await detox.init(params);
+  } catch (e) {
```
noomorph: I don't think this is a good idea to duplicate this particular functionality, because we already do have one:

https://github.com/wix/Detox/blob/32dec10754f2e350a0ac95d64ed49edf8c210a83/detox/src/DetoxExportWrapper.js#L83-L91

detox/src/index.test.js
yedidyak: This checks that the error thrown by detox internally is logged and then
rethrown

noomorph:
@yedidyak, the best I can suggest at this moment is to move your tests from `jest-jasmine2` to `jest-circus` (see [init-circus.js](https://github.com/wix/Detox/blob/master/detox/test/e2e/init-circus.js) and [config-circus.json](https://github.com/wix/Detox/blob/master/detox/test/e2e/config-circus.json)).

Screenshot from 2020-01-17 16-03-17

Compare the two outputs between them in the case when `detox.init()` fails. From the point of clarity, `jest-circus` here is the clear winner. :)

PR #1872, opened 2020-01-23
Removed non needed logs for detox instruments android

valentynberehovyi:
Removed unnecessary logs while trying to check instruments for android

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/instruments/reflected/InstrumentsReflected.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/instruments/reflected/InstrumentsReflected.java
@@ -1,7 +1,6 @@
 package com.wix.detox.instruments.reflected;
 
 import android.content.Context;
-import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


detox/android/detox/src/main/java/com/wix/detox/instruments/reflected/InstrumentsReflected.java
```diff
@@ -34,8 +33,6 @@
             methodGetInstanceOfProfiler = profilerClass.getDeclaredMethod("getInstance", Context.class);
             methodStartRecording = profilerClass.getDeclaredMethod("startProfiling", Context.class, configurationClass);
         } catch (Exception e) {
-            Log.i("DetoxInstrumentsManager", "InstrumentsRecording not found", e);
```
d4vidi: I hate to be a nag but if we're going to remove the log altogether, better make the catch more explicit (i.e. of a ClassNotFound exception) and _keep_ the log for other exception types

Issue #1445, opened 2019-06-12
Move process.log into a single file in the root of the artifacts folder


LeoNatan:
As discussed offline, the logic should just output all process logs into a single process.log, rather than cutting it into chunks (with potential to mess up).

~The predicate should be changed into `process==$processName`, where processName is taken before start from the `CFBundleExecutable` Info.plist of the .app specified in package.json.~

LeoNatan:
The `process==${}` has been implemented already in #1877.

noomorph:
@d

d4vidi:
This was about `device.log` (previously named `process.log`...). I don't think there's a case for this anymore.

noomorph:
Regarding the complete device (app) log – looks like it can be done in the scope of the next Artifacts refactor, in **late** stages of Remote Detox - #3743. 

Related to #3092.

PR #1877, opened 2020-01-26
Changed log stream predicate from processImagePath to process==

rotemmiz:
**Description:**
Changed `log stream` predicate from processImagePath to process==. This predicate suites detox much more since it is constant between installs, unlike processImagePath which changes on every reinstallation.


rotemmiz:
This fixes some of the issues presented in https://github.com/wix/Detox/issues/1445

LeoNatan:
Just a note: #1445 refers to the internal predicate, rather than the hint predicate.

Issue #1761, opened 2019-11-14
Move to Github actions instead of Jenkins


yershalom:
**Is your feature request related to a problem? Please describe.**
Nope

**Describe the solution you'd like**
Recently, Github added new feature called actions. The purpose is to define yml files to do the same thing as Jenkins but stay inside github. It also make easier the user management.

**Describe alternatives you've considered**
Jenkins that already is running.

**Additional context**
Also MacOS runners - https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners

rotemmiz:
I'm don't worry re iOS builds, they work pretty good on all SaaS CIs, those are the Android emulators that worry me, none of the services we tried (Circle, Travis, and Bitrise) run Android emulators with sufficient hardware acceleration, and the results are super slow tests (60 minutes and up). We should first find a solution for Android, they if it works, iOS will surely will.

yershalom:
Thanks for your quick answer @rotemmiz,
Can we do it in parallel to Jenkins?

rotemmiz:
You mean both Jenkins and GitHub?

yershalom:
Yes, and github won't fail prs

rotemmiz:
Sure, I'm eager to try it

yershalom:
I like it when you are eager

LeoNatan:
Wix is a legacy account, so we can't use Actions anymore. 🤦‍♂️

jakeleboeuf:
FWIW, I'd be very interested to see an example of a working Github Action workflow example :)

edvinasbartkus:
Github Actions working well for my React Native iOS app detox testing. Here is my workflow with recently added caching:

https://gist.github.com/edvinasbartkus/1e3118dd13c44f43988fcd40e8308db3

jakeleboeuf:
> Github Actions working well for my React Native iOS app detox testing. Here is my workflow with recently added caching:
> 
> https://gist.github.com/edvinasbartkus/1e3118dd13c44f43988fcd40e8308db3

Perfect- works like a charm!

henrymoulton:
@edvinasbartkus have you attempted using Detox + Android at all (in GitHub actions or other CI/CD environment?)

edvinasbartkus:
@henrymoulton not myself, but while researching workflow recipies I did see some successful workflows for Android. Here is one that I have bookmarked on the open-source project:
[...]

PR #1884, opened 2020-01-30
Add example for GitHub Actions

alexsegura:
- [x] This is a small change 

---

**Description:**

This adds an example workflow to run Detox on iOS with GitHub Actions.
I can also provide an example for Android with [android-emulator-runner](https://github.com/ReactiveCircus/android-emulator-runner), but I never managed to run it successfully.

### DIFF
--- a/docs/Guide.RunningOnCI.md
+++ b/docs/Guide.RunningOnCI.md
@@ -157,14 +157,14 @@ workflows:
 
 GitHub Actions is a handy CI service that runs directly inside GitHub. They provide free macOS VMs with HAXM installed.
 
-You can run Detox on GitHub by creating a new workflow. For example, to run iOS tests, create a `.github/workflows/detox.yml` like below:
+You can run Detox on GitHub by creating a new workflow. For example, to run both iOS and Android tests, create a `.github/workflows/detox.yml` like below:
 
 ```yml
 name: Detox
 on: [push, pull_request]
 jobs:
-  detox:
-    name: Detox Tests
+  ios:
+    name: iOS Simulator Tests
     runs-on: macOS-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


rotemmiz:
Great, thanks for this example. We wanted to give it a try and move Detox to github actions as well, this will surely help.
Regarding Android, I heard that some had a successful android e2e with Github Actions (though I am very doubtful after bad experience with Circle CI, Travis CI and bitrise handling x86 Emulators).
We have an issue open about that (re)opened about that
https://github.com/wix/Detox/issues/1761

and here's the gist of someone who managed to run Detox with Android Emulators on Github Actions:
https://gist.github.com/edvinasbartkus/4c09d85ac6f2a2f9eec9d9c8deac5e61

alexsegura:
I have an open pull request on my app if you want to take a look coopcycle/coopcycle-app#414

The emulator seems to start, but the app never finishes starting. Check the logs of my builds. 
I had to manually write `hw.lcd.width`/`hw.lcd.height` in AVD's `config.ini` to make it start (see #1504). 
Also, I had some troubles with Proguard, I disabled it.

d4vidi:
@alexsegura thanks a lot for this! seems that in your CI the emulators fail to load. I think it'd be best to hold this PR until we have a truly working example, or we'd be giving a false impression of that detox (Android) works well on Github actions.

alexsegura:
@rotemmiz @d4vidi finally got it working 🙂 

The main problem was related to [clear text traffic disabled](https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md#problem-the-app-loads-but-tests-fail-to-start-in-sdk--28) on API Level > 27.

I think it's possible to use a `network_security_config.xml`, but I haven't found how to enable it only for release builds created for Detox. In the meantime, I added a step with this a quick'n'dirty patch:

```
brew install xmlstarlet
xmlstarlet ed --inplace --ps --insert "/manifest/application" --type attr -n "android:usesCleartextTraffic" -v "true" android/app/src/main/AndroidManifest.xml
```

Also, as the `reactivecircus/android-emulator-runner` action takes care of creating & spinning an emulator, I use `emulator-5554` as the device name.

d4vidi:
@alexsegura just to be sure I got it right... you're using an attached device but it's name is `emulator-5554`?

alexsegura:
@d4vidi exactly. As I rely on [reactivecircus/android-emulator-runner](https://github.com/ReactiveCircus/android-emulator-runner) for the painful work of launching an emulator properly, I'm using the emulator it creates. 

I think using ready-made actions is totally in the philosophy of GitHub Actions, we should avoid writing code to launch the emulator. Instead, if we find limitations, we should contribute to the already-existing action 😉 
> An action is a reusable unit of code.

d4vidi:
@alexsegura traditionally, Detox works best when managing (and launching) the emulators on its own, and it can (should) also handle attached devices. Have you considered using Detox for this?

alexsegura:
Ok, I will test directly with Detox.

PR #1885, opened 2020-01-30
code: cleanup redundant artifacts code

noomorph:
- [x] This is a small change 

---

**Description:**

1. There was code in the artifacts manager that has never been in use since 2018, so I removed a good part of it.

2. `pathBuilder` in the configuration object, IMO, should be already some class instance, not a string. ArtifactsManager should not take this responsibility for resolving paths and modules. I move that to `configuration.js`.

3. Unit tests fixes.


rotemmiz:
More lines deleted than added == automerge

LeoNatan:
puny


You are all puny.

noomorph:
image

PR #1887, opened 2020-01-31
Fix android instruments isConnected issue

valentynberehovyi:
This fix resolves an issue for android with app termination and launching for instruments recording.


rotemmiz:
Is it tested somehow? Do you run instruments recording with Android e2e?

valentynberehovyi:
> Is it tested somehow? Do you run instruments recording with Android e2e?

It was tested locally and I was able to run instruments recording with Android e2e.
The main problem here was what earlier detox instruments for android was tested without app termination like now.

PR #1903, opened 2020-02-12
DetoxJUnitRunner for Android

valentynberehovyi:
Added DetoxJUnitRunner for Android, it enables ability instruments artifact recording for android.

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
@@ -120,8 +120,8 @@ class InstrumentsRecordingStateActionHandler(
 
     override fun handle(params: String, messageId: Long) {
         val recordingPath = JSONObject(params).opt("recordingPath")
-        if (recordingPath != null) {
-            instrumentsManager.startRecordingAtLocalPath(recordingPath as String)
+        if (recordingPath is String) {
+            instrumentsManager.startRecordingAtLocalPath(recordingPath)
         } else {
             instrumentsManager.stopRecording()
         }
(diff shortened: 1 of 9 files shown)


detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -119,9 +119,9 @@ class InstrumentsRecordingStateActionHandler(
 ) : DetoxActionHandler {
 
     override fun handle(params: String, messageId: Long) {
-        val recordingPath = JSONObject(params).optString("recordingPath", null)
+        val recordingPath = JSONObject(params).opt("recordingPath")
```
d4vidi:
```suggestion
        val recordingPath = JSONObject(params).opt("recordingPath") as String
```

docs/Introduction.Android.md
```diff
@@ -72,7 +72,7 @@ android {
   defaultConfig {
       // ...
       testBuildType System.getProperty('testBuildType', 'debug')  // This will later be used to control the test apk build type
-      testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
+      testInstrumentationRunner 'com.wix.detox.DetoxJUnitRunner'
```
d4vidi: I can't help but wonder whether posting this in the docs could be a bit too soon. Shouldn't we get this through some mileage first?

valentynberehovyi: Agree.

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
valentynberehovyi: > 

This will produce crash as JSONObject$1 will not cast to String.
Just pushed more elegant code for this case.

PR #1909, opened 2020-02-17
Relax androidx.annotation version constraint

### LINKED ISSUE #1898: Android build fails with React Native v0.61 due to version conflict
#### Describe the bug

The Android build fails with React Native v0.61.5 due to a version conflict with `androidx.annotation:annotation`.

#### To reproduce

✓ I have tested this issue on the latest Detox release and it still reproduces.

Steps to reproduce:

1. `npx react-native init AwesomeProject --version 0.61.5`
2. `cd AwesomeProject`
3. `yarn add detox --dev`
4. [Add Detox to the Android project](https://github.com/wix/Detox/blob/15.1.4/docs/Introduction.Android.md#2-add-detox-dependency-to-an-android-project)
5. [Add Kotlin](https://github.com/wix/Detox/blob/15.1.4/docs/Introduction.Android.md#3-add-kotlin)
6. [Create the test class](https://github.com/wix/Detox/blob/15.1.4/docs/Introduction.Android.md#4-create-android-test-class)
7. [Add the Android configuration to `package.json`](https://github.com/wix/Detox/blob/15.1.4/docs/Introduction.Android.md#5-add-android-configuration)
8. `npx detox build --configuration android.emu.debug`

The build then fails with:

```
Could not determine the dependencies of task ':app:preDebugAndroidTestBuild'.
> Could not resolve all task dependencies for configuration ':app:debugAndroidTestRuntimeClasspath'.
   > Could not resolve androidx.annotation:annotation:{strictly 1.0.0}.
     Required by:
         project :app
      > Cannot find a version of 'androidx.annotation:annotation' that satisfies the version constraints: 
           …
```

This issue does not reproduce with the [React Native demo project](https://github.com/wix/Detox/tree/15.1.4/examples/demo-react-native) because [it explicitly depends on `androidx.appcompat:appcompat:1.1.0`](https://github.com/wix/Detox/blob/15.1.4/examples/demo-react-native/android/app/build.gradle#L69) whereas [the default React Native template does not](https://github.com/facebook/react-native/blob/v0.61.5/template/android/app/build.gradle#L181-L192).

The same issue has been reported in https://github.com/wix/Detox/issues/1534, https://github.com/wix/Detox/issues/1537, https://github.com/wix/Detox/issues/1583, and https://github.com/wix/Detox/issues/1631.

#### Environment

- Detox: v15.1.4
- React Native: v0.61.5
- Node.js: v12.15.0
- OS: macOS

sonicdoe:
Relaxes the `androidx.annotation:annotation` version constraint. [Looking at the release notes](https://developer.android.com/jetpack/androidx/releases/annotation#version_110_3), Detox doesn’t seem to use any of the new features exclusive to v1.1.0.

Resolves #1898.

### DIFF
--- a/detox/android/detox/build.gradle
+++ b/detox/android/detox/build.gradle
@@ -74,7 +74,7 @@ dependencies {
     api 'androidx.test:runner:1.2.0'
     api 'androidx.test:rules:1.2.0'
     api 'androidx.test.ext:junit:1.1.1'
-    api 'androidx.annotation:annotation:1.1.0'
+    api 'androidx.annotation:annotation:1.0.0'
 
     // Version is the latest; Cannot sync with the Github repo (e.g. android/android-test) because the androidx
     // packaging version of associated classes is simply not there...


d4vidi:
@sonicdoe I now realize that by 'relaxation' in #1898 you were talking about the actual version rather than the constraint posed by it. Since that is the case, I'd rather not move fwd with this PR because generally speaking, I'd rather keep Detox' deps up to date with the latest versions (at least by default).

sonicdoe:
I’m calling it “relaxation” because Gradle would consider v1.0.0 a required version (which means that it should minimally be v1.0.0) but that can be upgraded. See [Simple version declaration semantics](https://docs.gradle.org/current/userguide/single_versions.html#simple_version_declaration_semantics) in Gradle’s docs.

That being said, I understand if you’d like to require the latest version. I just wanted to open a pull request to show one possible solution to #1898. Feel free to close it.

d4vidi:
Yes I see your point there, thank you. In reality, it's a bit more complex because what really constitutes what the constraint would be from the user perspective is not only gradle, but also the way we translate it to the `pom.xml` inside the `detox.aar`. So I'll have to revisit this more carefully, I suppose.

sonicdoe:
The original issue, https://github.com/wix/Detox/issues/1898, has been fixed with https://github.com/wix/Detox/pull/1950.

d4vidi:
@sonicdoe indeed so, thanks for pointing all of this out for us!

PR #1910, opened 2020-02-17
feat(local-cli): using cosmiconfig to remove requirement on package.j…

Debens:
…son key

- [x ] This is a small change 

---

**Description:**

It has been long pointed out that the package.json holds much responsibility in the node world, this is from Ryan Dahl himself. The larger a projects gets, the more clutter it can accumulate. 

From this a lot of libraries are now choosing to enable developers to split from the package.json and resolve config from individual files. This single config responsibility can aid in code reviewing, separation of concerns and enabling develops to write config in a language of their choice, from json, to js, to yaml. 

[cosmiconfig](https://www.npmjs.com/package/cosmiconfig) is widely popular util library to enable such a thing without being invasive. This change is backwards compatible and as such would not require any immediate change to the docs.

### DIFF
--- a/detox/__tests__/setupJest.js
+++ b/detox/__tests__/setupJest.js
@@ -20,12 +20,17 @@ function callCli(modulePath, cmd) {
   });
 }
 
-function mockPackageJson(mockContent) {
-  jest.mock(path.join(process.cwd(), 'package.json'), () => ({
-    detox: mockContent
+function mockDetoxConfig(mockContent) {
+  jest.mock('cosmiconfig', () => ({
+    cosmiconfigSync: (key) => ({
+      search: () => ({ config: {
+          detox: mockContent
+        }[key]
+      })
+    })
   }));
 }
 
-global.mockPackageJson = mockPackageJson;
+global.mockDetoxConfig = mockDetoxConfig;
 global.callCli = callCli;
 global.IS_RUNNING_DETOX_UNIT_TESTS = true;
--- a/detox/local-cli/build.test.js
+++ b/detox/local-cli/build.test.js
@@ -10,7 +10,7 @@ describe('build', () => {
   });
 
   it('runs the build script if there is only one config', async () => {
-    mockPackageJson({
+    mockDetoxConfig({
       configurations: {
         only: {
           build: 'echo "only"'
@@ -23,7 +23,7 @@ describe('build', () => {
   });
 
   it('runs the build script of selected config', async () => {
-    mockPackageJson({
+    mockDetoxConfig({
       configurations: {
         only: {
           build: 'echo "only"'
@@ -39,7 +39,7 @@ describe('build', () => {
   });
 
   it('fails with multiple configs if none is selected', async () => {
-    mockPackageJson({
+    mockDetoxConfig({
       configurations: {
         only: {
           build: 'echo "only"'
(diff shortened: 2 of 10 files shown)


d4vidi:
This is pretty cool, overall. Things that come to mind, however:
1. I'm not a fan of keeping configuration under hidden files, which cosmicconfig resolves to by default. On the other hand, I _do_ believe only `package.json` and `detox.config.js` would come into play here, but I could be wrong.
2. How well does this work on Windows?

Last but not least, @Debens, alonside the current PR changes, could you effectively extract Detox' config onto a `detox.config.js` file so that we could verify it effectively works as part of the CI proces?

@noomorph any thoughts?

Debens:
By default search will resolve a number of [possible configuration files](https://github.com/davidtheclark/cosmiconfig#cosmiconfigoptions), searching up from `process.cwd()`, which is much like the current default behaviour in place.

I believe windows is supported by default, their [changelog](https://github.com/davidtheclark/cosmiconfig/blob/master/CHANGELOG.md) sort of implies this with a single reference to windows in a very old version. [Trawling through their code](https://github.com/davidtheclark/cosmiconfig/blob/master/src/Explorer.ts) you find they are an avid user of `path.resolve`, which is traditionally the biggest issue when it comes to differing file systems. But I don't have access to windows machine to confirm this, so I sort of have to leave that open.

I'll update the PR at some point this morning to do as you suggest.

It might also be worth noting a potential downside of this would be potential added complexity in setting up the spec life-cycle if someone did choose split out the config.

```
beforeAll(async () => {
    await detox.init(config, { initGlobals: false });
}, 300000);

afterAll(async () => {
    await detox.cleanup();
});
```

noomorph:
In reality, the first step here is not to use some npm package, but to create a breaking change in Detox.

Examine this code:
https://github.com/wix/Detox/blob/25f6e8bcf457d30ef191f84759cc48b205d45d26/detox/local-cli/templates/jest.js#L11-L30

Find two key lines there:

```js
const config = require('../package.json').detox; 
...
await detox.init(config); 
```

Detox is being configured twice: in `local-cli/test.js` inside `detox` npm package, and `init.js` of the user project. That is the heart of the actual issue, which prevents us from embracing any common syntactic sugar solution, whatever it is. :man_facepalming: 

I view the next sequence of steps:

1. We change the `init` method signature to `detox.init(overrides?: DetoxSettings)` and release a breaking major version with the updated blank project templates. The configuration name hint will be taken either from the environment variable `configuration` or the CLI argument `--configuration` like it is implemented nowadays.
2. We replace the current hardcoded require(`${process.cwd()}/package.json`) with cosmiconfig with whatever suits you.
3. Optionally, I'd move also the configuration locator step from `local-cli` to `detox.init()` implementation, so it works regardless of the CLI wrapper, which is bound to Mocha and Jest only. I think it would be still good if Detox CLI gets less smart responsibility, acting more like a tool for passing arguments down, as-is to Mocha/Jest.

noomorph:
@Debens, you are welcome to start working on the new `detox.init()` signature in a separate pull request.

Debens:
What you have mentioned here is what I was referring to in my previous comment, and why I mentioned in the opening of how it would be recommended that no documentation would be amended to reflect this change. Instead continue to leave the templates and documentation to refer to having the configuration in the `package.json` and support this as default. In this first iteration it would instead give power to the developer initialise detox however they wish if they were so inclined to dig a little deeper with no breaking change.

I worry giving power to `detox` to resolve the configuration as increasing its blackox nature, and there is something to be said about how it would be the cli's job to resolve the parameters and call `detox`. 

Saying that tough, I do agree with you on having multiple places for config resolution to be worrying, however, I see this as pre existing condition and such could still be resolved with an addition following PR including the the breaking change you mention. Something I would still be happy to attempt

noomorph:
This issue is not stale - I plan to take it right after we entirely migrate to `jest-circus` with a new Environment class.

noomorph:
Continuing in #2050. Your contribution will be registered, don't worry. Thanks again for raising this issue with external configs and providing a part of the solution.

Issue #1882, opened 2020-01-28
Web support via Puppeteer


awinograd:
**Is your feature request related to a problem? Please describe.**

We deploy a react native web app in addition to ios/android. Ideally we'd be able to test the app using the same harness as our native RN apps.

**Describe the solution you'd like**

I've started implementing a POC `PuppeteerDriver` and web specific actions/matchers. I have some basic assertions / waitFor / actions working (though the code is currently a mess). If this is something your project is interested in having in master, I'd like to discuss approach / implementation details to come up with something that'd be acceptable.

**Describe alternatives you've considered**

There are alternative frameworks (e.g. Cypress) that are specific to web. Test could be 2x written for the web platform specifically.

I could stub the detox globals with ones backed by cypress / puppeteer without integrating directly into detox. Maintain the api, but have a different implementation.

LeoNatan:
Hello,

Since we have no way of supporting or maintaining this in the future, I think this would be best handled as an external driver, which you can maintain properly.

Do you know what is needed to support such a driver in Detox currently? We will be willing to accept a PR with API to be able to provide your own driver, which will then open the door to third party drivers. 

What do you think?
Thanks

awinograd:
@LeoNatan, I understand not wanting to support the plugin long term. Supporting external drivers sounds like a great path forward to me. Do you think that `configuration.type` could be set to a path on disk to a driver and change 

https://github.com/wix/Detox/blob/444b8d30fe32afab4b2811b84ef4297bb18983b6/detox/src/Detox.js#L55

to something like:

```
    const DeviceDriverClass = DEVICE_CLASSES[this._deviceConfig.type] || require(this._deviceConfig.type);
```

LeoNatan:
@noomorph @rotemmiz Can you pls advise? Thanks

rotemmiz:
We already had some discussion about such an API here https://github.com/wix/Detox/issues/1091, though I don't remember the details (I will read the thread again later tomorrow).
I would really be glad if this is pushed forward!

awinograd:
Thanks for linking the relevant issue! The proposed api in there looks like it would work. I've started an initial investigation into trying to add support and registering the driver + extending the base class are definitely the two main things i've run into.

My implementation is moving forward... Got a lot of the basic stuff working (most matchers, most assertions, screenshot plugin, video recording plugin in progress... etc). Code isn't in the best condition, but my first approach was to get something that works and then hopefully cleanup from there.

[...]

PR #1919, opened 2020-02-20
Add dynamic require for driver plugins

awinograd:
- [ ] This is a small change 
- [x] This change has been discussed in issue https://github.com/wix/Detox/issues/1882 and the solution has been agreed upon with maintainers.

---

**Description:**

This PR adds support for external drivers as discussed in https://github.com/wix/Detox/issues/1882#issuecomment-584090146.

```
"ios.sim.release": {
        "binaryPath": "...",
        "build": "...",
        "type": "my-custom-driver-package-name",
        "device": {
	  // can be pretty flexible (according to the device driver), add whatever device specification needed
        }
      },
```


detox/src/Detox.js
```diff
@@ -52,7 +52,7 @@ class Detox {
     this._client = new Client(sessionConfig);
     await this._client.connect();
 
-    const DeviceDriverClass = DEVICE_CLASSES[this._deviceConfig.type];
+    const DeviceDriverClass = DEVICE_CLASSES[this._deviceConfig.type] || require(this._deviceConfig.type);
```
awinograd: One thing I don't like about this direct fallback to require is you get a weird error about module resolution if you have a typo for one of the built-in drivers. In other words, the `if` on the line below is sort of useless...

@rotemmiz, what are your thoughts about requiring third party drivers to be prefixed in the configuration?

Something like:

```
"ios.sim.release": {
        "binaryPath": "...",
        "build": "...",
        "type": "plugin:my-custom-driver-package-name",
        "device": {
	  // can be pretty flexible (according to the device driver), add whatever device specification needed
        }
      },
```

```
    let DeviceDriverClass = DEVICE_CLASSES[this._deviceConfig.type];
    if (!DeviceDriverClass) {
      if (this._deviceConfig.type.startsWith('plugin:')) {
            DeviceDriverClass = require(this._deviceConfig.type);
      }
      throw new Error(`'${this._deviceConfig.type}' is not supported`);
    }
```

awinograd:
One other thing that would be great to resolve as part of this PR is the `binaryPath` issue discussed in https://github.com/wix/Detox/issues/1882#issuecomment-582023492.

Essentially, I want a way to avoid this line which is currently mandatory but doesn't make sense in the context of a website which doesn't have a binaryPath. (Currently I'm using binary path to specify the base web url but have to add a leading `/` to make it absolute and avoid this logic. e.g. `/https://example.com`)

https://github.com/wix/Detox/blob/475d10efc51b2e120d241f4b37b08566029f2319/detox/src/devices/Device.js#L18

rotemmiz:
Re binaryPath, let's make sure it is only used where needed (push it into the drivers maybe).
We can introduce `appPath` in Device instead, that will be used for binaryPath in iOS/Android drivers, and for url in your driver.
WDYT?

awinograd:
@rotemmiz thanks for the direction. I just pushed a commit that pushes absolute path logic into the driver layer (removing from device).

I wasn't sure exactly how you wanted me to handle `binaryPath` vs `appPath`

Did you want to totally remove `binaryPath` as an option from the config (a breaking change)?

Or add a new `appPath` field to config that takes precedence over and deprecates `binaryPath` for future removal.

In the grand scheme of things, it's currently not a huge issue to keep the `binaryPath` naming even though it perhaps isn't as accurate for my web driver, so I'm also happy to punt on that until the next major version is on its way.

Let me know how you'd like to proceed!

rotemmiz:
Hhy @awinograd, seems like a few unit tests failed, and coverage is missing as well.

awinograd:
@rotemmiz, tests + coverage fixed

rotemmiz:
Hey @awinograd, sorry for the delay here, I was trying to think and consult on how to proceed here.
So the things we think are still missing to make sure we handle this PR correctly are the following:
1. A new section in documentation, explaining how to create such a plugin, and most importantly, a section about how to use one.
2. A good integration test to make sure we don't break this on the next commit we make for Detox. So I suggest you add another project (either in examples dir on someplace else, but we just need to make sure it runs in CI as well), and add a very simple setup (a package.json, short e2e test, and an empty stub of a driver that just returns resolved promises on anything).

WDYT?

awinograd:
@rotemmiz no need to apologize! I appreciate your guidance through the PR process.

1. Yes I can write up a doc on custom drivers
1. Do you want me to configure the setup on circleci, or would a script in `examples/custom-driver/package.json` be sufficient for you to take it and get it on circleci?

rotemmiz:
Although it makes little sense to add the project to one of the platform specific test project CI builds, I believe it will be the fastest and easiest, and you'll be able to do it independently.

Try adding your new project to the list here, and see
https://github.com/wix/Detox/blob/master/scripts/demo-projects.android.sh

awinograd:
@rotemmiz, sorry for the delay in getting back to this. I just pushed an example project (🤞 that the CI passes!) 

Still need to work on the Doc, but wanted you to take a look at f014adc as I'm not happy with what i've done there. It's related to our discussion above (https://github.com/wix/Detox/pull/1919#issuecomment-589166755) about `binaryPath`. The old testing code assumes that `prepare` will throw if the binary path doesn't exist on disk which is an issue when binaryPath is a URL. I wasn't sure what the best path forward is, so just pushed something that passes but is sort of a hacky workaround. Open to suggestions there

awinograd:
Just put up a first pass at a new Guide around Third party drivers. I'm curious to know what areas you think may be interesting to include that i've left off. As of now, it's a pretty high level summary.

awinograd:
The CI is failing due to

> 19:05:05 /home/jenkins/.jenkins/workspace/detox-demo-projects-android-59-9-pr/examples/demo-puppeteer/node_modules/puppeteer/.local-chromium/linux-722234/chrome-linux/chrome: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory

I see two potential solutions:

1. Install the missing package(s)
1. Change the demo to use a trivial, noop driver rather than `detox-puppeteer`

Let me know which direction you prefer. If it's (1), I'll need help

awinograd:
I went ahead an chose path (2) since it seemed like simplest way to test the actual integration without worrying about the internal plugin details. If you'd like to go down path (1) and restore `examples/demo-puppeteer` let me know. It's just a simple `revert` away

awinograd:
Most recent test failure seems unrelated to my changes

rotemmiz:
Good call going with 2, we wouldn't want detox CI to be more complex than what it already is, this demo you added is there only to showcase the API and make sure we don't break it unintentionally

rotemmiz:
Though the first error is probably not related (couldn't found the cause scrolling with my phone), the second (in demo projects build)  looks related to your work

awinograd:
@rotemmiz , you're right the 2nd error is related to my test. It looks like it got introduced when merging in master since the detox major version was bumped and `demo-plugin` was installing `detox` from npm rather than as a symlink via lerna.

🤞 this next run passes. It's working locally for me

awinograd:
Woohoo! Green tests 😄

awinograd:
@rotemmiz , just merged master to resolve conflicts. Anything left to do to get this PR wrapped up?

awinograd:
doesn't seem as if test failures are related to me, but could be wrong

rotemmiz:
Not sure where this comes from, retriggered your build

lerna.json
```diff
@@ -9,6 +9,8 @@
     "examples/demo-react-native",
     "examples/demo-react-native-detox-instruments",
     "examples/demo-react-native-jest",
+    "examples/demo-puppeteer",
```
noomorph, line 12: Where is `demo-puppeteer` project? I cannot find it in the committed files.

awinograd: My mistake. It was a package previously in this PR but it was removed in favor of `demo-plugin`. Fixed in a16ff9b6

awinograd:
I see a new coveralls ci process has been added to the PR and has failed. https://coveralls.io/builds/29961248 

Should I be paying attention to it?

noomorph:
@awinograd, thanks for telling. I'll look into it.

detox/src/devices/Device.test.js
```diff
@@ -815,7 +815,7 @@ describe('Device', () => {
   async function launchAndTestBinaryPath(configuration) {
     const device = schemeDevice(configurationsMock.pathsTests, configuration);
 
-    await device.prepare();
+    await device.prepare({ checkBinaryPath: true });
```
rotemmiz, line 818: Seems to me like `checkBinaryPath` doesn't really belong in here,  is there any function call without `checkBinaryPath:true` ?
If this is to separate the web driver from iOS/Android drivers, then maybe this logic should reside in the drivers themselves.

detox/src/devices/Device.test.js
```diff
@@ -478,7 +478,7 @@ describe('Device', () => {
 
       await device.installApp('newAppPath');
 
-      expect(driverMock.driver.installApp).toHaveBeenCalledWith(device._deviceId, 'newAppPath', undefined);
+      expect(driverMock.driver.installApp).toHaveBeenCalledWith(device._deviceId, 'newAppPath', device._deviceConfig.testBinaryPath);
```
rotemmiz, line 481: Why did this change from `undefined`  to `testBinaryPath` ?

docs/Guide.ThirdPartyDrivers.md
```diff
@@ -0,0 +1,62 @@
+# Third Party Drivers
+
+Detox comes with built-in support for running on Android and iOS by choosing a driver type in your detox configurations. For example, the following configuration uses the "ios.simulator" driver.
```
rotemmiz, line 3: "as of v16.(3?)"
We need to remember to add it

docs/Guide.ThirdPartyDrivers.md
```diff
@@ -0,0 +1,62 @@
...
+```
+3. Run detox while specifying the name of your new configuration `detox test --configuration detox-driver-package`
+
+## Writing a new driver
+
+### Anatomy of a driver
+
+The architecture of a driver is split into a few different pieces. Understanding the [overall architecture of Detox](https://github.com/wix/Detox/blob/master/docs/Introduction.HowDetoxWorks.md#architecture) will help with this section
+
+1. The Device Driver - code runs on the Detox Server. It implements the details for the
```
rotemmiz, line 36: Not the "Detox Server", but "Detox tester, within the test runner context"

detox/src/devices/Device.test.js
awinograd: I'm not ecstatic about how I had to get the tests passing. If I remember correctly, I was fighting against code coverage, but perhaps the real solution is to add a separate test for `getAbsoluteBinaryPath`

detox/src/devices/Device.test.js
awinograd: Previously this expectation was relying on implementation details of the `Device` class. Specifically, it knew that testBinaryPath was not passed along to `newAppPath` if `prepare` has not yet been called. That's because previously, the binary path instance properties were set on the instance during `prepare` after doing the absolute path logic I have pushed down into the driver layer.

That seemed a bit odd to me, since in all real uses of Device (as far as I'm aware) prepare will always be called before using the other methods. This method expects the testBinaryPath as the last argument, it just previously happened to not be defined yet

awinograd:
@rotemmiz thanks for the review! I've done a first pass at addressing your comments

detox/src/devices/Device.test.js
rotemmiz: So does this mean you can get rid of checkBinaryPath: true?

rotemmiz:
Can't yet understand why all builds failed, I will try figuring it out tomorrow morning.
Also, do we still need the  checkBinaryPath flag in function calls?

detox/src/devices/Device.test.js
awinograd: sorry I missed this reference. yes we can get rid of it

awinograd:
@rotemmiz just updated the PR to remove the remaining reference to `checkBinaryPath`. unit tests are passing locally for me.

thanks for taking a look at the remaining failures.

noomorph:
@awinograd, the issue with Coveralls was finally solved - could you please merge `master` in?

awinograd:
@noomorph . Sure thing. Just merged master

awinograd:
Coveralls ran but failed here. 

Jenkins CI is green again!

noomorph:
@awinograd, thanks for reporting. I have found why the coverage was below 70ish% and pushed the fix. Could you please merge `master` again? Now it should pass the required coverage.

awinograd:
merged!

d4vidi:
Just sayin'... https://github.com/wix/Detox/pull/1919/files#r407250431 😄

examples/demo-plugin/e2e/init.js
```diff
@@ -0,0 +1,31 @@
...
+
+jasmine.getEnv().addReporter(adapter);
+
+// This takes care of generating status logs on a per-spec basis. By default, jest only reports at file-level.
+// This is strictly optional.
+jasmine.getEnv().addReporter(specReporter);
+
+// This will post which device has assigned to run a suite, which can be useful in a multiple-worker tests run.
+// This is strictly optional.
+jasmine.getEnv().addReporter(assignReporter);
```
d4vidi, line 15: @noomorph another thing to keep up to date with respect to jasmine deprecation

awinograd:
Anything else I can do here?

rotemmiz:
🥳

awinograd:
Amazing! Thanks for all of your help getting this change developed & merged! Excited to see it released officially 🎊

rotemmiz:
Thank you for having such persistency in having this PR perfected.
We are happy this change is in.

d4vidi:
W00t! Don't forget to bump a minor version for the next release :)

justinwyer:
Amazing stuff @awinograd this is super stuff 😍thanks so much for driving this!

d4vidi:
@awinograd @justinwyer please be advised that we're about to introduce a minor yet breaking change to the plugged-in drivers in #2214. The change it subtle and we will report the details in the release itself, but it will not be release as a major version.

Issue #1912, opened 2020-02-18
allow video codec to be specified


cs01:
**Is your feature request related to a problem? Please describe.**
When recording video, Detox does not not allow for the codec of the video to be specified. This makes us run an extra step to convert the video to a format supported by web browsers, since the default format, hevc, cannot play in the browser. We are viewing test failures through a web based application, so we need to get the video to something like h.264.

**Describe the solution you'd like**
Add support for an additional argument specifying the encoding. This is supported by xcode, so it is as simple as passing another argument to the shell command. The command would optionally have ` --codec $CODEC` (i.e. `--codec h264`) appended to it:

https://github.com/wix/Detox/blob/5dfb9caa601acaf9c825bfdfd3b92611326100b5/detox/src/devices/ios/AppleSimUtils.js#L236

**Describe alternatives you've considered**
Record video, then manually convert it. Below, the video is converted to h.264
```
>> ffmpeg -i hevc_encoded.mp4 -vcodec libx264 -acodec aac h264_encoded.mp4
```

**Additional context**
```
>> xcrun simctl io --help
Set up a device IO operation.
Usage: simctl io <device> <operation> <arguments>

Supported operations:
...

	recordVideo [--codec=<codec>] [--display=<display>] [--mask=<policy>] [--force] <file or url>
	    Records the display to a QuickTime movie at the specified file or url.
	    --codec      Specifies the codec type: "h264" or "hevc". Default is "hevc".
...
```

LeoNatan:
This sounds interesting, but unfortunately, we won't have the bandwidth to implement this internally. Would you mind submitting a PR for this?

cs01:
I would be happy to.

cs01:
While making these changes it occurred to me that having types, such as with TypeScript, would make me more comfortable that I'm not breaking things or making mistakes. Is TypeScript used by the Detox team during development of Detox's js code?

noomorph:
See configuration.js, ADB.js and Android vide recording plugin source code files. The pull request should be simple enough.

As for TypeScript, this is a pain point in the team, but tl;dr - no, we use only JSDoc for some type checking and that's it.

noomorph:
@cs01, open a new issue if PR #1920 seems not enough.

PR #1920, opened 2020-02-20
feat(artifacts): codec option for simulator video recording

cs01:
- [x] This change has been discussed in issue #1912  and the solution has been agreed upon with maintainers.

---

**Description:**

Adds `codec` option to artifacts JSON config (in package.json), e.g.:

```json
{
  "detox": {
    "artifacts": {
      "plugins": {
        "video": {
          "simulator": {
            "codec": "h264"
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/detox/src/artifacts/video/SimulatorRecordVideoPlugin.js
+++ b/detox/src/artifacts/video/SimulatorRecordVideoPlugin.js
@@ -14,14 +14,14 @@ class SimulatorRecordVideoPlugin extends VideoArtifactPlugin {
   }
 
   createTestRecording() {
-    const { context, appleSimUtils } = this;
+    const { api, context, appleSimUtils } = this;
     const temporaryFilePath = temporaryPath.for.mp4();
     let processPromise = null;
 
     return new Artifact({
       name: 'SimulatorVideoRecording',
       start: async () => {
-        processPromise = appleSimUtils.recordVideo(context.deviceId, temporaryFilePath);
+        processPromise = appleSimUtils.recordVideo(context.deviceId, temporaryFilePath, api.userConfig.simulator);
       },
       stop: async () => {
         if (processPromise) {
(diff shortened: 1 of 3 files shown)


detox/src/artifacts/video/SimulatorRecordVideoPlugin.js
```diff
@@ -11,6 +11,7 @@ class SimulatorRecordVideoPlugin extends VideoArtifactPlugin {
     super(config);
 
     this.appleSimUtils = config.appleSimUtils;
+    this.codec = config.codec;
```
cs01: The config does not currently have the codec passed to it. Advise on the best way to do this would be helpful.

LeoNatan: @noomorph @rotemmiz Pls advice.

detox/src/devices/ios/AppleSimUtils.js
```diff
@@ -233,7 +233,12 @@ class AppleSimUtils {
   }
 
   recordVideo(udid, destination) {
-    return exec.spawnAndLog('/usr/bin/xcrun', ['simctl', 'io', udid, 'recordVideo', destination]);
+    const args = ['simctl', 'io', udid, 'recordVideo', destination];
+    if (codec) {
+      args.push('--codec');
+      args.push(codec);
```
LeoNatan: Pls test what happens if user passes a bad argument (like "vp9") here; will Detox catch a failure in the simctl command and complain to the user, or silently ignore this? The former sounds like a better option.

cs01: It does fail and return an error code. Not sure if detox continues on afterwards or bails at that point. I'll have to test.
```
>> /usr/bin/xcrun simctl io booted recordVideo /tmp/video.mp4 -f --codec hsadf
(help text printed)
...
>> echo $?
117
```

noomorph: No worries, that should not be a problem. Detox will print a warning from stdout/stderr of `xcrun` .

cs01:
Thank you!

PR #1926, opened 2020-02-23
(Android) Introduce ANR monitor to produce warn-logs

### LINKED ISSUE #1925: Report thread-dump when ANR's are in effect
Detox needs to introduce a way to deal with [ANR's](https://developer.android.com/topic/performance/vitals/anr).

Though their coming-into-existence is _not_ a test-framework matter per se, their instantiation (i.e. when alerted by Android) evidently causes all tests to bulkly-fail and with no clear hint as to why they've appeared in the first place. The latter fact makes it very hard to trace back when running on CI agents, in particular. Detox needs to step up :-)

Have some mechanism that would detect and log a thread-dump when ANR's are in-effect, perhaps even provide the [system's traces file](https://developer.android.com/topic/performance/vitals/anr#pull_a_traces_file) as an artifact.

N/A

**Additional context**
N/A

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1925 and the solution has been agreed upon with maintainers.

---

**Description:**

Resolves #1925. Introduces usage of the very interesting [ANR-WatchDog project](https://github.com/SalomonBrys/ANR-WatchDog) for the sake of ANR-state proximation. The ends result is a warning on the Detox (tester) side with a main-thread stacktrace, when main-thread blockage is detected, alongside a complete thread-dump on the device logs.

Integration of it all was pretty straightforward, but testing this (e2e) was a bit of a pain since logs are not easy to intercept.

### DIFF
--- a/detox/src/Detox.js
+++ b/detox/src/Detox.js
@@ -1,7 +1,7 @@
 const _ = require('lodash');
 const util = require('util');
 const logger = require('./utils/logger');
-const log = require('./utils/logger').child({ __filename });
+const log = logger.child({ __filename });
 const Device = require('./devices/Device');
 const IosDriver = require('./devices/drivers/IosDriver');
 const SimulatorDriver = require('./devices/drivers/SimulatorDriver');
@@ -57,6 +57,7 @@ class Detox {
     }
 
     this._client = new Client(sessionConfig);
+    this._client.setNonresponsivenessListener(this._onNonresnponsivenessEvent.bind(this));
     await this._client.connect();
 
     const DeviceDriverClass = DEVICE_CLASSES[this._deviceConfig.type];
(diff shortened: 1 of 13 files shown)


detox/src/Detox.test.js
```diff
@@ -158,10 +172,11 @@ describe('Detox', () => {
   });
 
   it(`handleAppCrash if client has a pending crash`, async () => {
+    client.getPendingCrashAndReset.mockReturnValueOnce('crash');
+
     Detox = require('./Detox');
     detox = new Detox({deviceConfig: validDeviceConfigWithSession});
     await detox.init();
-    detox._client.getPendingCrashAndReset.mockReturnValueOnce('crash'); // TODO: rewrite to avoid accessing private fields
```
d4vidi: TODO ✅

detox/src/Detox.js
```diff
@@ -57,6 +57,7 @@ class Detox {
     }
 
     this._client = new Client(sessionConfig);
+    this._client.setNonresponsivenessListener(this._onNonresnponsivenessEvent.bind(this));
```
d4vidi: @noomorph I was far from sure as to whether this is the right level for this work-scope. I'd be happy to get your insights on this (for example, should this be part of the client's implementation?)

PR #1929, opened 2020-02-24
feat(artifacts): add onAppReady hook

### LINKED ISSUE #1914: Record device logs even when a beforeAll fails
**Describe the bug**
Detox does not produce device logs when an inner `beforeAll()` (inside the scope of a single describe) fails to complete. In such cases - running on private projects, only the `testDone.png` artifact was available, even though `--record-logs all/failed` was used.

**Expected behavior**
Device logs should be produced in this scenario.

 - Detox: 15.2.2 (latest)
 - React Native: 0.61.5
 - Node: v10.15.0
 - Device: Pixel 3
 - OS: API 28
 - Test-runner (select one): Jest+jest-circus

noomorph:
**Description:**

Partially resolves #1914  .

* Addresses `detox.init()` timeout issues
* Prevents a test runner from hanging when things go awry.
* Adds `onAppReady` hook to artifacts manager
* Adds automatic screenshot if: `appNotReady.png` (when there was an issue with launching the app).


LeoNatan:
What is "app ready"?

LeoNatan:
"Hook that is called after app sends ready signal"

So the first idle moment that Detox sees?

LeoNatan:
In the Detox Next, the test runner sends "ready" right away. So it will not work as you expect.

detox/runners/jest/DetoxAdapterCircus.js
```diff
@@ -35,7 +35,7 @@ class DetoxAdapterCircus extends CircusTestEventListenerBase {
 
   _onTestStart(event) {
     const { test } = event;
-    if (test.mode === 'skip' || test.mode === 'todo') {
+    if (test.mode === 'skip' || test.mode === 'todo' || test.errors.length > 0) {
```
noomorph: Here I am making sure that artifacts manager does not receive messages about `test start` when it is skipped because of the outer timeout.

detox/runners/jest/DetoxAdapterImpl.js
```diff
@@ -38,12 +38,16 @@ class DetoxAdapterImpl {
   }
 
   testComplete({status, timedOut}) {
-    const _test = {
-      ...this._currentTest,
-      status,
-      timedOut,
-    };
-    this._enqueue(() => this._afterEach(_test));
+    if (this._currentTest) {
```
noomorph: If outer hooks (e.g.: `beforeAll`) fail with a timeout, we still get `afterEach` called in Detox. :scream: 
Now, I am making sure we call `detox.afterEach()` only for those tests that actually ran.

detox/src/Detox.js
```diff
@@ -32,75 +32,44 @@ class Detox {
     );
 
     const {artifactsConfig, deviceConfig, session} = config;
+    this._artifactsConfig = artifactsConfig;
     this._deviceConfig = deviceConfig;
     this._userSession = deviceConfig.session || session;
     this._client = null;
     this._server = null;
-    this._artifactsManager = new ArtifactsManager(artifactsConfig);
+    this._artifactsManager = null;
```
noomorph: All is `null` at the beginning, so I know for sure what was initialized and cleaned up at a specific point of time.

detox/src/Detox.js
```diff
@@ -32,75 +32,44 @@ class Detox {
...
-      ...userParams,
-    };
-
-    if (!this._userSession) {
-      this._server = new DetoxServer({
-        log: logger,
-        port: new URL(sessionConfig.server).port,
-      });
+  init(userParams) {
+    if (!this._initPromise) {
```
noomorph: I just think that it is common sense to abort `detox.init()` promise if it is not done, but we are calling `detox.cleanup()`. Why should this promise be unresolved, if we can reject it? :man_shrugging:

detox/src/Detox.js
```diff
@@ -32,75 +32,44 @@ class Detox {
...
-
-    await this._artifactsManager.onInit();
-
-    return this;
+    return this._initPromise;
   }
 
   async cleanup() {
-    await this._artifactsManager.onBeforeCleanup();
+    await this._assertNoPendingInit().catch(_.noop);
```
noomorph: So, now cleanup tries to abort `detox.init` if it was still somewhere in progress.

detox/src/Detox.js
```diff
@@ -32,75 +32,44 @@ class Detox {
...
+
+    if (this._artifactsManager) {
+      await this._artifactsManager.onBeforeCleanup();
+      this._artifactsManager = null;
```
noomorph: Nullifying now everything, to fail fast if somebody tries to manipulate a particular Detox instance after the cleanup.

detox/src/Detox.js
```diff
@@ -109,14 +78,19 @@ class Detox {
...
 
     if (argparse.getArgValue('cleanup') && this.device) {
       await this.device.shutdown();
     }
+
+    this.device = null;
   }
 
   async beforeEach(testSummary) {
+    await this._assertNoPendingInit();
```
noomorph: Thank you, `jest-jasmine` :nauseated_face: :rage:  .
If test runner proceeds to run individual tests with their hooks after a timeout, `beforeEach` now will be mercilessly throwing errors.

detox/src/Detox.js
```diff
@@ -127,6 +101,8 @@ class Detox {
   }
 
   async afterEach(testSummary) {
+    await this._assertNoPendingInit();
```
noomorph: Added for symmetry with `beforeEach`, why not?

detox/src/Detox.js
```diff
@@ -144,6 +120,70 @@ class Detox {
...
+      throw new Error(`'${this._deviceConfig.type}' is not supported`);
+    }
+
+    const deviceDriver = new DeviceDriverClass({
+      client: this._client,
+    });
+
+    Object.assign(this, deviceDriver.matchers);
+
+    this.device = new Device({
```
noomorph: Assigning instances to properties as soon as they are constructed.
It makes it easier to work with artifacts in cleanup phase when init is not really done.

detox/src/Detox.test.js
```diff
@@ -83,8 +83,9 @@ describe('Detox', () => {
 
     detox = new Detox({deviceConfig: validDeviceConfig});
     await detox.init();
+    const device = detox.device;
     await detox.cleanup();
-    expect(detox.device.shutdown).toHaveBeenCalledTimes(1);
+    expect(device.shutdown).toHaveBeenCalledTimes(1);
```
noomorph: After cleanup `detox.device` now is null, so yeah... tests had to be updated.

detox/src/Detox.test.js
```diff
@@ -95,12 +96,21 @@ describe('Detox', () => {
     expect(() => detox.cleanup()).not.toThrowError();
   });
 
+  it(`Calling detox.init() twice returns the same promise`, async () => {
```
noomorph: I can't really explain why I specifically would want a single promise to be returned, but anyway, we use each instance of Detox class only once, so why not? I mean, calling the global `detox.init` works with `DetoxExportWrapper` which is responsible for the creation of Detox instances. Hence, technically, each Detox **instance** is `init`-ed only once, so I don't feel like this change is a mistake.

detox/src/DetoxExportWrapper.js
```diff
@@ -35,7 +34,18 @@ class DetoxExportWrapper {
       Detox.none.initContext(global);
     }
 
-    this[_detox] = await DetoxExportWrapper._initializeInstance(config, params);
+    Detox.none.setError(null);
+
+    try {
+      this[_detox] = DetoxExportWrapper._createInstance(config);
+      await this[_detox].init(params);
```
noomorph: This line is super important.
Without assigning the Detox instance to `this[_detox]`, we can't effectively call `detox.cleanup()`, surprise! :boom:
That was a fat bug responsible for why we had issues with artifacts not appearing on test runner timeouts.

detox/src/DetoxExportWrapper.js
```diff
@@ -35,7 +34,18 @@ class DetoxExportWrapper {
...
+    } catch (err) {
+      Detox.none.setError(err);
```
noomorph: :sob: 
On the other side, we can't automatically set `this[_detox] = Detox.none` anymore. :man_shrugging: 
Is that bad, you will ask me? Yes and not exactly.

Please keep in mind that all this juggling with `Detox.none vs this[__detox]` all of a sudden is our desperate attempts to work around `jest-jasmine` bug when tests just move forward, even if detox.init() failed with a timeout.

Jest Circus solutions don't need this workaround. Mocha is fine also.
So, now, I apply another workaround. See:

```js
try {
  await adapter.beforeEach();
} catch (err) {
  await detox.cleanup();
  throw err;
}
```

By this, even after timeout error, when we recklessly continue to run, the first throwing beforeEach will result into detox.cleanup(). That's why all the consequent tests which don't plan to stop :roll_eyes:  will be failing with our Detox.none workaround.

I can't say I love all this story, but let's deprecate Jest-Jasmine support, and we might throw out some mere 150-300 lines of weird support code.

detox/src/DetoxExportWrapper.js
```diff
@@ -58,48 +68,29 @@ class DetoxExportWrapper {
...
-      return instance;
-    } catch (err) {
-      Detox.none.setError(err);
-      log.error({ event: 'DETOX_INIT_ERROR' }, '\n', err);
-
-      if (instance) {
-        await instance.cleanup();
-      }
+  static _createInstance(detoxConfig) {
+    if (!detoxConfig) {
```
noomorph: No try-catch here, it moved upwards.

detox/src/artifacts/ArtifactsManager.js
```diff
@@ -134,10 +139,6 @@ class ArtifactsManager {
     await this._callPlugins('descending', 'onSuiteEnd', suite);
   }
 
-  async onInit() {
```
noomorph: No more `onInit`, because no plugin was actually using it, :open_mouth: mea culpa :blush:

detox/src/artifacts/ArtifactsManager.js
```diff
@@ -111,6 +112,10 @@ class ArtifactsManager {
     await this._callPlugins('plain', 'onLaunchApp', appLaunchInfo);
   }
 
+  async onAppReady(appInfo) {
```
noomorph: This way I can tracking if the app launch was not successful and, for instance, deliver `appNotReady.png` screenshot.

noomorph: Otherwise, if let's say `detox.init()` fails with timeout in the upper `beforeAll`, users will be left scratching their heads with almost empty artifacts folder, which is not something I really wish. That's why I am adding it.

detox/src/artifacts/screenshot/ScreenshotArtifactPlugin.js
```diff
@@ -12,6 +12,15 @@ class ScreenshotArtifactPlugin extends TwoSnapshotsPerTestPlugin {
     return this.api.preparePathForArtifact(`${name}.png`, testSummary);
   }
 
+  async onBeforeCleanup(e) {
+    if (this.context.isAppReady === false) {
+      this._hasFailingTests = true;
```
noomorph: Hacky way, I admit, but this is because I have a not flexible inheritance model with these `super.onBeforeCleanup`.

The idea is to make sure that the unresponsive state of the app is considered a failing test, **even if no test was run**. How come?

Because if the outer `beforeAll` fails, then we don't get no `detox.beforeEach()` and `detox.afterEach()`. No beforeEach-afterEach, ====> no tests. But the run is not successful. So this way I at least make sure that if user runs Detox with `--take-screenshots failing`, we'll still pull that `appNotReady.png` from the device.

If I don't put that flag, we'll just remove that screenshot from the device, destroying the evidence. :smile:

I think, this place can be improved, but can't currently wrap my head around it. Also, I think a better fix would require more heavy refactoring in ArtifactsManager, and I am afraid to touch more. The PR is already a few hundred lines of code.

noomorph: By the way, why `=== false`? Because `isAppReady` can be `undefined, false, true`. If we never ran app and we get inside detox.cleanup(), it is still a legit case. That's why I chose those three states.

detox/src/devices/Device.js
```diff
@@ -86,6 +86,11 @@ class Device {
 
     await this.deviceDriver.waitUntilReady();
     await this.deviceDriver.waitForActive();
+    await this.deviceDriver.emitter.emit('appReady', {
```
noomorph: Cannot put emitting inside that `waitForActive`, because it is used elsewhere (e.g., matchFace, unmatchFace).

I am hesitant to put that in `waitUntilReady` because I suspect there might be a chance something goes awry with `waitForActive`, so I want to emit as late as possible, to not miss a problem with the app.

detox/src/errors/DetoxRuntimeError.js
```diff
@@ -1,14 +1,15 @@
...
+    const formattedMessage = _.compact([
+      message,
+      hint && `HINT: ${hint}`,
+      debugInfo
+    ]).join('\n\n');
 
-  toString() {
-    return super.toString().replace(/^Error: /, '\n');
+    super(formattedMessage);
+    this.name = 'DetoxRuntimeError';
```
noomorph: This is an idiomatically correct way to extend `Error` (using `this.name =`). So, I fixed DetoxRuntimeError implementation.

detox/src/devices/drivers/AndroidDriver.test.js
```diff
@@ -91,8 +91,8 @@ describe('Android driver', () => {
         await promise;
         fail('Expected an error and none was thrown');
       } catch (e) {
-        expect(e.message).toContain('DetoxRuntimeError: Failed to run application on the device');
-        expect(e.message).toContain(`Native stacktrace dump: ${mockInstrumentationLogsParserClass.INSTRUMENTATION_STACKTRACE_MOCK}`);
+        expect(e.toString()).toContain('DetoxRuntimeError: Failed to run application on the device');
```
noomorph: It was incorrect from my side to put `DetoxRuntimeError:` inside Error.prototype.message.
I started seeing at some places `DetoxRuntimeError: DetoxRuntimeError: ...`, so I decided to fix it.

detox/test/e2e/init.js
```diff
@@ -22,7 +22,12 @@ beforeAll(async () => {
 }, 300000);
 
 beforeEach(async () => {
-  await adapter.beforeEach();
+  try {
+    await adapter.beforeEach();
+  } catch (err) {
+    await detox.cleanup();
```
noomorph: This is a trick for `jest-jasmine`, to set to `Detox.none`, when tests run forward while `detox.init()` has not ended well. I could have put that `detox.cleanup()` inside `adapter.beforeEach()`, but my gut feeling tells me that better to leave it exposed to the user. Moreover, this trick is specific to jest-jasmine only.

detox/src/Detox.test.js
noomorph: Ah, right. And then, I am able to understand that init is actually happening till now, so I can reject init promise.

d4vidi:
>In the Detox Next, the test runner sends "ready" right away. So it will not work as you expect. In fact, on iOS, "ready" has no meaning and I am contemplating removing it.

Isn't that a bug? ready should be sent, well.. when the app is ready...

detox/runners/jest/DetoxAdapterCircus.js
d4vidi: So Jest runs the tests nonetheless even with Circus but at also delivers errors as an indication?

detox/src/Detox.js
d4vidi: I understand and like this approach. Nevertheless we now have in this class these 3 members, all related to init:
 - `_isInitialized`
 - `_initPromise`
 - `_rejectInit`
I can't help but think we ought to replace those with a single ref to an external init controller of some sort, to which all of this be delegated, managing and holding all necessary flags and init state indicators. The code here is already entangled as it is.

detox/src/Detox.js
d4vidi: So this is here in order to handle timeouts in the global init, in particular?

detox/src/Detox.test.js
```diff
@@ -95,12 +96,21 @@ describe('Detox', () => {
...
+
   it(`Not passing --cleanup should keep the currently running device up`, async () => {
     Detox = require('./Detox');
     detox = new Detox({deviceConfig: validDeviceConfig});
     await detox.init();
+    const device = detox.device;
     await detox.cleanup();
-    expect(detox.device.shutdown).toHaveBeenCalledTimes(0);
+    expect(device.shutdown).toHaveBeenCalledTimes(0);
   });
```
d4vidi: The additional coverage is good, but I kind'a wish you'd cover the multiple async init rejection techniques.

It would make it both tested and also better documented (for example, if you specify explicitly that before-each should throw if init failed due to a _timeout_, etc.)

detox/src/DetoxExportWrapper.js
d4vidi: Fixes https://github.com/wix/Detox/issues/1914?

d4vidi:
@noomorph Having gone over ~50% of the code changes, I'm starting to recognize that it in fact sports two changes that are only loosely dependant: app-ready hook, and the (great) improvement of init failure handling (alongside the fix of #1914?).

This is too much to take in all at once, both for me as a reviewer, and I think - for the project as a whole. Would you consider breaking the PR into two?

d4vidi:
In addition, with all of the hacking work I see here that involves jasmine -- I think we ought to start considering a deprecation plan (i.e. making circus mandatory through a major version bump). Or is your PR to Jest blocking? Anyhow, IMO it should prove working in our major internal projects first.

detox/src/devices/Device.js
d4vidi: It appears `waitForActive()` is scarce on Android. @rotemmiz why this difference between platforms?

detox/src/devices/drivers/AndroidDriver.js
```diff
@@ -174,8 +174,14 @@ class AndroidDriver extends DeviceDriverBase {
   _getInstrumentationCrashError() {
     return new DetoxRuntimeError({
       message: 'Failed to run application on the device',
-      hint: 'Most likely, your main activity has crashed prematurely',
-      debugInfo: 'Native stacktrace dump: ' + this.instrumentationStackTrace,
+      hint: this.instrumentationStackTrace
+        ? 'Most likely, your main activity has crashed prematurely.'
+        : 'Most likely, your tests have timed out and called detox.cleanup() ' +
+          'while it was waiting for "ready" message (over WebSocket) ' +
+          'from the instrumentation process.',
```
d4vidi: Ouch :)
Could we explain in more simply words to the baffled user?

noomorph:
Splitted into 4 smaller PRs:

#1971
#1970
#1968
#1967

PR #1939, opened 2020-02-27
cleartextTrafficPermitted should be android:usesCleartextTraffic="true"

akiwarheit:
- [Y] This is a small change 

---

Doco update


d4vidi:
So weird... thanks :)

PR #1944, opened 2020-03-05
Introduce unified custom logger to log user events using bunyan

d4vidi:
- [ ] This is a small change 

---

**Description:**
Introduce a unified logging solution that overrides `console` functions implementation. Goal here is twofold:

1. Wire user logs onto bunyan, such that (in addition to its nice optional features) the log would be associated with the `pid` -- as in parallel-exec envs all user logs are currently scarce without it.
2. Pave way for thus bundling user logs into the timeline trace artefact.

#### Implementation
Initially, the plan was to simply grab the logs as-is and just rewire through bunyan. This turned out to be a bit of a dirty but mostly an imperfect solution - especially in Jest (which already provide a custom logger), which logs the log-origin (which gets broken in this approach) and also log in multiple lines and thus ident to start (which by default overrides the bunyan pid prefix).

Hence, eventually, solution was to introduce a fully blown custom logger. The ends result is pretty nice.

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -130,7 +130,7 @@ module.exports.builder = {
     boolean: true,
     default: true,
     group: 'Execution:',
-    describe: `Use Detox' custom console logging implementation. Disabling will fallback to node.js / test-runner's (e.g. Jest / Mocha) implementation.`,
+    describe: `Use Detox' custom console-logging implementation, for logging Detox (non-device) logs. Disabling will fallback to node.js / test-runner's implementation (e.g. Jest / Mocha).`,
   },
 };
 
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -125,7 +125,13 @@ module.exports.builder = {
   'device-launch-args': {
     group: 'Execution:',
     describe: 'Custom arguments to pass (through) onto the device (emulator/simulator) binary when launched.'
-  }
+  },
+  'use-custom-logger': {
+    boolean: true,
+    default: true,
+    group: 'Execution:',
+    describe: `Use Detox' custom console logging implementation. Disabling will fallback to node.js / test-runner's (e.g. Jest / Mocha) implementation.`,
+  },
 };
 
 const collectExtraArgs = require('./utils/collectExtraArgs')(module.exports.builder);
@@ -224,6 +230,7 @@ module.exports.handler = async function test(program) {
         (hasCustomValue('record-performance') ? `--record-performance ${program.recordPerformance}` : ''),
         (program.artifactsLocation ? `--artifacts-location "${program.artifactsLocation}"` : ''),
         (program.deviceName ? `--device-name "${program.deviceName}"` : ''),
+        (program.useCustomLogger ? `--use-custom-logger "${program.useCustomLogger}"` : ''),
       ]),
       ...getPassthroughArguments(),
     ]).join(' ');
@@ -284,6 +291,7 @@ module.exports.handler = async function test(program) {
         'reportSpecs',
         'readOnlyEmu',
         'deviceLaunchArgs',
+        'useCustomLogger'
       ]),
       DETOX_START_TIMESTAMP: Date.now(),
     };
--- a/detox/src/utils/customConsoleLogger.js
+++ b/detox/src/utils/customConsoleLogger.js
@@ -2,19 +2,17 @@ const util = require('util');
 const path = require('path');
 const callsites = require('./callsites');
 
-function callsitesToStack() {
-  return callsites().slice(2).reduce((result, callsite) => {
-    const filename = callsite.getFileName() ? path.relative(process.cwd(), callsite.getFileName()) : '';
-    const filePointer = filename ? `${filename}:${callsite.getLineNumber() || '?'}:${callsite.getColumnNumber() || '?'}` : '<unknown>';
-    const functionName = callsite.getFunctionName() || '<unknown>';
-    return result.concat(`\r    at ${functionName} (${filePointer})\n`)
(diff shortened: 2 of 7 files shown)


detox/package.json
```diff
@@ -41,6 +41,7 @@
     "@babel/core": "^7.4.5",
     "bunyan": "^1.8.12",
     "bunyan-debug-stream": "^1.1.0",
+    "callsites": "^3.1.0",
```
rotemmiz: this package is 5 lines of code, what if you paste them into detox?

d4vidi: done :)

docs/APIRef.DetoxCLI.md
```diff
@@ -83,6 +83,7 @@ Initiating your test suite. <sup>[[1]](#notice-passthrough)</sup>
 | --gpu                                         | [Android Only] Launch Emulator with the specific -gpu [gpu mode] parameter. |
 | --device-launch-args | A list of passthrough-arguments to use when (if) devices (Android emulator / iOS simulator) are launched by Detox.<br />**Note: the value must be specified after an equal size (`=`) and inside quotes.** Usage example:<br />`--device-launch-args="-http-proxy http://1.1.1.1:8000 -no-snapshot-load"` |
 | --no-color                                    | Disable colors in log output |
+| --use-custom-logger | Use Detox' custom console-logging implementation. Disabling will fallback to node.js / test-runner's (e.g. Jest / Mocha) implementation.<br />*Default: true* |
```
LeoNatan: It's important to note that this is only about the Detox node process, not the app process logging.

detox/src/utils/customConsoleLogger.js
```diff
@@ -0,0 +1,56 @@
+const util = require('util');
+const path = require('path');
+const callsites = require('./callsites');
+
+function callsitesToStack() {
+  return callsites().slice(2).reduce((result, callsite) => {
```
noomorph: Why can't we use something like `new Error().stack` for that?

detox/src/utils/customConsoleLogger.js
```diff
@@ -0,0 +1,56 @@
...
+
+function override(consoleLevel, bunyanFn) {
+  console[consoleLevel] = (...args) => {
+    bunyanFn({ event: 'USER_LOG' }, getOrigin(), '\n', util.format(...args));
+  };
+}
+
+function overrideTrace(consoleLevel, bunyanFn) {
+  console[consoleLevel] = (...args) => {
+    bunyanFn({ event: 'USER_LOG' }, getOrigin(), '\n  Trace:', util.format(...args), '\n', callsitesToStack());
```
noomorph: Are you replicating Jest logger behavior here?

detox/src/utils/customConsoleLogger.js
```diff
@@ -0,0 +1,56 @@
...
+  return callsites().slice(2).reduce((result, callsite) => {
+    const filename = callsite.getFileName() ? path.relative(process.cwd(), callsite.getFileName()) : '';
+    const filePointer = filename ? `${filename}:${callsite.getLineNumber() || '?'}:${callsite.getColumnNumber() || '?'}` : '<unknown>';
+    const functionName = callsite.getFunctionName() || '<unknown>';
+    return result.concat(`\r    at ${functionName} (${filePointer})\n`)
+  }, '');
+}
+
+function getOrigin() {
+  const callsite = callsites()[2];
```
noomorph: What is this magic number 2, can we name it?

detox/local-cli/test.js
```diff
@@ -125,7 +125,13 @@ module.exports.builder = {
   'device-launch-args': {
     group: 'Execution:',
     describe: 'Custom arguments to pass (through) onto the device (emulator/simulator) binary when launched.'
-  }
+  },
+  'use-custom-logger': {
```
noomorph: @d4vidi , I would rather invert logic, e.g.: `--no-console-patch`.

noomorph:
The idea itself is excellent, kudos, @d4vidi! For quite a long time I was considering bringing this functionality.

detox/local-cli/test.js
d4vidi: I was actually going back and forth with this, I'll revisit

detox/src/utils/customConsoleLogger.js
d4vidi: it's because this is handled two-levels deep compared to the frame with the original user's call. I will name this indeed, you're right!

detox/src/utils/customConsoleLogger.js
d4vidi: Almost. It's similar but different.

detox/src/utils/customConsoleLogger.js
d4vidi: This in essence what callsites does!

noomorph: I was referring to a complex error string formatting logic, this is not what callsites does.

d4vidi: ok, you got me - I got too excited about using callsites so I abused it a bit. 'Changed it now as per your suggestion :D

detox/local-cli/test.js
d4vidi: 'decided not to change seeing the a `no` prefix is something mocha omits and recognizes as a false. So for example, a `--no-bugs` will end up being passed in as `--bugs false` by mocha. All in all, that differs from Jest and I preferred having them unified. And yes, I'm pretty sure the current `--no-color` arg is broken.

d4vidi:
@noomorph merging for now - will reiterate if you have any additional concerns

PR #1954, opened 2020-03-11
Create prebuilt Detox framework

LeoNatan:
Overview of changes:

- Switched to Swift 5 for ABI stability
- Require Xcode 10.2, iOS 12.2 minimum so that Swift 5 exists on the platform
- iOS artifacts will no longer be packed during prepublish or in the main CI script
  - Instead, there is a new script `package:ios`, which will be called in a dedicated, child CI build
- During postinstall, instead of building the framework from the bundled sources, the framework will be extracted directly from the bundled prebuilt framework tbz
  - In dev mode, the old way remains as is; Detox framework is built from the sources
  - No changes to loading mechanism inside Detox because the prebuilt framework is extracted to the same location as before, under ~/Library/Detox/ios

After this is merged, I will create a new child task in Jenkins, set it to collect the two artifacts (source and prebuilt framework tbz files), and finally, during release, the artifacts will be used.


detox/scripts/build_universal_framework.sh
```diff
@@ -19,7 +21,7 @@ echo TEMP_DIR = "${TEMP_DIR}"
 
 # Step 0. Xcode version
 
-XCODEVERSION=$(xcodebuild -version | grep -oEi "([0-9]*\.[0-9]*)")
```
rotemmiz: 🤦‍♂️

LeoNatan: ?
Bad regex is bad anywhere.

scripts/ci.publish.js
```diff
@@ -40,11 +40,7 @@ function projectSetup() {
 function prePublishToNpm() {
   logSection('Prepublish');
 
-  log('Gathering up iOS artifacts...');
-  process.chdir('detox');
-  const {packageIosSources} = require('../detox/scripts/pack_ios');
-  packageIosSources();
-  process.chdir('..');
+  //No longer need to gather up iOS artifacts here, as those are created in a child build on CI.
 }
```
d4vidi: well if there's nothing left here, can it be removed?

LeoNatan: I don't know who calls it and didn't want to investigate. All of this will change soon with Next anyway.

d4vidi: Not sure what you mean. This is an internal func used only once here in the script...

LeoNatan: Cool, removed.

PR #1966, opened 2020-03-24
fix(windows): unescaped regexp issue, unit tests

noomorph:
- [x] This is a small change 

---

**Description:**

1. Fixed implementation of `callsites.stackdump` - added RegExp escaping. Otherwise, I see issue on Windows 10:

```
  ● callsites › stack-dump › should slice according to end-frame arg

    SyntaxError: Invalid regular expression: /C:\Users\noomo\Projects\wix\Detox\detox\/: \ at end of pattern
        at new RegExp (<anonymous>)

      16 |     .slice(2 + endFrame) // 2 = 1 for 'Error:' prefix and 1 for this function's frame
      17 |     .join('\n')
    > 18 |     .replace(new RegExp(cwd, 'g'), '');
         |              ^
      19 | };
      20 |

      at Function.stackdump (src/utils/callsites.js:18:14)
      at callStackDumpFromWrapperFn (src/utils/callsites.test.js:23:64)
      at callStackDumpFromTwoWrapperFn (src/utils/callsites.test.js:24:57)
      at Object.<anonymous> (src/utils/callsites.test.js:38:25)
```

Most likely, it is apt to occur in a real situation as well, not unit tests only.

For the solution to replace all, I use the suggestion option: https://stackoverflow.com/a/1145525
I think it is not a performance-critical code, because it is not invoked hundreds of times, right? Detox calls this function only to facilitate debugging when something goes wrong, I assume.

2. Fixed unit tests so they pass both on POSIX and Win32 platforms.

### DIFF
--- a/detox/src/utils/callsites.js
+++ b/detox/src/utils/callsites.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const {escapeRegExp} = require('lodash');
 const cwd = process.cwd() + path.sep;
 
 // Taken from https://github.com/sindresorhus/callsites
@@ -15,6 +16,5 @@ module.exports.stackdump = (endFrame = 0) => {
     .split('\n')
     .slice(2 + endFrame) // 2 = 1 for 'Error:' prefix and 1 for this function's frame
     .join('\n')
-    .split(cwd)
-    .join('');
+    .replace(new RegExp(escapeRegExp(cwd), 'g'), '');
 };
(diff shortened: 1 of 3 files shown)


d4vidi:
Kudos for testing on Windows _and_ getting this fixed!
As for the path replacement solution... well, I'll approve if we have to but really, it's not very elegant, performance-wise, to say the least...

PR #1968, opened 2020-03-24
fix: handling detox.init() timeout

noomorph:
This pull request is extracted out of https://github.com/wix/Detox/pull/1929

- [x] This change has been discussed in issue #1914 and the solution has been agreed upon with maintainers.

---

**Description:**

1. Fixes `detox.cleanup()`, so it can access a pending Detox instance in the midst of `detox.init()` and actually clean it up.
2. Jest-Circus: prevents false executions of `detox.beforeEach` and `detox.afterEach` in the case when `beforeAll` hook failed with a timeout. The problem is, that even after an error in `beforeAll` hook we are getting `_onTestStart` and `_onTestDone` methods of `DetoxAdapterCircus.js` called, albeit the tests' body is surprisingly not executed, just receiving events where `testStart` is called with a test with `errors.length > 0` already.
3. Jest-Jasmine: `detox.beforeEach()` will throw an error on attempt to call it while `detox.init()` is happening. This way we are working around Jest-Jasmine bug when tests proceed 😲 after timeouts, no matter what.
4. Mocha: tested on Mon, Apr 6, 2020, worked well and in an expected way.


detox/src/Detox.js
```diff
@@ -32,75 +32,44 @@ class Detox {
...
+    this._artifactsConfig = artifactsConfig;
     this._deviceConfig = deviceConfig;
     this._userSession = deviceConfig.session || session;
     this._client = null;
     this._server = null;
-    this._artifactsManager = new ArtifactsManager(artifactsConfig);
+    this._artifactsManager = null;
+    this._isInitialized = false;
+    this._initPromise = null;
+    this._rejectInit = _.noop;
```
d4vidi: (Quoting my previous notes)

I understand and like this approach. Nevertheless we now have in this class these 3 members, all related to init:
 - _isInitialized
 - _initPromise
 - _rejectInit

I can't help but think we ought to replace those with a single ref to an external init controller of some sort, to which all of this be delegated, managing and holding all necessary flags and init state indicators. The code here is already entangled as it is :-/

detox/src/Detox.js
```diff
@@ -32,75 +32,44 @@ class Detox {
...
-
-    await this._artifactsManager.onInit();
-
-    return this;
+    return this._initPromise;
   }
 
   async cleanup() {
-    await this._artifactsManager.onBeforeCleanup();
+    await this._assertNoPendingInit().catch(_.noop);
```
d4vidi: really wish you'd cover this logic using unit tests

detox/src/Detox.js
```diff
@@ -127,6 +101,8 @@ class Detox {
   }
 
   async afterEach(testSummary) {
+    await this._assertNoPendingInit();
```
d4vidi: c&p: really wish you'd cover this logic using unit tests

detox/src/Detox.js
```diff
@@ -109,14 +78,19 @@ class Detox {
...
 
     if (argparse.getArgValue('cleanup') && this.device) {
       await this.device.shutdown();
     }
+
+    this.device = null;
   }
 
   async beforeEach(testSummary) {
+    await this._assertNoPendingInit();
```
d4vidi: c&p: really wish you'd cover this logic using unit tests

detox/src/DetoxExportWrapper.js
```diff
@@ -35,7 +34,18 @@ class DetoxExportWrapper {
       Detox.none.initContext(global);
     }
 
-    this[_detox] = await DetoxExportWrapper._initializeInstance(config, params);
+    Detox.none.setError(null);
+
+    try {
+      this[_detox] = DetoxExportWrapper._createInstance(config);
+      await this[_detox].init(params);
```
d4vidi: It took a while for me to understand why this is indeed necessary but I got there eventually 😄 
Maybe adding some notes here would help for future reference! As you mentioned, this change is very important and its effect is two-fold (timeouts - handle screenshots, jasmine hack).

d4vidi: Could we manifest this behavior using an E2E test?

detox/src/devices/drivers/AndroidDriver.js
```diff
@@ -174,8 +174,14 @@ class AndroidDriver extends DeviceDriverBase {
...
     return new DetoxRuntimeError({
       message: 'Failed to run application on the device',
-      hint: 'Most likely, your main activity has crashed prematurely',
-      debugInfo: 'Native stacktrace dump: ' + this.instrumentationStackTrace,
+      hint: this.instrumentationStackTrace
+        ? 'Most likely, your main activity has crashed prematurely.'
+        : 'Most likely, your tests have timed out and called detox.cleanup() ' +
+          'while it was waiting for "ready" message (over WebSocket) ' +
+          'from the instrumentation process.',
+      debugInfo: this.instrumentationStackTrace
```
d4vidi: yes! thank you 😸

noomorph: For clarity, do we need any improvement here?

detox/src/utils/Deferred.test.js
```diff
@@ -0,0 +1,75 @@
...
+  describe('static properties', () => {
+    it('should have "pending" status', () =>
+      expect(Deferred.PENDING).toBe('pending'));
+
+    it('should have "resolved" status', () =>
+      expect(Deferred.RESOLVED).toBe('resolved'));
+
+    it('should have "rejected" status', () =>
+      expect(Deferred.REJECTED).toBe('rejected'));
+  });
```
d4vidi: Why is this needed?

detox/src/Detox.js
```diff
@@ -23,6 +24,9 @@ const DEVICE_CLASSES = {
   'android.attached': AttachedAndroidDriver,
 };
 
+const _initHandle = Symbol('_initHandle');
+const _assertNoPendingInit = Symbol('_assertNoPendingInit');
```
d4vidi: What's the benefit of using a symbol here? 🤔

detox/src/Detox.js
```diff
@@ -168,20 +168,25 @@ class Detox {
...
     return this;
   }
 
-  async _assertNoPendingInit() {
-    if (!this._isInitialized) {
-      this._rejectInit(new DetoxRuntimeError({
-        message: 'Aborted detox.init() execution, and now running detox.cleanup()',
-        hint: 'Most likely, your test runner is tearing down the suite due to the timeout error',
-      }));
+  [_assertNoPendingInit]() {
```
d4vidi: I mean, the `Deffer` solution is good but I was hoping for a richer construct that could provide an implementation for this. wdyt?

detox/src/Detox.test.js
```diff
@@ -105,6 +105,22 @@ describe('Detox', () => {
...
+    await expect(detoxInitPromise).rejects.toThrowError(/Aborted detox.init/);
+  });
+
+  it(`Calling detox.afterEach() before detox.init() completes, throws error`, async () => {
+    Detox = require('./Detox');
+    detox = new Detox({deviceConfig: validDeviceConfig});
+    const detoxInitPromise = detox.init();
+    await expect(detox.afterEach(testSummaries.failed())).rejects.toThrowError(/Aborted detox.init/);
+    await expect(detoxInitPromise).rejects.toThrowError(/Aborted detox.init/);
+  });
```
d4vidi: ✅

detox/src/utils/Deferred.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+        it('should still have that rejected status', () =>
+          expect(deferred.status).toBe(Deferred.REJECTED));
+
+        it('should still have that rejected promise', () =>
+          expect(deferred.promise).rejects.toThrowError('RejectionTest'));
+      });
+    });
+  });
+});
```
d4vidi: amaze!

detox/src/Detox.js
noomorph: Actually I want to hide these and not only these properties from the users of Detox, so there is zero chance they use unless they explicitly ask us in Issues and we consider that.

Since user can get a Detox instance through `const detoxInstance = await detox.init();`, I think this is quite important to not expose more than we really want to.

detox/src/Detox.js
noomorph: @d4vidi, sorry, but your suggestion is very foggy and blurry, I can't really understand what you mean. Could you elaborate more?

detox/src/utils/Deferred.test.js
noomorph: Why not? Does it break some principles of good design or something? I just prefer to explicitly test existence of those three static properties, to not mix that concern into other tests.

detox/src/Detox.js
noomorph: E.g. I would hide also `_client`, `_server`, `_artifactsManager` from the external access.

detox/src/Detox.js
d4vidi: Sorry, I'll try to shed more light! 😄 ❤️ 
Actually, `Deferred` _is_ sort of what I meant. However, I was suggesting a class/object that would handle all init-status related concerns (i.e. an init-controller). That would include `assertNoPendingInit` in its API. As it holds all init-related state, it'd be easier to implement:
```
if (state === this.PENDING) { // i.e. not Deferred.PENDING
   throw new DetoxRuntimeError(...);
}
```

and in Detox:
```
this.initController.assertNoPendingInit()
```

That would also dismiss the need for most of the currently public API in `Deferred`, such as `status` and `promise`, I believe. Although, you _can_ keep `Deferred` and have it used by the `"InitController"` since `Deferred` seems rather useful in its own...

d4vidi: But anyways, I'm OK with whatever you decided, shouldn't block you from moving fwd.

detox/src/Detox.js
noomorph: Covered with test: `Calling detox.afterEach() before detox.init() completes, throws error`

detox/src/Detox.js
noomorph: Covered with test: `Calling detox.afterEach() before detox.init() completes, throws error`

detox/src/Detox.js
noomorph: Not relevant, as the conversation has already moved to `Deferred` discussion.

detox/src/Detox.js
noomorph: Covered, thanks.

detox/src/devices/Device.js
```diff
@@ -270,6 +270,7 @@ class Device {
 
   async _cleanup() {
     await this.deviceDriver.cleanup(this._deviceId, this._bundleId);
+    this.deviceDriver = null;
```
noomorph, line 273: Added nullifying to be on the safe side after cleanup starts.

detox/src/DetoxExportWrapper.js
noomorph: Added minimal e2e tests for Jasmine and Circus in the last commits, though they don't specifically cover 2 from the point of the assertions.

detox/src/Detox.js
noomorph: I have investigated the situation, and it seems that extracting some `DetoxInitController` would result in a considerable refactoring worth a separate PR. I think we indeed should pay attention to the bloating of `Detox.js` and start cleaning it up, but just not in this PR.

PR #1971, opened 2020-03-24
feat(artifacts): take appNotReady.png screenshot if needed

noomorph:
- [x] This is a small change 

---

**Description:**

Supplements #1914 . Will be taking an automatic screenshot `appNotReady.png`, when there was an issue with launching the app, when it never signalled back that it is ready and active.

@d4vidi, here the question is mostly whether `device.driver.emitter` makes you smell _Feature envy_, one of the code smells, or it is fine?

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -86,6 +86,11 @@ class Device {
 
     await this.deviceDriver.waitUntilReady();
     await this.deviceDriver.waitForActive();
+    await this.deviceDriver.emitter.emit('appReady', {
+      deviceId: this._deviceId,
+      bundleId: _bundleId,
+      pid: processId,
+    });
 
     if(params.detoxUserNotificationDataURL) {
       await this.deviceDriver.cleanupRandomDirectory(params.detoxUserNotificationDataURL);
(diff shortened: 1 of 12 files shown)


detox/src/devices/Device.js
```diff
@@ -86,6 +86,11 @@ class Device {
 
     await this.deviceDriver.waitUntilReady();
     await this.deviceDriver.waitForActive();
+    await this.deviceDriver.emitter.emit('appReady', {
```
d4vidi: why not tested in device unit tests?

d4vidi:
>here the question is mostly whether device.driver.emitter makes you smell Feature envy
😄 
TBH it's weird for me to begin with that the emitter is a strict property of the driver. Feels like it should be an object shared among the designated emitters. You know what I mean?

detox/src/devices/Device.js
noomorph: added a test, thanks.

noomorph:
I have rewritten emitters code according to what you suggested earlier, so, now we have a single emitter instance per one Detox, and it is passed down to devices and device drivers.

PR #1974, opened 2020-03-25
Fixed output of build_framework.ios.sh on failures

silyevsk:
- [X] This is a small change 

---

**Description:**
This is a fix for missing logs on `build_universal_framework.sh` failures. (`build_framework.ios.sh` runs with `bash -e` which means it fails right away on `build_universal_framework.sh` failure, so no log is printed)


LeoNatan:
Thanks
The `-e` was a recent addition and it didn't account for that.

PR #1980, opened 2020-03-30
feat(config): configure which screenshots to take

noomorph:
**Action:**

I decided to add the `takeWhen` config for `screenshot` plugin in package.json for those who explicitly want to configure which automatic screenshots to take, and which not:

```json
  "detox": {
    "artifacts": {
      "plugins": {
        "screenshot": {
          "takeWhen": {
            "testDone": true
          }
        }
      }
    }
  }
```

**Motivation:**

1.  `appNotReady.png` screenshot (https://github.com/wix/Detox/pull/1971) is hopefully going to be one of the first we will be adding to the plugin.
2. a `testStart` screenshot of the current test is usually identical to a `testDone` of the previous test.


d4vidi:
@noomorph kudos. Can you also put this on Detox docs wherever seems apt? Among other things -- what's the default value for each field?

noomorph:
@d4vidi , please check #1998.

PR #1982, opened 2020-03-30
Limit emu's skin config patching to <28

d4vidi:
- [x] This is a small change 
- [ ] This change has been discussed in issue #x and the solution has been agreed upon with maintainers.

---

**Description:**

Restricts the emulator skin configuration patching discussed in #1157 to old emulators (version < 28), as -- according to some experimentation, emulator binaries seem to work without it in version 28 and newer.

On internal build platforms, we've seen this error happening, but not as described in #1157 - as it takes place in env's running tests using parallel-execution.

The fix accounts for both cases, assuming the patch-work is no longer needed.

### DIFF
--- a/detox/src/devices/drivers/EmulatorDriver.js
+++ b/detox/src/devices/drivers/EmulatorDriver.js
@@ -18,6 +18,8 @@ const DetoxEmulatorsPortRange = {
 };
 
 const ACQUIRE_DEVICE_EV = 'ACQUIRE_DEVICE';
+const EMU_BIN_STABLE_SKIN_VER = 28;
+const EMU_BIN_VERSION_DETECT_EV = 'EMU_BIN_VERSION_DETECT';
 
 class EmulatorDriver extends AndroidDriver {
   constructor(config) {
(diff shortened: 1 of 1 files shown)


Overtorment:
@d4vidi when I update from 16.0.1 to 16.0.2 my tests on travisci break with error

```
""/home/travis/android-sdk/emulator/emulator" -version" failed with code = 127, stdout and stderr:
detox[11585] ERROR: [exec.js/EXEC_FAIL, #1] 
detox[11585] ERROR: [exec.js/EXEC_FAIL, #1] /home/travis/android-sdk/emulator/qemu/linux-x86_64/qemu-system-x86_64: error while loading shared libraries: libXcursor.so.1: cannot open shared object file: No such file or directory
```

it worked fine on 16.0.0. and 16.0.1
here's the full log and travis config: https://travis-ci.com/github/BlueWallet/BlueWallet/builds/157474607

do you want me to open another ticket for that or, maybe, I'm doing something wrong..?

Overtorment:
PS. came here from https://github.com/wix/Detox/issues/1157

d4vidi:
@Overtorment :sigh:
I'd appreciate a separate report. Oh travis...

PR #2007, opened 2020-04-16
Handle logging in async function

rynop:
- [x] This is a small change 
---

**Description:**

Sorry I didn't file an issue cuz this was such a small change. 

When using mocha, and trying to log within a hook/test `detox/src/utils/customConsoleLogger.js:: getOrigin()` blows up because `userCallsite.getFileName()` is `undefined` and therefore `path.relative()` explodes.

Simple reproduce:
```
  before: async () => {
    console.log('**********BLAH');
```

PR is not the most elegant solution, but not sure there is much that can be done when using https://github.com/sindresorhus/callsites

Without this fix, I can not use Detox (and I want to, cuz really like it :)

### DIFF
--- a/detox/src/utils/customConsoleLogger.js
+++ b/detox/src/utils/customConsoleLogger.js
@@ -11,7 +11,7 @@ function getStackDump() {
 
 function getOrigin() {
   const userCallsite = callsites()[USER_STACK_FRAME_INDEX];
-  const filename = path.relative(process.cwd(), userCallsite.getFileName());
+  const filename = userCallsite && userCallsite.getFileName() ? path.relative(process.cwd(), userCallsite.getFileName()) : '?';
   return `at ${filename}:${userCallsite.getLineNumber() || '?'}:${userCallsite.getColumnNumber() || '?'}`;
 }
 
--- a/detox/src/utils/customConsoleLogger.js
+++ b/detox/src/utils/customConsoleLogger.js
@@ -12,8 +12,10 @@ function getStackDump() {
 function getOrigin() {
   const userCallsite = callsites()[USER_STACK_FRAME_INDEX];
   const callsiteFilename = userCallsite && userCallsite.getFileName();
+  const callsiteLine = userCallsite && userCallsite.getLineNumber();
+  const callsiteCol = userCallsite && userCallsite.getColumnNumber();
   const filename = callsiteFilename ? path.relative(process.cwd(), callsiteFilename) : '<unknown>';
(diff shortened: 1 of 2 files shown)


detox/src/utils/customConsoleLogger.js
```diff
@@ -11,7 +11,7 @@ function getStackDump() {
 
 function getOrigin() {
   const userCallsite = callsites()[USER_STACK_FRAME_INDEX];
-  const filename = path.relative(process.cwd(), userCallsite.getFileName());
+  const filename = userCallsite && userCallsite.getFileName() ? path.relative(process.cwd(), userCallsite.getFileName()) : '?';
```
noomorph, line 14: Could you split this line into two with
```js
const callsiteFilename = userCallsite && userCallsite.getFileName();`
```
beforehand?

d4vidi:
I must emphasize what @noomorph is asking regarding tests. This is TDD code -- need to add a failing test that acknowledges this specific case.

Also, please don't use `?` for the file name, but an `<unknown>` notation.

Other than that: thank you for this! 😄

detox/src/utils/customConsoleLogger.js
rynop: For sure, NP.

rynop:
Fixed up and test added!

detox/src/utils/customConsoleLogger.js
```diff
@@ -11,7 +11,8 @@ function getStackDump() {
 
 function getOrigin() {
   const userCallsite = callsites()[USER_STACK_FRAME_INDEX];
-  const filename = path.relative(process.cwd(), userCallsite.getFileName());
+  const callsiteFilename = userCallsite && userCallsite.getFileName();
+  const filename = callsiteFilename ? path.relative(process.cwd(), callsiteFilename) : '<unknown>';
   return `at ${filename}:${userCallsite.getLineNumber() || '?'}:${userCallsite.getColumnNumber() || '?'}`;
```
noomorph, line 16: I am afraid that in case when userCallsite is indeed falsy, the attempt to get line and column number will fail as well. If you have started already guarding against this edge case, make sure to make it consistent here too.

rynop: It didn't in my use case (I was surprised as well). I'll add logic tho.

Leaving test case as is cuz the line/col is obv non-deterministic (given my use case findings).  The update code will handle both scenarios.

noomorph:
@rynop, I am sorry for the delay - the last update from you somehow slipped between the cracks, and I have not noticed it. I'll review your PR soon.

rynop:
@noomorph no worries!

noomorph:
@rynop, thank you! Great job! :+1:

PR #2009, opened 2020-04-18
feat(runner): new integration with jest-circus

### LINKED ISSUE #1661: Take screenshot after test failure, before afterEach is executed
It is common to put additional logic in `afterEach` step (to return to primary state, make additional repetitive checks, etc). 
With current implementation, when the test fails, `afterEach` is being executed and only THEN screenshot is taken. It means that other screen could be opened already, so screen where bug happened is not visible anymore. This way it is harder to analyze bug failures, as you do not have corresponding screenshot.

*Please* implement taking screenshot at the start of afterEach statement.

noomorph:
New projects by default will be using a new test runner integration between Detox and `jest-circus`.

## Motivation

1. Initiate Detox as early as possible, even before we know what tests we are going to run. That potentially enables discarding `--testNameRegexPattern=(^(^()^())` workaround for Jest, as in this position we will be able to decide in runtime which tests we want to skip, and which - not. Also, the initialization errors won't be bound to some specific test - less clutter and misleading information for users.

1. Compared to previous `jest-circus` integration, we remove that a bit alien `global.detoxCircus` environment variable. Now, registering the listeners happens right in a user-defined environment class which derives from `DetoxCircusEnvironment`. On the other hand, we add a global `detox` variable, because of the very same sandboxing issues (to avoid creating many isolated Detox instances with every next `require('detox')` - see `detox/src/index.js` for the actual place).

1. Together with removing `detoxCircus` variable, and moving concerns into environment class, at the end of the day, we gain freedom from `./init.js` environment setup files. We don't depend on user-defined hooks (`beforeAll`, `beforeEach`, `afterEach`, `afterAll`) - now we run above that. That's why the new DetoxCircusEnvironmnent has less complexity because it does not mediate between things happening in user-facing code (`setupFilesAfterEnv: ['init.js']`) and jest-circus' `handleTestEvent` events.

1. We used to run `afterEach` lifecycle actions as a first thing in the consequent test's `beforeEach` (or `detox.cleanup()` if there were no next tests). That led sometimes to issues when failures in `detox.afterEach()` (hence, incomplete execution of all planned routines) of the previous test could impact the behavior of `detox.beforeEach()` of the current one.

1. It resolves #1661 - artifacts plugins become able to react ASAP to specific failure inside tests, before any hook (e.g. `afterEach`) executes. That should improve the relevance of post-error screenshots, as one of the consequences.

## Description

Instead of a well-known `e2e/init.js` file, you are going to have `e2e/environment.js` file.

```js
const {
  DetoxCircusEnvironment,
  SpecReporter,
  WorkerAssignReporter,
} = require('detox/runners/jest-circus');

class CustomDetoxEnvironment extends DetoxCircusEnvironment {
  constructor(config) {
    super(config);

    // Can be safely removed, if you are content with the default value (=300000ms)
    this.initTimeout = 300000;

    // This takes care of generating status logs on a per-spec basis. By default, Jest only reports at file-level.
    // This is strictly optional.
    this.registerListeners({
      SpecReporter,
      WorkerAssignReporter,
    });
  }
}

module.exports = CustomDetoxEnvironment;
```

The entire class is optional if you don't plan to use the customizations above.

Jest's `e2e/config.json` has the following changes:

```diff
 {
-  "setupFilesAfterEnv": ["./init.js"],
-  "testEnvironment": "node",
+  "testEnvironment": "./environment",
+  "testRunner": "jest-circus/runner",
+  "testTimeout": 120000,
   "reporters": ["detox/runners/jest/streamlineReporter"],
   "verbose": true
 }
```

#### Miscallenous

CI scripts are targeting `jest-circus` by default. `jest-jasmine` is tested only for the edge case with `detox.init()` timeout.


d4vidi:
You are my hero

detox/local-cli/templates/jest.js
```diff
@@ -12,32 +13,21 @@ const config = require('../package.json').detox;
...
-    // Workaround for the 'jest-jasmine' runner (default one): if 'beforeAll' hook above fails with a timeout,
-    // unfortunately, 'jest' might continue running other hooks and test suites. To prevent that behavior,
-    // adapter.beforeEach() will throw if detox.init() is still running; that allows us to run detox.cleanup()
-    // in that emergency case and disable calling 'device', 'element', 'expect', 'by' and other Detox globals.
-    // If you switch to 'jest-circus' runner, you can omit this try-catch workaround at all.
-
-    await detox.cleanup();
-    throw err;
-  }
+  await adapter.beforeEach();
```
d4vidi, line 30: This is awesome!!!

detox/runners/jest/DetoxAdapterJasmine.js
```diff
@@ -1,9 +1,15 @@
...
+const log = require('../../src/utils/logger').child({ __filename });
 
 class DetoxAdapterJasmine /* extends JasmineReporter */ {
   constructor(detox) {
     this._adapter = new DetoxAdapter(detox, DetoxAdapterJasmine._describeInitError);
+    this._printDeprecationWarning = _.once(() => {
+      log.warn(
+        '"jest-jasmine2" support is going to be dropped soon.\n' +
+        'Please follow the migration guide: https://github.com/wix/Detox/blob/master/docs/Guide.Migration.md#1630');
+    });
```
d4vidi, line 12: I have to ask - is this really necessary? At what intermediate phase will we still have jasmine (pre-deprecation) Will it not be alright to kill it and release a major version?

detox/local-cli/templates/jest.js
d4vidi: Also, if this is the case - perhaps we should consider simplifying the API for getting the `env`, or whichever proxy-object we decide proper here in order to add a listener? (the `detoxCircus.getEnv()` shape of an API was just a means to stay coherent with jasmine's `jasmine.getEnv()`...)

d4vidi:
@noomorph sorry for repeating myself, but really - do we really need to keep jasmine as 'legacy'? Can we not just deprecate in a major version and move on?

noomorph:
@d4vidi, I really really want to rewrite adapter in a very breaking manner, so I would go an extra mile postponing a major release to not breakchange two times in a row, with jasmine and with adapters.

detox/local-cli/templates/jest.js
noomorph: Nice idea, let's discuss that soon.

noomorph: Done.

detox/runners/jest/DetoxAdapterJasmine.js
noomorph: I think that the safest bet is to code-freeze old Jest integrations and gradually migrate to the new DetoxCircusEnvironment. After 1-2 months, we'll release a major version that drops the old Jest integrations.

rotemmiz:
@noomorph I still see there is no documentation re the huge change in #2027, just making sure its going to be a part of this PR.

noomorph:
@rotemmiz, that's right, because #2027 was being treated as a part of this one. I will be working on the docs today. I am not bananas not to supply such PRs without docs, hehe, don't worry.

I'll ask you to review user-facing places on Sunday.

rotemmiz:
This is so important, we should have done this kind of thing a long time ago, tapping in at the configuration space instead of the user space.
Can you please clarify regarding point no 2, why would this type of setup disregard the module cache?

d4vidi:
side note -
>Initiate Detox as early as possible, even before we know what tests we are going to run.

That would also enable a fix for #2097, as the device's platform would already be known before `it` calls register tests

wlindner:
With this new integration with jest-circus. How should I launch the app with the permissions I need?

I got my tests working again by overriding the `initDetox` function in my `CustomDetoxEnvironment` like this:

```
  async initDetox() {
    const instance = await this.detox.init(undefined, {launchApp: false});
    await instance.device.launchApp({permissions: {notifications: 'YES'}});
    return instance;
  }
```

Is this the best way to define permissions for my device?

noomorph:
@wlindner, I have what to say. I'll answer tomorrow.

noomorph:
@wlindner, at the moment, not much time has passed since this feature entered production, hence I can't say for sure what the best method is.

## Another approach

In one of our projects we use another approach since the initialization phase is more complicated, e.g.: starting a mock server and some CI preparations.

### detox.config.js

```js
module.exports = {
  // ...
  behavior: {
    init: { launchApp: false },
  },
  // ...
};
```

### init.js

```js
beforeAll(async () => {
  // launch mock server and other stuff
  await device.launchApp({permissions: { /* ... */ }});
});
``` 

### jest.config.js

```diff
module.exports = {
...
+  setupFilesAfterEnv: ["./init.js"],
...
};
``` 

## Pros and cons 

### initDetox()

**Pros.** If you keep all your initialization steps in the protected `initDetox`, then in case if the app cannot start correctly, the failure phase is going to be much shorter and cleaner - for example, at the moment, Detox simply skips all the tests and promptly exits with a failure.

**Cons.**  Any potential timeout errors of `initDetox()` are handled by handwritten wrapper functions (bypassing Jest timeout mechanisms), and there's a chance you might have sometimes subpar experience in unstudied aspects (use with sourcemaps, etc). Also, although there is no intention to break `initDetox(): Promise<Detox>` signature, I cannot give guarantees either that its signature won't undergo some changes later.

### beforeAll

**Pros.** The `beforeAll` hook approach looks a bit cleaner and totally in line with Jest and Detox approach, and less dependent on our recent semi-documented `initDetox` and `cleanupDetox` virtual methods.

**Cons.** Most likely, if you have `SpecReporter` enabled, you'll have a bit more messy and verbose output if there is a failure in the very beginning. If your setup is very simple, you might feel that it is redundant to create one more file and add it to `setupFilesAfterEnv`, but that's a purely aesthetical reason with no practical reasoning underneath.

## Conclusion

Try out `beforeAll` approach, if this is acceptable for you.

If you nevertheless prefer overriding the `initDetox` protected method, your code excerpt looks valid to me, hence you are free to do that either.

PR #2013, opened 2020-04-21
Remove iOS only flag for workers flag

yedidyak:
- [x] This is a small change 

---

**Description:**
Remove iOS only flag for workers flag


rotemmiz:
Thanks @yedidyak

d4vidi:
Lol can't believe this stayed out of date for so long. Thanks!

Issue #1857, opened 2020-01-14
[WIP] Research jammed adb commands (aka adb jam)


d4vidi:
On Detox runs on CI (both OSS and internal projects') we occasionally see that `adb` commands get stuck "forever", midway through runs of severals suites on the available devices (parallel-execution). When that happens, the running test becomes frozen and eventually fails due to the global test time-out definition. Most often we see that the app install itself is the one that hangs, but that's not consistent.

From some discussions with associated devs at Google, the main culprit here is `adb`. We've been given by them tools to investigate that, and we need to apply them for the sake of research and hopefully, help them fix whichever issues we find, thus substantially decrease stability issues.

d4vidi:
Fortunately, we haven't seen this happening in a long time -- esp. since we switched to more powerful macs on our CI (i.e. since buildkite).

PR #2019, opened 2020-04-26
Adb install wtf

d4vidi:
- [ ] This is a small change 
- [ ] This change is associated with issue #1857 and the solution has been agreed upon with maintainers.

---

**Description:**
Hopefully mitigate issues we've been having with ADB, namely the problem previously coined as  _adb jam_ (see #1857). The purposed strategy is - at least for google emulators, to get around usage of `adb install` in particular<sup>*</sup>, and rather push the binaries onto the device and install them "manually" using the package manager.

> _* we don't know for sure the problem's domain is restricted to `adb install`, but it does seem that - at least for the majority of cases, the problem resides primarily there._

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -132,6 +132,12 @@ module.exports.builder = {
     group: 'Execution:',
     describe: `Use Detox' custom console-logging implementation, for logging Detox (non-device) logs. Disabling will fallback to node.js / test-runner's implementation (e.g. Jest / Mocha).`,
   },
+  'force-adb-install': {
+    boolean: true,
+    default: false,
+    group: 'Execution:',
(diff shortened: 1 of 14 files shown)


netshade:
Because this path is now using `pm install` via the `adb shell`, this now causes an issue where, for devices, *all* runtime permissions are accepted ( see the presence of the `-r` flag here https://github.com/wix/Detox/blob/master/detox/src/devices/drivers/android/tools/ADB.js#L92-L96 ) , whereas for emulators, it seems like no runtime permission affordances are granted  ( https://github.com/wix/Detox/blob/master/detox/src/devices/drivers/android/tools/ADB.js#L107-L109 ).  It seems like the prior behavior is what would be desirable?

d4vidi:
@netshade you're probably right - and I will revisit this. We were aware of that this fix might be tricky and so for now you can pass in the `--force-adb-install true` flag to Detox in order to go back to the old impl.

PR #2021, opened 2020-04-27
config(docs): rm outdated parallelization note

freewil:
- [x] This is a small change 

---

**Description:**
detox supports `--workers` flag to enable parallelization with jest as noted in CLI docs: https://github.com/wix/Detox/blob/master/docs/APIRef.DetoxCLI.md


rotemmiz:
Thanks!

PR #2025, opened 2020-04-29
fix: use bugfixed telnet-client version

noomorph:
- [x] This is a small change 

---

**Description:**

I see CI build stuck with telnet connecting for 1 hour in TeamCity.

Checked source code of the former telnet-client version and it appears...

...they don't throw any error on timeout.

P. S. Cannot bump to the latest, because they have changed licensing to LGPLv3, and I am not legally sure if we can use it.


d4vidi:
We might be able to soon replace these commands using the `adb emu` command set

PR #2027, opened 2020-04-30
feat(jest): new DetoxCircusEnvironment

noomorph:
**Description:**

Fixes long-standing issues with `afterEach-beforeEach` transition hack.

More detailed description - to be done.


d4vidi:
@noomorph insanely exciting! A few notes though:
1. Could you please rewrite an elaborate description of the final scope of the PR?
2. What are the next steps, release-wise? As I understand it, this is a breaking change...? We now require jest >= 26?
3. Will you please walk us (team) through the fundamentals of this change? It's a bit difficult to see the forest for the trees here :)

noomorph:
@d4vidi, thanks. I'll elaborate:

1. I have merged these two (#2009, #2027) because I feel that it is a bit early to push such a deep change without a beta-testing period. Hence, #2009 does not deprecate (yet) existing jest-jasmine or jest-circus adapter integrations.
2. The new environment will be available for those who migrate and those who create a new Detox project. After testing internally for a month at least, we can release a breaking change, where we'll remove the old adapters and demand using the new environment. Yes, it will require `jest >= 26`.
3. Let's schedule, I think that would be helpful.

PR #2011, opened 2020-04-19
Introduce safety time-margins between launches of multiple android emu's

d4vidi:
- [x] This is a small change 
- [ ] This change has been discussed in issue #<?> and the solution has been agreed upon with maintainers.

---

**Description:**
Add `sleep()`s in between Android emulator launches so as to heuristically prevent absolute simultaneous launches of emulators of the same AVD.

This aims at improving 2 things:
1. The Android emulator binary doesn't seem to handle many launches taking place at the same time very well. In particular, seems launches touch some of the configuration, and - if race conditions are met, that evidently causes some emulators either boot from scratch (overlook snapshots) due to an alleged config change or possibly boot with the wrong specs.
2. The well-known `adb jam` problem (see #1857): making tests run timeline vary can help avoid launching multiple `adb install` commands simultaneously, which seems to be the soft spot of `adb` with respect to that specific bug.

PR #2035, opened 2020-05-05
wip

d4vidi:
- [ ] This is a small change 

---

# Description:

A rephrase of the work in #2011: more deterministic and thorough. Introduces 2 fundamental changes:

#### Forced sequential launch of emulators (i.e. await launch if a launch is in session).

Implemented by including the emulator spawn in the lock-file scoped handler function (i.e. don't unlock until emu starts loading).
This aims at solving the following problem: The Android emulator binary doesn't seem to handle many launches taking place at the same time very well. In particular, seems launches touch some of the configuration, and - if race conditions are met, that evidently causes some emulators either boot from scratch (overlook snapshots) due to an alleged config change or possibly boot with the wrong specs.

#### Post-launch delay (todo)
 
Avoid moving forward to ADB commands until all emulators have completed spawning.
This addresses the now well-know adb-jam problem (see #1857), which following #2011, I now believe that reproduces when an emulator launches while an adb command is in-flight.

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -92,6 +92,7 @@
       "EmulatorTelnet.js",
       "EmulatorExec.js",
       "EmulatorLauncher.js",
+      "EmulatorAllocator.js",
       "DeviceDriverBase.js",
       "GREYConfiguration.js",
       "src/utils/environment.js",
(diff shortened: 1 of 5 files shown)


d4vidi:
https://jenkins-oss.wixpress.com/job/detox-android-62-x-master.safe-boot-emus/

d4vidi:
Was closed for the same reason as its predecessor: adb jam problem ensues in a more frequent way.

PR #2037, opened 2020-05-05
DetoxInstruments' sampling interval configuration

valentynberehovyi:
This PR allows to configure sampling interval for detox instruments from configuration.

```diff
"plugins": {
        "instruments": {
               "enabled": true,
+              "samplingInterval": 100
        }
}
```

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
@@ -120,9 +120,11 @@ class InstrumentsRecordingStateActionHandler(
 ) : DetoxActionHandler {
 
     override fun handle(params: String, messageId: Long) {
-        val recordingPath = JSONObject(params).opt("recordingPath")
+        val json = JSONObject(params)
+        val recordingPath = json.opt("recordingPath")
         if (recordingPath is String) {
-            instrumentsManager.startRecordingAtLocalPath(recordingPath)
+            val samplingInterval = json.optLong("samplingInterval", 250)
+            instrumentsManager.startRecordingAtLocalPath(recordingPath, samplingInterval)
         } else {
             instrumentsManager.stopRecording()
         }
--- a/detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
@@ -118,12 +118,15 @@ class InstrumentsRecordingStateActionHandler(
         private val instrumentsManager: DetoxInstrumentsManager,
         private val wsClient: WebSocketClient
 ) : DetoxActionHandler {
+    companion object {
+        const val DEFAULT_SAMPLING_INTERVAL = 250L
+    }
 
     override fun handle(params: String, messageId: Long) {
         val json = JSONObject(params)
         val recordingPath = json.opt("recordingPath")
         if (recordingPath is String) {
-            val samplingInterval = json.optLong("samplingInterval", 250)
+            val samplingInterval = json.optLong("samplingInterval", DEFAULT_SAMPLING_INTERVAL)
             instrumentsManager.startRecordingAtLocalPath(recordingPath, samplingInterval)
         } else {
             instrumentsManager.stopRecording()
--- a/detox/android/detox/src/main/java/com/wix/detox/DetoxJUnitRunner.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/DetoxJUnitRunner.java
@@ -34,12 +34,25 @@ public class DetoxJUnitRunner extends AndroidJUnitRunner {
         monitor.addLifecycleCallback(lifecycleCallback);
     }
 
+    @Override
+    public void onDestroy() {
+        instrumentsManager.stopRecording();
+        instrumentsManager = null;
+        lifecycleCallback = null;
+
+        super.onDestroy();
+    }
+
     private void onBeforeAppOnCreate(Application app, Bundle arguments) {
         if (DetoxInstrumentsManager.supports()) {
             final String recordingPath = arguments.getString("detoxInstrumRecPath");
             if (recordingPath != null) {
+                final long samplingInterval = Long.parseLong(
+                        arguments.getString("detoxInstrumSamplingInterval", "250")
+                );
+
                 instrumentsManager = new DetoxInstrumentsManager(app);
-                instrumentsManager.startRecordingAtLocalPath(recordingPath);
+                instrumentsManager.startRecordingAtLocalPath(recordingPath, samplingInterval);
             }
         }
     }
--- a/detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
+++ b/detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
@@ -166,6 +166,15 @@ object DetoxActionHandlersSpec : Spek({
                 verify(instrumentsManager).startRecordingAtLocalPath(eq("/MockPath"), eq(100500L))
             }
 
+            it("should start recording with path and default samplingInterval") {
+                uut().handle("{\"recordingPath\":\"/MockPath\"}", messageId)
+                verify(instrumentsManager)
+                        .startRecordingAtLocalPath(
+                                eq("/MockPath"),
+                                eq(InstrumentsRecordingStateActionHandler.DEFAULT_SAMPLING_INTERVAL)
+                        )
+            }
+
             it("should stop recording without path") {
                 uut().handle("{\"recordingPath\":null}", messageId)
                 verify(instrumentsManager).stopRecording()
--- a/detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
+++ b/detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
@@ -162,8 +162,8 @@ object DetoxActionHandlersSpec : Spek({
             }
 
             it("should start recording with path") {
-                uut().handle("{\"recordingPath\":\"/MockPath\"}", messageId)
-                verify(instrumentsManager).startRecordingAtLocalPath(eq("/MockPath"))
+                uut().handle("{\"recordingPath\":\"/MockPath\", \"samplingInterval\":100500}", messageId)
+                verify(instrumentsManager).startRecordingAtLocalPath(eq("/MockPath"), eq(100500L))
             }
 
             it("should stop recording without path") {
--- a/detox/ios/Detox/DetoxInstrumentsManager.h
+++ b/detox/ios/Detox/DetoxInstrumentsManager.h
@@ -8,13 +8,17 @@
 
 #import <Foundation/Foundation.h>
 
+@interface DetoxInstrumentsConfiguration : NSObject
+
+@end
 
 @interface DetoxInstrumentsManager : NSObject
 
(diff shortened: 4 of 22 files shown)


detox/ios/Detox/DetoxInstrumentsManager.h
```diff
@@ -8,12 +8,13 @@
...
 #import <Foundation/Foundation.h>
 
+
 @interface DetoxInstrumentsManager : NSObject
 
 + (NSURL*)defaultURLForTestName:(NSString*)testName;
 
-- (void)startRecordingAtURL:(NSURL*)URL;
-- (void)continueRecordingAtURL:(NSURL*)URL;
+- (void)startRecordingAtURL:(NSURL*)URL andSamplingInterval:(NSTimeInterval)samplingInterval;
```
LeoNatan, line 16: Instead of passing a sampling interval, the method should receive a `DTXProfilingConfiguration` object with the configuration overridden as needed. Selector name should be `startRecordingWithConfiguration:`; the URL can now be set directly on the DTXProfilingConfiguration object.

Another option is to change this method to `startRecordingAtURL:configuration:` and pass an NSDictionary with config values, and then create the `DTXProfilingConfiguration` internally.

This way we can extend this to support full configuration setting from json config.

detox/ios/Detox/DetoxInstrumentsManager.m
```diff
@@ -252,21 +252,23 @@ - (id)_configForDetoxRecordingWithURL:(NSURL*)URL
...
 	return config;
 }
 
-- (void)startRecordingAtURL:(NSURL*)URL
+- (void)startRecordingAtURL:(NSURL*)URL andSamplingInterval:(NSTimeInterval)samplingInterval
 {
-	dtx_log_info(@"Starting recording at %@", URL);
-	[_recorderInstance startProfilingWithConfiguration:[self _configForDetoxRecordingWithURL:URL]];
+	dtx_log_info(@"Starting recording at %@ with samplingInterval %f", URL, samplingInterval);
+	id config = [self _configForDetoxRecordingWithURL:URL andSamplingInterval:samplingInterval];
```
LeoNatan, line 263: This should be moved either on the outside, if the method was changed to `startRecordingWithConfiguration :`, or this method should receive an `NSDictionary` with configuration and translate that while creating the `DTXProfilingConfiguration` object.

detox/ios/Detox/DetoxManager.m
```diff
@@ -90,11 +90,14 @@ - (instancetype)init
...
 	[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(_appDidEnterBackground:) name:UIApplicationDidEnterBackgroundNotification object:nil];
-	
-	NSString* recordingPath = [NSUserDefaults.standardUserDefaults objectForKey:@"recordingPath"];
+
+	NSUserDefaults *userDefaults = NSUserDefaults.standardUserDefaults;
+	NSString* recordingPath = [userDefaults objectForKey:@"recordingPath"];
 	if(recordingPath != nil)
 	{
-		[self _handlePerformanceRecording:NSDictionaryOfVariableBindings(recordingPath) isFromLaunch:YES completionHandler:nil];
+		NSString* samplingInterval = [userDefaults objectForKey:@"samplingInterval"];
```
LeoNatan, line 98: Not sure having "samplingInterval" as the launch argument is a good idea, it's too generic. This is also true about "recordingPath".

detox/ios/Detox/DetoxInstrumentsManager.h
valentynberehovyi: I though about this as we'll may have many different, but decided to do this when we'll have 3+ params for recording.

detox/ios/Detox/DetoxInstrumentsManager.m
valentynberehovyi: Agreed

detox/ios/Detox/DetoxManager.m
valentynberehovyi: Are you talking about naming problem or about the idea of passing launch arguments as general idea?

detox/ios/Detox/DetoxInstrumentsManager.h
LeoNatan: Let's do it now. It will be needed once we really start using the feature.

detox/ios/Detox/DetoxManager.m
LeoNatan: I mean naming. There is no escape from launch arguments for now because they have to be passed very early on so that recording starts as early as possible. But "recordingPath" and "samplingInterval" are quite generic, so there may be collision with user arguments.

detox/ios/Detox/DetoxInstrumentsManager.m
```diff
@@ -234,10 +250,10 @@ - (instancetype)init
 	return self;
 }
 
-- (id)_configForDetoxRecordingWithURL:(NSURL*)URL andSamplingInterval:(NSTimeInterval)samplingInterval
+- (DetoxInstrumentsConfiguration *)configurationFromProps:(NSDictionary *)props
```
LeoNatan, line 253: Do you mind changing this to `configurationWithDictionary:`

detox/ios/Detox/DetoxInstrumentsManager.m
```diff
@@ -55,6 +55,22 @@ typedef NS_ENUM(NSUInteger, __DTXEventStatus) {
 WEAK_IMPORT_ATTRIBUTE
 @interface DTXProfiler : NSObject @end
 
+@interface DetoxInstrumentsConfiguration ()
```
LeoNatan, line 58: Is `DetoxInstrumentsConfiguration` needed if all it does is hold the DTXProfilerConfiguration object? Why not just return that and pass it to recording manager?

detox/ios/Detox/DetoxInstrumentsManager.m
vberegovoy: Thats fine. I'll change.

detox/ios/Detox/DetoxInstrumentsManager.m
vberegovoy: Because we'll need to specify the configuration type somehow. If we'll just pass the `id` type so anyone outside may pass the any id into it.

LeoNatan: So why not make the entry method startReacordingWithDictionary: and internally create the configuration?

valentynberehovyi: Better to separate configuration creation from starting/continue logic.

detox/android/detox/src/main/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -120,9 +120,11 @@ class InstrumentsRecordingStateActionHandler(
 ) : DetoxActionHandler {
 
     override fun handle(params: String, messageId: Long) {
-        val recordingPath = JSONObject(params).opt("recordingPath")
+        val json = JSONObject(params)
+        val recordingPath = json.opt("recordingPath")
         if (recordingPath is String) {
-            instrumentsManager.startRecordingAtLocalPath(recordingPath)
+            val samplingInterval = json.optLong("samplingInterval", 250)
```
d4vidi, line 126: could you set 250 as a const in a `companion object`?

detox/android/detox/src/test/java/com/wix/detox/DetoxActionHandlersSpec.kt
```diff
@@ -162,8 +162,8 @@ object DetoxActionHandlersSpec : Spek({
             }
 
             it("should start recording with path") {
-                uut().handle("{\"recordingPath\":\"/MockPath\"}", messageId)
-                verify(instrumentsManager).startRecordingAtLocalPath(eq("/MockPath"))
+                uut().handle("{\"recordingPath\":\"/MockPath\", \"samplingInterval\":100500}", messageId)
+                verify(instrumentsManager).startRecordingAtLocalPath(eq("/MockPath"), eq(100500L))
```
d4vidi, line 166: add another (similar) test to verify the default (250)?

detox/android/detox/src/main/java/com/wix/detox/DetoxJUnitRunner.java
```diff
@@ -34,12 +34,25 @@ public void onApplicationLifecycleChanged(Application app, ApplicationStage stag
...
     }
 
+    @Override
+    public void onDestroy() {
+        instrumentsManager.stopRecording();
+        instrumentsManager = null;
+        lifecycleCallback = null;
+
+        super.onDestroy();
+    }
```
d4vidi, line 44: ?? ❤️

detox/src/artifacts/instruments/android/AndroidInstrumentsPlugin.test.js
```diff
@@ -0,0 +1,100 @@
...
+    };
+    await plugin.onTestStart(testSummary);
+    await plugin.onBeforeLaunchApp(event);
+    expect(event.launchArgs.detoxInstrumRecPath).toBe(mockedDevicePathBuilder.buildTemporaryArtifactPath());
+    expect(event.launchArgs.detoxInstrumSamplingInterval).toBeUndefined();
+  });
+
+  it('should build launchArgs with started test and samplingInterval from config', async () => {
+    const plugin = new AndroidInstrumentsPlugin({
+      api: {
```
noomorph, line 65: This is an optional TODO, but the preparation in every test looks really long. 
To shorten the tests, I would suggest initializing `pluginConfig` variable in `beforeEach`:

```js
pluginConfig = {
  api: mockedApi,
  client: mockedClient,
  devicePathBuilder: mockedDevicePathBuilder,
  userConfig: { enabled: true },
};
```

Then, in test preparations, you will be able to explicitly state what is different in the config that you are testing:

```js
pluginConfig.userConfig.samplingInterval = 100500;
```

instead of many lines like:

```js
   const plugin = new AndroidInstrumentsPlugin({
      api: {
        ...mockedApi,
        userConfig: {
          enabled: true
        }
      },
      client: mockedClient,
      devicePathBuilder: mockedDevicePathBuilder
    });
```

detox/src/artifacts/instruments/ios/SimulatorInstrumentsPlugin.js
```diff
@@ -14,6 +14,10 @@ class SimulatorInstrumentsPlugin extends InstrumentsArtifactPlugin {
 
     if (this.testRecording) {
       event.launchArgs['recordingPath'] = this.testRecording.temporaryRecordingPath;
+      event.launchArgs['samplingInterval'] =
+        this.api.userConfig.samplingInterval
+          ? Number(this.api.userConfig.samplingInterval) / 1000.0
+          : 0.25;
```
noomorph, line 20: Looks a bit disturbing, in a sense that you put an inline logic with divisions and default values here, instead of using `prepareSamplingInterval`, :confused: ? Should not the same logic be applied if you send JSON over Websocket in SimulatorInstrumentsPlugin?

detox/android/detox/src/main/java/com/wix/detox/DetoxJUnitRunner.java
valentynberehovyi: This one was missed before. So better cleaning for self late than never.

valentynberehovyi: > ?? ❤️

This one was missed before. So better cleaning for self late than never.

PR #2045, opened 2020-05-07
Apply missing flags to "pm install": reinstall, permissions, test

d4vidi:
- [x] This is a small change 
- [ ] This change has been discussed in issue #2019 and the solution has been agreed upon with maintainers.

---

**Description:**
Add missing flags to recently instated `adb shell pm install` command (see #2019) - the same ones as used to be specified for `adb install`:
- `-r` to allow for the replacement of existing apps (hence no need to manually uninstall)
- `-t` to allow for test packages (seems useless, but nevertheless)
- `-g` to auto-allow for all runtime permissions.

This doesn't fix any reported issue in particular, but rather addresses [this comment](https://github.com/wix/Detox/pull/2019#issuecomment-623632126) in the original PR.

### DIFF
--- a/detox/src/devices/drivers/android/AndroidDriver.js
+++ b/detox/src/devices/drivers/android/AndroidDriver.js
@@ -84,11 +84,6 @@ class AndroidDriver extends DeviceDriverBase {
     await uninstallHelper.uninstall(deviceId, bundleId);
   }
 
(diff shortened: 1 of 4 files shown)


netshade:
:+1: You're awesome.

PR #2061, opened 2020-05-13
Fix and clarify troubleshooting "The app loads but tests fail to start in SDK >= 28" on android

imownbey:
- [x] This is a small change 

---

**Description:**
Just updates the Getting Started for Android guide.

Previously this said to use `<domain-config cleartextTrafficPermitted="false">` which is how you explicitly set domains to not allow cleartext traffic but for Detox we want the opposite.  I also tried to make the readme section a little more explicit on exactly what to do.

I also removed the "untested" part of this because I have tested and confirmed it works.


d4vidi:
Thanks for this! I'll embed in our example apps to use this, and further update the docs.

PR #2068, opened 2020-05-17
update docs to include latest kotlin version

fvalela1:
following the current docs, the `detox build -c android.emu.release`, it will fail saying:
```
The Android Gradle plugin supports only Kotlin Gradle plugin version 1.3.10 and higher.
The following dependencies do not satisfy the required version:
root project 'example' -> org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.0
```

- [x] This is a small change 

---

**Description:**

### DIFF
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -86,7 +86,7 @@ If your project does not already support Kotlin, add the Kotlin Gradle-plugin to
 ```groovy
 buildscript {
     // ...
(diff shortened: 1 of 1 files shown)


d4vidi:
Sorry, I can't accept this change. Soon there will be kotlin 1.3.99 or 1.4.0... I will never be able to keep up, I'm afraid -- hence, the comment next to the version

PR #2070, opened 2020-05-18
fix(mocha): use .mocharc when possible

### LINKED ISSUE #2057: Configuration via mocha.opts is deprecated. Use RC files or package.json instead.
When I execute Detox tests I see the console message:
```
detox[88089] ERROR: at internal/process/warning.js:32:3
 (node:88089) DeprecationWarning: Configuration via mocha.opts is DEPRECATED and will be removed from a future version of Mocha. Use RC files or package.json instead.
```

On https://mochajs.org/#mochaopts it says:

> mocha.opts file support is DEPRECATED and will be removed from a future version of Mocha. We recommend using a configuration file instead.

It looks like the Detox configuration options do not yet allow a mocha configuration file because only the `mocha.opts` option is described here:
https://github.com/wix/detox/blob/master/docs/APIRef.Configuration.md#mocha

I would like to get a solution to use a Mocha configuration file to be able to upgrade Mocha.

Maybe it's possible to use Jest instead of Mocha. But Mocha works well for us so far.

**Additional context**
I use the following versions:
- "detox": "^16.4.1",
- "mocha": "^7.1.1",

noomorph:
- [x] This change has been discussed in issue #2057 and the solution has been agreed upon with maintainers.

Resolves #2057 

---

**Description:**

1. In `detox` section of `package.json`, when `test-runner` is `mocha`, and `runner-config` field points to a file with an extension different from `.opts`, Detox CLI in `detox test` command will use `--config` option instead of `--opts`.

2. `detox init -r mocha`  generates a template with `e2e/.mocharc` file instead of the deprecated `e2e/mocha.opts`.

### DIFF
--- a/detox/local-cli/init.js
+++ b/detox/local-cli/init.js
@@ -26,6 +26,7 @@ module.exports.handler = async function init(argv) {
       createMochaFolderE2E();
       patchDetoxConfigInPackageJSON({
         runner: 'mocha',
+        runnerConfig: 'e2e/.mocharc'
       });
       break;
     case 'jest':
@@ -76,7 +77,7 @@ function createFile(filename, content) {
 
 function createMochaFolderE2E() {
   createFolder('e2e', {
-    'mocha.opts': mochaTemplates.runnerConfig,
+    '.mocharc': mochaTemplates.runnerConfig,
     'init.js': mochaTemplates.initjs,
     'firstTest.spec.js': mochaTemplates.firstTest
   });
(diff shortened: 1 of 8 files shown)


sonicdoe:
Does Mocha actually support `.mocharc`? [The documentation only lists `.mocharc.js`, `.mocharc.json`, etc.](https://mochajs.org/#configuring-mocha-nodejs) but no extension-less `.mocharc`.

noomorph:
Apparently, yes, if you specify that explicitly.

https://github.com/mochajs/mocha/blob/1a4646dd2f7b35010c2488edcb2d5f4cfb2f14a8/lib/cli/config.js#L82

Though, indeed better if the extension were `.mocharc.json`.

sonicdoe:
Ah, you’re right 👍 That’s even documented in [Custom Locations](https://mochajs.org/#custom-locations):

> Mocha will use the file's extension to determine how to parse the file, and will assume JSON if unknown.

PR #2074, opened 2020-05-20
Fix spelling of developers in android quick start

imownbey:
- [x] This is a small change 

---

**Description:**
Just noticed this spelling error :)


rotemmiz:
Thanks!

PR #2086, opened 2020-05-26
hotfix(android): respect --headless option in Mocha and other runners

### LINKED ISSUE #2046: Issue with the --headless argument not working
**Describe the bug**
The --headless argument is not working since the arguments parser is verifying for === "true" while the argument will actually be boolean true. I have added in the local-cli/test.js `(program.headless ? '--headless "true"' : ''),` and this fixes the issue. But most likely this issue affects other parts of the application as well.

I am trying to run it in docker for CI and this is what I get:
```text
Error: Command failed: node_modules/.bin/mocha --opts test/integration/mocha/mocha.opts --configuration android.emu.release --cleanup --grep :ios: --invert --headless --use-custom-logger "true" "test/integration/specs"

Fatal: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

**To Reproduce**

- [x ] I have tested this issue on the latest Detox release and it still reproduces

Steps to reproduce:
1. Install latest detox with mocha runner.
2. Attempt to run detox test --headless.
3. Notice in EmualtorExec.js that it is not "true" but rather true.

**Expected behavior**
The --headless command should properly work and set the -no-window argument for calling the emulator.

**Workaround**

Add in init.js of mocha before the require of detox this to temporarly apply fix until this is fixed in detox:
```js
// @TODO: Remove once https://github.com/wix/Detox/issues/2046 is fixed.
if (process.argv.includes('--headless')) {
  process.argv[process.argv.indexOf('--headless')] = '--headless=true';
  console.log(process.argv);
}

/// ...requires
```

 - Detox: 16.4, tested 16.2 as well
 - React Native: 0.62.2: does not matter
 - Node: 12.16.3
 - Device: desktop / docker: does not matter
 - OS: Ubunut 18.04: does not matter
 - Test-runner (select one): Mocha

noomorph:
Resolves #2046.

### DIFF
--- a/detox/src/devices/drivers/android/tools/EmulatorExec.js
+++ b/detox/src/devices/drivers/android/tools/EmulatorExec.js
@@ -42,8 +42,8 @@ class LaunchCommand extends ExecCommand {
       '-verbose',
       '-no-audio',
       '-no-boot-anim',
-      argparse.getArgValue('headless') === 'true' ? '-no-window' : '',
-      argparse.getArgValue('readOnlyEmu') === 'true' ? '-read-only' : '',
+      `${argparse.getArgValue('headless')}` === 'true' ? '-no-window' : '',
+      `${argparse.getArgValue('readOnlyEmu')}` === 'true' ? '-read-only' : '',
       options.port ? `-port` : '',
       options.port ? `${options.port}` : '',
       ...deviceLaunchArgs,
@@ -64,7 +64,8 @@ class LaunchCommand extends ExecCommand {
       return gpuArgument;
     }
 
-    if (argparse.getArgValue('headless')) {
+    const headless = `${argparse.getArgValue('headless')}` === 'true';
+    if (headless) {
       switch (os.platform()) {
         case 'darwin':
           return 'host';


archfz:
@noomorph no tests? that means in the future this might break again

noomorph:
@archfz, thanks for caring, seriously, but:

1. That is being released as a **hotfix**, not as a **fix**. I am displeased this issue has gotten to my radar so late, more than 18 days passed since the report, and I want you to have it as soon as it is possible. My apologies for being late here,, and thanks for your report!

2. The `import-argparse-everywhere-and-use-it` approach has plagued many places in Detox, and during the last 1.5 years, I have been crusading to make it perish in a series of separate PRs. What was edited here, is one of the places that have survived after configuration subsystem refactors - not surprising since there were too many.

It is no good here to wrap booleans inside template strings (like I just did in #2086) and hurry up to shout "hurray!". A long-living approach will be to make a proper constructor injection with type-strict parameters (boolean, in this case) and cover that `headless` case with 2 tests: when `headless === false` and when `headless === true`.

Otherwise, testing `true, "true", 1, 0, undefined, null, ""` in this class simply violates a single responsibility approach, and I am strongly against polluting the codebase with short-lived (and conceptually incorrect) unit tests.

If you insist, I can create an internal issue for this refactor in Android drivers so that you can monitor when this refactor lands.

Cheers.

PR #2094, opened 2020-06-01
Apply a clear, by-the-book impl. for taps and multi-taps on Android

d4vidi:
- [ ] This is a small change 

---

**Description:**

Improve implementation of `tap()` and `multiTap()` on Android so as to consider potential gesture-handler listening to complex gestures - double-taps in particular, that can hijack tap-events if not applied by-the-book.

Motivated for fixing now in the wake of related bugs recently found while integrating test-suites of newly joining internal projects (for execution on Android emulators) -- where an interesting sequence of taps has been applied in the beginning of each test.

---
As for implementation, most work was around creating e2e tests that would uncover the faulty behaviour. That required the creation of a custom native view, now rendered as a part of the test-app's actions screen (but only on Android), that would provide a simple impl. of a tappable text view that changes only when *double-tapped*, specifically.

The core of the matter was twofold:
1. Plain taps lacked the proper post-tap delay (was too short) such that consequent taps (coming from additional calls to `tap()` at test-level) can be mistakingly considered a continuation of them; The end-result being that 2 taps in a row can be mistakingly considered part of a single double-tap gesture.
2. Multi-tap handlers, on the other hand, has an inter-tap delay that was accumulatively too long so as to mistakingly prevent the multiple taps from _becoming_ a double/triple/etc. tap sequence, as they should.

### DIFF
--- a/detox/test/android/app/src/main/java/com/example/DoubleTapsTextViewManager.java
+++ b/detox/test/android/app/src/main/java/com/example/DoubleTapsTextViewManager.java
@@ -1,0 +1,61 @@
+package com.example;
+
+import android.graphics.Color;
+import android.view.GestureDetector;
+import android.view.Gravity;
+import android.view.MotionEvent;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.ViewGroup.LayoutParams;
+import android.widget.FrameLayout;
+import android.widget.TextView;
+
+import com.example.utils.DoubleTapListenerStub;
+import com.facebook.react.uimanager.SimpleViewManager;
+import com.facebook.react.uimanager.ThemedReactContext;
+
+import androidx.annotation.NonNull;
+
+public class DoubleTapsTextViewManager extends SimpleViewManager<ViewGroup> {
+    private static class ViewState {
+        int taps = 0;
+    }
+
+    @NonNull
+    @Override
+    public String getName() {
+        return "DetoxDoubleTapsTextView";
+    }
+
+    @NonNull
+    @Override
+    protected ViewGroup createViewInstance(@NonNull ThemedReactContext reactContext) {
+        final ViewState viewState = new ViewState();
+
+        final FrameLayout layout = new FrameLayout(reactContext);
+        layout.setLayoutParams(new FrameLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
+
+        final TextView textView = new TextView(reactContext);
+        textView.setTag("doubleTappableText");
+        textView.setLayoutParams(new FrameLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.WRAP_CONTENT, Gravity.CENTER));
+        textView.setTextAlignment(View.TEXT_ALIGNMENT_CENTER);
+        textView.setGravity(Gravity.CENTER);
+        textView.setText("Double-Taps: "+viewState.taps);
+        textView.setTextColor(Color.BLUE);
+
+        final GestureDetector gestureDetector = new GestureDetector(reactContext, new GestureDetector.SimpleOnGestureListener());
+        gestureDetector.setOnDoubleTapListener(new DoubleTapListenerStub() {
+            @Override
+            public boolean onDoubleTap(MotionEvent e) {
+                viewState.taps++;
+                textView.setText("Double-Taps: "+viewState.taps);
+                return true;
+            }
+        });
+        textView.setOnTouchListener((v, event) -> gestureDetector.onTouchEvent(event));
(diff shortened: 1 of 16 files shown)


detox/test/android/app/src/main/java/com/example/NativeModule.java
```diff
@@ -118,39 +113,4 @@ public void crashMainThread() {
             throw new RuntimeException("Simulated crash (native)");
         });
     }
-
-    /**
```
d4vidi, line 122: not directly related, just some refactoring

detox/test/android/app/src/main/java/com/example/utils/ViewSpies.java
```diff
@@ -0,0 +1,62 @@
...
+     * {@link ReactFindViewUtil#addViewListener(ReactFindViewUtil.OnViewFoundListener)}.
+     * However, it is found to be a bit buggy since it removes all listeners immediately
+     * after being called (i.e. while iterating) with no thread-sync mechanisms to protect it.
+     * If real life (CI), we've genuinely seen that it throws ConcurrentModificationException exceptions,
+     * on occasions (and why wouldn't it? - our demo app's ActionsScreen has multiple views subscribing and
+     * called concurrently; could it be that not always everything is run in the main thread?).
+     * Therefore, we use here {@link ReactFindViewUtil#addViewsListener(ReactFindViewUtil.OnMultipleViewsFoundListener, Set)},
+     * which is too generic but nevertheless allows us to better control when we are to be removed.
+     */
+    abstract class BaseViewSpy implements ReactFindViewUtil.OnMultipleViewsFoundListener {
```
d4vidi, line 23: Result of aforementioned refactoring (not directly related)

detox/test/src/Screens/ActionsScreen.js
```diff
@@ -58,7 +64,7 @@ export default class ActionsScreen extends Component {
           delayLongPress={1200}
           onLongPress={this.onButtonPress.bind(this, 'Long Press With Duration Working')}
         >
-          <Text style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }}>Long Press Me 1.5s</Text>
+          <Text testID='longTappable' style={{ color: 'blue', marginBottom: 20, textAlign: 'center' }}>Long Press Me 1.5s</Text>
```
d4vidi, line 67: A result of an intermediate state. Need to remove.

detox/test/android/app/src/main/java/com/example/DoubleTapsTextViewManager.java
```diff
@@ -0,0 +1,57 @@
...
+import android.view.ViewGroup;
+import android.view.ViewGroup.LayoutParams;
+import android.widget.FrameLayout;
+import android.widget.TextView;
+
+import com.example.utils.DoubleTapListenerStub;
+import com.facebook.react.uimanager.SimpleViewManager;
+import com.facebook.react.uimanager.ThemedReactContext;
+
+public class DoubleTapsTextViewManager extends SimpleViewManager<ViewGroup> {
```
guyca, line 17: Can you extend `ReactTextViewManager`? Last time I checked, extending base view managers wasn't possible, perhaps that changed? Then you wouldn't have to mess around layout params and styles in this class.

detox/test/android/app/src/main/java/com/example/DoubleTapsTextViewManager.java
```diff
@@ -0,0 +1,57 @@
...
+        gestureDetector.setOnDoubleTapListener(new DoubleTapListenerStub() {
+            @Override
+            public boolean onDoubleTap(MotionEvent e) {
+                viewState.taps++;
+                textView.setText("Double-Taps: "+viewState.taps);
+                return true;
+            }
+        });
+        textView.setOnTouchListener((v, event) -> gestureDetector.onTouchEvent(event));
+        textView.setOnClickListener(v -> {});
```
guyca, line 52: does this listener serve any purpose?

detox/test/android/app/src/main/java/com/example/DoubleTapsTextViewManager.java
d4vidi: I tried, tbh, and it's too much of a headache. I don't need the manager + shadow-node in this case, since I don't want things such as having the layout and content controlled by the JSX -- it's actually somethings I need to avoid here, and keep content changes "instant".

detox/test/android/app/src/main/java/com/example/DoubleTapsTextViewManager.java
d4vidi: Yes! idk why but my touch listener wasn't invoked without it! perhaps I should note that! 🙏

PR #1473, opened 2019-06-24
Fail test when waitFor().withTimeout() reaches timeout and assertion fails

LeoNatan:
Also removed no longer needed comments in docs regarding the behavior of waitFor().

PR #2095, opened 2020-06-01
Update waitFor docs

rcscott:
- [x] This is a small change 

---

**Description:**

#1473 fixed `waitFor().withTimeout()` so that an exception is raised when the assertion fails. That PR updated other sections of the doc to remove the note about using `expect()` but it looks like this line was missed.


LeoNatan:
Thanks

Issue #877, opened 2018-08-08
 GitLab CI Detox integration


burhanyilmaz:
Is there any boilerplate  `GitLab CI` for detox?
I found this https://gist.github.com/adam-stasiak/1b2dd63f3d0963a38d46aaea73631072. it is not work :/

noomorph:
Frankly speaking, I don't think that this is a properly created issue:

> Is there any boilerplate GitLab CI for detox?
> I found this https://gist.github.com/adam-stasiak/1b2dd63f3d0963a38d46aaea73631072. it is not work :/

However, as you might see in the documentation, there is no boilerplate for `GitLab CI` currently there:
https://github.com/wix/detox/blob/master/docs/Guide.RunningOnCI.md

So, the answer is *no*, there is no such boilerplate in the official detox documentation.

However, you are welcome:

1) to work on GitLab CI integration with detox;
2) figure out what is wrong for that script;
3) fix the issues;
4) create a tiny sample project with Detox+GitLab CI intergration;
5) open-source it to GitHub as a reference;
6) and, finally, contribute to the documentation:

https://github.com/wix/detox/blob/master/docs/Guide.RunningOnCI.md

Thanks.

PR #2111, opened 2020-06-09
Update Gitlab CI section for running on CI guide

jarvisluong:
- [x ] This is a small change 

---

**Description:**

As mentioned in https://github.com/wix/Detox/issues/877, it would be nice to add a section for example running on gitlab CI

### DIFF
--- a/docs/Guide.RunningOnCI.md
+++ b/docs/Guide.RunningOnCI.md
@@ -152,3 +152,30 @@ workflows:
     - _detox_tests
     after_run: []
 ```
+
+### • Running Detox on [GitlabCI](https://docs.gitlab.com/ee/ci/README.html) - Android Only
+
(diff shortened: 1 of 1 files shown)


rotemmiz:
Wrote an entire manifesto here, it all got deleted :(((((( ugh
Anyway, sorry for the long delay in handling this PR. I just went over it, and although I believe it is written well, and uses good examples (AOSP image!), I believe that there should be an easier option than using an external cloud provider. If I'm not mistaken, Circle CI's macOS VMs run with nested virtualization support, and might actually work out of the box (slower maybe, but its worth checking, or at least noting).

jarvisluong:
> Wrote an entire manifesto here, it all got deleted :(((((( ugh
> Anyway, sorry for the long delay in handling this PR. I just went over it, and although I believe it is written well, and uses good examples (AOSP image!), I believe that there should be an easier option than using an external cloud provider. If I'm not mistaken, Circle CI's macOS VMs run with nested virtualization support, and might actually work out of the box (slower maybe, but its worth checking, or at least noting).

Thanks for the comment! IMO circle CI is a different CI provider, and they don't provide their VM to be used directly. I think this guide is meant for people who has to choose Gitlab CI due to specific reasons.

rotemmiz:
⁦the effects of sleeping too little🤦‍♂️⁩, sorry about the confusion.

PR #2114, opened 2020-06-10
Fix typo in APIRef.DetoxCLI.md

Fishbowler:
- [x] This is a small change 

---

Fix a teeny tiny typo.


rotemmiz:
Thanks!

PR #2126, opened 2020-06-14
detoxrc key fix in documentation

mendyEdri:
`detox build --configuration ios` will fail due missing key "ios" in .detoxrc.json file. 
In this `getting started` documentation it says to use the same key as the package.json configuration "ios.sim.debug" that won't build.

- [ ] This is a small change 

---

**Description:**

### DIFF
--- a/docs/Introduction.GettingStarted.md
+++ b/docs/Introduction.GettingStarted.md
@@ -130,7 +130,7 @@ Either way the config should look like this:
 ```json
 {
   "configurations": {
-    "ios.sim.debug": {
+    "ios": {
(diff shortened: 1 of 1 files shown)


rotemmiz:
This is a new part of the doc, I think it's not aligned with the previous documentation.
I would prefer it the other way around, to retain the ios.sim.debug key, and fix it in the second config.

mendyEdri:
Yeah I just saw it’s in multiple places like that, so I deleted the PR, thanks :)

rotemmiz:
It would be good if you can just do it the other way around, change `ios` to `ios.sim.debug`

PR #2130, opened 2020-06-14
Address minor tech debts in main test suite

d4vidi:
- [x] This is a small change 

---

**Description:**
Address minor tech debts to tidy-up test code.


LeoNatan:
Please wait with those. There have been many changes in DetoxNext, which will conflict with changes here.

PR #2132, opened 2020-06-14
Add localhost to android network_security_config.xml (both in example and docs) to fix tests stuck on attached devices

alesharik:
- [x] This is a small change 

---

**Description:**
This PR adds `localhost` to android `network_security_config` in example app and docs. This rule is required to run tests on attached device (because `adb reverse` exposes ports on localhost) and its absence is discouraged for newbies


d4vidi:
thanks!

PR #2134, opened 2020-06-16
Introduce basic support for notification-triggers in device.launchApp()

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1192 and the solution has been agreed upon with maintainers.

---

**Description:**
Partially implements #1192 by adding support for the core use case, as conveyed by this API:
```js
    const userNotification = {
      payload: {
        userData1: 'userDataValue1',
        userData2: 'userDataValue2',
      },
    };
    await device.launchApp({ newInstance: true, userNotification });
```

Gaps to fill in after merge:
- [ ] Add support for handling notification when app is in background (i.e. move to foreground with notification data)
- [ ] Add support for `sendUserNotification()` API

PR #2138, opened 2020-06-18
Introduce basic support for all user-notification API's in all relevant use cases

### LINKED ISSUE #1192: Introduce basic support for push notifications mocking on Android
### Description

Define and implement an API for mocking push notification invocation events on Android, so as to introduce parity with the equivalent API for iOS.

This has long been a missing functionality on Android compared to iOS, as - unlike on iOS, there's no skeletal definition of notification invocation events. Rather, on Android, notifications fire pending intents, originally produced upon notification arrival handling *by the app itself*; All the while there's also no restriction as to how and when notifications show, and what the nature of the registered pending intent is.

The general direction is towards formalizing a reduced intents API, to be used by the JS test code.

#### Phase 1

In this basic-support phase, we limit the functionality to a naive one where the same intent is always used*, and the user is only able to provide payload, using the API's already defined for iOS.

This boils down to 3 types of use cases:

```js
    const userNotification = {
      payload: {
        userData1: 'userDataValue1',
        userData2: 'userDataValue2',
      },
    };

    // #1 - most common
    await device.launchApp({ newInstance: true, userNotification });

    // #2 - `launchApp` used for a running app (a not very appealing API overall)
    await device.launchApp({ newInstance: false, userNotification });

    // #3 - The dedicated function
```
> Note: while `sendUserNotification()` is an API that's not very idiomatic on Android, we use it here nonetheless, assuming we will abandon all of this in phase 2 (see below).

#### Phase 2

For the full vision and more versatile Android-idiomatic solution, see #2141.

---
<sup> * This refers to a simply, launcher-like intent, with user payload "flat"-set in its extras -- much like the way the default FCM impl. handles *message* notifications on the device.</sup>

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #1192 and the solution has been agreed upon with maintainers.

---

**Description:**
Resolves #1192, by filling the missing gaps - as described in #2134, namely, by adding support for these two API usage cases:

```js
await device.launchApp({ newInstance: false, userNotification });
```

and:
```js
```

---
In the process, I've also cleaned up and aligned the way URL-based launches are done. Formerly, the driver used to hold a transient pending-url property for the time between `deliverPayload` and `launchApp()` are called (by `device.launchApp()`). This has been removed, and `deliverPayload` is now completely ignored in this use case, as it is effectively irrelevant for Android.

### DIFF
--- a/detox/src/android/espressoapi/Detox.js
+++ b/detox/src/android/espressoapi/Detox.js
@@ -46,6 +46,18 @@ class Detox {
     };
   }
 
+  static startActivityFromNotification(dataFilePath) {
+    if (typeof dataFilePath !== "string") throw new Error("dataFilePath should be a string, but got " + (dataFilePath + (" (" + (typeof dataFilePath + ")"))));
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.Detox"
+      },
+      method: "startActivityFromNotification",
+      args: [dataFilePath]
+    };
+  }
+
   static extractInitialIntent() {
     return {
       target: {
@@ -53,6 +65,17 @@ class Detox {
         value: "com.wix.detox.Detox"
       },
       method: "extractInitialIntent",
+      args: []
+    };
+  }
+
+  static getAppContext() {
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.Detox"
+      },
+      method: "getAppContext",
       args: []
     };
   }
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -56,7 +56,7 @@ class Device {
     const _bundleId = bundleId || this._bundleId;
     if (this._isAppInBackground(params, _bundleId)) {
       if (hasPayload) {
-        await this.deviceDriver.predeliverPayload({...params});
+        await this.deviceDriver.deliverPayload({...params, delayPayload: true});
       }
     }
 
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -19,14 +19,6 @@ class Device {
     await this.deviceDriver.prepare();
   }
 
-  createPayloadFileAndUpdatesParamsObject(key, launchKey, params, baseLaunchArgs) {
-    const payloadFilePath = this.deviceDriver.createPayloadFile(params[key]);
-    baseLaunchArgs[launchKey] = payloadFilePath;
-    //`params` will be used later for `deliverPayload`, so remove the actual notification and add the file URL
-    delete params[key];
-    params[launchKey] = payloadFilePath;
-  }
-
   async launchApp(params = {newInstance: false}, bundleId) {
     const payloadParams = ['url', 'userNotification', 'userActivity'];
     const hasPayload = this._assertHasSingleParam(payloadParams, params);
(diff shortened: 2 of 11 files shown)


detox/src/devices/Device.js
```diff
@@ -64,7 +56,7 @@ class Device {
     const _bundleId = bundleId || this._bundleId;
     if (this._isAppInBackground(params, _bundleId)) {
       if (hasPayload) {
-        await this.deviceDriver.deliverPayload({...params, delayPayload: true});
```
LeoNatan, line 67: `delayPayload` is needed for [iOS](https://github.com/wix/Detox/blob/d3a90729f50d5920371029d9e9debe85577f1047/detox/ios/Detox/DetoxManager.m#L269). Why was it removed?
As I said in our internal conversation, timing and these params are very important for the correct behavior of iOS, so please don't remove stuff.

d4vidi: I originally added `delayPayload` for Android, and it was a bit of a hack. With `predeliver` its much cleaner, but nevertheless - iOS remains unaffected in both cases, so no worries!

d4vidi: Hmm just spotted the native usage of `delayPayload` on iOS. Strikes me by surprise. I'm still looking into alternatives, then. In any case, let's not forget that this is the purpose of code reviews. Let's also try to figure out why this change hasn't broken any test - unit or e2e: this is a signal of that coverage is missing.

LeoNatan: It’s hard to test this in Detox E2E. The native is unit tested.

detox/src/devices/Device.js
```diff
@@ -104,6 +96,14 @@ class Device {
...
   }
 
+  _createPayloadFileAndUpdatesParamsObject(key, launchKey, params, baseLaunchArgs) {
+    const payloadFilePath = this.deviceDriver.createPayloadFile(params[key]);
+    baseLaunchArgs[launchKey] = payloadFilePath;
+    //`params` will be used later for `predeliverPayload`, so remove the actual notification and add the file URL
+    delete params[key];
+    params[launchKey] = payloadFilePath;
+  }
+
```
d4vidi, line 106: :boyscout:

detox/src/android/espressoapi/Detox.js
```diff
@@ -57,6 +69,17 @@ class Detox {
...
+  static getAppContext() {
+    return {
+      target: {
+        type: "Class",
+        value: "com.wix.detox.Detox"
+      },
+      method: "getAppContext",
+      args: []
+    };
+  }
```
d4vidi, line 81: 🙄

detox/src/devices/drivers/android/AndroidDriver.js
```diff
@@ -122,14 +120,18 @@ class AndroidDriver extends DeviceDriverBase {
     return pid;
   }
 
-  async deliverPayload(params) {
-    const {delayPayload, url} = params;
+  async deliverPayload(params, deviceId) {
+    if (params.delayPayload) {
+      return;
+    }
```
d4vidi, line 126: `delayPayload` is irrelevant for Android, it seems. We ignore everything when it is set; rather, we look at the explicit params provided to `launchApp()` (where we used to get the previously cached url, previously)

detox/src/devices/drivers/android/AndroidDriver.test.js
```diff
@@ -136,41 +155,191 @@ describe('Android driver', () => {
     });
   });
 
-  describe('Notification data handling (user-notification-data-URL arg)', () => {
-    const launchArgs = Object.freeze({
+  describe('URL runtime delivery handling', () => {
```
d4vidi, line 158: Now coverage to existing api's (i.e. launch with url & sendURL() flow) - but aligned with a modified implementation (as previously mentioned).

PR #2139, opened 2020-06-19
fix(jest-circus): reinitialize custom logger on every Environment.setup

noomorph:
- [x] This is a small change 

---

**Description:**

Fixes a regression bug that has popped after DetoxCircusEnvironment rewrite.

Before the fix (regression):

Screenshot from 2020-06-19 19-11-35

After the fix:

Screenshot from 2020-06-19 19-11-11


d4vidi:
OMJ. Approved, I suppose 😅

PR #2144, opened 2020-06-22
ci: nvm install 14

noomorph:
- [x] This is a small change 

---

**Description:**

Upgrading to the next Node.js LTS version (14.x) might be useful, especially if we try to gather code coverage reports via V8 built-in mechanism. Our current 10.x has already entered maintenance mode, and sometimes I see different behavior for error stack traces between Node versions.

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -2,7 +2,9 @@
 
 source $(dirname "$0")/logger.sh
 
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
 nvm install 14
+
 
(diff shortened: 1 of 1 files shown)


LeoNatan:
Shouldn't this be preinstalled on CI machines by @igorgn ?

igorgn:
> Shouldn't this be preinstalled on CI machines by @igorgn ?

Installing nvm version is literally takes ~5 seconds. Though there's some issues with installing it here (which I'm going to fix), updating CI agents with another node version every time is definitely not the most convenient way.

LeoNatan:
If it's only 5 seconds, that's fine 👍

PR #2145, opened 2020-06-23
Update docs across the board with respect to notification support on Android

### LINKED ISSUE #1192: Introduce basic support for push notifications mocking on Android

d4vidi:
- [ ] This is a small change 
- [ ] This change has been discussed in issue #1192 and the solution has been agreed upon with maintainers.

---

**Description:**
Updates the missing documentation with respect to changes introduced in #2134 and #2138, both associated with #1192 (i.e. android notification mocking support). Resolves #1192.

### DIFF
--- a/docs/APIRef.DeviceObjectAPI.md
+++ b/docs/APIRef.DeviceObjectAPI.md
@@ -29,7 +29,7 @@ The value will be `undefined` until the device is properly _prepared_ (i.e. in `
 - [`device.uninstallApp()`](#deviceuninstallapp)
 - [`device.openURL(url)`](#deviceopenurlurl-sourceappoptional)
 - [`device.sendUserNotification(params)`](#devicesendusernotificationparams)
-- [`device.sendUserActivity(params)`](#devicesenduseractivityparams)
+- [`device.sendUserActivity(params)` **iOS Only**](#devicesenduseractivityparams)
 - [`device.setOrientation(orientation)`](#devicesetorientationorientation)
 - [`device.setLocation(lat, lon)`](#devicesetlocationlat-lon)
 - [`device.setURLBlacklist([urls])`](#deviceseturlblacklisturls)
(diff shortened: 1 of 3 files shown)


d4vidi:
@LeoNatan please review so as to make sure this makes sense from on iOS perspective

docs/APIRef.DeviceObjectAPI.md
```diff
@@ -80,37 +82,36 @@ Detox uses [AppleSimUtils](https://github.com/wix/AppleSimulatorUtils) to implem
...
-await device.launchApp({url: url, newInstance: true}); //Launch a new instance of the app
-await device.launchApp({url: url, newInstance: false}); //Reuse existing instance
+await device.launchApp({url});
+await device.launchApp({url, newInstance: true}); // Launch a new instance of the app
+await device.launchApp({url, newInstance: false}); // Reuse existing instance
 ```
-This will launch the app from background and handle the deep link.
 
-Read more in [here](APIRef.MockingOpenWithURL.md).
+Read more [here](APIRef.MockingOpenWithURL.md). Go back to subsection 1 to read about the full effect of the `newInstance` argument.
```
LeoNatan, line 90: > Go back to subsection 1 to read about the full effect of the `newInstance` argument.

This is unnecessary for every section. People should read the guide if they want to understand how things work.

d4vidi: You're right, yet this gives a certain cue as to what the reader should pay attention to (the final behavior depends on `newInstance`). I've found it confusing a bit myself, reading this at first.

PR #2164, opened 2020-06-29
Consolidate (and split) setup onto a single getting-start guide [ci skip]

### LINKED ISSUE #2115: Unify platforms on the getting-started guide
While Android support is now mature, the [getting-started guide](https://github.com/wix/Detox/blob/master/docs/Introduction.GettingStarted.md) still address the two platforms in a non-equivalent fashion, making the setup process cluttered and less comprehensible.

Unify descriptions for both platforms. For example, the prerequisites section should refer the reader to two equivalent guides of the platforms (android already has a dedicated one, iOS' needs to be extract to such).

In this context, we can update the android support page as well.

I think it's time to remove this from the template

**Additional context**
N/A

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #2115 and the solution has been agreed upon with maintainers.

---

**Description:**
Resolves #2115. The change in a nutshell is as follows:
- Extraction of iOS specific instructions into equivalent android guides (dev-env, configuration+general setup)
- Generalization of getting-started guide with pointers to delegated guides where necessary
- Slight reorganization of the getting-start guide.

In order to review, consider simply going over the getting-started as a new detox user - as the diffs are impossible to interpret.

---
Rules of thumb I followed:
- DRY: tried my best not to have anything repeated. I managed to do so for the most part, except for two similar explanations regarding the configuration files resolution.
- (Somewhat related): Unify what's possible onto the main guide, and delegate onto platform-specific guides where things should be done in a platform-specific way. In the future, this would make it fairly easy to make the leap onto a single page with cascaded tabs (like react native's).
- Did my best to keep a workflow that makes sense: At any given point in the guide - tried not to address any worked-out detail if it hasn't been mentioned and explained, previously.
- Tried to be user-friendly about things 😄

### DIFF
--- a/docs/Guide.Mocha.md
+++ b/docs/Guide.Mocha.md
@@ -1,0 +1,35 @@
+# Mocha setup guide
+
+This guide describes how to install [Mocha](mochajs.org) as a test runner to be used by Detox for running the E2E tests.
+
+**Disclaimer:**
+
+* Here we focus on installing Detox on _new projects_. If you're migrating a project with an existing Detox installation, please apply some common sense while using this guide.
+
+
+
+## Installation
+
+### 1. Install Mocha :coffee:
+
+Before starting with Mocha setup, be sure to complete the preliminary sections of the [Getting Started](Introduction.GettingStarted.md) guide.
+
+```sh
+npm install mocha --save-dev --no-package-lock
+```
+
+### 2. Set up test-code scaffolds :building_construction:
+
+```sh
+detox init -r mocha
+```
+
+> **Note:** errors occurring in the process may appear in red.
+
+If things go well, you should to have this set up:
+
+- An `e2e/` folder in your project root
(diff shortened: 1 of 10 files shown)


LeoNatan:
Great!

docs/Guide.Mocha.md
```diff
@@ -0,0 +1,35 @@
+# Mocha setup guide
+
+This guide describes how to install [Mocha](mochajs.org) as a test runner to be used by Detox for running the E2E tests.
```
rotemmiz, line 3: I would mention that Mocha is a less preferred test runner to use, for two reasons. The first being its lack of multi worker support, and the second is the much more advanced integration with Jest Circus, and also some behaviours or bugs are only solvable with Circus.

rotemmiz: So if you want a 2018 retro feeling of Detox, use Mocha :)

d4vidi: I couldn't agree more, but will take care of it on a different scope 🙏

PR #2165, opened 2020-06-29
Integrate test-butler with ANR disabling

d4vidi:
- [ ] This is a small change 

---

**Description:**
Integrate Linkedin's awesome test-butler, which - among many other things, disables ANR's 🚀  (takes the assumption that a stock-android image is used... lucky us).
First creating a draft just to see how the CI responds.

### DIFF
--- a/detox/test/android/app/build.gradle
+++ b/detox/test/android/app/build.gradle
@@ -26,7 +26,7 @@ android {
             abiFilters 'armeabi-v7a', 'x86', 'x86_64'
         }
         testBuildType System.getProperty('testBuildType', 'debug')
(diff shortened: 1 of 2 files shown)


rotemmiz:
So this is just a demo of how to integrate right? Every project should do that on its own? Add link to docs/warm recommendation?

d4vidi:
@rotemmiz good question, I don't know yet. I wanna to gain mileage (and with that confidence), to see how deep we integrate this into the library. The main blocker is the need for aosp emulators. One compromise to consider, then, is to bundle this into Detox' existing custom test-runner (overriding) implementation, used for instruments, and instruct users to use that _always_.

PR #2168, opened 2020-06-30
Fixed issue with "undefined" as value passed to android instrum launch args

valentynberehovyi:


detox/android/detox/src/main/java/com/wix/detox/DetoxJUnitRunner.java
```diff
@@ -47,9 +49,15 @@ private void onBeforeAppOnCreate(Application app, Bundle arguments) {
         if (DetoxInstrumentsManager.supports()) {
             final String recordingPath = arguments.getString("detoxInstrumRecPath");
             if (recordingPath != null) {
-                final long samplingInterval = Long.parseLong(
-                        arguments.getString("detoxInstrumSamplingInterval", "250")
```
rotemmiz, line 51: } catch (NumberFormatException e) {
                   //log that a default is being used because input is invalid, but no stack trace, you don't want that stacktrace.
```

valentynberehovyi: Thanks! Changed.

PR #2169, opened 2020-06-30
Find devices for 'android.attached' based on 'adbName' pattern

DmytroKost:
For running parallel tests on attached devices

- [x] This is a small change 

---

**Description:**

I want to be able to run parallel tests on multiple attached Android devices (e.g. Genymotion). Currently, the `android.attached` configuration type accepts only one `adbName`.


noomorph:
@d4vidi, let's look together into this.

noomorph:
Preliminary speaking, looks good to me, except maybe for 1-2 very minor things. I'll give it a try locally.

d4vidi:
nice one - hope to have a look soon!

DmytroKost:
I noticed that when enabling artifacts (logs, videos), Detox fails to save them when running in parallel mode.

```
tests_1                | detox[188] DEBUG: [exec.js/SPAWN_CMD, #14] [pid=260] /opt/sdk/platform-tools/adb shell "logcat --pid=1930 -f /sdcard/62339204_0.log"
tests_1                | detox[188] DEBUG: [DetoxServer.js/CANNOT_FORWARD] role=testee not connected, cannot fw action (sessionId=c2967c73-58e6-887c-834b-4cba63136e9b)
tests_1                | detox[188] ERROR: [exec.js/SPAWN_STDERR, #14] error: more than one device/emulator


tests_1                | detox[174] ERROR: [exec.js/EXEC_FAIL, #22] ""/opt/sdk/platform-tools/adb"  shell du /sdcard/62743283_0.mp4" failed with code = 1, stdout and stderr:
tests_1                | 
tests_1                | detox[174] ERROR: [exec.js/EXEC_FAIL, #22] 
tests_1                | detox[174] ERROR: [exec.js/EXEC_FAIL, #22] error: more than one device/emulator
```

From the output I can see that it doesn't pass `-s device_name`.
That doesn't happen when running tests on local emulators. Did I miss something? I haven't yet spent time investigating the cause of the issue.

d4vidi:
@DmytroKost I've revisited this now and I feel there are 2 things to consider.

1. Implementation
Should we decide to accept this as a _feature_ (and I personally think it could definitely be the right evolution for _attached_ devices support!), there are two drawbacks in putting support inside a driver -- bloatedness of drivers and unit-test coverage. Therefore, we could only accept this after refactoring the device lookup onto a separate, full-tested (tdd) finder class - much like [FreeEmulatorFinder](https://github.com/wix/Detox/blob/51b51c94e64dfa0322cfe19da2695d4a38b9cac8/detox/src/devices/drivers/android/emulator/FreeEmulatorFinder.js#L6), which is the equivalent one for emulators. See how it's tested, [here](https://github.com/wix/Detox/blob/ea973989ea762dc3270231f70f7f377e1b910b8a/detox/src/devices/drivers/android/emulator/FreeEmulatorFinder.test.js#L1).

2. Configuration contract
I'm still not certain whether an adb-names array is the right way to represent this data. For that, I'll need @noomorph's insights...

noomorph:
@d4vidi, I'm about to review this today or maximum — tomorrow. 

As for your question 2 — well, I was considering also RegExp-based solution. This way we can elegantly solve many cases:

```
"emulator-\\d+" // use only spawned emulators 
".*" // use anything that is free
"device_1|device_2|device_3" // use this kinda array of devices
``` 

What do you think?

DmytroKost:
@d4vidi Yes, I can do that. Any insight into why artifacts don't work properly with multi-attached devices? https://github.com/wix/Detox/pull/2169#issuecomment-653050857

noomorph:
Cannot deliver my promise to review it the PR yesterday, obviously, have to repromise that for tomorrow.
Cannot really say without a deeper look, @DmytroKost. Sorry for the long wait, thanks for the patience.

DmytroKost:
@d4vidi @noomorph I've made the following changes:

1. Implemented `FreeAdbDeviceFinder` and added unit tests for it.
1. Extracted a parent class`FreeDeviceFinderBase` from `FreeEmulatorFinder` and reused it for  `FreeAdbDeviceFinder`. Same for the unit tests.
1. Implemented lookup strategy based or `adbName` being a regular expression.
1. Fixed the problem with artifacts by emitting a `bootDevice` event with the `devideId`.

noomorph:
@DmytroKost , if you don't mind, could you remerge master again into your branch, please?

DmytroKost:
> @DmytroKost , if you don't mind, could you remerge master again into your branch, please?

Done.

noomorph:
@DmytroKost, I think it makes sense to add:

1. A mention about the RegExp matching in the docs:

    a. [Introduction.Android.md](https://github.com/wix/Detox/blob/cbf6966b640948db7c92035229294a6621e5319c/docs/Introduction.Android.md#2-apply-detox-configuration)

        Connect to already-attached android device. The device should be listed in the output of adb devices command under provided name. Use this type to connect to Genymotion emulator.

    b. [APIRef.Configuration.md](https://github.com/wix/Detox/blob/78bc945bc5dbf15c116cceb790c8ae5c9aac9af4/docs/APIRef.Configuration.md#device-configuration) — see example with `android.att.release`, maybe you could add right there an example of how you can specify multiple devices.

2. An example in the configs:
https://github.com/wix/Detox/blob/cbf6966b640948db7c92035229294a6621e5319c/examples/demo-react-native/detox.config.js

At the very least, let's add some `emulator-.*` configuration with `android.attached` type, that can be used for testing that your PR works. I wonder, maybe it will make sense to start running a smoke test with that attached configuration in CI, @d4vidi , what do you think?

DmytroKost:
@noomorph Updated the docs and the `detox.config.js` example as requested.

noomorph:
@DmytroKost , see my suggestions in https://github.com/DmytroKost/Detox/pull/1

d4vidi:
Sorry, 'been overly booked in the last couple of days. I'll try my best to review asap.

DmytroKost:
@noomorph Thank you for the feedback. I'll wait for the review from @d4vidi before making changes so that I don't have to go back and forth.

d4vidi:
'Promise to review this this week, probably within a day or 2!

examples/demo-react-native/detox.config.js
```diff
@@ -51,6 +51,14 @@ module.exports = {
       "device": {
         "avdName": "Pixel_API_28"
       }
+    },
+    "android.att.release": {
```
d4vidi, line 55: Perhaps change to `android.genymotion.release`? a bit confusing to see both `attached` and the `emulator-.*` regexp in the same config

detox/src/devices/drivers/android/emulator/FreeEmulatorFinder.test.js
```diff
@@ -37,24 +37,7 @@ describe('FreeEmulatorFinder', () => {
...
-    expect(adbDevicesHelper.lookupDevice).toHaveBeenCalledWith(expect.any(Function));
-  });
-
-  it('should return the matched emulator', async () => {
-    const adbName = 'mock-adb-name';
-    adbDevicesHelper.lookupDevice.mockReturnValue(adbName);
-
-    const deviceName = await uut.findFreeDevice();
-    expect(deviceName).toEqual(adbName);
-  });
```
d4vidi, line 55: Would rather keep these here, and thus keep the test base-class agnostic

detox/src/devices/drivers/android/FreeDeviceFinderBase.js
```diff
@@ -0,0 +1,20 @@
+const AdbDevicesHelper = require('./tools/AdbDevicesHelper');
+
+class FreeDeviceFinderBase {
+  constructor(adb, deviceRegistry) {
+    this._adbDevicesHelper = new AdbDevicesHelper(adb);
+    this._deviceRegistry = deviceRegistry;
```
d4vidi, line 6: No need in holding this here IMO.

docs/APIRef.Configuration.md
```diff
@@ -63,7 +63,7 @@ Please find the [Detox example app](/examples/demo-react-native/detox.config.js)
         "build": "...",
         "type": "android.attached",
         "device": {
-          "adbName": "YOGAA1BBB412",
+          "adbName": "localhost:555[56]",
```
d4vidi, line 66: I think we need to keep the previous example as-is (perhaps expand to the 'array' form) and _add_ the new option as `android.genymoition.release` right after it

detox/src/devices/drivers/android/AttachedAndroidDriver.js
```diff
@@ -14,24 +23,35 @@ class AttachedAndroidDriver extends AndroidDriver {
...
-              `You can also try restarting adb with 'adb kill-server && adb start-server'.`
-      });
-    }
+    const adbNamePattern = _.isPlainObject(deviceQuery) ? deviceQuery.adbName : deviceQuery;
+
+    const adbName = await this._allocateDevice(adbNamePattern);
 
     await this.adb.apiLevel(adbName);
     await this.adb.unlockScreen(adbName);
+    await this.emitter.emit('bootDevice', { deviceId: adbName });
```
d4vidi, line 32: @noomorph does this make sense here?

DmytroKost: This step is necessary for the artefact managers to [register](https://github.com/wix/Detox/blob/master/detox/src/artifacts/templates/plugin/ArtifactPlugin.js#L106) the `deviceId`. And use it later to call ADB commands for that specific device when running in parallel mode https://github.com/wix/Detox/pull/2169#issuecomment-653050857

d4vidi: 👍 I'm double-checking seeing that we do not effectively _boot_ anything here

examples/demo-react-native/detox.config.js
noomorph: I asked @DmytroKost first of all to add it for the testing purposes. I am sure that Genymotion emulators do not follow `emulator-<port>` naming (!). But we can add something Genymotionish as well.

But maybe it's indeed good to rename it to `android.any.release` (att -> any).

DmytroKost: We connect to Genymotion via SSH tunnel, the names of devices will be `localhost:5555`, `localhost:5556`, .etc.

I can make an example for `android.genymotion.release` to demonstrate that, and one for `android.any.release` with a `.*` mark to demonstrate connection to any attached device/emulator. I thought `detox.config.js` for also for demo purposes. But since it's for testing, I'm not sure I should make any changes to it.

I also mentioned the example with a remote device in https://github.com/DmytroKost/Detox/blob/uic/adb-name-array/docs/APIRef.Configuration.md#device-configuration

noomorph: @d4vidi, your call.

detox/src/devices/drivers/android/attached/FreeAdbDeviceFinder.test.js
```diff
@@ -0,0 +1,17 @@
+const FreeAdbDeviceFinder = require('./FreeAdbDeviceFinder');
+
+describe('FreeAdbDeviceFinder', () => {
+  it("should test candidate's adbName against a regular expression", async () => {
+    const mockRegExpTest = jest.fn();
```
noomorph, line 5: Maybe I am missing something in this life, but can't we just duck test it, something like this?

```js
const onlyDigits =  "^\\d+$";

expect(await finder.isDeviceMatching({ adbName: '5555' }, onlyDigits).toBe(true);
expect(await finder.isDeviceMatching({ adbName: 'emulator-5555' }, onlyDigits).toBe(false);
```

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -51,7 +46,7 @@ class EmulatorDriver extends AndroidDriver {
     await this._avdValidator.validate(avdName);
     await this._fixEmulatorConfigIniSkinNameIfNeeded(avdName);
 
-    const adbName = await this._allocateDevice(avdName);
+    const adbName = await this.allocateDevice(avdName) || this._createDevice();
```
noomorph, line 49: I have reasons to believe that you mistyped something here, probably you meant:

```js
const adbName = await this.allocateDevice(avdName) || await this._createDevice();
```

because otherwise you risk writing a Promise inside `adbName` variable.

image

detox/src/devices/drivers/android/emulator/FreeEmulatorFinder.test.js
```diff
@@ -1,94 +1,32 @@
...
-      };
-
-      await uut.findFreeDevice();
-
-      const matcherFn = resolveMatcherFn();
-      expect(await matcherFn(device)).toEqual(false);
-    });
-
-    const resolveMatcherFn = () => adbDevicesHelperObj().lookupDevice.mock.calls[0][0];
+  const createDevice = (avdName, type = 'emulator') => ({
```
noomorph, line 26: I suggest moving out of `detox/src/devices/drivers/android/exec/ADB.js` those two classes:

* DeviceHandle
* EmulatorHandle

either to one file (e.g.., `DeviceHandle.js`), or two appropriate ones, let's say inside the folder `detox/src/devices/drivers/android/tools`.

Then you will be just able to use them in both of your tests, for `FreeDeviceFinder` and `FreeEmulatorFinder`.

Emulator: 

```js
const candidate = new EmulatorHandle('emulator-5555\tdevice');
jest.spyOn(candidate, 'queryName').mockImplementation(async () => 'mockAvdName');
expect(await uut.isDeviceMatching(candidate, 'wrongAvdName')).toBe(false);
```

Genymotion:

```js
const candidate = new DeviceHandle('192.168.60.101:6666\tdevice');
```

Attached device:

```js
const candidate = new DeviceHandle('MOCK_SERIAL\tdevice');
```

Incidentally, if you take a look at the implementation, you'll see we check for `candidate.type === 'emulator'` first not only for fun, but also because non-emulator device handles simply do not have this method, `async queryName()`, so it is not legit to have it always in your test doubles.

I believe the aforementioned steps will bring more cohesion into unit tests codebase because the reader will understand what those `candidate`s are in reality.

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.js
```diff
@@ -0,0 +1,40 @@
...
+
+  async findFreeDevice(deviceQuery) {
+    const { devices } = await this.adb.devices();
+    for (const candidate of devices) {
+      const isBusy = this.deviceRegistry.isDeviceBusy(candidate.adbName);
+      if (isBusy) {
+        log.debug({ event: DEVICE_LOOKUP_LOG_EVT }, `Device ${candidate.adbName} is busy, skipping...`);
+        continue;
+      }
+
```
noomorph, line 19: @d4vidi, while I was reviewing unit tests and the candidate class implementation, which is `DeviceHandle` class (or `EmulatorHandle` subclass), I've been struck by a question:

**Why don't we check that `candidate.status === 'device'` or at least `candidate.status !== 'offline'`?** Does not that makes some sense and adds a bit of safety for the edge-cases?

detox/src/devices/drivers/android/emulator/FreeEmulatorFinder.test.js
noomorph: P. S. Since you use candidates in 3 unit tests, it might make sense to create some `__mocks__/handles.js` helper file that exports one handle of each kind, like those mentioned above.

docs/APIRef.Configuration.md
noomorph: Agree. Here we can expand that to array, like `YOGAA1BBB412|YOGAA1BBB413` and add a Genymotion example separately.

examples/demo-react-native/detox.config.js
noomorph: @d4vidi , @DmytroKost , let's change it to ".*" and rename to `any` instead of `att`. It is just for local testing at the moment.

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
```diff
@@ -0,0 +1,87 @@
...
+      constructor(adb, deviceRegistry) {
+        super(adb, deviceRegistry);
+        this.isDeviceMatching = mockIsDeviceMatching;
+      }
+    }
+    uut = new FreeDeviceFinderImpl(mockAdb, mockDeviceRegistry);
+  });
+
+  it('should return null when there are no devices', async () => {
+    mockNoAdbDevices();
```
noomorph, line 23: Can't we just use `mockAdbDevices([]);`?

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
```diff
@@ -0,0 +1,87 @@
...
+    mockAllDevicesFree();
+    mockAnyDeviceMatches();
+    expect(await uut.findFreeDevice()).toEqual(null);
+  });
+
+  it('should return null when all devices busy', async () => {
+    mockOneAdbDevice();
```
noomorph, line 30: Can't we just use `mockAdbDevices([createDevice()]);`?

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
```diff
@@ -0,0 +1,87 @@
...
+      candidate,
+      createDevice('device3'),
+    ]);
+    mockAllDevicesFree();
+    mockNthDeviceMatches(2);
+    expect(await uut.findFreeDevice(deviceQuery)).toEqual(candidate.adbName);
+    expect(mockIsDeviceMatching).toHaveBeenCalledWith(candidate, deviceQuery);
+  });
+
+  it('should throw when `isDeviceMatching` is not overridden in a sub-class', async () => {
```
noomorph, line 60: Let's make it the first test in the suite.

Also, you can use `uut.findFreeDevice.restoreOriginalImplementation()` or something like that in Jest. No need to recreate it, I think. Look in the docs, it should be there.

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
```diff
@@ -0,0 +1,87 @@
...
+  const mockAdbDevices = (devices) => mockAdb.devices.mockResolvedValue({ devices });
+  const mockNoAdbDevices = () => mockAdbDevices([]);
+  const mockOneAdbDevice = () => mockAdbDevices([createDevice()]);
+
+  const mockAllDevicesBusy = () => mockDeviceRegistry.isDeviceBusy.mockReturnValue(true);
+  const mockAllDevicesFree = () => mockDeviceRegistry.isDeviceBusy.mockReturnValue(false);
+
+  const mockAnyDeviceMatches = () => mockIsDeviceMatching.mockResolvedValue(true);
+  const mockNoDeviceMatches = () => mockIsDeviceMatching.mockResolvedValue(false);
+  const mockNthDeviceMatches = (n) => {
```
noomorph, line 77: Can't we make it like `mockDeviceMatchesName`, where you make a very simple implementation like `async (candidate, adbName) => candidate.adbName === adbName`? The nth call logic looks not so readable.

detox/src/devices/drivers/android/attached/FreeAdbDeviceFinder.test.js
DmytroKost: Changed

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
DmytroKost: Fixed

detox/src/devices/drivers/android/emulator/FreeEmulatorFinder.test.js
DmytroKost: Done

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.js
DmytroKost: Added a check for 'offline' devices

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
DmytroKost: Changed

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
DmytroKost: Changed

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
DmytroKost: > Let's make it the first test in the suite.

Done

> Also, you can use uut.findFreeDevice.restoreOriginalImplementation() or something like that in Jest. No need to recreate it, I think. Look in the docs, it should be there.

I simplified the mocking logic. However, I left the `uut` instantiation in `beforeEach` because other unit tests follow same approach (15 test suits, e.g. _src/devices/drivers/android/AndroidDriver.test.js_).

detox/src/devices/drivers/android/tools/FreeDeviceFinderBase.test.js
DmytroKost: Done

DmytroKost:
> We are approaching the finish line, let's make one more effort and we'll be there. 👍 Good job.

Effort made :)

detox/src/devices/drivers/android/tools/DeviceHandle.js
```diff
@@ -0,0 +1,22 @@
+class DeviceHandle {
+  constructor(deviceString) {
+    const [adbName, status] = deviceString.split('\t');
+    this.type = this.inferDeviceType(adbName);
+    this.adbName = adbName;
+    this.status = status;
+  }
+
+  inferDeviceType(adbName) {
```
noomorph, line 9: Could you mark it by underscore convention as private (`_inferDeviceType`), please?

detox/src/devices/drivers/android/FreeDeviceFinderBase.js
noomorph: Is the concern resolved here, @d4vidi?

examples/demo-react-native/detox.config.js
noomorph: Any progress here? @d4vidi, is the `.*` suggestion okay?

docs/APIRef.Configuration.md
noomorph: So, shall we? @d4vidi

noomorph:
What's left is:

1. To decide finally what about those examples in configs and docs - we have two disputable places, AFAIC. That should be very quick to solve.
2. My small request to make `inferDeviceType` privateish `_inferDeviceType`.
3. @d4vidi had some unresolved comments for `FreeDeviceFinderBase.js ` and `FreeEmulatorFinder.test.js`, but honestly speaking, the files look good enough for me. Do we need anything else?

For me, looks like 99% is done.

detox/src/devices/drivers/android/tools/DeviceHandle.js
DmytroKost: Changed

d4vidi:
Well it appears many things have changes here since I checked last... 😆 
I'll go over it again from scratch

d4vidi:
Our CI is acting up. Ignore the results for now. I'll keep you posted.

examples/demo-react-native/detox.config.js
d4vidi: Why `.*`? does it really make sense to match all-the-things? How about we negate `emulator-`? `(?!(emulator))`, I believe...

Anyways on 2nd thought no point in adding an extra genymotion config here - I agree.

DmytroKost: Or iso negating, assert that it's a remotely connected device with [this pattern](https://github.com/DmytroKost/Detox/blob/uic/adb-name-array/detox/src/devices/drivers/android/tools/DeviceHandle.js#L14)

d4vidi:
So besides my minor reply to one of the comments, here are the 2 current concerns:
1. Seems that somehow the PR is wreaking havoc on our CI - [2491](https://jenkins-oss.wixpress.com/job/multi-detox-pr/2491/), [2490](https://jenkins-oss.wixpress.com/job/multi-detox-pr/2490/). We must figure out what's going on there before moving forward, of course.
2. I do not currently agree with the current unit-testing approach over the `FreeXYZFree` class family. It doesn't make sense to me that the tests associated with the concrete classes (`FreeEmulatorFinder`, `FreeAdbDeviceFinder`) only cover the logic of the protected `isDeviceMatching()` method <sup>*</sup>. My opinion is that unit tests should only invoke public API's, and should be as inheritance+private API's agnostic as possible. In that sense, The coverage over `FreeDeviceFinderBase` is completely unnecessary and should be cloned and transferred (twice, separately) to the unit tests of each of the concrete class' tests.

@DmytroKost if you find 2 to be too heavy at this point, let's focus and 1 and we'll sort it out later.

---
\* should be changed to `_isDeviceMatching`

DmytroKost:
@d4vidi I can remove unit tests for `FreeEmulatorFinder` and `FreeAdbDeviceFinder`. Both implementations of `isDeviceMatching` are straight-forward and the value of unit-testing them is low.

Any clue why CI fails? Those failing E2E tests look flaky to me. Otherwise, why would different tests fail from different runs?

d4vidi:
I find the tests strictly for `isDeviceMatching` useless for a different reason -- because it is a protected method. They should be bundled with calls to `findFreeDevice`.

d4vidi:
As per CI - trying to figure out myself right now. Away from the scope of this PR, the builds seem stable...

DmytroKost:
> I find the tests strictly for `isDeviceMatching` useless for a different reason -- because it is a protected method. They should be bundled with calls to `findFreeDevice`.

I'm not sure what you're suggesting

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -63,6 +58,10 @@ class EmulatorDriver extends AndroidDriver {
     return adbName;
   }
 
+  async doAllocateDevice(deviceQuery) {
+    return this.freeDeviceFinder.findFreeDevice(deviceQuery);
```
d4vidi, line 62: @DmytroKost missing an `await` here
```suggestion
    return await this.freeDeviceFinder.findFreeDevice(deviceQuery);
```

noomorph:
@DmytroKost , I think @d4vidi suggests to copy `FreeDeviceFinderBase.test.js` twice and delete. The copies will be called `FreeDeviceFinder.test.js` and `FreeEmulatorFinder.test.js`. Both copies will run very similar test suites except for a few emulator-specific and device-specific cases, e.g.: `FreeDeviceFinder` is indifferent about whether a device is an emulator or not —  it just should match the specified regular expression, while `FreeEmulatorFinder` cares exclusively about emulators and a correct AVD name; however, both test suites will make sure that those finders omit offline and busy devices — that is the common part for both implementations.

I hope I've managed to convey the message.

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
noomorph: @d4vidi, why do you want an explicit `await`  there?

detox/src/devices/drivers/android/attached/AttachedAndroidDriver.js
```diff
@@ -0,0 +1,35 @@
...
+    await this.adb.apiLevel(adbName);
+    await this.adb.unlockScreen(adbName);
+    await this.emitter.emit('bootDevice', { coldBoot: false, deviceId: adbName, type: 'device' });
+
+    this._name = adbName;
+    return adbName;
+  }
+
+  async doAllocateDevice(deviceQuery) {
+    return this.freeDeviceFinder.findFreeDevice(deviceQuery);
```
d4vidi, line 31:
```suggestion
    return await this.freeDeviceFinder.findFreeDevice(deviceQuery);
```

d4vidi:
@DmytroKost please rebase your branch over Detox' latest master. I believe this should solve CI issues.

DmytroKost:
@d4vidi I merged `FreeAdbDeviceFinder` to `FreeDeviceFinder` with the following changes:

-  Added an underscore to `_isDeviceMatching`
-  Renamed `FreeDeviceFinderBase` to `FreeDeviceFinder`
-  Removed `FreeAdbDeviceFinder`
-  `FreeDeviceFinder._isDeviceMatching` tests candidate's adbName against a regex by default (copied the implementation from `FreeAdbDeviceFinder._isDeviceMatching`)
-  `FreeEmulatorFinder` extends `FreeDeviceFinder` and overrides `_isDeviceMatching`
-  Moved the unit test from `FreeAdbDeviceFinder.test.js` to `FreeDeviceFinder.test.js` and removed `FreeAdbDeviceFinder.test.js`
-  `FreeEmulatorFinder.test.js` tests only the `_isDeviceMatching` override through the public method `findFreeDevice`

DmytroKost:
@d4vidi Rebase on master

examples/demo-react-native/detox.config.js
DmytroKost: @d4vidi @noomorph Reverted changes to the `detox.config.js` file. Moved the example with `android.genymotion.release` to `docs/APIRef.Configuration.md`. What do you think?

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
d4vidi, line 49: @DmytroKost this might be the CI stability issue: `this._createDevice()` must be called from within `allocateDevice`, or the device wouldn't be locked... I'll try to fix that myself, in a bit...

d4vidi:
```suggestion
    const adbName = await this.allocateDevice(avdName);
```

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -63,6 +58,10 @@ class EmulatorDriver extends AndroidDriver {
     return adbName;
   }
 
+  async doAllocateDevice(deviceQuery) {
+    return await this.freeDeviceFinder.findFreeDevice(deviceQuery);
```
d4vidi, line 62:
```suggestion
    return await this.freeDeviceFinder.findFreeDevice(deviceQuery) || this._createDevice();
```

d4vidi:
Green at last!!!

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
d4vidi: Good point - `createDevice()` is relevant for emulators only as it synth's an emulator's name with a randomly allocated port

noomorph:
WOOT!!

PR #2174, opened 2020-07-01
Rework DetoxJUnitRunner checkers to reprioritize extra reflection check

valentynberehovyi:


rotemmiz:
Can you briefly explain what the problem was?

valentynberehovyi:
> Can you briefly explain what the problem was?

As was discussed with @d4vidi - DetoxJUnitRunner may be extended for TestButler and will be used for not only detox-instruments use cases. So to avoid non-needed reflection access for another DetoxJUnitRunner use case we decided to re-prioritize checks.

d4vidi:
Even regardless of test butler, this is better practice - restrict making intrusive `instruments` checks to when the user has explicitly asked for it. No need in failing over a potential instruments bug in a system where it hasn't been applied.

detox/android/detox/src/main/java/com/wix/detox/DetoxJUnitRunner.java
```diff
@@ -46,9 +46,9 @@ public void onDestroy() {
     }
 
     private void onBeforeAppOnCreate(Application app, Bundle arguments) {
-        if (DetoxInstrumentsManager.supports()) {
-            final String recordingPath = arguments.getString("detoxInstrumRecPath");
-            if (recordingPath != null) {
+        final String recordingPath = arguments.getString("detoxInstrumRecPath");
+        if (recordingPath != null) {
+            if (DetoxInstrumentsManager.supports()) {
```
d4vidi, line 51: @valentynberehovyi can u double check that `getString` returns null if key doesn't exist?

valentynberehovyi: From the android framework description.
```
   /**
     * Returns the value associated with the given key, or null if
     * no mapping of the desired type exists for the given key or a null
     * value is explicitly associated with the key.
     *
     * @param key a String, or null
     * @return a String value, or null
     */
    @Nullable
    public String getString(@Nullable String key)
```

So null is expected here.

PR #2185, opened 2020-07-05
Add support for Detox Recorder in Detox CLI

LeoNatan:
This adds support for [DetoxRecorder](https://github.com/wix/DetoxRecorder) in Detox CLI.

Now, when running `detox recorder`, CLI will test to see if Recorder is installed and forward that binary the commands. Otherwise behave the same as before.

### DIFF
--- a/detox-cli/cli.js
+++ b/detox-cli/cli.js
@@ -15,14 +15,29 @@ if (fs.existsSync(detoxPackageJsonPath)) {
       { stdio: 'inherit' });
     process.exit(result.status);
   } else {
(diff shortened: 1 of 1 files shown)


LeoNatan:
How to release a new CLI release?

rotemmiz:
It should be automatic, Lerna recognises that there's a diff in this project and bumps version/releases.

LeoNatan:
You mean in CI?

LeoNatan:
I changed the external CLI, the global one, not the local-cli.

PR #2188, opened 2020-07-07
Test butler app apk

d4vidi:
- [ ] This is a small change 

---

# Description

Add the missing piece in integrating `Test-Butler` into Detox' test app, and with that - enable its integration in _any_ detox project (although the solution isn't pretty, just yet). This in fact is based on the initial work done in #2165.

## More Details

In order for test butler to work, it's not enough to just apply the `aar` into Detox app, because it actually consists of 2 separate parts:
1. A custom, system-level APK that needs to be preinstalled on the tested device.
2. An API that enables / controls it (this is the `aar` part).

_Naturally, without the APK, API usage will fail, and ANR's alert disabling will not be set-up._

With pure Android-native projects, by using the Android-Gradle plugin's `androidTestUtil`, the APK is just automagically preinstalled when running `gradle connectedCheck` (or one of its flavor-suited dependant-tasks). With Detox, we do things ourselves, which makes efficiently installing the APK dynamically, a challange. This PR solves it, then, in a way that is twofold.

###  1. Installing the APK

Assuming the APK is available somewhere in the system, Detox needs to allow for its installation on each running Android device, be it real or an emulator. Note that sometimes the emulator is booted by Detox itself...
Also, we want to install it once and move on, and **not follow** the rules of reinstalling the app, etc.

To work that out, I've allowed for an array of binary paths to be configured under a new field called `utilBinaryPaths` inside Detox' config, that goes alongside the usual things such as `binaryPath` and `testBinaryPath`.

### 2. Downloading the APK

Preferably, the APK should be prebaked into the device image. But that's not always possible (e.g. SaaS CI's). As an alternative, it can be added to the git repo, but that is discourage.

So the fully-blown solution here is to download it manually from bintray. And quite importantly, that should be done only once, before everything starts. Because detox currently does not allow for a global initialization (rather, a distributed, per-run init is done), as an original (yet hackish) solution, I've utilized our fresh ability to create javascript based configuration files (see inside).


noomorph:
I have two questions here.

1. A minor one. Could we use an array for `binaryPath` property itself, without creating one more property (`utilBinaryPaths`)? I don't see a good way either (if anything), but I feel it is right to raise this question anyway.
1. A major one. Don't you think that the new behavior is a bit inconsistent with the `reuseApp` flag — the installation seemingly happens in any case? Also, I think I see no way to delete those utility binaries except for if you run `adb uninstall` manually and directly, not via Detox...

🤔

d4vidi:
@noomorph thanks for these super important questions! I think that though there are 2, the answer is one.

As you know, in Detox, we take ownership of various aspect of managing the test devices, as a service to our users. For example, on Android, apart from allocating and booting emulators (which in its own is a pretty big responsibility), we take care of disabling all native animations, for the user's sake.

This feature continues this line. By allowing for _util_ binaries to be set up, we provide yet another means to help users bring their devices to a setup suitable for their environment. To even take this a step further, we could one day consider providing a proper hook where users would have complete access to the allocated devices and tweak whatever they wish in there. But that would be an overkill for what we _currently_ need, I feel.

So in that sense, these util-APK's are not *app* APK's, and they are not to be tested as such. Hence, they are out of scope when it comes to things like `--reuse`, or even app reinstall. In the driver implementation, I even went as far as to precheck whether the APK's have already been installed on the device, so as to avoid reinstalling them (they don't need to be cleaned-up, etc., as they are utils). Uninstalling them doesn't make all that much sense, too - if they're already there, great, it's what we want! :)
Last but not least - again going with the same line of thought here, util binaries therefore exist regardless of whether we support a single _app_ binary, or several ones in `binaryPath`... To me, that's a different matter.

d4vidi:
@noomorph that said, I do see a use case that could challenge what I'm suggesting: An update to an existing util APK. This is a case where a user _would_ like to reinstall it. Although, in most cases, the emulators start fresh, and if they don't -- the user should be admin-ing their state.

detox/src/devices/drivers/android/EmulatorDriver.js
```diff
@@ -73,9 +72,7 @@ class EmulatorDriver extends AndroidDriver {
       binaryPath,
       testBinaryPath,
     } = this._getInstallPaths(_binaryPath, _testBinaryPath);
-
-    const installHelper = new AppInstallHelper(this.adb, this.fileXfer);
-    await installHelper.install(deviceId, binaryPath, testBinaryPath);
+    await this.appInstallHelper.install(deviceId, binaryPath, testBinaryPath);
```
d4vidi, line 75: Related indirectly: use the super-class' instance, now that it's there.

detox/src/devices/Device.js
```diff
@@ -180,6 +181,19 @@ class Device {
     await this.deviceDriver.uninstallApp(this._deviceId, _bundleId);
   }
 
+  async installUtilBinaries() {
+    const paths = this._deviceConfig.utilBinaryPaths;
+    if (paths) {
+      if (!_.isArray(paths)) {
+        throw new DetoxRuntimeError({
```
noomorph, line 188: @d4vidi, I think this is very far from its origination. What if we move that validation inside `detox/configuration/*` layer?

detox/src/devices/Device.js
```diff
@@ -180,6 +181,19 @@ class Device {
...
+          message: `utilBinaryPaths specified in Detox configuration is invalid: ${paths}`,
+          hint: 'Configuration must be applied as an Array of paths',
+        });
+      }
+      await this.deviceDriver.installUtilBinaries(this._deviceId, paths);
```
noomorph, line 193: Can't we make 1-size-fits-all `this.deviceDriver.installApp(deviceId, binaryPath)` signature?

Is there any fundemantal difference between `this.deviceDriver.installUtilBinaries` and `this.deviceDriver.installApp` except for spread arguments?

detox/src/devices/drivers/android/AndroidDriver.js
```diff
@@ -71,8 +74,16 @@ class AndroidDriver extends DeviceDriverBase {
...
-    const uninstallHelper = new AppUninstallHelper(this.adb);
-    await uninstallHelper.uninstall(deviceId, bundleId);
+    await this.appUninstallHelper.uninstall(deviceId, bundleId);
+  }
+
+  async installUtilBinaries(deviceId, paths) {
+    for (const path of paths) {
+      const packageId = await this.getBundleIdFromBinary(path);
+      if (!await this.adb.isPackageInstalled(deviceId, packageId)) {
+        await this.appInstallHelper.install(deviceId, path);
```
noomorph, line 84: What if the device has older version installed? Should we overwrite, hence update it?

detox/src/devices/Device.js
d4vidi: Yes - will revisit!

detox/src/devices/Device.js
d4vidi: I took the time to think about it and decided not to because the use-case and impl. here are different, and because I like being explicit (it makes it easier to make changes, later...). To prove my point, `installApp` currently expects two binaries and as a contract uses a default if no test-binary is specified, and I don't want to break it.

detox/src/devices/drivers/android/AndroidDriver.js
d4vidi: See my answer to the previous comment... also, I guess I'm allergic to single functions taking many arguments which result in various `if`'s, etc.

d4vidi: As per version overwrite, I admit I'm not entirely comfortable about it as well, but this isn't a real-life use case ATM. I will try to scratch my head again and see whether I could make it feasible nonetheless.

d4vidi:
@noomorph one thing's for sure, this PR screams that we really need to refactor for a global detox setup and teardown -- much more than before.

detox/src/devices/Device.js
d4vidi: updated! see https://github.com/wix/Detox/pull/2188/commits/936e164e3822f09d005d3bf7a48316fef98add0c

noomorph: I see, but let's refine how the error will be looking on the user screen. Should be neutral, terse, and user-friendly.

detox/src/devices/Device.js
noomorph: @d4vidi , let's check again we are on the same page:

> 1. To prove my point, installApp currently expects two binaries and as a contract uses a default if no test-binary is specified.

First of all, isn't it a bad contract, considering that iOS simply does not need this quirk at all? I reckon that the optional `testBinary` parameter is not generic, rather quite platform- and implementation-specific, no?

On the other hand, I have considered the difficulty of moving these parts, and currently, I don't see a cleaner solution than the one you had suggested. Although, I feel like a heavy technical debt is forming here. Hence, I want to ask you in a roadmap context (and not of this PR):

Don't you think that a wiser decision would be to reimplement eventually `Device` class by splitting it into several stateful platform-specific implementations such as `SimulatorDevice`, `AttachedAndroidDevice`, `AndroidEmulatorDevice` and so on? The fact that we have a monolithic one-size-fits-all `Device` that has a complex state, yet it manages (with a struggle, albeit) to somehow function by juggling stateless `*DeviceDriver` methods has been saddening me for quite long. Making Device subclasses does not cancel code reuse, anyway, but the current way of handling is making configuration messy while making superficial visibility that is uniform.

Won't it be just relieving to have
```
{ type: 'ios.simulator' } -> gets handled by SimulatorDevice, that has its own schema
{ type: 'android.emulator' } -> gets handled by EmulatorDevice, that has its own and another schema
```
?

detox/src/devices/drivers/android/AndroidDriver.js
noomorph: 👍  @d4vidi , probably you are right here. Just trying to wrap my head around how we publish this API to the general public... 🤔  The current implementation level feels like protected/internal, and we are adding one more thing to the public Device class and its configuration, which by this PR design is not intended to work on both platforms. 🤯 I still feel like we might add some tech debt by this, this is legit, but eventually we should think on total rewriting Device class.

d4vidi: I thought this over and realized that in order to properly solve this, we need to have global init & teardown. Only then, will I be able to make sure that the APK is installed _once_ - overwriting any existing version, and then refrain from doing so in each suite.

I'll work out an issue explaining what needs to be done and what is blocked by it (this isn't the only thing, of course).

d4vidi: As per platform, well - there's technically no one to take this on iOS. I'm as unsatisfied as you are.

detox/src/devices/Device.js
d4vidi: ok, will apply all your suggested fixes asap 🙏

detox/src/devices/Device.js
d4vidi: Honestly, it's like you've read my mind! I went back and forth with this idea in my head a fair share of times.
I definitely agree the the state-less-ness of the drivers has no value. Seriously, passing in the `deviceId` in each call? this is a high-level coding environment... we should apply more advanced techniques (sorry for mocking).

As per the lack of their necessity altogether? I can't clearly conclude I agree with _that_. There's value in delegating the platform-specifics onto a different domain, making the `Device` still inheritable potentially for other (more profitable?) needs. But being a bloated class that it is right now? probably a bad idea.

detox/src/devices/Device.js
d4vidi: please revisit 😄 https://github.com/wix/Detox/pull/2188/commits/d875883b843e85079406d8bbdb2f5fb4b37d3c4a

PR #2191, opened 2020-07-07
ci(coveralls): use nyc instead of Jest built-in coverage

noomorph:
- [x] This is a small change 

---

**Description:**

Bypass Jest's built-in coverage calculation, because it is not designed to calculate coverage globally, in files that `environment` requires very early. Instead, I suggest using NYC, since it is agnostic to Jest's internals and works globally for everything that is required on run.


d4vidi:
@noomorph can you fix our coverage badge in the main README? It is still associated with coveralls. Can it be aligned with nyc reports?

PR #2194, opened 2020-07-08
Switch to downloading test-butler APK in jest's global-setup

d4vidi:
- [x] This is a small change 

---

**Description:**
Fixes some wrong-doings of #2188 🤭
Explanation: I've intially had plans of using the globally set temp dir (i.e. $TMPDIR) for this, such that I would have to dynamically alter the path with its resolved value. This is not needed with the current impl after all, and hence global-setup can suffice.


d4vidi:
@noomorph you think we should restore detox' config onto `package.json` - as it was, as well?

rotemmiz:
I'm quite worried about this kind of implementation. What if there's no internet connection? (FB run their CI disconnected from the internet for example).

rotemmiz:
Can't we just provide the apk?

d4vidi:
This is strictly for detox' test app.

d4vidi:
@rotemmiz see https://github.com/wix/Detox/blob/66ba64ecbe2632da0dd6dbf1ce230aa1b2a022c5/docs/Introduction.Android.md#8-test-butler-support-optional

rotemmiz:
Nice!

PR #2198, opened 2020-07-09
Detox Next (iOS, Phase 2)

### LINKED ISSUE #2182: Improve debug-sync print timings
Debug-sync only works on iOS and only for invokes. If an app starts and never goes idle, there is no way to find about it automatically.
* we can manually add `'detoxPrintBusyIdleResources': 'YES'` to launch args but this is not an automatic mechanism, but rather a debugging mechanism.

- Implement output to log on app launch, not only during slow Invokes (can be done regardless of DetoxNext Phase 2)
- Adding debug-sync by default (only during/after DetoxNext Phase 2, since the current mechanism is not crash free).

### LINKED ISSUE #2199: Use new sync status format in JS in Detox Next Phase 2
The new format is just a string which displays the information. For example,

```
The system is busy with the following tasks:

• Delayed perform selector: “_accessibilitySetUpQuickSpeak” on “<DetoxApp: 0x7ff4eac04c10>”
• Delayed perform selector: “_tryAndHandleError:” on “<RCTCxxBridge: 0x7ff4eaf362c0>”

• 9 work blocks on dispatch queue “<OS_dispatch_queue_main: com.apple.main-thread>”

• Run Loop (<Main Run Loop>)
• Run Loop (<JS Run Loop>)

• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Animation (“UIView animation with duration: “0.25” delay: “0.1898720026016235””)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)
• Runloop Perform Block (“<JS Run Loop>”)

• Timers: (
    "Timer with fire date: \U201c2020-07-09 21:16:02 +0300\U201d interval: \U201c0\U201d repeats: \U201cNO\U201d"
)
```

### LINKED ISSUE #2469: Jest-Jasmine and (older) Jest-Circus deprecation in favour of a newer implementation
Detox 19 is dropping support for the older Jest adapters.

### How to tell if you are affected by this issue?

Check your `e2e/init.js` (or the file you use in `setupFilesAfterEnv` strings array in your Jest config). If you have calls to `adapter.beforeEach` or `adapter.afterEach`, like shown below, then yes, you are affected:

```js
beforeEach(async function() {
  await adapter.beforeEach();
});

afterAll(async () => {
  await adapter.afterAll();
  await detox.cleanup();
});
```

### What shall you do to migrate to the new `jest-circus` integration?

Follow this [Jest setup guide](https://github.com/wix/Detox/blob/master/docs/Guide.Jest.md). _This section might be amended later._

### Why this is happening?

At the moment, Detox has three different ways to integrate with Jest, which is too much to support at once. Only the strongest shall survive, and that is the latest `jest-circus` integration introduced in 2020 with `detox@16.9.2`.  

Overall, it is more future-proof since it provides fine-grained control over the tests' lifecycle. To be more specific, it dramatically improves error handling (e.g., immediate screenshots on test failure) and provides more stability (e.g., in cases when `detox.init()` call fails early and some others).

### LINKED ISSUE #2487: Allow CLI args/flags as environment variables
I apologize in advance if this feature/enhancement is already asked. I can't find it on the issues and I don't know the right term/query to search.

Not really a detox problem. But, the official Detox + Cucumber integration is not working. Because cucumber does not allow CLI args it does not recognize to pass through. There is already an [ongoing discussion here](https://github.com/cucumber/cucumber-js/issues/1474).

To solve this issue, all CLI args should have an environment variable counterpart.
They should be prefixed with `DETOX_` so that they will not conflict with other variables.

For example.

`--configuration platform.flavor` is equal to `DETOX_CONFIGURATION=platform.flavor`
`--headless` is equal to `DETOX_HEADLESS_MODE=1`

The second solution I'm thinking of is, allow these args to be set into the `detxorc` files so that no need for the extra `DETOX_*` environment variables, and instead it's up to the user to set these variables.

For example.
`.detoxrc.js`
```
module.exports = {
  testRunner: 'jest',
  runnerConfig: 'e2e/config.json',
  detox: {
	// Something like this?
	args: {
    	headless: process.env.MY_DETOX_HEADLESS_VALUE,
		defaultConfiguration: process.env.MY_DETOX_CONFIGURATION || 'android.emu.release',
	},
    behavior: {
      init: {
        exposeGlobals: false,
        reuse: true,
      },
    },
  },
  configurations: {
    'ios.sim.release': {
      type: 'ios.simulator',
      build: 'xcodebuild -workspace ios/rnsandbox.xcworkspace -scheme rnsandbox -configuration Release -sdk iphonesimulator -derivedDataPath ios/build',
      binaryPath: './ios/build/Build/Products/Debug-iphonesimulator/rnsandbox.app',
      device: {
        type: 'iPhone 12'
      }
    }
  }
}
```

**Additional context**
None

LeoNatan:
iOS:

- [x] Remove Earl Grey in favor of Detox Sync
- [x] Implement URL blacklist feature in Detox Sync
- [x] Implement synchronization enable/disable in Detox Sync
- [x] Change JS to use new sync status format (#2199)
- [x] Decouple sync status from isReady request in JS (#2182)
- [x] Profile DetoxSync and improve performance
- [x] Implement touch indicator

Closes #2199
Closes #2182

JS:

- [x] Drop launchApp support in detox.init() (TODO: create an issue)
- [x] Prefixed environment variables (#2487)
- [x] Add migration notice about fix in launchArgs quoting (for iOS)
- [x] Write migration notes for JS part (1 - `super(config, context)`, 2 - `launchApp`)
- [x] add `--capture-view-hierarchy` flag to CLI

Docs:
- [x] add `uiHierarchy` docs.

Postponed for next minor release:
- Move community TypeScript definitions to Detox repo
- Write a JS helper for reminding users to launch the app when they forget

Postponed for next major release:
- Drop deprecated Jest adapters (#2469)
- Update Troubleshooting.RunningTests.md with this common issue.
- https://github.com/wix/Detox/blob/DetoxNext/docs/APIRef.DeviceObjectAPI.md - update launch args with blacklist example

Closes #2469
Closes #2487


LeoNatan (commented):
-

scripts/ci.publish.js
```diff
@@ -1,4 +1,4 @@
-/* tslint:disable: no-console */
+/* tslint:disable: no-console */
```
LeoNatan, line 1: CDD

examples/demo-react-native/e2e/init.js
```diff
@@ -3,6 +3,9 @@ const adapter = require('detox/runners/mocha/adapter');
 
 before(async () => {
   await detox.init();
+  await detox.launchApp({
+    newInstance: true,
```
LeoNatan, line 7: This is not needed, it will launch a new instance here regardless.

examples/demo-react-native-jest/e2e/setup.js
```diff
@@ -0,0 +1,3 @@
+beforeAll(async () => {
+  await device.relaunchApp();
```
LeoNatan, line 2: This should be `launchApp()`, not the deprecated API.

examples/demo-native-ios/e2e/example.spec.js
```diff
@@ -1,6 +1,6 @@
 describe('Example', () => {
   beforeEach(async () => {
-    await device.launchApp({ newInstance: true });
+    await device.launchApp();
```
LeoNatan, line 3: Should have `newInstance: true` here.

LeoNatan:
@noomorph When you add a warning about app not launched in init, remember pls to update Troubleshooting.RunningTests.md with this common issue. Thanks

rotemmiz (approved):

PR #2214, opened 2020-07-15
Bump top-level matchers api to upper Detox layer (where they should be)

### LINKED ISSUE #2012: [feature] Element Screenshot

none. 

**Additional context**

d4vidi:
- [ ] This is a small change 

---

## Description

Relates to #2012: lays some preliminary ground work so that we could introduce it with proper API.

### Dive-in

Looking at the current software architecture of Detox at a high-level perspective, we see two main **distinct** units, top-to-bottom:
1. A framework for providing `expect` and element-interactions as a top-level, user-facing API (e.g. `element(by.id(testId)).tap()`), which in turn uses a distinct, platform-specific node-side (JS) implementation that eventually directly triggers device-side logic using `invoke`.
2. A framework for providing element-less, device-scope, top-level, user-facing APIs (e.g. `device.takeScreenshot()`), which are provided mainly by the (currently) singleton `Device` object, that in turns delegates platform-specific logic to Drivers.

While the two global _units_ are, in principle, agnostic, we've nevertheless created an artificial coupling between them, as the top-level implementation of unit 1 is created via the drivers of unit 2, due to convenience ([example](https://github.com/wix/Detox/blob/78bc945bc5dbf15c116cceb790c8ae5c9aac9af4/detox/src/devices/drivers/android/AndroidDriver.js#L36)). Unfortunately, this type of an unnecessary coupling - as a rule of thumb, damages flexibility, and we can see it in the case of elements' screenshot.

The pseudo-code implementation of taking an element screenshot is as follows:
a. `Invoke` an on-test-device, native method that creates an on-device `.png` file.
b. Pull the image back to the host through the driver (e.g. using `adb`) and provide the path back to the caller.

Ideally, this top-level API for doing this would be amazing:
```js
const screenshotFilePath = await element(by.id('fancy-element-id')).takeScreenshot()`
```

But due to the coupling mentioned earlier, this cannot be done, as the underlying `Element.takeScreenshot()` function that would implement this -- originally generated by driver, cannot access the `device` in _part b_ of the pseudo-code (!) It more so artificially forces us onto this kind of an API, which is largely undesired as it is both not friendly and would encourage code duplication and single-responsibility breakage in Detox itself:

```js
const screenshotFilePath = await device.takeScreenshot( element(by.id('fancy-element-id')) );
```

If we want to nevertheless stick to the API that we really want - and with that, empower high flexibility in Detox (yes, gimme all those things!), we need to break the coupling.

**This is the scope of this PR:** 100% Separating drivers and matchers, and their creation. Instead of having the matchers created by the drivers, we delegate the proper matcher resolution to a new util called `matchersRegistry` (equilvant to `driverRegistry`), which is utilized in the Detox class (aka our Detox _main_ layer).

Unfortunately, the main drawback is a minor breakage of external-drivers implementation: Whichever plugin-driver implementations that exist out there in the wild -- they _still provide the matchers through the driver._ Fortunately, I've made the migration pretty easy. Given a plugin `driver.js`, users will have to apply these minor changes:
```diff
class Expect {
-  constructor(invocationManager) {
+  constructor({ invocationManager }) {
     this._invocationManager = invocationManager;
  }
}

class PluginDriver {
  constructor() {
-    this.matchers = new Expect(new invocationManager());
+    // no this.matchers init here...
  }
}

-module.exports = PluginDriver;
+module.exports = {
+  DriverClass: PluginDriver,
+  ExpectClass: Expect,
+}
```

### DIFF
--- a/detox/src/Detox.js
+++ b/detox/src/Detox.js
@@ -8,11 +8,12 @@ const DetoxRuntimeError = require('./errors/DetoxRuntimeError');
 const AsyncEmitter = require('./utils/AsyncEmitter');
 const MissingDetox = require('./utils/MissingDetox');
 const Client = require('./client/Client');
+const { InvocationManager } = require('./invoke');
 const DetoxServer = require('./server/DetoxServer');
 const ArtifactsManager = require('./artifacts/ArtifactsManager');
-
 const log = logger.child({ __filename });
 const driverRegistry = require('./devices/DriverRegistry').default;
+const matchersRegistry = require('./matchersRegistry');
 
 const _initHandle = Symbol('_initHandle');
 const _assertNoPendingInit = Symbol('_assertNoPendingInit');
@@ -144,12 +145,12 @@ class Detox {
     this._client.setNonresponsivenessListener(this._onNonresnponsivenessEvent.bind(this));
     await this._client.connect();
 
+    const invocationManager = new InvocationManager(this._client);
     const deviceDriver = driverRegistry.resolve(this._deviceConfig.type, {
       client: this._client,
+      invocationManager,
       emitter: this._eventEmitter,
     });
-
-    Object.assign(this, deviceDriver.matchers);
 
     this.device = new Device({
       deviceConfig: this._deviceConfig,
@@ -158,18 +159,23 @@ class Detox {
       sessionConfig,
     });
 
-    if (behaviorConfig.exposeGlobals) {
-      Object.assign(Detox.global, {
-        ...deviceDriver.matchers,
-        device: this.device,
-      });
-    }
-
     this._artifactsManager = new ArtifactsManager(this._artifactsConfig);
     this._artifactsManager.subscribeToDeviceEvents(this._eventEmitter);
     this._artifactsManager.registerArtifactPlugins(deviceDriver.declareArtifactPlugins());
 
     await this.device.prepare();
+
+    const matchers = matchersRegistry.resolve(this.device, {
+      invocationManager,
+    });
+    Object.assign(this, matchers);
+
+    if (behaviorConfig.exposeGlobals) {
+      Object.assign(Detox.global, {
+        ...matchers,
+        device: this.device,
+      });
+    }
 
     if (behaviorConfig.reinstallApp) {
       await this.device.uninstallApp();
--- a/detox/src/devices/Device.test.js
+++ b/detox/src/devices/Device.test.js
@@ -111,6 +111,11 @@ describe('Device', () => {
     expect(device.name).toEqual('mock-device-name-from-driver');
   });
 
+  it('should return the type from the configuration', async () => {
+    const device = validDevice();
+    expect(device.type).toEqual('ios.simulator');
+  });
+
   it('should return an undefined ID for an unprepared device', async() => {
     const device = validDevice();
     expect(device.id).toBeUndefined();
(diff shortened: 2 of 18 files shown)


rotemmiz:
How does it relate to #2012?

noomorph:
@rotemmiz, Amit wants to create a convenient DSL for taking element screenshots. He'll explain in depth if you want, but anyway I think that's the right direction.

d4vidi:
@rotemmiz I'll write down an elaborate explanation asap!

rotemmiz:
I am still not convinced that this is an important change, though I am OK with whatever makes you happy :)

The API you mentioned ` await element(by.id('fancy-element-id')).takeScreenshot()` is achievable without this change, as a new action (similar to the newly added `element(matcher).getAttributes()`. Or am I missing something?

detox/src/Detox.js
```diff
@@ -144,33 +145,38 @@ class Detox {
...
-    }
-
     this._artifactsManager = new ArtifactsManager(this._artifactsConfig);
     this._artifactsManager.subscribeToDeviceEvents(this._eventEmitter);
     this._artifactsManager.registerArtifactPlugins(deviceDriver.declareArtifactPlugins());
 
     await this.device.prepare();
 
+    const matchers = matchersRegistry.resolve(this.device, {
+      invocationManager,
```
noomorph, line 169: @d4vidi, so, your plan is to inject `device` or a similar handle there, to ensure that the matchers will be able to pull the files from the device, after the screenshot is done?

d4vidi:
@rotemmiz 
>The API you mentioned await element(by.id('fancy-element-id')).takeScreenshot() is achievable without this change
It can be defined and worked out to take the screenshot on the device, as you normally would, such that a a file will be created on the device (part A of my pseudo-code. But it would not be able to pull the file back to the host, as it cannot access the `device` (part B of the peuso-code).

>though I am OK with whatever makes you happy
1. ❤️ 
2. I put some more thought into it - it's not fair to say the two _distinct units_ should be hard-separated; They don't _have_ to be. Rather, the problem here is specifically with how in "unit 2"'s lowest layers (i.e. device driver), we currently create the top-level API of "unit 1" (i.e. the element/expect API's). That is a sheer violation of basic software architecture practices, and should be fixed even regardless of screenshots or any specific feature. The element screenshot simply surfaces that specific code-smell.
3. What would really make me happy is to separate all of our top-level API's onto an API layer, and have **everything** handled by a lower-layer implementation, which hard the freedom to be platform-specific where needed. In other words, make the top-level API layer very thin, and extending the device/driver layers to serve _all_ API's - not just `device.xyz()` methods.

detox/src/Detox.js
d4vidi: Bingo! The restored flexibility would allow us to do so 💪

detox/src/devices/Device.test.js
```diff
@@ -111,6 +111,11 @@ describe('Device', () => {
     expect(device.name).toEqual('mock-device-name-from-driver');
   });
 
+  it('should return the type from the configuration', async () => {
+    const device = validDevice();
```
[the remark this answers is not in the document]
d4vidi: Good point. With this implementation it will return the relative-path as-is, which might be weird, but that wouldn't be the driver's fault (rather, the whole concept's "fault"). This is why I suggest to keep as-is. Nevertheless, it is overrideable by the driver's implementor.

PR #2220, opened 2020-07-20
fix: biometric unmatching timeout

christianchown:
- [x] This is a small change 

Related PR: https://github.com/wix/Detox/pull/1605

Currently attempting to _unmatch_ biometrics always times out. This is because of the `waitForActive` line: as the unmatch command allows the user to retry (up to 3 times), the biometric prompt is still active, so the device driver remains inactive, and `waitForActive` never resolves.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -148,7 +148,6 @@ class Device {
 
   async unmatchFace() {
     await this.deviceDriver.unmatchFace(this._deviceId);
-    await this.deviceDriver.waitForActive();
   }
 
   async matchFinger() {
@@ -158,7 +157,6 @@ class Device {
 
   async unmatchFinger() {
     await this.deviceDriver.unmatchFinger(this._deviceId);
-    await this.deviceDriver.waitForActive();
   }
 
   async shake() {


LeoNatan:
Hmm, but what happens on the third time? Does Detox behave well? Is there a test in the suite for failing biometrics several times? If not, can I please ask you to add one? Thanks

E1213180:
Hello @LeoNatan - it can never really hit the 3rd time, as the promise never resolves after the firs, as the device does not become active due to the UI prompt remaining visible. Currently you'd have to do this in code (and is exactly what I'm currently doing as a workaround):

```
device.unmatchFinger();
await sleep(1);

device.unmatchFinger();
await sleep(1);

await device.unmatchFinger();

// can now proceed
```
Adding a test to the suite won't help, as the tests (naturally) only check calls against the mocked DeviceDriverBase.

LeoNatan:
I’m sorry, I don’t understand. After three failures, the test would continue because of biometric failure, is this not correct? So my question is, what happens if, after your change, one of the “unmatch” API is called three times? Is a sleep necessary there also? If so, this change isn’t fully thought out. If not, a test can be added to ensure it works.

E1213180:
Hello, after the change, no sleep is necessary, as the driver's unmatchXXX promise doesn't resolve until the action has taken place.

I'm not fully following where in the repo a test can be added: I've seen https://github.com/wix/Detox/blob/master/detox/src/devices/Device.test.js but these only work with the mock device. Can you show me where you would like the test, or (if in that file) give me a clue as to the form it should take?

E1213180:
gotcha, will look at

LeoNatan:
Thank you!

LeoNatan:
Hello,

Any news on this PR?

LeoNatan:
Hello,
Any update here?
Thanks

PR #2222, opened 2020-07-21
Element screenshot (Android)

### LINKED ISSUE #2012: [feature] Element Screenshot

d4vidi:
- [ ] This is a small change 
- [x] This change has been discussed in issue #2012 and the solution has been agreed upon with maintainers.

---

**Description:**
Relates to #2012: introduces the fully-blown (almost*) element screenshot taking solution for Android, using an _interaction_ API.

**Implementation notes:**
After some research, files on Android are a bit of a headache, and really - there's no need for a file on the device. I therefore took @valentynberehovyi's advice and skipped file saving altogether, and instead return the bitmap as the invocation's result (base-64 encoded, of course). I tested this even with an element that takes up almost the entire screen on a full HD device (1920x1080), and it went smooth.

**P.S.**
PR `2222`, best ever

---
\* see caveats in documentation


detox/android/detox/src/main/java/com/wix/detox/common/ViewScreenshot.kt
```diff
@@ -0,0 +1,24 @@
+package com.wix.detox.common
+
+import android.graphics.Bitmap
+import android.graphics.Canvas
+import android.view.View
+import java.io.ByteArrayOutputStream
+
+class Screenshot(private val bitmap: Bitmap) {
```
d4vidi, line 8: maybe change to `ScreenshotResult`

detox/android/detox/src/main/java/com/wix/detox/espresso/EspressoDetox.java
```diff
@@ -34,10 +34,13 @@
 public class EspressoDetox {
     private static final String LOG_TAG = "detox";
 
-    private static final String METHOD_GET_ACTIVITY = "getCurrentActivity";
+    public static Object perform(ViewInteraction interaction, ViewAction action) {
+        interaction.perform(action);
 
-    public static ViewInteraction perform(ViewInteraction interaction, ViewAction action) {
-        return interaction.perform(action);
```
d4vidi, line 40: This never made sense! `interaction.peform()` always returns the interaction itself. `invoke` cannot parse an object, and therefore the returned result has _always_ been `null`.

With this new implementation, it's possible to return a result from the `action` itself. In the future, this should be useful for implementing `getAttributes()` @rotemmiz

detox/src/android/expect.js
```diff
@@ -115,14 +116,22 @@ class SwipeAction extends Action {
...
 class Interaction {
   constructor(invocationManager) {
     this._call = undefined;
     this._invocationManager = invocationManager;
   }
 
   async execute() {
-    await this._invocationManager.execute(this._call);
+    const resultObj = await this._invocationManager.execute(this._call);
+    return resultObj ? resultObj.result : undefined;
```
d4vidi, line 134: This enables effectively using the result returned from invoke action on the device

detox/src/android/expect.test.js
```diff
@@ -125,53 +128,108 @@ describe('expect', () => {
...
-    await e.element(e.by.id('ScrollView799')).swipe('down', 'fast');
-    await e.element(e.by.id('ScrollView799')).swipe('up', 'slow');
-    await e.element(e.by.id('ScrollView799')).swipe('left', 'fast');
-    await e.element(e.by.id('ScrollView799')).swipe('right', 'slow');
-    await e.element(e.by.id('ScrollView799')).swipe('down', 'fast', 0.9);
-    await e.element(e.by.id('ScrollView799')).swipe('up', 'slow', 0.9);
-    await e.element(e.by.id('ScrollView799')).swipe('left', 'fast', 0.9);
-    await e.element(e.by.id('ScrollView799')).swipe('right', 'slow', 0.9);
-    await e.element(e.by.id('ScrollView799')).atIndex(1);
+  describe('element interactions', () => {
```
d4vidi, line 131: more boyscouting here than really related to this feature, except that I wanted more specific tests related to _this_ specific interaction (screenshot taking)

d4vidi:
@ethanshar

d4vidi:
@rotemmiz how about `element(matcher).captureBitmap()`?

rotemmiz:
Also great

d4vidi:
@rotemmiz done!

LeoNatan:
@rotemmiz As opposed to screenshot, which is not a bitmap?

“Bitmap” is a bad terminology. It should be “snapshot”.

rotemmiz:
That's why I suggested that `view` will be a part of the function name. We want to get a bitmap of the view. I think it should be with little ambiguity as possible. 
Why do you think snapshot is a better terminology?

LeoNatan:
I think we should be using the artifact system for this, and so “bitmap” is a very specific term. Images won’t be bmp, probably.

LeoNatan:
View is also something we avoid everywhere, so we shouldn’t use that. `snapshot()` is enough, and perhaps provide file name.

detox/src/utils/easy-file-io.js
```diff
@@ -0,0 +1,13 @@
+const tempfile = require('tempfile');
+const fs = require('fs');
+
+function saveRawBase64Data(dataBase64, { filePath, fileSuffix }) {
+  const _filePath = filePath || tempfile(fileSuffix);
+  const data = Buffer.from(dataBase64, 'base64');
+  fs.writeFileSync(_filePath, data);
```
noomorph, line 7: How do you clean up temporary files after the test session ends?

noomorph: Any good sense to bother with Buffer conversions?
```js
fs.writeFileSync('hello.txt', 'SGVsbG8sIHdvcmxkIQ==', 'base64')
fs.readFileSync('hello.txt', 'utf8'); // 'Hello, world!'
```

d4vidi:
@rotemmiz I tend to agree with @LeoNatan:
1. View = implementation details. Our API's offer an abstraction.
2. Bitmap is good, but might be a bit too technical. I favoured it because it hints of that we do not effectively grab pixels off the screen - but rather **redraw** the element onto a bitmap. We should probably come up with something between _bitmap_ and _screenshot_. I'm open to suggestions.

As per _snapshot_ - I'm highly against that because the term is used in _snapshot_ testing. Moreover, this feature in particular, aims at providing a means for app-devs to take _screenshots_ in order to perform _snapshot_ tests. We should really avoid the term here.

LeoNatan:
So the feature is intended for snapshot testing, but snapshot is not a good term? I don't understand.

LeoNatan:
Apple uses "screenshot" terminology:

https://developer.apple.com/documentation/xctest/xcuiscreenshotproviding

detox/src/utils/easy-file-io.js
d4vidi: 😻 😻 😻

detox/src/utils/easy-file-io.test.js
```diff
@@ -0,0 +1,54 @@
...
+      fs = require('fs');
+
+      jest.mock('tempfile');
+      tempfile = require('tempfile');
+      tempfile.mockReturnValue(tempFilePath);
+
+      uut = require('./easy-file-io');
+    });
+
+    it('should save raw data to a given file', () => {
```
noomorph, line 23: Where do you use this scenario, Amit? @d4vidi

detox/src/android/expect.js
```diff
@@ -268,6 +277,12 @@ class Element {
     this._selectElementWithMatcher(this._originalMatcher._avoidProblematicReactNativeElements());
     return await new ActionInteraction(this._invocationManager, this, new SwipeAction(direction, speed, percentage)).execute();
   }
+
+  async captureBitmap() {
+    // TODO this should be moved to a lower-layer handler of this use-case
+    const resultBase64 = await new ActionInteraction(this._invocationManager, this, new CaptureBitmapAction()).execute();
+    return ezFile.saveRawBase64Data(resultBase64, { fileSuffix: '.detox.element.png' });
```
noomorph, line 284: Could not we just move `detox/src/artifacts/utils/temporaryPath.js` to `detox/src/utils`, import it:

```js
const temporaryPath = require('../utils/temporaryPath');
```

and just use in `captureBitmap`:
```js
const resultBase64 = await new ActionInteraction(this._invocationManager, this, new CaptureBitmapAction()).execute();
const pngPath = temporaryPath.for.png();
await fs.writeFile(pngPath, resultBase64, 'base64');
return pngPath;
```

and not create yet another mini-module, please?

detox/src/utils/easy-file-io.test.js
d4vidi: Right. I'll remove support for that.

detox/src/android/expect.js
d4vidi: I considered doing that a number of times, but `temporaryPath` seemed very artifact oriented, seeing that the API set basically pairs 1:1 with the artifacts types we have (image, log, video, perf recording). If you find it apt nevertheless, I *will* use it here.

The mini-module was a compromise as we do not have a lower-layer logic-provider for this, just yet. I wanted to keep the logic here down to the bare minimum. 'Makes it easier to test (when in a separate module), too. wdyt?

noomorph: I just don't see what it is here to test. :man_shrugging:

The test idea here would be to check whether we write a received (mocked) base64 string to some temporary file.

If we extract logic to a separate file, it feels like we are going to write very similar tests (hence, 2 times the same) - for `expect.js`'s `takeScreenshot` and that minimodule.

I don't feel like expanding the test logic by extra 3-4 lines is going to hurt its legibility:

```js
ezFileIo = require('../utils/easy-file-io');
ezFileIo.saveRawBase64Data.mockReturnValue('/mock/file/path');
// ...
```

```js
jest.mock('fs-extra').mock('../utils/temporaryPath');
require('../utils/temporaryPath').for.png.mockReturnValue('mock.png');
// ...
expect(require('fs-extra').writeFile).toHaveBeenCalledWith('mock.png', invokeResultInBase64, 'base64');
```

d4vidi:
@noomorph I'll take your advice on this and get rid of easy-file-io :)

noomorph:
@d4vidi, :pray:  I sincerely hope I have managed to convey why I considered easy-file-io impractical a bit in a sense of inflating code and unit tests base with minimal benefits in return. Good luck with the field testing of the feature! Hope it works out as our guys expect.

detox/src/android/expect.test.js
```diff
@@ -209,27 +210,32 @@ describe('expect', () => {
...
+
+      fs = require('fs');
 
       _element = e.element(e.by.id('FancyElement'));
     });
 
     it('should take and save the screenshot in a temp-file', async () => {
       await _element.takeScreenshot();
-      expect(ezFileIo.saveRawBase64Data).toHaveBeenCalledWith(invokeResultInBase64, { fileSuffix: '.detox.elem-screenshot.png' })
+      expect(fs.writeFileSync).toHaveBeenCalledWith(tempFilePath, invokeResultInBase64, 'base64');
```
noomorph, line 232: Are you sure it is the best idea to block Node.js process? Use `fs-extra` and `await fse.writeFile()` maybe?

d4vidi: Fixed!

d4vidi:
@noomorph I generally agree. Again, in this case, this came up trying to work-around the fact we don't have proper abstractions in the expectation/matching realms, namely - with a proper place to put this logic in (I still don't think the `Element.takeScreenshot()` is the right place, at least no as-is).

But after pointing this out to me, I've come to terms that this can only be worked-out using proper refactoring.

PR #2226, opened 2020-07-23
Store element-screenshot results as artifacts

### LINKED ISSUE #2012: [feature] Element Screenshot

d4vidi:
- [x] This is a small change 
- [x] This change has been discussed in issue #2012 and the solution has been agreed upon with maintainers.

---

**Description:**
Resolves #2012. Applies the final piece of this puzzle: the administration of artifacts management we have over image-files generated by using this feature.


detox/src/Detox.js
```diff
@@ -167,6 +167,7 @@ class Detox {
 
     const matchers = matchersRegistry.resolve(this.device, {
       invocationManager,
+      emitter: this._eventEmitter,
```
d4vidi, line 170: such flexibility, much design 😄

d4vidi:
@noomorph with `device.takeScreenshot()` - do we e2e-test that the artifact is indeed generated, and with the proper name? If not, should we? It's tricky to do so, but it can be done I believe.

detox/src/Detox.js
noomorph: image

d4vidi: exactly! :doge:

docs/APIRef.ActionsOnElement.md
```diff
@@ -242,12 +242,16 @@ jestExpect(multipleMatchedElements.elements.length).toBe(5);
 jestExpect(multipleMatchedElements.elements[0].identifier).toBe('FirstElement');
 ```
 
-### `takeScreenshot()` Android Only
+### `takeScreenshot(name)` Android Only
```
noomorph, line 245: I'd suggest using `takeScreenshot([name])` square brackets convention for marking optional args.

docs/APIRef.ActionsOnElement.md
```diff
@@ -242,12 +242,16 @@ jestExpect(multipleMatchedElements.elements.length).toBe(5);
...
 
-Take a screenshot of the native view associated with the element in question. Useful for highly-focused visual comparison tests (i.e. comparison between elements rather than of complete screen, with visual "noise").
+Takes a screenshot of the native view associated with the element in question, and schedules putting it to the [artifacts folder](https://github.com/wix/Detox/blob/master/docs/APIRef.Artifacts.md#enabling-artifacts) upon completion of the current test. Useful for highly-focused visual comparison tests (i.e. comparison between elements rather than of [complete screens](https://github.com/wix/Detox/blob/master/docs/APIRef.DeviceObjectAPI.md#devicetakescreenshotname) with "visual noise").
+
+`name (optional)` - Name of the final file to store as the image artifact. For example, `my-text-field` would result in a file named `my-text-field.png`. In case the name isn't provided, Detox would self-generate a distinct name, instead (though not a very descriptive one).
```
noomorph, line 249: > would self-generate a distinct name, instead (though not a very descriptive one).

generates a random alphanumeric name instead.

noomorph:
@d4vidi, I am sorry in advance for piggybacking, but could you also make `device.takeScreenshot(name)` optional in a sense that it will be generating a random name for the screenshot, to align the APIs? What do you think?

noomorph:
@d4vidi, side thought — do we **noop** `element().takeScreenshot()`for iOS?

noomorph:
Looks good, overall.

noomorph:
Regarding:
> Resolves #2012

Just musing, whether we consider #2012 to **resolved** without having iOS feature parity... ? 🤔

docs/APIRef.ActionsOnElement.md
d4vidi: np

d4vidi:
@noomorph https://github.com/wix/Detox/pull/2226/commits/3d6063f16cd5828708eaf53dc9d0a8c118488989
✅  Piggybacking 
✅  iOS no-op (actually, decided to throw an error because `takeScreenshot()` should return a value
✅  Device-API doc
❌  Not resolving #2012 - let's open a separate issue for that, as we've done in the past with Android

PR #2231, opened 2020-07-27
Allow for setting rn-context load time in DetoxTest

### LINKED ISSUE #2197: waited a whole minute for the new RN-context
**Describe the bug**

Running metro bundle that takes a bit more than one minutes cause the app to crash 
A good solution is a flag to increase this timeout or to change it.
**To Reproduce**

- [ ] I have tested this issue on the latest Detox release and it still reproduces

Steps to reproduce:
1. have large project bundle (js) 
2. run detox test with debug configuration

**Expected behavior**

Tests to run.
**Screenshots**

**Device and Verbose Detox Logs**

```
    java.lang.RuntimeException: waited a whole minute for the new RN-context
        at com.wix.detox.reactnative.ReactNativeLoadingMonitor.awaitNewRNContext(ReactNativeLoadingMonitor.kt:56)
        at com.wix.detox.reactnative.ReactNativeLoadingMonitor.getNewContext(ReactNativeLoadingMonitor.kt:21)
        at com.wix.detox.reactnative.ReactNativeExtension.awaitNewReactNativeContext(ReactNativeExtension.kt:99)
        at com.wix.detox.reactnative.ReactNativeExtension.waitForRNBootstrap(ReactNativeExtension.kt:58)
        at com.wix.detox.DetoxManager.initReactNativeIfNeeded(DetoxManager.java:137)
        at com.wix.detox.DetoxManager.access$000(DetoxManager.java:26)
        at com.wix.detox.DetoxManager$2.run(DetoxManager.java:79)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at com.wix.detox.Detox$1.run(Detox.java:173)
        at java.lang.Thread.run(Thread.java:919)
```
 - Detox: 16.8.0
 - React Native: 0.62.0
 - Node: 10.16.0
 - Device: emulator pixel 3a xl
 - OS: Android 10
 - Test-runner: `jest-circus`

d4vidi:
- [x] This is a small change 

---

**Description:**
Resolves #2197. Migrates Detox' idle-policy config to a more generic config structure (that handles idle-policy as well), such that it can also hold a custom timeout for waiting for RN's loading to complete. That enables, for example, the customization of the timeout based on whether the app running in debug mode rather than release - where this can take longer for large apps. See [here](https://github.com/wix/Detox/pull/2231/files#diff-8c8eebd6512cc7aeed7d36a9bdc0b170R32).

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/config/DetoxConfig.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/config/DetoxConfig.kt
@@ -1,11 +1,11 @@
 package com.wix.detox.config
 
 class DetoxConfig {
-    var idlePolicyConfig: DetoxIdlePolicyConfig? = DetoxIdlePolicyConfig()
-    var rnContextLoadTimeoutSec = 60
+    @JvmField var idlePolicyConfig: DetoxIdlePolicyConfig = DetoxIdlePolicyConfig()
+    @JvmField var rnContextLoadTimeoutSec = 60
 
     fun apply() {
-        idlePolicyConfig?.apply()
+        idlePolicyConfig.apply()
     }
 
     companion object {
--- a/detox/android/detox/src/main/java/com/wix/detox/config/DetoxConfig.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/config/DetoxConfig.kt
@@ -1,0 +1,14 @@
+package com.wix.detox.config
+
+class DetoxConfig {
+    var idlePolicyConfig: DetoxIdlePolicyConfig? = DetoxIdlePolicyConfig()
+    var rnContextLoadTimeoutSec = 60
+
+    fun apply() {
+        idlePolicyConfig?.apply()
+    }
+
+    companion object {
+        lateinit var CONFIG: DetoxConfig
+    }
+}
--- a/detox/android/detox/src/main/java/com/wix/detox/reactnative/ReactNativeLoadingMonitor.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/reactnative/ReactNativeLoadingMonitor.kt
@@ -5,6 +5,8 @@ import android.util.Log
 import com.facebook.react.ReactApplication
 import com.facebook.react.ReactInstanceManager
 import com.facebook.react.bridge.ReactContext
+import com.wix.detox.common.DetoxErrors
+import com.wix.detox.config.DetoxConfig
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
@@ -13,7 +15,8 @@ private const val LOG_TAG = "DetoxRNLoading"
 open class ReactNativeLoadingMonitor(
         private val instrumentation: Instrumentation,
         private val rnApplication: ReactApplication,
-        private val previousReactContext: ReactContext?) {
+        private val previousReactContext: ReactContext?,
+        private val config: DetoxConfig = DetoxConfig.CONFIG) {
     val countDownLatch = CountDownLatch(1)
 
     fun getNewContext(): ReactContext? {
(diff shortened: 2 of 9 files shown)


d4vidi:
@swabbass

detox/android/detox/src/main/java/com/wix/detox/config/DetoxConfig.kt
```diff
@@ -0,0 +1,14 @@
+package com.wix.detox.config
+
+class DetoxConfig {
+    var idlePolicyConfig: DetoxIdlePolicyConfig? = DetoxIdlePolicyConfig()
```
swabbass, line 4: if the ctor does not throw exceptions, then no need for Nullable type, you can remove `?` if calling constructor of `DetoxIdlePolicyConfig` does not yield null.

swabbass: and can be val if the reference is not changing

swabbass: as i see the usage you are creating an object and use the setter 
```
class DetoxConfid(val idlePolicyConfig: DetoxIdlePolicyConfig = DetoxIdlePolicyConfig(),var rnContextLoadTimeoutSec:Int = 60){

fun apply() {...}
companio object {.....}
```

detox/android/detox/src/main/java/com/wix/detox/reactnative/ReactNativeLoadingMonitor.kt
```diff
@@ -13,7 +15,8 @@ private const val LOG_TAG = "DetoxRNLoading"
 open class ReactNativeLoadingMonitor(
         private val instrumentation: Instrumentation,
         private val rnApplication: ReactApplication,
-        private val previousReactContext: ReactContext?) {
+        private val previousReactContext: ReactContext?,
+        private val config: DetoxConfig = DetoxConfig.CONFIG) {
```
swabbass, line 19: you are assigning lateinit var, did u check that all flows that have access to this config is after it is being initialized?

swabbass:
> * [x]  This is a small change
> * [ ]  This change has been discussed in issue #<?> and the solution has been agreed upon with maintainers.
> 
> **Description:**
> Resolves #2197. Migrates Detox' idle-policy config to a more generic config structure (that handles idle-policy as well), such that it can also hold a custom timeout for waiting for RN's loading to complete. That enables, for example, the customization of the timeout based on whether the app running in debug mode rather than release - where this can take longer for large apps. See [here](https://github.com/wix/Detox/pull/2231/files#diff-8c8eebd6512cc7aeed7d36a9bdc0b170R32).
THX A LOT Wating for this to release to use it.
not a reviewer but wrote my thoughts on kotlin code might be useful :],

detox/android/detox/src/main/java/com/wix/detox/config/DetoxConfig.kt
d4vidi: oops, right about the null heh

d4vidi: Ok I kept fields as `var` and applied `@JvmField` on all fields... this is what I really wanted to begin with... thanks for pointing this out!!!

detox/android/detox/src/main/java/com/wix/detox/reactnative/ReactNativeLoadingMonitor.kt
d4vidi: Yes - it must be set as early as in `Detox.runTest()`, and if that's not the case - I want the _uninitialized_ error to be thrown.

Issue #2073, opened 2020-05-20
Support artifact collection of test retries (`jest.retryTimes(x)`)


rotemmiz:
When running a test with `jest.retryTimes(x)`, the artifact manager will create a folder for the first retry, but will fail recording the next retries (if they happen).

Example:
```
detox[4447] WARN:  [Artifact.js/MOVE_FILE_EXISTS] cannot overwrite: "~/temp/buildTmp/d4efa026-0524-4cf9-a6da-a5c3fba4c3bc.detox.png" => "artifacts/ios.sim.release.2020-05-17 09-18-53Z/✗ testName/testDone.png"
```

I suggest we add the retry number to the folder name, 
retry no 1: `test name`
retry no 2: `test name - retry 2`
retry no 2: `test name - retry 3`

d4vidi:
@viktorijasujetaite and I have taken a 2nd look on some tests where retry is applied and we noticed the opposite - namely, that only the artifacts from last retry prevail.

rotemmiz:
Could be different between platforms, anyway, the dir names should solve either issue we have there.

viktorijasujetaite:
> @viktorijasujetaite and I have taken a 2nd look on some tests where retry is applied and we noticed the opposite - namely, that only the artifacts from last retry prevail.

I would actually think that artifacts of [Retry #1] are recorded, because for latter retries (#2 & #3) we see "cannot overwrite" warning.

d4vidi:
@viktorijasujetaite again, this is probably iOS-specific, and different on Android. In any case, solving this would solve it for both platforms

PR #2242, opened 2020-07-29
feat(jest): support artifact collection for multiple retries with jest.retryTimes(n)

noomorph:
- [x] This change has been discussed in issue #2073 and the solution has been agreed upon with maintainers.

---

**Description:**

Adds support `jest.retryTimes(n)` when used with `jest-circus`.

For example, if a `Flaky test` fails 2 times and then passes on the third, the artifacts will be saved like this:

```
artifacts/<auto-generated-dir>/✗ Flaky test/...
artifacts/<auto-generated-dir>/✗ Flaky test (2)/...
artifacts/<auto-generated-dir>/✓ Flaky test (3)/...
```

This pull request has unit and E2E tests for the feature.


detox/src/artifacts/utils/ArtifactPathBuilder.test.js
```diff
@@ -73,6 +73,20 @@ describe(ArtifactPathBuilder, () => {
...
+
+      expect(asPosixPath(artifactPath)).toBe('/tmp/subdir/✗ test (2)/1.log');
+    });
+
+    it('should append (3) invocations count for a test that passes for the third time', () => {
+      const testSummary = {title: '', fullName: 'test', status: 'passed', invocations: 3 };
+      const artifactPath = pathBuilder.buildPathForTestArtifact('1.log', testSummary);
+
+      expect(asPosixPath(artifactPath)).toBe('/tmp/subdir/✓ test (3)/1.log');
+    });
```
d4vidi, line 89: Just wondering: isn't this test redundant?...

detox/test/e2e/27.retries.js
```diff
@@ -0,0 +1,17 @@
...
+  });
+
+  it('should fail twice and pass once', async () => {
+    const matcher = --counter > 0
+      ? by.text('Not existing')
+      : by.text('Sanity');
+
+    await element(matcher).tap();
+  });
+});
```
d4vidi, line 17: Nice, but here too (much like with screenshot tests) we lack the testing of the final artifacts...

detox/test/e2e/config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   "rootDir": "../..",
   "testEnvironment": "./test/e2e/environment.js",
   "testRunner": "./test/node_modules/jest-circus/runner",
-  "setupFilesAfterEnv": ['./test/e2e/init-coverage.js'],
+  "setupFilesAfterEnv": [],
```
d4vidi, line 5: Not really related, right?

detox/test/e2e/27.retries.js
d4vidi: Also, I didn't see anything related in the build's artifacts. Have I been looking in the wrong place?

detox/src/artifacts/utils/ArtifactPathBuilder.test.js
noomorph: I adhere to the blackbox approach, @d4vidi.
In my opinion, there were two ways to implement this.
Either we omit ` (${invocations})` suffix, when the status is `passed` or we add it nevertheless.
Therefore, I **fixate** the behavior and confirm that I as an author intend it to be this way.

detox/test/e2e/config.js
noomorph: Yeah, but I've noticed that this file is redundant already. It should have gone with my other Coveralls fixes, but I missed it.

detox/test/e2e/27.retries.js
noomorph: Hmmm... well, I can add checks, theoretically.

detox/src/artifacts/utils/ArtifactPathBuilder.test.js
d4vidi: Actually, I see now that I missed the difference in `status` value. You can ignore my comment.

detox/test/e2e/27.retries.js
d4vidi: So action items here:
1. Rename to `27.retries.test.js`
2. Add artifacts validation?

noomorph: Thanks for pointing out, fixing.

PR #2244, opened 2020-07-31
Update APIRef.Artifacts.md

uloco:
all preset default to keepOnlyFailedArtifacts: false

- [ ] This is a small change 

---

**Description:**


noomorph:
Wow, @uloco, that's a good catch! Çok teşekkürler! 🙏 thx 👍

PR #2266, opened 2020-08-10
Update capitalization on field names to current

lukewlms:
- [ ] This is a small change 

---

**Description:**


noomorph:
Thanks! That should have been updated for a very very long time! :+1

PR #2269, opened 2020-08-12
feat(cli): add -R, --retries <count> option

### LINKED ISSUE #2073: Support artifact collection of test retries (`jest.retryTimes(x)`)

noomorph:
Resolves #2073 for users who temporarily cannot rewrite their test suites to make them adhere to `jest.retryTimes(n)` guidelines. The idea is to keep re-running failed test suites (or **files**, to be more precise) before either we run out of attempt or succeed with those flaky tests.

This is not a recommended method since it assumes, for example, that if in a 100-tests suite even 1 test fails, we will be re-running the entire test file with those 100 tests again.

### To-dos

- [x] ~decide what do we do if users write into the same `artifacts-location` since it is apt to generate suppressed I/O errors in the logs (that is the case when there is no autogenerated `ios.sim.release.2020-08-12 13-02-42Z` suffix, only `artifacts/` folder directly).~ Now, we support this case: you'll just see (1) (2) (3) prefixes.
- [x] ~preceed this PR with the one that fixes `env.PATH` resolution in `detox-cli` npm package (not `local-cli` one).~ 
- [x] ~remove remnants of env.PATH PR from this PR~
- [x] ~check `package.json` dependencies (maybe some of them should be optional or peer dependencies)~
- [x] tidy "Retrying ..." message
- [x] ~check if we can add Mocha support with 5-15 lines~ (we cannot because Mocha does not support multiple reporters yet) disable the feature for Mocha
- [x] extract the piggybacking out from JestVerboseReporter
- [x] add `-R` test case for `local-cli/test.test.js`
- [x] cover `splitArgv.js` file with tests
- [x] fix long-standing `--debug-synchronization e2e/01.sanity.test.js` issue (incorrect boolean handling)
- [x] update CLI docs


detox/runners/jest/DetoxStreamlineJestReporter.js
```diff
@@ -91,6 +94,16 @@ class DetoxStreamlineJestReporter extends JestVerboseReporter {
       return reporterName === 'default';
     });
   }
+
+  async onRunComplete(_contexts, { testResults }) {
+    await super.onRunComplete();
+
+    const failedFiles = testResults
```
noomorph, line 101: @d4vidi, that might be the most controversial point here, but if we add the logic here, it is going to be accessible immediately for all Detox + Jest users. 🤷‍♂️

d4vidi: Oh man. Not a fan of this. I think this deserves an independent reporter.

detox/src/devices/DeviceRegistry.js
```diff
@@ -59,6 +67,18 @@ class DeviceRegistry {
 
     this._lockfile.write(newState);
   }
+
+  static ios() {
```
d4vidi, line 71: what about... `createIos()`, `createAndroid()`? plain `ios` is a bit under-verbose

detox/package.json
```diff
@@ -47,6 +48,7 @@
     "funpermaproxy": "^1.0.1",
     "get-port": "^2.1.0",
     "ini": "^1.3.4",
+    "jest-cli": "^26.2.2",
```
rotemmiz, line 51: Why is jest-cli a dependency? It might cause issues for users with different versions (wonky node_modules mess, where this supercedes the actual jest dependencies the users configured for themselves)

detox/runners/jest/DetoxStreamlineJestReporter.js
noomorph: Okay, will do.

detox/package.json
noomorph: Moved to peerDependencies.

PR #2277, opened 2020-08-14
Add binaryPath troubleshooting

lukewlms:
`binaryPath` is *not* the path that Detox should use for the binary.  Rather, Detox uses this as a base path and tries to transform it to determine its own test binary path via the internal function `getTestApkPath`.  This tripped me up and also a number of other users:

https://stackoverflow.com/questions/53984667/detox-app-could-not-be-found-did-you-run-gradlew-assembleandroidtest/63414420#63414420
https://stackoverflow.com/questions/58951816/detox-build-successful-but-detox-test-fails/63414275#63414275

Adding to troubleshooting here to try and help future users solve this more quickly.

- [ ] This is a small change 

---

**Description:**


rotemmiz:
Thanks!

noomorph:
I think this was not the best solution anyway. We should extend DetoxRuntimeError hint message with this FAQ suggestion to try out testBinaryPath.

d4vidi:
I agree with @noomorph

PR #2294, opened 2020-08-25
Add reference to second instance of com.example

lukewlms:
- Also added note on where to find the package name, since RN devs may not work frequently with Android config

- [x] This is a small change (docs only)


rotemmiz:
Thanks

PR #2299, opened 2020-09-01
Swipe in screen coordinate space, rather than element coordinate space

### LINKED ISSUE #2298: Swipe takes percentage of the element, not screen. Small elements cannot be swiped
Problem is that with Detox17 swipe is almost useless, especially with small elements, as detox swipe only within bounds of that element. This results in action sheets that cannot be closed and similar issues

In function `swipe(direction, speed, percentage)` percentage stands for % of the screen

In a specific case where action sheet top handle could not be swiped, I tried swiping the whole action sheet but it's not by design to close it in such way.

LeoNatan:
Closes #2298


rotemmiz (approved):
Failed on an unrelated Android test, looks good otherwise.

PR #2300, opened 2020-09-02
Debug sync will be activated on all WS calls, not only on invoke.

rotemmiz:
- [ ] This is a small change 

---

**Description:**
Previously, `--debug-synchronization` was only triggered on `actions.Invoke()` commands. This PR adds support for debug sync over all slow websocket calls.
Still needs to polish over edge cases.


LeoNatan:
Also consider not waiting for isReady before starting to send the sync queries. Right now there is a big limitation that it won’t print status unless the app was idle at least once.

rotemmiz:
@LeoNatan debug sync will now be triggered on every call sent through ws, isReady is one of them. Did you mean something more than that?

LeoNatan:
I’m not sure. I just want us to be able to debug launch sync issues.

rotemmiz:
We got it then

PR #2362, opened 2020-09-24
Add new e2e scenarios that reproduce known issues

noomorph:
**Description:**

There were numerous internal bug reports complaining about 17.6.0 visibility logic rework on iOS, so this pull request aims to add a few cases to make sure the next fix version will include fixes both to problems of 17.5.5 and 17.6.0.

| № | Assertion                                                                         |                                                         Screenshot                                                        |
|:-:|-----------------------------------------------------------------------------------|:-------------------------------------------------------------------------------------------------------------------------:|
| 1 | should be able to scroll tab bar strip with a gradient overlay                    |  |
| 2 | should be able to tap on the main ScrollView regardless of the on-screen keyboard |  |
| 3 | should be able to tap on the badged button                                        |  |
| 4 | should not crash after an attempt to tap an element with an out-of-bounds index:  `await element(by.text('Index')).atIndex(100500).tap();` | moved to #2372 |

To be continued.


detox/test/e2e/28.visibility.test.js
```diff
@@ -13,9 +13,27 @@ describe('Visibility Cases', () => {
...
 
-  it('should be able to tap on the main ScrollView regardless of the on-screen keyboard', async () => {
+  it(':ios: should be able to tap on the main ScrollView in a blind way, on the center, when the on-screen keyboard is enabled', async () => {
     await element(by.id('screenScroll')).tap();
     await element(by.id('inputExample')).tap();
+
+    const { frame: { x, y, width, height } } = await element(by.id('screenScroll')).getAttributes();
+
+    await element(by.id('screenScroll')).tap({
+      x: Math.floor(x + 0.5 * width),
```
LeoNatan, line 23: The coordinates are in the element's coordinate space, so it doesn't make sense to add x and y.

LeoNatan (changes requested):

detox/test/e2e/28.visibility.test.js
noomorph: Fixed.

noomorph: I still have the error:

```
  ● Visibility Cases › :ios: should be able to tap on the main ScrollView in a blind way, on the center, when the on-screen keyboard is enabled

    Test Failed: View “<RCTScrollView: 0x7ff1cd536f60>” is not visible: view does not pass visibility threshold (75%)
    TIP: To print view hierarchy on failed actions/matches, use loglevel verbose and above.

      19 |     const { elementSafeBounds: bounds } = await element(by.id('screenScroll')).getAttributes();
      20 |
    > 21 |     await element(by.id('screenScroll')).tap({
         |                                          ^
      22 |       x: Math.floor(0.5 * bounds.width),
      23 |       y: Math.floor(0.5 * bounds.height),
      24 |     });

      at Object.it.only (test/e2e/28.visibility.test.js:21:42)
```

LeoNatan: 🤷‍♂️ Is it visible?

noomorph: Yes, but the algorithm thinks it is not. Showing the problem on the example of the internal E2E tests.

| Human | Machine |
|:-----:|:-------:|
|DETOX_VISIBILITY_RCTScrollView  0x7fcaf9dd0af0 _SCREEN|DETOX_VISIBILITY_RCTScrollView  0x7fcaf9dd0af0 _TEST|

LeoNatan: Probably not enough percentage of visibility. Other than implementing a visibility policy setting, I don't see what else can be done here.

noomorph: Can we check absolute value of visible area? E.g., if it is about 40pxx40px area (or whatever Apple HIG say) AND 50% of area is visible, then tap can be executed... 🤔

LeoNatan: I don’t understand what that means.

LeoNatan: Also, keyboard has different proportions on different devices.

noomorph: How do you compare against 75% threshold? What is your formula to calculate the total area vs visible area?

LeoNatan: Count visible pixels, divide by total.

noomorph: When you count visible pixels, can you tell what the biggest visible area region is, and what its dimensions?

detox/test/e2e/28.visibility.test.js
```diff
@@ -24,6 +24,16 @@ describe('Visibility Cases', () => {
     });
   });
 
+  it(':ios: should be able to dismiss the keyboard by scrolling down in the ScrollView', async () => {
```
LeoNatan, line 27: This won't magically work on its own. It requires that the screen add support for this in its code.

LeoNatan:
Now that #2375 has landed, please merge master and try running your tests. Tap to dismiss keyboard should work.

detox/test/e2e/28.visibility.test.js
noomorph: Got it. I've removed the test until the better times come. 😄

noomorph:
@LeoNatan , have you had a chance to look at the case number 3?

LeoNatan:
@noomorph That's a bad implementation. The badge should be part the button hierarchy. There is nothing I can do about it.

PR #2393, opened 2020-10-08
fix(jest): allow `-t, --testNamePattern` overriding

### LINKED ISSUE #1881: Pick tests by the name
**Problem**
Detox allows run specific tests by passing a path/fileName only. E.g. `detox test e2e/testSuite.spec.js`.
There is no way to run a specific test list from the suite. E.g. `testSuite.spec.js` script is
```
it('test name 1', () => {...});
it('test name 2', () => {...});
it('test name 3, () => {...});
it('test name 4, () => {...});
...
```
I want to run `test name 1` and `test name 3` tests only (I'm not talking about development stage where I can use `it.only`)

**Usage**
I have a huge number of e2e tests that take a lot of time. I can't run all of them on every PR/commit coz it will significantly increase development cycle time. Instead, I want to create a specific list of the tests that will run (kind of sanity suite). 

**Solution**
If Detox runs with Jest, I can run a specific list of the tests by their name using `--testNamePattern` (https://jestjs.io/docs/en/cli#--testnamepatternregex). Currently, it's occupied by a regexp that Detox sets `--testNamePattern=^((?!:android:).)*$` for iOS ad `--testNamePattern=^((?!:ios:).)*$` for Android, so the value I pass is ignored.

noomorph:
**Description:**

Resolves #1881.

E.g.:

```bash
detox test -c ios.sim.release -t hello.*tap
```

as expected, translates to:

```bash
configuration="ios.sim.release" ... jest ... --testNamePattern "hello.*tap"
```

instead of the former:

```bash
configuration="ios.sim.release" ... jest ...  --testNamePattern '^((?!:android:).)*$'  --testNamePattern "hello.*tap"
```

/cc @andy-zy

### DIFF
--- a/detox/local-cli/test.js
+++ b/detox/local-cli/test.js
@@ -131,14 +131,14 @@ function prepareJestArgs({ cliConfig, runnerArgs, runnerConfig, platform }) {
   const platformFilter = getPlatformSpecificString(platform);
 
   return {
-    argv: {
+    argv: _.omitBy({
       color: !cliConfig.noColor && undefined,
       config: runnerConfig.runnerConfig /* istanbul ignore next */ || undefined,
       testNamePattern: platformFilter ? shellQuote(`^((?!${platformFilter}).)*$`) : undefined,
       maxWorkers: cliConfig.workers,
 
       ...passthrough,
-    },
+    }, _.isUndefined),
 
     env: _.omitBy({
       ..._.pick(cliConfig, _.compact([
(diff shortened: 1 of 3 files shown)


d4vidi:
@noomorph sorry for reviewing this late; I was just wondering, perhaps a better strategy would be to append the user's custom regexp to detox' automatically generated one, rather than omitting it altogether? I'm just thinking that from the user's perspective, the two concepts are non-overlapping: As a user, while providing a custom pattern of my own, I would still like it run strictly on the platform associated with my configuration...

PR #2395, opened 2020-10-08
DETOX_ARGV_OVERRIDE env var

noomorph:
**Description:**

E.g., you have a failing build on CI, but there's:
1. --loglevel verbose, but you want `--loglevel trace`.
2. there is one suite failing and you want to re-run it ASAP, e.g.: `e2e/03.actions.test.js`
3. your build is hanging, and without adding `--forceExit` to Jest you are unable to get the artifacts from CI.

With this `$DETOX_ARGV_OVERRIDE` environment variable, now all the above is possible in one click:

image

The `detox test` command will merge in those args from $DETOX_ARGV_OVERRIDE variable, as if you were typing that from a script.

See the excerpt for the updated CLI guide:

#### DETOX_ARGV_OVERRIDE

If you happen to be troubleshooting Detox tests inside a complex script, or a failing CI build (e.g., on TeamCity or Jenkins), there is an escape-hatch feature for running Detox with some extra CLI args just by setting the `DETOX_ARGV_OVERRIDE` environment variable before re-running it again.

```
> export DETOX_ARGV_OVERRIDE="--forceExit -w 1 --testNamePattern='that hanging test' e2e/sanity/login.test.js"
> bash scripts/ci.e2e.sh
  # ... some output ...
  > detox test -c ios.sim.release -l verbose --workers 3
    # ...
    configuration=ios.sim.release ... jest --maxWorkers 1 --forceExit --testNamePattern='that hanging test' e2e/sanity/login.test.js
```

Consider the example above, where `DETOX_ARGV_OVERRIDE` forces Detox to run Jest in a single worker mode with a forceful exit (after 1 second) only for a selected test in a specific file.

As you might see, the idea of `DETOX_ARGV_OVERRIDE` is quite similar to [NODE_OPTIONS](https://nodejs.org/api/cli.html#cli_node_options_options) except for the fact you use it not for regular flows, but for forced ad-hoc patching of a failing Detox configuration to save your time.

Please avoid using it in your regular flows – instead, use Detox configuration files (`.detoxrc.js`) as your primary choice.


noomorph:
Hmm... maybe I'll rename it to $DETOX_OPTS, to allude to NODE_OPTS. 🤔

d4vidi:
Oh geez. This would make a very powerful feature... perhaps _too_ powerful. Was there a particular use case to have driven this?

PR #2399, opened 2020-10-09
Configurable launchArgs

noomorph:
**Description:**

This pull request adds two things.

### detox.configurations[conf].launchArgs

```diff
 "ios.sim.release": {
   "type": "ios.simulator",
   "binaryPath": "ios/build/Build/Products/Release-iphonesimulator/example.app"
+  "launchArgs": {
+    "detoxPrintBusyIdleResources": "YES"
+  },
   "device": {
     "type": "iPhone 11 Pro"
   },
 },
```

So that you can add some default `launchArgs` if your app configuration needs them all the time.

### DETOX_APP_LAUNCH_ARGS

E.g., you need to see what's wrong with the busy idle resources, so instead of going somewhere to patch your configuration, you just can re-run locally on your machine, when troubleshooting:

```bash
DETOX_APP_LAUNCH_ARGS="-detoxDebugVisibility YES" detox test -c ios.sim.release
```

and this extra option just automatically jumps in, along with the other ones.

### Conclusion

What do you think, guys?


LeoNatan:
I don’t understand why detoxDebugVisibility should be enabled in CI?

noomorph:
@LeoNatan, to debug cases when the visibility check fails only on CI. AFAIR, there was one like that already.

LeoNatan:
How will you collect from CI’s Deksyop?

noomorph:
Well, it might become possible in `globalTeardown` phase, if we eventually build our internal Detox preset for Jest.

Plus, the purpose was not limited to CI.
I just want to be able to re-run Detox locally with extra launch args (usually, yes, for troubleshooting) but without rummaging where do they launch their app in the project.

noomorph:
@LeoNatan, I've edited my example according to your request.

LeoNatan:
Some care needs to be put with future expansions of Detox, for example supporting multiple apps. 

Why is this necessary as an environment variable, compared to making a branch with changed arguments in init file?

noomorph:
@LeoNatan, the current Detox implementation fails miserably at handling multiple applications at once, so until a big rewrite in JS part, I don't think it is a subject for the discussion at the moment – and the said rewrite will 100% result into breaking changes for the public API.

And if you think that launchArgs break many-apps concept, then why does not having "binaryPath" alongside conflict with the concept? It is also a dealbreaker for many-to-many relations between apps and devices.

As for why, I'll come from another side – we already have for a long time those `--device-launch-args ""` CLI support, so why can't we have both, for customizing app launches, for the balance?

Of course, while making a branch is always an option, for example, it steals my time when debugging 3-4 projects in one day, and also it complicates life for people who submit new bug reports to Detox Github Issues. I mean, when we request people to send more logs and artifacts, it is a bit problematic to explain how in one step they can extract all the necessary evidence after they run. Telling:

> Please re-run your Detox with `DETOX_APP_LAUNCH_ARGS="-detoxPrintBusyIdleResources YES" detox test -c <your-config> --loglevel trace" and submit your report here.

sounds easier for me than explaining to go somewhere and change their app launch args, especially given that with the current Detox + Jest Circus implementation the app launch happens implicitly, within `CustomDetoxEnvironment`, unless you set `behavior.launchApp = false` and add your custom `beforeAll(() => device.launchApp({ ...options }))` step.

LeoNatan:
Path is an old issue which has been discussed in the past how to resolve. But you are suggesting let's pile on more stuff just because some old ones exist. 

At some point feature creep is a Bad Thing™, and it doesn't sound to me like a compelling feature. 🤷‍♂️

The entire auto launch is also a Bad Thing™, one that we discussed with @rotemmiz to remove from init() process. App launching should be entirely the responsibility of the test.

noomorph:
> Path is an old issue which has been discussed in the past how to resolve. But you are suggesting let's pile on more stuff just because some old ones exist.

Yes, let's pile, since anyway that fits the unfortunate existing direction of the project circa 2017, which we haven't managed to drastically change, so instead of struggle with it, I opt to adopt it and leverage the most of it.

> At some point feature creep is a Bad Thing™, and it doesn't sound to me like a compelling feature. 🤷‍♂️

I don't feel like this particular feature is bloating the system and makes its design more complex.
The configuration workflow `JS/JSON (detoxrc) < environment variables < command line args` is ingrained now everywhere in Detox. Here I just have aligned the blank spot on the map to fit the general trend.

I can't say our official troubleshooting techniques are easy to follow – there are so many toggles and switches in the system (debug synchronization, loglevel, artifacts, advanced launch args), so it becomes not trivial for a newcomer to figure out how to make Detox spit the necessary debug info, as it is scattered and dependent on many factors.

And I don't try to fight this toggles-n-switches system at the moment, I am just trying to make sure all the handles are available for immediate use, first of all. If we are talking about a well-designed and easy-to-use system, then all Detox needs is to provide a single diagnostics mode (ON/OFF), which just works and gives you every detail you need, so you can relax and turn off your brain, but this is a feature we can't really afford in the scope of this month, I am afraid, yet still, there's a lot of pending bug fixing work.

And since we have not approached that holy grail, I am fixing issues within the current paradigm, even if that means to provide certain shortcuts for quicker troubleshooting.

> The entire auto launch is also a Bad Thing™, one that we discussed with @rotemmiz to remove from init() process. App launching should be entirely the responsibility of the test.

Here I totally agree, and I think we can implement that in the next Detox 18, if we all share this vision. 👍

LeoNatan:
If love to. I think one of the main issues in making this change is communicating it to the community. If we just stop launching apps, people will cry because virtually all tests will break. If we throw a warning now and stop launching in the future, that won’t change a thing because people are reliably lazy and don’t read logs, even when their tests are failing. 🤦‍♂️ This is why an initial design and implementation is very important, and just piling on features is a Bad Thing™, and we should pay concern to at least things we’ve though about, such as multiple apps.

noomorph:
I have no issue with releasing breaking changes when necessary.

The problem here is that although I've been working on Detox from 7.x up till now, 17.x, those Bad Things™ remained intact in spite of many other things changing. I just don't believe in the multiple-apps feature without a prior heavy internal and external redesign.

Looking at our priorities distribution (perf project, internal preset) I am very doubtful we'll arrive at the feature you mention during this 2020. 😑 I hope I am wrong.

rotemmiz:
I like this feature, but we need to make sure it actually solves the problem not only with static launch args. I think it makes a lot of sense, especially when the config is a js file that can dynamically resolve launch args (mock server ports, etc.), though I am not sure it solves it all.

The main usecase we see internally is wrapping `device.launchApp()` just to add in all the required launch args in one place, this of course, solves that, especially with the example you showed.
I suggest you check how dynamic resolution works during global setup and during suite setup, so we'll have good understanding of setting up dynamic launch args through `.dtxrc.js`

rotemmiz:
Regarding @LeoNatan's concern, well, although I agree with you, I think that whatever we throw into `configuration` is really only for one application.
If and when we'd want to support multiple apps in we'd need multiple instances, that can be represented as an array/map inside a single configuration, each should probably take instance should take its configuration from a different config in that array/map.
I agree this is not an elaborate design, and we might come up with a totally different way here, but I still think the suggested implementation fits.

d4vidi:
Guys I'm sorry -- 'didn't go over the whole thread here. However, as a bottom line I agree this feature adds value and I'm for it. 'Only concern I have is exactly as @rotemmiz has described it: dynamic key/value resolution. I've found myself already, in the past, at situations where I was swizzling `device.launchApp()` for projects, incorporating dynamic resolution logic. I suppose a `.dtxrc.js` would do the trick here, albeit being a bit implicit to my taste. Perhaps a more elegant way is around the corner, but it shouldn't block moving forward here.

detox/src/errors/DetoxConfigErrorBuilder.js
```diff
@@ -87,6 +87,15 @@ class DetoxConfigErrorBuilder {
     });
   }
 
+  malformedAppLaunchArgs() {
```
d4vidi, line 90: What about unit tests here?

detox/src/configuration/composeDeviceConfig.js
```diff
@@ -29,10 +36,24 @@ function validateUtilBinaryPaths({ errorBuilder, rawDeviceConfig }) {
...
+    rawDeviceConfig.launchArgs = _.chain({})
+      .thru(() => parse(cliConfig.appLaunchArgs, {
+        configuration: {
+          'short-option-groups': false,
+        },
+      }))
+      .omit(['_', '--'])
+      .defaults(rawDeviceConfig.launchArgs)
+      .omitBy(value => value === false)
+      .value();
```
d4vidi, line 54: This part really wants to have its own function :-)

detox/src/configuration/composeDeviceConfig.js
```diff
@@ -29,10 +36,24 @@ function validateUtilBinaryPaths({ errorBuilder, rawDeviceConfig }) {
...
+  if (cliConfig.appLaunchArgs) {
+    rawDeviceConfig.launchArgs = _.chain({})
+      .thru(() => parse(cliConfig.appLaunchArgs, {
+        configuration: {
+          'short-option-groups': false,
+        },
+      }))
+      .omit(['_', '--'])
+      .defaults(rawDeviceConfig.launchArgs)
+      .omitBy(value => value === false)
```
d4vidi, line 53: Will this not also excessively omit values intentionally set to `false`?
```js
launchArgs: {
   verbose: false,
   someCustomDebug: false,
}
```

etc.

detox/src/configuration/collectCliConfig.js
```diff
@@ -17,6 +17,7 @@ function collectCliConfig({ argv }) {
     configuration: get('configuration'),
     debugSynchronization: get('debug-synchronization'),
     deviceLaunchArgs: get('device-launch-args'),
+    appLaunchArgs: get('DETOX_APP_LAUNCH_ARGS'),
```
d4vidi, line 20: I'm not entirely sure I follow, here: the expectation is that `DETOX_APP_LAUNCH_ARGS` would be set as an env-var by user? If so, why overrule the option of a cli arg?

d4vidi: Also, in any case -- should we not cover this at some integrative/e2e test point?

detox/src/configuration/composeDeviceConfig.js
noomorph: @d4vidi, there is no reason to set those values to false since anyway they become converted to strings. If you want "false", you write `"false"`, not `false`.

The key idea about this `false` was to enable dropping launchArgs from CLI **(!)**, e.g. `--no-someCustomDebug` will just throw out the corresponding key-value from the default `launchArgs` map.

What we can do, however, is to throw an error on an attempt to use non-strings in that object (except for `false`, obviously).

detox/src/configuration/composeDeviceConfig.js
noomorph: Thanks for telling me about that secret wish!
Ho ho ho, where is my Santa suite 🎅 ?! 🦌 🦌 🦌

detox/src/errors/DetoxConfigErrorBuilder.js
noomorph: Ooooh, right. 😄 I've really forgotten I had restored snapshot tests there... 😊 

and... pssst, man...



Just make sure to not tell Leo about those snapshot tests. 😆

detox/src/configuration/collectCliConfig.js
noomorph: It is covered in `local-cli/test.test.js`, but you are right — I indeed should change it to `--app-launch-args` property **plus** `DETOX_APP_LAUNCH_ARGS` environment variable. 

The point is that I'd like eventually to rewrite all environment variable names to `DETOX_*` to scope ourselves from the other applications running on users' computers and CI. That's why I thought that we should already give some start for this process so that in next `detox@18` all env vars follow that convention.

detox/src/configuration/composeDeviceConfig.js
noomorph: Moved to a separate function.

detox/src/configuration/composeDeviceConfig.js
noomorph: I assume we can close that — also I've added an extra validation to validate against non-string values in that launchArgs object, so this situation you have described now is just impossible.

d4vidi: Yes, I spotted a `--no-someCustomDebug` equivalent in your unit tests, and thus realized what you were trying to do. Anyways, seems like we're good to go!

detox/src/errors/DetoxConfigErrorBuilder.js
d4vidi: hehe that is probably the weirdest reference I've ever come across

detox/src/configuration/composeDeviceConfig.js
d4vidi: rofl thanks 🎅🏻

PR #2405, opened 2020-10-12
Support scroll view transforms

LeoNatan:


LeoNatan:
The detection for "not scrollable" has changed. Now, a scroll is attempted always; if the scrollview content offset does not change for some amount of touch moves, the scroll operation is failed.

PR #2407, opened 2020-10-13
Fix generation bug, causing overloaded UiDevice.pressKeyCode to malform

### LINKED ISSUE #2381: device.getUiDevice().pressKeyCode method does not work.
**Describe the bug**

There is a bug in https://github.com/wix/Detox/blob/master/detox/src/android/espressoapi/UIDevice.js#L153-L180

If you want to use `device.getUiDevice().pressKeyCode` method as below, 
```js
  const uidevice = device.getUiDevice();
  await uidevice.pressKeyCode(24);
```
It will give you an error "**keyCode should be a number, but got [object Object] (object)**".

**To Reproduce**

- [x] I have tested this issue on the latest Detox release and it still reproduces

 - Detox: 17.7.1
 - React Native: 0.61.5
 - Node: 12.18.3
 - Device: android 10 Essential PH-1
 - OS: Ubuntu 16.04
 - Test-runner (select one): `jest-circus` 

**Solution:**
The bug is caused by an incorrect `arguments` length condition and reusage. Please change the `pressKeyCode` method as below: 
```js
  static pressKeyCode(element, keyCode) {
    function pressKeyCode1(element, keyCode) {
      if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
      return {
        target: element,
        method: "pressKeyCode",
        args: [{
          type: "Integer",
          value: keyCode
        }]
      };
    }

    function pressKeyCode2(element, keyCode, metaState) {
      if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
      if (typeof metaState !== "number") throw new Error("metaState should be a number, but got " + (metaState + (" (" + (typeof metaState + ")"))));
      return {
        target: element,
        method: "pressKeyCode",
        args: [{
          type: "Integer",
          value: keyCode
        }, {
          type: "Integer",
          value: metaState
        }]
      };
    }

    if (arguments.length === 1) throw new Error("keyCode should not be null.");

    if (arguments.length === 2) {
      return pressKeyCode1.apply(null, arguments);
    } 

    return pressKeyCode2.apply(null, arguments);
  }

```
I know the original code was auto-generated from the Android source file, then something went wrong in the `generation` sub-project. I have not got a chance to check it.

d4vidi:
- [x] This is a small change 

---

**Description:**
Resolves #2381.

### DIFF
--- a/detox/src/android/espressoapi/UIDevice.js
+++ b/detox/src/android/espressoapi/UIDevice.js
@@ -151,7 +151,7 @@ class UiDevice {
   }
 
   static pressKeyCode(element, keyCode) {
-    function pressKeyCode1(keyCode) {
+    function pressKeyCode2(element, keyCode) {
       if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
       return {
         target: element,
@@ -163,7 +163,7 @@ class UiDevice {
       };
     }
 
-    function pressKeyCode2(keyCode, metaState) {
+    function pressKeyCode3(element, keyCode, metaState) {
       if (typeof keyCode !== "number") throw new Error("keyCode should be a number, but got " + (keyCode + (" (" + (typeof keyCode + ")"))));
       if (typeof metaState !== "number") throw new Error("metaState should be a number, but got " + (metaState + (" (" + (typeof metaState + ")"))));
       return {
@@ -179,12 +179,12 @@ class UiDevice {
       };
     }
 
-    if (arguments.length === 1) {
-      return pressKeyCode1.apply(null, arguments);
-    }
-
     if (arguments.length === 2) {
       return pressKeyCode2.apply(null, arguments);
+    }
+
+    if (arguments.length === 3) {
+      return pressKeyCode3.apply(null, arguments);
     }
   }
 
--- a/detox/src/android/espressoapi/UiDeviceProxy.js
+++ b/detox/src/android/espressoapi/UiDeviceProxy.js
@@ -1,5 +1,5 @@
 const uiDevice = require('./UIDevice');
-const uiAutomaton = require('./UIAutomator');
+const uiAutomator = require('./UIAutomator');
 const invoke = require('../../invoke');
 
 class UiDeviceProxy {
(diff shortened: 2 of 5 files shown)


generation/index.js
```diff
@@ -7,39 +7,45 @@ const espressoFilesToDownload = {
...
+  const downloadedEspressoFilesMap =
+    Object
+      .entries(espressoFilesToDownload)
+      .reduce(function (obj, [fullyQualifiedClass, dest]) {
+          obj[downloadEspressoFileByClass(fullyQualifiedClass)] = dest;
+          return obj;
+        }, {}
+      );
 
-const androidFiles = {
```
d4vidi, line 34: no actual changes here, just pushed all the logic into a function

detox/src/android/espressoapi/UiDeviceProxy.js
```diff
@@ -1,5 +1,5 @@
 const uiDevice = require('./UIDevice');
-const uiAutomaton = require('./UIAutomator');
+const uiAutomator = require('./UIAutomator');
```
d4vidi, line 2: typo fix

detox/src/android/espressoapi/UIDevice.js
```diff
@@ -151,7 +151,7 @@ class UiDevice {
   }
 
   static pressKeyCode(element, keyCode) {
-    function pressKeyCode1(keyCode) {
+    function pressKeyCode2(element, keyCode) {
```
d4vidi, line 154: This is the result of the new generator invocation. I've temporarily enabled and then redisabled the generation of uidevice.js (see changes in `generation/index.js`)

noomorph: Does that mean that the new implementation is incorrect to some extent?

detox/src/android/espressoapi/UiDeviceProxy.js
noomorph: 👍

detox/src/android/espressoapi/UIDevice.js
d4vidi: What I meant was that I've regenerated UIDevice locally after the changes and then had it removed from generation's files list

PR #2427, opened 2020-10-19
Improve content of native errors shown on the JS test runner

d4vidi:
- [ ] This is a small change 

---

**Description:**
Often enough, errors sent from the device's native fail to convey proper meaning to test user. Usually (perhaps always), this happens because the test runner side (JS) only logs the top-most wrapping error, and not the very original one, specified by the error's `cause` field (recursively).

This attempts at fixing that by drilling down to the root cause of errors, before sending them to the test-runner side of Detox. As an example of why this works, it is verified in the recently added slow-tap e2e test in the actions suite.

### DIFF
--- a/detox/src/client/Client.test.js
+++ b/detox/src/client/Client.test.js
@@ -9,6 +9,7 @@ describe('Client', () => {
   let WebSocket;
   let Client;
   let client;
+  let bunyan;
   let log;
 
   beforeEach(() => {
@@ -18,8 +19,12 @@ describe('Client', () => {
     jest.mock('../utils/argparse');
     argparse = require('../utils/argparse');
 
+    jest.mock('bunyan')
+    bunyan = require('bunyan');
+
     Client = require('./Client');
     log = require('../utils/logger');
+    log.level.mockReturnValue(bunyan.DEBUG)
   });
 
   it(`reloadReactNative() - should receive ready from device and resolve`, async () => {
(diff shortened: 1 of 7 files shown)


d4vidi:
Yikes, I wasn't planning on merging this just yet. Could you guys review nonetheless?

noomorph:
Generally, looks good, but I'll re-check soon one more time.

detox/src/client/Client.test.js
```diff
@@ -279,14 +284,41 @@ describe('Client', () => {
...
+
+  it(`execute() - "testFailed" result should throw without a view hierarchy`, async () => {
     await connect();
-    client.ws.send.mockReturnValueOnce(response("testFailed",  {details: "this is an error"}, 1));
+    client.ws.send.mockReturnValueOnce(response("testFailed",  {details: "this is an error", viewHierarchy: undefined}, 1));
     const call = invoke.call(invoke.IOS.Class('GREYMatchers'), 'matcherForAccessibilityLabel:', 'test');
     try {
       await client.execute(call);
     } catch (ex) {
       expect(ex).toBeDefined();
```
d4vidi, line 320: @noomorph note that this assertion was never enough. The test passes but the actual error was coming from the comparison of `log.level() <= bunyan.DEBUG` with `log.level()` returning `undefined`, and NOT the actual error being expected here. The additional mocks (see above) and tests are hence needed here, and thus fix this situation.

Issue #2428, opened 2020-10-20
Add offsetPositionX and offsetPositionY to `swipe(direction, speed, percentage)`


rotemmiz:
One use case: Trying to swipe a fullscreen scrollviews with header on top it, to trigger reload, will fail.
Detox tries to swipe from to farthest position possible, so that the swipe will be maximal.

We want to add `offsetPositionX` and `offsetPositionY` to both platforms to mitigate such issues.

LeoNatan:
I’d like us to also consider changing all our legacy x/y APIs to be normalized x/y (percentage of view instead of points). I think tap is the only one that remains with this. Perhaps phase it over several releases?

LeoNatan:
@rotemmiz @d4vidi ^

noomorph:
Unblocks CollapsingHeader tests in https://github.com/wix-private/mobile-guild/issues/169

noomorph:
Resolved by #2432.

PR #2432, opened 2020-10-23
Add x,y coordinate support for swipe action

noomorph:
- [x] This change has been discussed in issue #2428 and the solution has been agreed upon with maintainers.
- [x] JS native implementation
- [x] iOS native implementation
- [x] Android native implementation
- [x] Fix code review comments
- [x] Ensure consistent parameter naming
- [x] Add a few more tests for various `fast=true/false, startPointX/Y=NaN/something, offset=NaN/0/0.5/1` tests
- [x] Re-cover Android implementation with unit-tests

---

**Description:**

Signature:

```js
/*
 * @param { 'up' | 'down' | 'left' | 'right' } direction - Direction to swipe
 * @param { 'fast' | 'slow' } speed
 * @param {number} [normalizedSwipeOffset] - swipe amount relative to the screen width/height
 * @param {number} [normalizedStartingPointX] - X coordinate of swipe starting point, relative to the view width
 * @param {number} [normalizedStartingPointY] - Y coordinate of swipe starting point, relative to the view height
 */
function swipe(direction, speed, normalizedSwipeOffset = NaN, normalizedStartingPointX = NaN, normalizedStartingPointY = NaN) { ... }
```

Usage:

```js
await element(by.id('ScrollView161')).swipe('up');
await element(by.id('ScrollView161')).swipe('down', 'slow');
await element(by.id('ScrollView161')).swipe('left', 'fast', 0.95);
await element(by.id('ScrollView161')).swipe('right', 'slow', NaN, 0.3);
await element(by.id('ScrollView161')).swipe('up', 'fast', undefined, 0.5, 0.95);
```

### DIFF
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
@@ -2,6 +2,7 @@ package com.wix.detox.espresso;
 
 import android.view.View;
 
+import com.wix.detox.common.DetoxErrors;
 import com.wix.detox.common.DetoxErrors.DetoxRuntimeException;
 import com.wix.detox.common.DetoxErrors.StaleActionException;
 import com.wix.detox.espresso.action.DetoxMultiTap;
@@ -10,6 +11,7 @@ import com.wix.detox.espresso.action.TakeViewScreenshotAction;
 import com.wix.detox.espresso.common.annot.MotionDir;
 import com.wix.detox.espresso.scroll.ScrollEdgeException;
 import com.wix.detox.espresso.scroll.ScrollHelper;
+import com.wix.detox.espresso.scroll.SwipeHelper;
 
 import org.hamcrest.Matcher;
 
@@ -20,13 +22,9 @@ import androidx.test.espresso.action.GeneralClickAction;
 import androidx.test.espresso.action.GeneralLocation;
 import androidx.test.espresso.action.GeneralSwipeAction;
 import androidx.test.espresso.action.Press;
-import androidx.test.espresso.action.Swipe;
+import androidx.test.espresso.action.ViewActions;
 
 import static androidx.test.espresso.action.ViewActions.actionWithAssertions;
-import static androidx.test.espresso.action.ViewActions.swipeDown;
-import static androidx.test.espresso.action.ViewActions.swipeLeft;
-import static androidx.test.espresso.action.ViewActions.swipeRight;
-import static androidx.test.espresso.action.ViewActions.swipeUp;
 import static androidx.test.espresso.matcher.ViewMatchers.isAssignableFrom;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static com.wix.detox.espresso.common.annot.MotionDefs.MOTION_DIR_DOWN;
@@ -168,67 +166,21 @@ public class DetoxAction {
         });
     }
 
-    private final static float EDGE_FUZZ_FACTOR = 0.083f;
-
     /**
      * Swipes the View in a direction.
      *
      * @param direction Direction to swipe (see {@link MotionDir})
      * @param fast true if fast, false if slow
-     *
+     * @param normalizedOffset or "swipe amount" between 0.0 and 1.0, relative to the screen width/height
+     * @param normalizedStartingPointX X coordinate of swipe starting point (between 0.0 and 1.0), relative to the view width
+     * @param normalizedStartingPointY Y coordinate of swipe starting point (between 0.0 and 1.0), relative to the view height
      */
-    public static ViewAction swipeInDirection(final int direction, boolean fast) {
-        if (fast) {
-            switch (direction) {
-                case MOTION_DIR_LEFT:
-                    return swipeLeft();
-                case MOTION_DIR_RIGHT:
-                    return swipeRight();
-                case MOTION_DIR_UP:
-                    return swipeUp();
-                case MOTION_DIR_DOWN:
-                    return swipeDown();
-                default:
-                    throw new RuntimeException("Unsupported swipe direction: " + direction);
-            }
-        }
-
-        switch (direction) {
-            case MOTION_DIR_LEFT:
-                return actionWithAssertions(new GeneralSwipeAction(Swipe.SLOW,
-                        translate(GeneralLocation.CENTER_RIGHT, -EDGE_FUZZ_FACTOR, 0),
-                        GeneralLocation.CENTER_LEFT, Press.FINGER));
-            case MOTION_DIR_RIGHT:
-                return actionWithAssertions(new GeneralSwipeAction(Swipe.SLOW,
-                        translate(GeneralLocation.CENTER_LEFT, EDGE_FUZZ_FACTOR, 0),
-                        GeneralLocation.CENTER_RIGHT, Press.FINGER));
-            case MOTION_DIR_UP:
-                return actionWithAssertions(new GeneralSwipeAction(Swipe.SLOW,
-                        translate(GeneralLocation.BOTTOM_CENTER, 0, -EDGE_FUZZ_FACTOR),
-                        GeneralLocation.TOP_CENTER, Press.FINGER));
-            case MOTION_DIR_DOWN:
-                return actionWithAssertions(new GeneralSwipeAction(Swipe.SLOW,
-                        translate(GeneralLocation.TOP_CENTER, 0, EDGE_FUZZ_FACTOR),
-                        GeneralLocation.BOTTOM_CENTER, Press.FINGER));
-            default:
-                throw new RuntimeException("Unsupported swipe direction: " + direction);
-        }
+    public static ViewAction swipeInDirection(final int direction, boolean fast, double normalizedOffset, double normalizedStartingPointX, double normalizedStartingPointY) {
+        SwipeHelper swipeHelper = SwipeHelper.getDefault();
+        return swipeHelper.swipeInDirection(direction, fast, normalizedOffset, normalizedStartingPointX, normalizedStartingPointY);
     }
 
     public static ViewAction takeViewScreenshot() {
         return new TakeViewScreenshotAction();
-    }
-
-    private static CoordinatesProvider translate(final CoordinatesProvider coords,
-                                                 final float dx, final float dy) {
-        return new CoordinatesProvider() {
-            @Override
-            public float[] calculateCoordinates(View view) {
-                float xy[] = coords.calculateCoordinates(view);
-                xy[0] += dx * view.getWidth();
-                xy[1] += dy * view.getHeight();
-                return xy;
-            }
-        };
     }
 }
--- a/detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
+++ b/detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
@@ -1,0 +1,96 @@
+package com.wix.detox.espresso.scroll
+
+import androidx.test.espresso.ViewAction
+import androidx.test.espresso.action.*
+import com.wix.detox.espresso.common.annot.*
+import com.wix.detox.espresso.utils.*
+import kotlin.math.abs
+import kotlin.math.min
+import kotlin.math.max
+
+private const val EDGE_FUZZ_FACTOR = 0.083
+
+private fun minMax(minValue: Double, value: Double, maxValue: Double) = max(minValue, min(value, maxValue))
+
+private fun ifNaN(value: Double, fallback: Double) = when {
+    value.isNaN() -> fallback
+    else -> value
+}
+
+typealias CreateSwipeAction = (
+        swiper: Swiper,
+        startCoordinatesProvider: CoordinatesProvider,
+        endCoordinatesProvider: CoordinatesProvider,
+        precisionDescriber: PrecisionDescriber
+) -> ViewAction
+
+class SwipeHelper(private val createAction: CreateSwipeAction) {
+
+    fun swipeInDirection(
+            direction: Int,
+            fast: Boolean,
+            normalizedSwipeAmount: Double,
+            normalizedStartingPointX: Double,
+            normalizedStartingPointY: Double
+    ): ViewAction {
+        val (edgeMin, edgeMax) = Pair(EDGE_FUZZ_FACTOR, 1.0 - EDGE_FUZZ_FACTOR)
+        val defaultNormalizedStartingPoint = Vector2D(0.5, edgeMin).rotate(direction, MOTION_DIR_DOWN).normalize()
+        val safeNormalizedStartPoint = Vector2D(
+                minMax(edgeMin, ifNaN(normalizedStartingPointX, defaultNormalizedStartingPoint.x), edgeMax),
+                minMax(edgeMin, ifNaN(normalizedStartingPointY, defaultNormalizedStartingPoint.y), edgeMax)
+        )
+
+        val safeSwipeAmount = minMax(0.0, ifNaN(normalizedSwipeAmount, 0.75), 1.0)
+        val startCoordinatesProvider = buildStartCoordinatesProvider(safeNormalizedStartPoint)
+        val endCoordinatesProvider = buildEndCoordinatesProvider(startCoordinatesProvider, direction, safeSwipeAmount)
+        val swiper = if (fast) Swipe.FAST else Swipe.SLOW
+
+        return this.createAction(swiper, startCoordinatesProvider, endCoordinatesProvider, Press.FINGER)
+    }
+
+    private fun buildStartCoordinatesProvider(normalizedStartPoint: Vector2D) = CoordinatesProvider { view ->
+        val xy = GeneralLocation.TOP_LEFT.calculateCoordinates(view)
+        xy[0] += (normalizedStartPoint.x * view.width).toFloat()
+        xy[1] += (normalizedStartPoint.y * view.height).toFloat()
+        xy
+    }
+
+    private fun buildEndCoordinatesProvider(startCoordinatesProvider: CoordinatesProvider, direction: Int, normalizedSwipeAmount: Double) =
+            CoordinatesProvider { view ->
+                val xy = startCoordinatesProvider.calculateCoordinates(view)
+
+                val screenEdge = Vector2D.from(
+                        view.context.resources.displayMetrics.widthPixels,
+                        view.context.resources.displayMetrics.heightPixels
+                )
+
+                val additionVector = Vector2D(0.0, normalizedSwipeAmount).rotate(direction, MOTION_DIR_DOWN)
+                val swipeEnd = Vector2D.from(xy)
+                        .add(screenEdge.scale(additionVector))
+                        .trimMax(0.0, 0.0)
+                        .trimMin(screenEdge.x, screenEdge.y)
+
+                xy[0] = swipeEnd.x.toFloat()
+                xy[1] = swipeEnd.y.toFloat()
+                xy
+            }
+
+    companion object {
+        @JvmStatic
+        val default = SwipeHelper { swiper: Swiper,
+                                    startCoordinatesProvider: CoordinatesProvider,
+                                    endCoordinatesProvider: CoordinatesProvider,
+                                    precisionDescriber: PrecisionDescriber ->
+            ViewActions.actionWithAssertions(
+                    GeneralSwipeAction(
+                            swiper,
+                            startCoordinatesProvider,
+                            endCoordinatesProvider,
+                            precisionDescriber
+                    )
+            );
+        }
+
+        const val edgeFuzzFactor = EDGE_FUZZ_FACTOR.toFloat()
+    }
+}
--- a/detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwipeHelperSpec.kt
+++ b/detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwipeHelperSpec.kt
@@ -1,0 +1,213 @@
+package com.wix.detox.espresso.scroll
+
+import android.content.Context
+import android.content.res.Resources
+import android.util.DisplayMetrics
+import android.view.View
+import androidx.test.espresso.ViewAction
+import androidx.test.espresso.action.*
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.whenever
+import com.wix.detox.espresso.common.annot.*
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+import kotlin.test.assertEquals
+
+data class SwipeArguments(
+        val direction: Int,
+        val fast: Boolean,
+        val normalizedSwipeAmount: Double,
+        val normalizedStartingPointX: Double,
+        val normalizedStartingPointY: Double
+)
+
+data class SwipeCoordinates(
+        val startX: Float,
+        val startY: Float,
+        val endX: Float,
+        val endY: Float
+)
+
+data class RectangleCalculator(val x: Int, val y: Int, val width: Int, val height: Int, val fuzz: Float) {
+    val left: Float get() = x + width * 0.0f
+    val center: Float get() = x + width * 0.5f
+    val right: Float get() = x + width * 1.0f
+    val top: Float get() = y + height * 0.0f
+    val middle: Float get() = y + height * 0.5f
+    val bottom: Float get() = y + height * 1.0f
+
+    val safeLeft: Float get() = x + width * (0f + fuzz)
+    val safeRight: Float get() = x + width * (1f - fuzz)
+    val safeTop: Float get() = y + height * (0f + fuzz)
+    val safeBottom: Float get() = y + height * (1f - fuzz)
+}
+
+@RunWith(Parameterized::class)
+class SwipeHelperTest(
+        val args: SwipeArguments,
+        val expected: SwipeCoordinates
+) {
+    lateinit var action: ViewAction
+    lateinit var swiper: Swiper
+    lateinit var startCoordinatesProvider: CoordinatesProvider
+    lateinit var endCoordinatesProvider: CoordinatesProvider
+    lateinit var precisionDescriber: PrecisionDescriber
+    lateinit var swipeResult: Any
+
+    private val swipeHelper = SwipeHelper { _swiper: Swiper,
+                                            _startCoordinatesProvider: CoordinatesProvider,
+                                            _endCoordinatesProvider: CoordinatesProvider,
+                                            _precisionDescriber: PrecisionDescriber ->
+        swiper = _swiper
+        startCoordinatesProvider = _startCoordinatesProvider
+        endCoordinatesProvider = _endCoordinatesProvider
+        precisionDescriber = _precisionDescriber
+        action = mock()
+        action
+    }
+
+    @Before
+    fun setup() {
+        swipeResult = swipeHelper.swipeInDirection(
+                args.direction,
+                args.fast,
+                args.normalizedSwipeAmount,
+                args.normalizedStartingPointX,
+                args.normalizedStartingPointY
+        )
+    }
+
+    @Test
+    fun shouldReturnAction() {
+        assertEquals(action, swipeResult)
+    }
+
+    @Test
+    fun shouldUseFingerPrecision() {
+        assertEquals(Press.FINGER, precisionDescriber)
+    }
+
+    @Test
+    fun shouldCalculateCorrectStartPoint() {
+        val mockView = setupMockView()
+        val startXY = startCoordinatesProvider.calculateCoordinates(mockView)
+        val expectedStart = Pair(expected.startX, expected.startY)
+        val actualStart = Pair(startXY[0], startXY[1])
+        assertEquals(expectedStart, actualStart)
+    }
+
+    @Test
+    fun shouldCalculateCorrectEndPoint() {
+        val mockView = setupMockView()
+        val endXY = endCoordinatesProvider.calculateCoordinates(mockView)
+        val expectedEnd = Pair(expected.endX, expected.endY)
+        val actualEnd = Pair(endXY[0], endXY[1])
+        assertEquals(expectedEnd, actualEnd)
+    }
+
+    companion object {
+        @JvmStatic
+        @Parameterized.Parameters(name = "{0}")
+        fun data(): Collection<Array<Any>> {
+            return listOf(
+                    arrayOf(
+                            SwipeArguments(MOTION_DIR_LEFT, true, Double.NaN, Double.NaN, Double.NaN),
+                            SwipeCoordinates(view.safeRight, view.middle, view.safeRight - 0.75f * screen.width, view.middle)
+                    ),
+                    arrayOf(
+                            SwipeArguments(MOTION_DIR_LEFT, true, 1.0, Double.NaN, Double.NaN),
+                            SwipeCoordinates(view.safeRight, view.middle, screen.left, view.middle)
+                    ),
+                    arrayOf(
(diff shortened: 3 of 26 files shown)


LeoNatan:
Implemented on iOS. Is this iOS only for now or will @d4vidi add impl for Android as well?

d4vidi:
I can get to it soon, should be fairly ez to set up

LeoNatan:
Great, thanks! @d4vidi

noomorph:
@d4vidi, @LeoNatan, need your advice regarding the swiped amount.

Look at the difference between the current iOS and Android implementations.

This happens because I calculate on Android the end coordinate not relative to the screen, but relative to the element itself. I have done that not really on purpose, just I was trying to follow the logic in the current Android implementation, and ended up in the place where I am.

So, iOS and Android behavior here differs. What would be more correct behavior, what do you think?

<table>
<tr>
<td width="50%">

</td>
<td width="50%">

</td>
</tr>
</table>

LeoNatan:
That's how it was on iOS too, but according to feedback, it had to be in screen coords. CC @viktorijasujetaite

noomorph:
@LeoNatan, thanks. I've been suspecting something like that. 😳 
Ok, will rewrite the logic on Monday, apparently.

LeoNatan:
How was it before your changes? The Espresso swipe…?

LeoNatan:
If you need to write your own logic for the start and end points, you can copy the logic from the iOS side. The logic should be the same.

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwipeHelperSpec.kt
```diff
@@ -0,0 +1,271 @@
...
+        it("should calculate right with given start X, Y and offset") {
+            val from = Pair(0.25, 0.75)
+            val to = Pair(0.5, 0.75)
+            val expected = Pair(from, to)
+            val actual = SwipeHelper.calculateSwipe(MOTION_DIR_RIGHT, 0.25, 0.25, 0.75)
+
+            assertPointsEqual(expected, actual)
+        }
+    }
+})
```
d4vidi, line 271: OMFG

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
```diff
@@ -168,67 +169,24 @@ public void perform(UiController uiController, View view) {
...
-                        GeneralLocation.TOP_CENTER, Press.FINGER));
-            case MOTION_DIR_DOWN:
-                return actionWithAssertions(new GeneralSwipeAction(Swipe.SLOW,
-                        translate(GeneralLocation.TOP_CENTER, 0, EDGE_FUZZ_FACTOR),
-                        GeneralLocation.BOTTOM_CENTER, Press.FINGER));
-            default:
-                throw new RuntimeException("Unsupported swipe direction: " + direction);
-        }
+    public static ViewAction swipeInDirection(final int direction, boolean fast, double amount, double startOffsetPercentX, double startOffsetPercentY) {
+        final boolean isInCompatibilityMode = (fast && amount == Double.NaN && startOffsetPercentX == Double.NaN && startOffsetPercentY == Double.NaN);
```
d4vidi, line 182: TBH, I was not familiar with the Java-ish representation of `NaN`. Are you 100% that this corresponds to whichever value that is provided by JS when offsets are not provided?

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+typealias Point = Pair<Double, Double>
+
+object SwipeHelper {
+    fun swipeFastInDirection(direction: Int): ViewAction? {
+        return when (direction) {
+            MOTION_DIR_LEFT -> ViewActions.swipeLeft()
+            MOTION_DIR_RIGHT -> ViewActions.swipeRight()
+            MOTION_DIR_UP -> ViewActions.swipeUp()
+            MOTION_DIR_DOWN -> ViewActions.swipeDown()
+            else -> throw RuntimeException("Unsupported swipe direction: $direction")
```
d4vidi, line 19: I think an `IllegalArgumentException` would be more suitable here

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun swipeCustomInDirection(direction: Int, fast: Boolean, offset: Double, startPositionX: Double, startPositionY: Double): ViewAction? {
+        val swiper = if (fast) Swipe.FAST else Swipe.SLOW
+        val (from, to) = calculateSwipe(direction, offset, startPositionX, startPositionY)
+        val start = translate(from)
+        val end = translate(to)
+        val swipeAction = GeneralSwipeAction(swiper, start, end, Press.FINGER)
+        return ViewActions.actionWithAssertions(swipeAction)
+    }
+
+    fun calculateSwipe(
```
d4vidi, line 32: `private` much?

d4vidi:
@noomorph I agree with @LeoNatan - swiping has to start somewhere inside the view bounds, but end as farther away as the _screen_ allows it.

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+            unsafeOffset: Double = Double.NaN,
+            unsafeStartX: Double = Double.NaN,
+            unsafeStartY: Double = Double.NaN
+    ): Pair<Point, Point> {
+        val isHorizontal = direction == MOTION_DIR_LEFT || direction == MOTION_DIR_RIGHT
+        val isVertical = direction == MOTION_DIR_UP || direction == MOTION_DIR_DOWN
+        val isDescending = direction == MOTION_DIR_UP || direction == MOTION_DIR_LEFT
+
+        if (!isHorizontal && !isVertical) {
+            throw RuntimeException("Unsupported swipe direction: $direction")
```
d4vidi, line 43: `IllegalArgumentException` 2.0

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+import com.wix.detox.espresso.common.annot.*
+import kotlin.math.min
+import kotlin.math.max
+import kotlin.math.sign
+
+typealias Point = Pair<Double, Double>
+
+object SwipeHelper {
+    fun swipeFastInDirection(direction: Int): ViewAction? {
+        return when (direction) {
```
d4vidi, line 14: Minor styling note. Can nicely utilize Kotlin's `=` here:
```
fun swipeFastInDirection(...) =
   when (direction) {
       ...
   }
```

d4vidi: BTW the IDE should have suggested that, try revisiting that

docs/APIRef.ActionsOnElement.md
```diff
@@ -116,7 +123,7 @@ await element(by.id('scrollView')).scrollTo('top');
 
 Simulates typing of the specified text into the element, using the system's builtin keyboard and typing behavior.
 
-On iOS, any element can be typed into, as long as it can become first responder and conforms to the [`UITextInput`](https://developer.apple.com/documentation/uikit/uitextinput) protocol.
+On iOS, any element can be typed into, as long as it can become first responder and conforms to the [`UITextInput`](https://developer.apple.com/documentation/uikit/uitextinput) protocol. Before typing the text, Detox attempts making the element the first responder, if it isn't already. If the element’s window is not key window, Detox attempts making it the key window.
```
d4vidi, line 126: "...Detox attempts [at] making the element the first responder..."?

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
```diff
@@ -168,67 +169,24 @@ public void perform(UiController uiController, View view) {
...
+
+        return (isInCompatibilityMode)
+            ? SwipeHelper.INSTANCE.swipeFastInDirection(direction)
```
d4vidi, line 185: I may be asking this simply because I don't understand this well enough just yet, but really - could we not give up on compatibility code and always use the new implementation, with proper fallbacks when offsets are not provided?

d4vidi: Also, in a way, compatibility is needed here in case `fast=false`, isn't it?

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
d4vidi: btw, this what android studio has to say about this
1. `isInCompatibilityMode` is always `false` (and why? -)
2. 
```
Comparison to 'Double.NaN' is always false 
 Inspection info: Reports any equality or inequality comparisons to Double.NaN or Float.NaN. Equality comparison to these values is always false. Instead, use the Double.isNaN() of Float.isNaN() methods instead.
```

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+            MOTION_DIR_LEFT -> ViewActions.swipeLeft()
+            MOTION_DIR_RIGHT -> ViewActions.swipeRight()
+            MOTION_DIR_UP -> ViewActions.swipeUp()
+            MOTION_DIR_DOWN -> ViewActions.swipeDown()
+            else -> throw RuntimeException("Unsupported swipe direction: $direction")
+        }
+    }
+
+    fun swipeCustomInDirection(direction: Int, fast: Boolean, offset: Double, startPositionX: Double, startPositionY: Double): ViewAction? {
```
d4vidi, line 23: Please consider keeping the names consistent. Going from `amount` in the external API to `offset` had me confused. In a sense, `startPointX` vs. `startOffsetPercentX` is also not trivial...

docs/APIRef.ActionsOnElement.md
LeoNatan: No, it attempts making the element the first responder.

That's correctly worded.

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+        val to = if (offset > 0.0) {
+            min(1.0, from + offset)
+        } else {
+            max(0.0, from + offset)
+        }
+
+        return Pair(from, to);
+    }
+
+    private fun translate(point: Point): CoordinatesProvider {
```
d4vidi, line 92:
```suggestion
    private fun translate(point: Point) = CoordinatesProvider { view ->
        val (dx, dy) = point
        val xy = GeneralLocation.TOP_LEFT.calculateCoordinates(view)
        xy[0] += dx.toFloat() * view.width
        xy[1] += dy.toFloat() * view.height
        xy
    }

```

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+        if (isDescending) {
+            offset *= -1
+        }
+
+        val startPrimary = safeProportion(
+                if (isHorizontal) unsafeStartX else unsafeStartY,
+                0.5 - sign(offset) * EDGE_FUZZ_FACTOR
+        )
+
+        val startSecondary = safeProportion(
```
d4vidi, line 56: Perhaps rename to `startPrimaryAxis`, etc.? it got me confused, thinking there might be two competing start points, or an intermediate point along the axis 😬

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
LeoNatan: JS provides a `null`, because a JSON cannot hold `NaN` values. 🤦‍♂️ So the JSON parser needs to be aware of this, and put `NaN` where appropriate, if the JS didn't provide a value.

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwipeHelperSpec.kt
```diff
@@ -0,0 +1,271 @@
...
+        assertEquals(expected.second.first, actual.second.first, 0.01)
+        assertEquals(expected.second.second, actual.second.second, 0.01)
+    }
+
+    describe("SwipeHelper") {
+        it("should calculate left swipe with default values") {
+            var from = Pair(0.5 + EDGE_FUZZ_FACTOR, 0.5)
+            var to = Pair(0.0, 0.5)
+            var expected = Pair(from, to)
+            var actual = SwipeHelper.calculateSwipe(MOTION_DIR_LEFT)
```
d4vidi, line 23: var=>val

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        var offset = safeProportion(unsafeOffset, 0.5 + EDGE_FUZZ_FACTOR);
+        if (isDescending) {
+            offset *= -1
+        }
+
+        val startPrimary = safeProportion(
+                if (isHorizontal) unsafeStartX else unsafeStartY,
+                0.5 - sign(offset) * EDGE_FUZZ_FACTOR
```
d4vidi, line 53: I may not understand this properly, but I'm sort of uncomfortable with this usage of `EDGE_FUZZ_FACTOR`. As I understand it, the fuzz_factor was originally meant for calculating a safe touch area in order to avoid starting the swipe from the very edge. If the start point is somewhere in the center across an axis, I suppose there's no need in taking the edge into account... or am I missing something here?

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
d4vidi: maybe even swipeAxis and dormantAxis? 😅

d4vidi:
Well @noomorph I'm sort of done reviewing this, for now - seems there's a lot to go over for now, so let's reiterate this when the time is right

docs/APIRef.ActionsOnElement.md
d4vidi: source?

LeoNatan: The English language.

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
noomorph: @d4vidi, well, `fast=true` triggers Espresso's built-in functions, while `fast=false` triggers some of our implementations. So, I felt safer about replacing our implementation rather than ditching the use of Espresso's facade functions while having no idea what's under the hood there.

noomorph:
@d4vidi , @LeoNatan

Now the implementations feel aligned, right?

<table>
<tr>
<td width="50%">

</td>
<td width="50%">

</td>
</tr>
</table>

LeoNatan (approved):

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
noomorph: image

@d4vidi , it works the same in Java.

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
noomorph: Okay, so, we decided to align iOS and Android behavior first, stepping on compatibility with previous Android implementation.

noomorph: Done.

detox/android/detox/src/main/java/com/wix/detox/espresso/DetoxAction.java
```diff
@@ -168,67 +166,21 @@ public void perform(UiController uiController, View view) {
...
-                        GeneralLocation.TOP_CENTER, Press.FINGER));
-            case MOTION_DIR_DOWN:
-                return actionWithAssertions(new GeneralSwipeAction(Swipe.SLOW,
-                        translate(GeneralLocation.TOP_CENTER, 0, EDGE_FUZZ_FACTOR),
-                        GeneralLocation.BOTTOM_CENTER, Press.FINGER));
-            default:
-                throw new RuntimeException("Unsupported swipe direction: " + direction);
-        }
+    public static ViewAction swipeInDirection(final int direction, boolean fast, double amount, double startOffsetPercentX, double startOffsetPercentY) {
+        SwipeHelper swipeHelper = SwipeHelper.Companion.getDefault();
```
d4vidi, line 179:
```suggestion
        SwipeHelper swipeHelper = SwipeHelper.getDefault();
```

(see usage of `@JvmStatic` below)

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    private fun getScreenEdge(view: View, direction: Int) = when (direction) {
+        MOTION_DIR_LEFT, MOTION_DIR_UP -> 0
+        MOTION_DIR_RIGHT -> view.context.resources.displayMetrics.widthPixels
+        MOTION_DIR_DOWN -> view.context.resources.displayMetrics.heightPixels
+        else -> throw DetoxErrors.DetoxIllegalArgumentException("Unsupported swipe direction: $direction")
+    }
+
+    companion object {
+        val default = SwipeHelper { swiper: Swiper,
```
d4vidi, line 97:
```suggestion
        @JvmStatic val default = SwipeHelper { swiper: Swiper,
```

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwiperHelperSpec.kt
```diff
@@ -0,0 +1,131 @@
...
+object SwiperHelperSpec: Spek({
+    describe("SwipeHelper") {
+        val view = mock<View>()
+
+        var viewX = 1000
+        var viewY = 2000
+        var viewWidth = 2000
+        var viewHeight = 1000
+        var screenWidth = 4000
+        var screenHeight = 5000
```
d4vidi, line 25: `var`=>`val` 🙏🏻

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwiperHelperSpec.kt
```diff
@@ -0,0 +1,131 @@
...
+                                        _precisionDescriber: PrecisionDescriber ->
+            swiper = _swiper
+            startCoordinatesProvider = _startCoordinatesProvider
+            endCoordinatesProvider = _endCoordinatesProvider
+            precisionDescriber = _precisionDescriber
+            action = GeneralSwipeAction(
+                    _swiper,
+                    _startCoordinatesProvider,
+                    _endCoordinatesProvider,
+                    _precisionDescriber
```
d4vidi, line 81: You can simplify this since we don't need an actual action:
```
action = mock()
```

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwiperHelperSpec.kt
```diff
@@ -0,0 +1,131 @@
...
+        fun screenBottom() = screenTop() + screenHeight
+        fun screenRight() = screenLeft() + screenWidth
+
+        beforeGroup {
+            val mockDisplayMetrics = mock<DisplayMetrics>()
+            mockDisplayMetrics.widthPixels = screenWidth
+            mockDisplayMetrics.heightPixels = screenHeight
+
+            val mockResources = mock<Resources>()
+            `when`(mockResources.displayMetrics).then { mockDisplayMetrics }
```
d4vidi, line 47: Sorry for nitpicking, but please switch to `whenever` instead of \`when\`. When comes from mockito's core rather than the mockito-kotlin wrapper, and is not as idiomatic.

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+        endCoordinatesProvider: CoordinatesProvider,
+        precisionDescriber: PrecisionDescriber
+) -> ViewAction
+
+class SwipeHelper(private val createAction: CreateSwipeAction) {
+
+    fun swipeInDirection(
+            direction: Int,
+            fast: Boolean = true,
+            offset: Double = Double.NaN,
```
d4vidi, line 26: 'really wish this would be `amount` 😬

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+            startPositionX: Double = Double.NaN,
+            startPositionY: Double = Double.NaN
```
d4vidi, line 28: The naming here confuses me, still, because it's nevertheless not really a position, but a ratio. Also, best if it stays consistent with the caller's `startOffsetPercentX` and `startOffsetPercentY`.

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+    ): ViewAction {
+        val unsafeStartingPoint = AgnosticPoint2D.fromDoubles(startPositionX, startPositionY, direction)
+        val startingPoint = getSafeStartingPoint(unsafeStartingPoint, direction)
+        val start = translateFrom(startingPoint, direction)
+        val end = translateTo(startingPoint, direction, offset)
```
d4vidi, line 33: Naming here also had me a bit confused, because `translate` in the context of points really makes you think you're getting a translated point. Could something like `getFromCoordsProvider()` work?

detox/android/detox/src/main/java/com/wix/detox/espresso/utils/AgnosticPoint2D.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import com.wix.detox.common.DetoxErrors
+import com.wix.detox.espresso.common.annot.MOTION_DIR_DOWN
+import com.wix.detox.espresso.common.annot.MOTION_DIR_LEFT
+import com.wix.detox.espresso.common.annot.MOTION_DIR_RIGHT
+import com.wix.detox.espresso.common.annot.MOTION_DIR_UP
+
+data class FloatingPoint(val x: Float, val y: Float)
+
+data class AgnosticPoint2D(val primary: Double, val secondary: Double) {
```
d4vidi, line 11: Nice usage of data classes! Much better than using `Pair`s

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwiperHelperSpec.kt
```diff
@@ -0,0 +1,125 @@
...
+import androidx.test.espresso.ViewAction
+import androidx.test.espresso.action.*
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.whenever
+import com.wix.detox.espresso.common.annot.*
+import org.spekframework.spek2.Spek
+import org.spekframework.spek2.style.specification.describe
+import kotlin.test.assertEquals
+
+object SwiperHelperSpec: Spek({
```
d4vidi, line 16: Typo: `Swiper`=>`Swipe`

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+        val swiper = if (fast) Swipe.FAST else Swipe.SLOW
+
+        return this.createAction(swiper, start, end, Press.FINGER)
+    }
+
+    private fun getSafeStartingPoint(unsafeStartingPoint: AgnosticPoint2D, direction: Int): AgnosticPoint2D {
+        val isDescendingMove = direction == MOTION_DIR_UP || direction == MOTION_DIR_LEFT
```
d4vidi, line 40: Would you mind moving this to `MotionDefs` itself? Sounds like a common util like the existing ones (e.g. `isVertical`)

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+import androidx.test.espresso.action.*
+import com.wix.detox.common.DetoxErrors
+import com.wix.detox.espresso.common.annot.*
+import com.wix.detox.espresso.utils.AgnosticPoint2D
+import kotlin.math.min
+import kotlin.math.max
+
+private const val EDGE_FUZZ_FACTOR = 0.083f
+
+typealias CreateSwipeAction = (
```
d4vidi, line 14: Yes!

detox/android/detox/src/main/java/com/wix/detox/espresso/utils/AgnosticPoint2D.kt
```diff
@@ -0,0 +1,36 @@
+package com.wix.detox.espresso.utils
```
d4vidi, line 1: Nice usage of data classes! Much better than using `Pair`s

detox/android/detox/src/main/java/com/wix/detox/espresso/utils/AgnosticPoint2D.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.wix.detox.espresso.common.annot.MOTION_DIR_RIGHT
+import com.wix.detox.espresso.common.annot.MOTION_DIR_UP
+
+data class FloatingPoint(val x: Float, val y: Float)
+
+data class AgnosticPoint2D(val primary: Double, val secondary: Double) {
+    fun toFloatingPoint(direction: Int): FloatingPoint {
+        val primaryF = primary.toFloat()
+        val secondaryF = secondary.toFloat()
+        val isHorizontal = direction == MOTION_DIR_LEFT || direction == MOTION_DIR_RIGHT
```
d4vidi, line 15: Can use `MotionDefs.isHorizontal(direction)`

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,114 @@
...
+            0.0 + EDGE_FUZZ_FACTOR
+
+        return AgnosticPoint2D(
+                safeProportion(unsafeStartingPoint.primary, safeEdge),
+                safeProportion(unsafeStartingPoint.secondary, 0.5)
+        )
+    }
+
+    private fun safeProportion(value: Double, nanFallback: Double): Double {
+        return if (value.equals(Double.NaN)) {
```
d4vidi, line 53:
```suggestion
        return if (value.isNaN()) {
```

detox/android/detox/src/main/java/com/wix/detox/espresso/utils/AgnosticPoint2D.kt
d4vidi, line 11: So if I understand correctly, this really represents motion, or a _translation_, along 2 axes - one being the primary and the other, secondary. Since the translation is given in ratio rather than actual points, it is agnostic.
Wdyt about things like `AgnosticVector2D` for a class name, then? maybe `ProportionateVector2D`? And `primaryAxis`, `secondaryAxis` for params?

detox/android/detox/src/test/java/com/wix/detox/espresso/scroll/SwiperHelperSpec.kt
```diff
@@ -0,0 +1,125 @@
...
+        val screenHeight = 5000
+
+        fun viewLeft() = viewX + viewWidth * 0f;
+        fun viewCenter() = viewX + viewWidth * 0.5f;
+        fun viewRight() = viewX + viewWidth * 1f;
+        fun viewTop() = viewY + viewHeight * 0f;
+        fun viewMiddle() = viewY + viewHeight * 0.5f;
+        fun viewBottom() = viewY + viewHeight * 1f;
+        fun viewFuzzH() = viewWidth * SwipeHelper.edgeFuzzFactor;
+        fun viewFuzzV() = viewHeight * SwipeHelper.edgeFuzzFactor;
```
d4vidi, line 34: no semicolons please 😄

d4vidi:
@noomorph amazing work overall! I hope you don't mind me nitpicking here and there 😂  Anyways it's mostly naming that bothered me in my way of understanding the code, so most comments are in that area. Other than that seem pretty good!

d4vidi (changes requested):
@noomorph amazing work overall! I hope you don't mind me nitpicking here and there 😂 Anyways it's mostly naming that bothered me in my way of understanding the code, so most comments are in that area. Other than that seem pretty good!

detox/android/detox/src/main/java/com/wix/detox/espresso/utils/AgnosticPoint2D.kt
noomorph: Yes. :) Thanks.

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
noomorph: Sure, I'll align the naming to iOS tomorrow.

detox/android/detox/src/main/java/com/wix/detox/espresso/utils/AgnosticPoint2D.kt
noomorph: Not relevant anymore — I've rewritten the implementation to use vectors.

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
noomorph: normalizedStartOffsetX, normalizedStartOffsetY probably

detox/android/detox/src/main/java/com/wix/detox/espresso/scroll/SwipeHelper.kt
```diff
@@ -0,0 +1,100 @@
...
+        xy[1] += (normalizedStartPoint.y * view.height).toFloat()
+        xy
+    }
+
+    private fun buildEndCoordinatesProvider(startCoordinatesProvider: CoordinatesProvider, direction: Int, normalizedSwipeAmount: Double) =
+            CoordinatesProvider { view ->
+                val xy = startCoordinatesProvider.calculateCoordinates(view)
+
+                val screenEdge = Vector2D.from(
+                        view.context.resources.displayMetrics.widthPixels,
```
d4vidi, line 63:
```suggestion
                        view.resources.displayMetrics.widthPixels,
```

d4vidi (approved):
Nice work! It's great to see how eventually you've used the generic algebric constructs to calculate both the absolute and normalized starting point, and better yet - the ending point.

PR #2434, opened 2020-10-27
Expand retry() API and improve exec() retry-logging

d4vidi:
- [x] This is a small change 
- [ ] This change has been discussed in issue #<?> and the solution has been agreed upon with maintainers.

---

**Description:**
Associated with the ongoing work on Genymotion-Cloud integration (#2429), where logging of command failures, and better-retry control, are required.

In particular, in the case where a command is retried due to repeated failures, instead of having an obscure log - simply showing the command and an overdue failure, as is today (in trace level):
```
detox[18311] DEBUG: [exec.js/EXEC_CMD, #2] "gmsaas" --format compactjson instances start --stop-when-inactive 5974e92c-5ec6-4576-a93e-2ceec4338f0b "Detox-1603890039526.4004"
(log stuck for ~1.5m... :-/)
detox[18311] ERROR: [exec.js/EXEC_FAIL, #2] ""gmsaas" --format compactjson instances start --stop-when-inactive 5974e92c-5ec6-4576-a93e-2ceec4338f0b "Detox-1603890039526.4004"" failed with code = 4, stdout and stderr:
detox[18311] ERROR: [exec.js/EXEC_FAIL, #2]
detox[18311] ERROR: [exec.js/EXEC_FAIL, #2] {"error": {"message": "API return unexpected code: 403. Error: {\"code\":\"TOO_MANY_RUNNING_VDS\",\"message\":\"Too many running virtual devices\"}", "details": ""}, "exit_code": 4, "exit_code_desc": "API_ERROR"}
```

With a proper argument to `exec` (i.e. `statusLog.retrying: true`), all retries and errors in between would be debug/trace-logged: (*)

```
detox[18311] DEBUG: [exec.js/EXEC_CMD, #2] "gmsaas" --format compactjson instances start --stop-when-inactive 5974e92c-5ec6-4576-a93e-2ceec4338f0b "Detox-1603890039526.4004"
detox[18311] TRACE: [exec.js/EXEC_TRY_FAIL, #2] {"error": {"message": "API return unexpected code: 403. Error: {\"code\":\"TOO_MANY_RUNNING_VDS\",\"message\":\"Too many running virtual devices\"}", "details": ""}, "exit_code": 4, "exit_code_desc": "API_ERROR"}

[...]

PR #2437, opened 2020-10-28
Merge execWithRetriesAndLogs's statusLog arg into options

d4vidi:
**Description:**
A small follow-up on #2434

### DIFF
--- a/detox/src/devices/drivers/android/exec/ADB.js
+++ b/detox/src/devices/drivers/android/exec/ADB.js
@@ -271,13 +271,16 @@ class ADB {
   }
 
   // TODO refactor the whole thing so as to make usage of BinaryExec -- similar to EmulatorExec
-  async adbCmd(deviceId, params, options) {
+  async adbCmd(deviceId, params, options = {}) {
     const serial = `${deviceId ? `-s ${deviceId}` : ''}`;
     const cmd = `"${this.adbBin}" ${serial} ${params}`;
-    const retries = _.get(options, 'retries', 1);
-    _.unset(options, 'retries');
+    const _options = {
+      ...options,
+      retries: options.retries || 1,
+    }
 
-    return execWithRetriesAndLogs(cmd, options, retries);
+    _.unset(options, 'retries');
+    return execWithRetriesAndLogs(cmd, _options);
   }
 
   /***
--- a/detox/src/devices/drivers/android/exec/ADB.js
+++ b/detox/src/devices/drivers/android/exec/ADB.js
@@ -277,7 +277,7 @@ class ADB {
     const retries = _.get(options, 'retries', 1);
     _.unset(options, 'retries');
 
-    return execWithRetriesAndLogs(cmd, options, undefined, retries);
+    return execWithRetriesAndLogs(cmd, options, retries);
   }
 
   /***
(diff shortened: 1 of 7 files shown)


detox/src/devices/drivers/android/exec/ADB.js
```diff
@@ -277,7 +277,7 @@ class ADB {
     const retries = _.get(options, 'retries', 1);
     _.unset(options, 'retries');
 
-    return execWithRetriesAndLogs(cmd, options, undefined, retries);
+    return execWithRetriesAndLogs(cmd, options, retries);
```
noomorph, line 280: You know, since the function is called "exec with **retries** and **logs**", and you have put the "logs" inside the options, maybe just put "retries" into the options as well, so we don't have to return to it anymore.

d4vidi: I was considering that, myself, and couldn't clearly decide. Perhaps you're right, although there was something about the explicitness of the retires/interval args. wdyt?

d4vidi: ok, done. tell me what you think!

noomorph: I bet it took some sweat, but it looks totally worth it. Kudos! Great job 👍👍

noomorph:
This change involves a lot of signatures changes. I'll make sure to review it as well tomorrow.

detox/src/devices/drivers/android/exec/ADB.js
```diff
@@ -271,13 +271,16 @@ class ADB {
...
   // TODO refactor the whole thing so as to make usage of BinaryExec -- similar to EmulatorExec
-  async adbCmd(deviceId, params, options) {
+  async adbCmd(deviceId, params, options = {}) {
     const serial = `${deviceId ? `-s ${deviceId}` : ''}`;
     const cmd = `"${this.adbBin}" ${serial} ${params}`;
-    const retries = _.get(options, 'retries', 1);
-    _.unset(options, 'retries');
+    const _options = {
+      ...options,
+      retries: options.retries || 1,
```
noomorph, line 279: I'm afraid that `|| 1` is not an equivalent transformation. Just so you are aware that you no longer will be able to nullify retry count, i.e. `{ retries: 0} -> { retries: 1 }`.

Nevertheless, this function has never been called with zero, so we are in a safe position.

detox/src/devices/drivers/android/exec/ADB.js
```diff
@@ -271,13 +271,16 @@ class ADB {
...
+    }
 
-    return execWithRetriesAndLogs(cmd, options, undefined, retries);
+    _.unset(options, 'retries');
```
noomorph, line 282: I don't think you longer need this `_.unset`.
AFAIR, I added it to make sure I don't pass a dirty object to `exec...` which had 5 scalar parameters at that time.

Let's remove that `_.unset` and inline that `_options` object:

```js
return execWithRetriesAndLogs(cmd, {
  retries: 1, // default value
  ...options
};
```

detox/src/devices/drivers/android/exec/ADB.js
d4vidi: What an excellent find. I will fix this + add a unit test!

detox/src/devices/drivers/android/exec/ADB.js
d4vidi: 🚀

PR #2447, opened 2020-11-02
Allow releasing prereleases from different branches

rotemmiz:
**Description:**

Loosen some of the constraints we currently have on releasing versions (up until now only from master), and allowing more flexible release process of all types of prereleases (patch, minor, major) from branches.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -13,7 +13,7 @@
     "generation",
     "."
   ],
-  "version": "17.10.6",
+  "version": "17.10.7-alpha.0",
   "npmClient": "npm",
   "command": {
     "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -31,8 +31,7 @@
     },
     "bootstrap": {
       "npmClientArgs": [
-        "--no-package-lock",
-        "--no-ci"
+        "--no-package-lock"
       ]
     }
   }
--- a/lerna.json
+++ b/lerna.json
@@ -31,7 +31,8 @@
     },
     "bootstrap": {
       "npmClientArgs": [
-        "--no-package-lock"
+        "--no-package-lock",
+        "--no-ci"
       ]
     }
   }
--- a/lerna.json
+++ b/lerna.json
@@ -13,7 +13,7 @@
     "generation",
     "."
   ],
-  "version": "17.10.7-alpha.0",
+  "version": "17.10.6",
   "npmClient": "npm",
   "command": {
     "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.4.0",
+  "lerna": "3.22.1",
   "packages": [
     "detox",
     "detox-cli",
--- a/lerna.json
+++ b/lerna.json
@@ -13,7 +13,7 @@
     "generation",
     "."
   ],
-  "version": "17.10.7-alpha.0",
+  "version": "17.10.7",
   "npmClient": "npm",
   "command": {
     "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.11.0",
+  "lerna": "3.22.1",
   "packages": [
     "detox",
     "detox-cli",
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "3.22.1",
+  "lerna": "2.11.0",
   "packages": [
     "detox",
     "detox-cli",
--- a/scripts/ci.publish.js
+++ b/scripts/ci.publish.js
@@ -49,7 +49,7 @@ function publishToNpm(npmTag) {
     log('SKIP NPM is set: Lerna-publishing without publishing to NPM');
   }
 
-  exec.execSync(`lerna publish --cd-version "${versionType}" --yes --npm-tag ${npmTag} ${npmTag === 'latest'? '': '--preid=alpha'} --skip-git ${(dryRun || skipNpm) ? '--skip-npm' : ''}`);
+  exec.execSync(`lerna publish --cd-version "${versionType}" --yes --dist-tag ${npmTag} ${npmTag === 'latest'? '': '--preid=alpha'} --skip-git ${(dryRun || skipNpm) ? '--skip-npm' : ''}`);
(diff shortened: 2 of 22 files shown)


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "lerna": "3.22.1",
+  "lerna": "2.11.0",
```
LeoNatan, line 2: Uffff

rotemmiz: It actually works with 3.22.1, but we first need to update CI machines

scripts/ci.release.js
```diff
@@ -29,10 +29,10 @@ function validateEnv() {
...
 	}
 
-	if (process.env.GIT_BRANCH !== ONLY_ON_BRANCH) {
-		log(`Release blocked: Not publishing on branch ${process.env.GIT_BRANCH}, which isn't ${ONLY_ON_BRANCH}`);
-		return false;
-	}
+	// if (process.env.GIT_BRANCH !== ONLY_ON_BRANCH) {
+	// 	log(`Release blocked: Not publishing on branch ${process.env.GIT_BRANCH}, which isn't ${ONLY_ON_BRANCH}`);
+	// 	return false;
+	// }
```
d4vidi, line 35: delete!

scripts/ci.release.js
```diff
@@ -5,11 +5,11 @@ const {log, logSection, getVersionSafe} = require('./ci.common');
 
 const isRelease = (process.env.RELEASE_VERSION_TYPE && process.env.RELEASE_VERSION_TYPE !== 'none');
 
-const ONLY_ON_BRANCH = 'origin/master';
+// const ONLY_ON_BRANCH = 'origin/master';
```
d4vidi, line 8: delete me!

scripts/ci.release.js
```diff
@@ -29,10 +29,10 @@ function validateEnv() {
 		return false;
 	}
```
d4vidi, line 30: What is this? Doesn't seem right... Can you also get that fixed? It's either the statement is wrong, or the log message is.

scripts/ci.publish.js
```diff
@@ -24,19 +24,19 @@ function validatePrerequisites() {
...
   }
 
-  const lernaVersion = exec.execSyncRead('lerna --version');
-  if (!lernaVersion.startsWith('2.')) {
-    throw new Error(`Cannot publish: lerna version isn't 2.x.x (actual version is ${lernaVersion})`);
-  }
+  // const lernaVersion = exec.execSyncRead('lerna --version');
+  // if (!lernaVersion.startsWith('2.')) {
+  //   throw new Error(`Cannot publish: lerna version isn't 2.x.x (actual version is ${lernaVersion})`);
+  // }
```
d4vidi, line 30: delete me! ❤️

d4vidi: On 2nd thought, why not keep this validation as-is?

noomorph: I just have expressed my confusion about why can't we move to Lerna 3, by using `npx lerna ...` local CLI.

scripts/ci.release.js
```diff
@@ -79,8 +79,18 @@ function versionTagAndPublish() {
 	log(`Great success, much amaze`);
 }
 
+function releaseNpmTag() {
+	if (process.env.GIT_BRANCH === 'master') {
+		return 'latest';
+	} else if (process.env.RELEASE_NPM_TAG) {
+		return process.env.RELEASE_NPM_TAG;
```
d4vidi, line 86: So this means we cannot release a tag from master... Do we not want to keep this option as an escape hatch? meaning, prioritize any value set for `RELEASE_NPM_TAG` over anything else (including the current branch)?

scripts/ci.release.js
```diff
@@ -79,8 +79,18 @@ function versionTagAndPublish() {
...
+	} else {
+		return process.env.GIT_BRANCH;
```
d4vidi, line 88: That's a good idea for a fallback, but might be a bit implicit. wdyt?

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.4.0",
+  "lerna": "3.22.1",
```
d4vidi, line 2: !!!!!!!

d4vidi: There's no `--cd-version` in 3.x so we need to rephrase the command execution for this to work, really

scripts/ci.publish.js
```diff
@@ -49,7 +49,7 @@ function publishToNpm() {
     log('SKIP NPM is set: Lerna-publishing without publishing to NPM');
   }
 
-  exec.execSync(`lerna publish --cd-version "${versionType}" --yes --skip-git ${(dryRun || skipNpm) ? '--skip-npm' : ''}`);
+  exec.execSync(`lerna publish --cd-version "${versionType}" --yes --dist-tag ${npmTag} ${npmTag === 'latest'? '': '--preid=alpha'} --skip-git ${(dryRun || skipNpm) ? '--skip-npm' : ''}`);
```
d4vidi, line 52: Are we not support to use `versionType=prerelease`?

scripts/ci.publish.js
rotemmiz: A bit more complicated than that, but looks like its working now

scripts/ci.publish.js
rotemmiz: prerelease == prepatch == 10.16.0 -> 10.16.1-alpha.0
preminor == 10.16.0 -> 10.17.0-alpha.0
premajor == 10.16.0 -> 11.0.0-alpha.0

npm tags are set to branch name, unless stated otherwise in build parameters

d4vidi: OK thanks! also 'noticed you updated this in the master-build's configuration on Jenkins 👍🏻 
What about the `alpha` part, which is static - wouldn't it make a fair use case that we would like to release `10.16.0-alpha.0` from branch `bigFeature1` and at the same time, `10.16.0-next.0` from branch `detoxNext`?

rotemmiz: was actually thinking the same thing :) We can improve on what we have of course

rotemmiz: I'll set the npm-tag as the prerelease tag

PR #2455, opened 2020-11-07
test(android): fix unit tests

noomorph:
- [x] This is a small change 

---

**Description:**

Trying to fix errors like these, when I run `./gradlew build` from `detox/android` folder:

image

Example of an error:

image

### DIFF
--- a/detox/test/android/gradle/wrapper/gradle-wrapper.properties
+++ b/detox/test/android/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-all.zip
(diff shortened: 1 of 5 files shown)


detox/test/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2.2-all.zip
```
noomorph, line 6: The project does not work locally on my machine unless I bump the Gradle version. 🤷 

Otherwise, I just keep seeing this:
https://stackoverflow.com/questions/61289461/java-lang-noclassdeffounderror-could-not-initialize-class-org-codehaus-groovy-v

noomorph:
Okay, closing it.



I've forgotten that Android is still in the ice age. 🤯 
Moved to JDK 8, got my issues fixed.

detox/test/android/gradle/wrapper/gradle-wrapper.properties
rotemmiz: You can upgrade to gradle 6.5+ but need to upgrade the android gradle plugin to 4.1.0+
https://developer.android.com/studio/releases/gradle-plugin

PR #2446, opened 2020-11-02
Introduce the fundamentals of Genymotion-cloud support

d4vidi:
- [ ] This is a small change 
- [ ] This change has been discussed in issue #2429 and the solution has been agreed upon with maintainers.

---

**Description:**
Introduce the fundamental support for Genymotion cloud emulators, enabling e2e execution over cloud device instances.

**There are still many loose ends to close here, and merging this does not in any way mean Genymotion-cloud is fully supported.**
To name a few:

- *Detox-based device instances teardown (a big todo).*
- An optimization of instance creation is required (i.e. usage of the `--no-wait` flag), with boot-up polling.
- Better tweaking of Detox' CI script, so as to support exec on both local and cloud emulators in each build.
- Fix broken `ci-unhappy` e2e flavor

**How to review these changes**
Start from the new, top-level `GenyCloudDriver` class, and drill down through its self-generated delegation -- mainly the 3 `service`-layer classes that allow for recipe querying, instance creation, deletion and lookup. In turn, they have delegations of their own (injected), such as a wrapper over the `gmsaas` executable, and an instance naming concern provider. Follow through to drill further down the chain to review these classes as well, last.

Be advised that per each such class, there's a unit test suite that's been written TDD with full coverage over the equivalent production code.

Issue #2429, opened 2020-10-21
Add support for Genycloud emulators for grand PoC


d4vidi:
**Is your feature request related to a problem? Please describe.**
For the sake of running internal-PoC's at Wix, we need to enable support for [Genycloud](https://cloud.geny.io/)-based emulators. This mainly suggests a dedicated driver should be written to support that.

The work for a Detox-PoC has already been done by @EdenGottlieb in branch `genymotion-support-2`. Now need to pick it up from where it was left off, in order to achieve complete functionality.

---
Task list in order to be ready for a contract signing and a PoC at scale:
- [x] Initial naive support so as to start getting numbers running on OSS CI. [#2446]
- [x] An optimization of instance creation (i.e. usage of the --no-wait flag), with boot-up polling. [#2466]
- [x] Inherent device instances tear-down on global tear-down context. [#2471, #2529, wix-private/shono/pull/106]
- [x] Improve instance leaks warning-logs by maintaining instance names in cleanup registry. [#2476]
- [x] Instance leakage warning in case of abrupt termination (research, first - maybe with `signal-exit`) [#2555]
- [x] Detox user-agent
- [x] Reimplement `LookupService.getInstance()` using new gmsaas `1.6.0` command: `gmsaas instance get <uuid>` (also, need to add a prevalidation of that version is >= 1.6.0) [#2546]

PR #2459, opened 2020-11-10
Genycloud optimize and teardown

d4vidi:
**Description:**
Work on #2429 continues here. This is the 2nd iteration, following #2446.
Still a WIP with many things to be done, yet already good enough for evaluating on CI.


detox/src/utils/ExclusiveLockfile.js
```diff
@@ -32,8 +32,11 @@ class ExclusiveLockfile {
    * @returns {Promise<any>}
    */
   async exclusively(fn) {
-    await this._lock();
+    if (this._isLocked) {
+      return await fn();
```
noomorph, line 36: ⚠️  No, wait, what are you doing?? If it is locked, you should stop at line `await this._lock()` ⚠️ That's the idea.

@d4vidi ??

detox/src/devices/DeviceRegistry.js
```diff
@@ -79,6 +89,12 @@ class DeviceRegistry {
       lockfilePath: environment.getDeviceLockFilePathAndroid(),
     });
   }
+
+  static forGenyCloudCleanup() {
```
noomorph, line 93: Wait, can't we stop multiplying the entities, and think, maybe, about a unified device registry?

detox/src/utils/ExclusiveLockfile.js
d4vidi: Wow thanks for going over this code so soon 😅 ❤️  
I think the terms here are misleading - the logic is a bit reversed: `_isLocked=true` means you've already "obtained the semaphore" and that you can do whatever you like. Only otherwise, you are forced to go into `this._lock()` and wait for your turn, inside.

d4vidi: Your instinct is correct, but do note that there's a unit test that validates this behavior (i.e. can't execute actions if not locked): https://github.com/wix/Detox/blob/611ae1aaa1f964407a10237554197c7fd95da6d8/detox/src/utils/ExclusiveLockfile.test.js#L50

detox/src/devices/DeviceRegistry.js
d4vidi: I think we should definitely unify the platform based separation (even added a TODO there 😄 )
But as for cleanup, for now we should keep it separated...

detox/src/utils/ExclusiveLockfile.js
noomorph: @d4vidi, no, there's something really odd here. First of all, why did you have to change that place at all? What happened that you had to change something in that class?

My initial idea was that there can be *no condition* in our source code, when two calls of `.exclusively(callback)` are being executed. Every call when it ends the execution, unlocks the file, so the next one can start.

Now, with your change, we can start having race conditions inside our source code, with the lock file. I am not convinced that this change of yours is safe. Could you elaborate more, please?

d4vidi: Oh yes I see now - you don't want potential simultaneous callers to step on each other's toes, eventually polluting the value of the shared resource.

Well, I have another solution, which isn't as nice, but will do nonetheless. Let's discuss 1x1.

noomorph: @d4vidi, well, I did not want it to happen just **accidentally**. Let's discuss it indeed.

d4vidi:
Closing in favor of smaller PR's, which I will open soon

PR #2460, opened 2020-11-10
Update doc: device.setLocation() is iOS only

clemblanco:


noomorph:
Thanks!

LeoNatan:
The link is now broken due to title change @clemblanco @noomorph

LeoNatan:
Fixed, thanks

PR #2466, opened 2020-11-15
Introduce optimization of genycloud instance allocation

d4vidi:
## Description
The 2nd piece of the Genymotion-cloud puzzle (see #2429).
Introduces a more efficient way of allocating cloud instances, as per the description below.

For each initializing driver -- associated with an initializing worker, in order to grab a hold over a cloud instance, one has to:
```
a. Check for a free, already-running instance and acquire it, OR otherwise:
b. Allocate a new instance, namely:
b.1. spin-up a new instance synchronously
b.2. adb-connect to it synchronously
```

Currently, all of `b` is done inside the _worker-synchronized_, locked scope of `DeviceRegistry.allocateDevice()`. That means that the allocation - though performed by workers running concurrently, is actually performed serially; Each worker thus holds back all other workers potentially doing the same job, for as long as the entire allocation process (which can take up to a whole minute!).

This PR therefore replaces `b` with the following adjustment:
```
b.1 acquire lock + spin-up a new instance *asynchronously* + unlock
b.2. wait for instance to be online (polling)
b.3. adb-connect to it synchronously
```

---
While the latter description concludes the general idea of the PR, it does not account for its side-effects.

The biggest side-effect is that unlike existing drivers, the Genymotion cloud one has to rely on the geny-cloud's instance UUID's for device-registry management. Existing drivers use the ADB-name. Namely, instead of having _the_ lockfile looking something like this:
```
['emulator-5554', 'emulator-5556']
```
For genymotion-cloud instances, it would looks like this:
```
['db5c1a74-461c-491c-bdd6-f70338a8f9cd', '77c584f2-5996-4bc6-85cc-e92d7ea3ac97']
```
> Remember that right after instance creation, ADB is not connected yet and hence, there _is_ no ADB name yet...

While this in its own is not a big change (replacing one type of string for another), it does not take into account the currently implicit duality in using ADB-name for a device ID:
1. It is used as the unique ID for device-registry management, as explained.
2. It is also the device name consequently used by the driver when accessing the device via ADB, throughout the session, and the usage is trivial (`deviceId`<=>`adbName`).

So while `#1` is easy to replace (i.e. use instance UUID's instead of ADB-names), it's change invalidates `#2`. Hence a side-effect where some translations have to take place. This boils down to these areas:
1. In the Genycloud layer of the driver (i.e. `GenyCloudDriver` and delegates), the allocation function returns the instance UUID for the device-registry, but ultimately returns back the whole `instance` for a `deviceId` (`uuid` and `adbName` properties included...).
2. In the `AndroidDriver` itself, the interchangeable usage of `deviceId` as an `adbName` is no longer straightforward. Rather, inside every relevant method, the `deviceId` is first explicitly translated to the `adbName`, if needed: The driver now allows for `deviceId` to either be a primitive or an object, but in the latter case - is has to hold an `adbName` property.
3. In the new `GenyCloudDriver`, the same is done whenever `adb` is used (but in a simplified way - `deviceId.adbName`).
4. ~In every access to the **device-registry** in the Genycloud realm - even from the base `AndroidDriver`, it needs to be translated from a device ID (i.e. the whole instance) to its uuid, first. This is done agnostically using a mediation layer in the form of a `DeviceRegistry` wrapper, called `GenyDeviceRegistryWrapper`, which the driver replaces the raw device-registry with.~ I've decided to kill this concept of an adaptor and instead move the responsibility of managing the device disposal (from the device-registry) to the unit the allocates it - namely, remove any sign of it from `AndroidDriver` and switch it to the concrete classes (commit dea182431a8da4412081b999d944b2a491591eb5). That makes these translation more explicit and easy to maintain.


detox/package.json
```diff
@@ -96,11 +96,19 @@
...
       "src/server/DetoxServer.js",
-      ".*Driver.js",
+      "DeviceDriverBase.js",
+      "AndroidDriver.js",
+      "AttachedAndroidDriver.js",
+      "EmulatorDriver.js",
+      "IosDriver.js",
+      "SimulatorDriver.js",
+      "AndroidDeviceAllocator.js",
+      "AttachedDeviceAllocator.js",
```
d4vidi, line 106: Genycloud driver is a driver that is _not_ ignored

detox/src/devices/DeviceRegistry.js
```diff
@@ -44,10 +44,14 @@ class DeviceRegistry {
     });
   }
 
-  isDeviceBusy(deviceId) {
+  includes(deviceId) {
```
d4vidi, line 47: usage of 'busy' here was wrong to begin with. this is a registry, it either includes or doesn't include a device.

detox/src/devices/DeviceRegistry.js
```diff
@@ -44,10 +44,14 @@ class DeviceRegistry {
...
     return this._lockfile.read().includes(deviceId);
   }
 
+  getRegisteredDevices() {
```
d4vidi, line 51: An optimization in case `include` needs to be checked repeatedly.

detox/src/devices/drivers/android/AndroidDriver.js
```diff
@@ -64,52 +63,42 @@ class AndroidDriver extends DeviceDriverBase {
...
-  }
-
-  /**
-   * @protected
-   * @return {Promise<string>} adbName of a free matching device
-   */
-  async doAllocateDevice(deviceQuery) {
-    throw Error('Not implemented!');
-  }
-
```
d4vidi, line 81: Putting this here was a bad idea to begin with. DeviceRegistry is not easy to used generically ATM.
Instead, a more permissive version of this code is now in `AndroidDeviceAllocator`.

detox/src/devices/drivers/android/AndroidDriver.js
```diff
@@ -289,27 +282,31 @@ class AndroidDriver extends DeviceDriverBase {
...
+    const pid = await this.adb.pidof(adbName, bundleId);
     if (!pid) {
       throw new Error('PID still not available');
     }
     return pid;
   }
+
+  _getAdbName(deviceId) {
+    return _.isObjectLike(deviceId) ? deviceId.adbName : deviceId;
+  }
```
d4vidi, line 309: As explained in the description, it is no longer safe to assume that `deviceId` and `adbName` are interchangeable.

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -61,11 +61,6 @@ class EmulatorDriver extends AndroidDriver {
     return adbName;
   }
 
-  async doAllocateDevice(deviceQuery) {
-    const freeEmulatorAdbName = await this._freeDeviceFinder.findFreeDevice(deviceQuery);
-    return freeEmulatorAdbName || this._createDevice();
-  }
```
d4vidi, line 67: Simplified and moved to `EmulatorDeviceAllocator`

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -147,16 +139,6 @@ class EmulatorDriver extends AndroidDriver {
...
   }
-
-  _createDevice() {
-    const {min, max} = DetoxEmulatorsPortRange;
-    let port = Math.random() * (max - min) + min;
-    port = port & 0xFFFFFFFE; // Should always be even
-
-    const adbName = `emulator-${port}`;
-    this.pendingBoots[adbName] = port;
-    return adbName;
```
d4vidi, line 158: Relocated to `EmulatorDeviceAllocator`

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -33,9 +29,10 @@ class EmulatorDriver extends AndroidDriver {
 
     const avdsResolver = new AVDsResolver(emulatorExec);
     this._avdValidator = new AVDValidator(avdsResolver, this._emuVersionResolver);
-    this._freeDeviceFinder = new FreeEmulatorFinder(this.adb, this.deviceRegistry);
 
-    this.pendingBoots = {};
```
d4vidi, line 38: Simplified: port to use for boot-up is returned from `EmulatorDeviceAllocator`

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -82,13 +77,10 @@ class EmulatorDriver extends AndroidDriver {
...
 
-  async _boot(avdName, adbName) {
-    const coldBoot = !!this.pendingBoots[adbName];
-
+  async _boot(avdName, adbName, bootPort) {
+    const coldBoot = !!bootPort;
     if (coldBoot) {
-      const port = this.pendingBoots[adbName];
-      await this._emuLauncher.launch(avdName, { port });
-      delete this.pendingBoots[adbName];
```
d4vidi, line 91: Simplified yet again

detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
```diff
@@ -0,0 +1,72 @@
...
+
+    this.instanceLookupService = instanceLookupService;
+    this.instanceLifecycleService = instanceLifecycleService;
+  }
+
+  async _doAllocateDevice(recipe) {
+    let { instance, isNew } = await this._doSynchronizedAllocation(recipe);
+
+    instance = await this._waitForInstanceBoot(instance);
+    instance = await this._adbConnectIfNeeded(instance);
```
d4vidi, line 17: Arguably, this may or may not be here. But that can be updated later on.

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -17,43 +17,43 @@ class GenyCloudDriver extends AndroidDriver {
 
     const exec = new GenyCloudExec();
     const instanceNaming = new InstanceNaming(); // TODO should consider a permissive impl for debug/dev mode. Maybe even a custom arg in package.json (Detox > ... > genycloud > sharedAccount: false)
-    this.recipeService = new RecipesService(exec, logger);
-    this.instanceLookupService = new InstanceLookupService(exec, instanceNaming, this.deviceRegistry);
-    this.instanceLifecycleService = new InstanceLifecycleService(exec, instanceNaming);
-    this.deviceQueryHelper = new DeviceQueryHelper(this.recipeService);
-    this.allocationHelper = new AllocationHelper(this.instanceLookupService, this.instanceLifecycleService);
+    this.deviceRegistry = new GenyDeviceRegistryWrapper(this.deviceRegistry);
```
d4vidi, line 20: The wrapper is installed as an agnostic adaptor, performing translations whenever the device-registry is used in common code.

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -17,43 +17,43 @@ class GenyCloudDriver extends AndroidDriver {
...
-    this._name = `GenyCloud-${adbName}`;
-    return adbName;
+    this._name = `GenyCloud:${instance.name} (${uuid} ${adbName})`;
+    return instance;
   }
 
-  async doAllocateDevice({ recipe, cookie }) {
-    const { instance, coldBooted } = await this.allocationHelper.allocateInstance(recipe.uuid);
-    cookie.coldBoot = coldBooted;
-    return instance.adbName;
```
d4vidi, line 53: No self implementation, no allocation helper. Everything is now in `GenyCloudDeviceAllocator`.

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -17,43 +17,43 @@ class GenyCloudDriver extends AndroidDriver {
...
+  async installApp({ adbName }, _binaryPath, _testBinaryPath) {
+    const {
+      binaryPath,
+      testBinaryPath,
+    } = this._getInstallPaths(_binaryPath, _testBinaryPath);
+    await this.appInstallHelper.install(adbName, binaryPath, testBinaryPath);
```
d4vidi, line 53: Not 100% related to this PR, but very important nonetheless. This has been empirically proved to greatly improve installation time (compared to the usage of `adb install` in base-class impl).

detox/src/devices/drivers/android/genycloud/exec/GenyCloudExec.js
```diff
@@ -7,12 +6,11 @@ class GenyCloudExec {
...
 
   getInstances() {
-    return this._exec(`instances list`);
+    return this._exec(`instances list -q`);
   }
 
   startInstance(recipeUUID, instanceName) {
-    // TODO should we use --no-wait so as to release workers lock "mutex" asap and then wait-for-boot as we do with google emulators?
-    return this._exec(`instances start --stop-when-inactive ${recipeUUID} "${instanceName}"`);
+    return this._exec(`instances start --stop-when-inactive --no-wait ${recipeUUID} "${instanceName}"`);
```
d4vidi, line 13: The core of everything. `--no-wait` means the request is responded immediately, without waiting for the device to become online.

detox/src/devices/drivers/android/genycloud/services/GenyInstanceLookupService.js
```diff
@@ -1,40 +1,39 @@
...
   }
 
-  async _getRelevantInstances(recipeUUID) {
+  async getInstance(instanceUUID) {
+    const instances = await this._getAllInstances();
+    return _.find(instances, (instance) => instance.uuid === instanceUUID);
+  }
+
+  async _getRelevantInstances() {
+    const takenDevices = this.deviceRegistry.getRegisteredInstanceUUIDs();
```
d4vidi, line 22: More efficient to read the content once, and reuse (instead of calling `isBusy` repeatedly).

detox/src/devices/drivers/android/genycloud/services/GenyInstanceLookupService.js
```diff
@@ -1,40 +1,39 @@
...
     const isRelevant = (instance) =>
-      instance.recipeUUID === recipeUUID &&
-      !instance.isTerminated() &&
+      (instance.isOnline() || instance.isInitializing()) &&
```
d4vidi, line 24: isInitializing is a fix for a rare case where an instance might be selected while being deleted

d4vidi:
I have an idea of how to possibly simplify the code. I'm switching back to draft until this is done. Probably tomorrow.

detox/src/devices/DeviceRegistry.js
noomorph: Although, it makes me wonder if we need at all the `includes` method after this change...? 🤔

detox/src/devices/DeviceRegistry.js
noomorph: Okay, but semantically — what "inclusion" means? It means that a worker can no longer take this device, it is used somewhere else.

detox/src/devices/drivers/android/AndroidDriver.js
noomorph: We no longer can call it `deviceId`, after this transformation. I suggest renaming it to `deviceHandle`.

detox/src/devices/drivers/android/emulator/EmulatorDeviceAllocator.js
```diff
@@ -0,0 +1,44 @@
...
+    await this.deviceRegistry.allocateDevice(async () => {
+      adbName = await this._freeDeviceFinder.findFreeDevice(deviceQuery);
+      if (!adbName) {
+        placeholderPort = this._allocateEmulatorPlaceholderPort();
+        adbName = `emulator-${placeholderPort}`;
+      }
+      return adbName;
+    });
+
+    return {
```
noomorph, line 29: Maybe that could use some class, like `EmulatorHandle extends DeviceHandle`.

detox/src/devices/drivers/android/emulator/EmulatorDeviceAllocator.js
```diff
@@ -0,0 +1,44 @@
...
+    let placeholderPort = null;
+    let adbName = null;
+
+    await this.deviceRegistry.allocateDevice(async () => {
+      adbName = await this._freeDeviceFinder.findFreeDevice(deviceQuery);
+      if (!adbName) {
+        placeholderPort = this._allocateEmulatorPlaceholderPort();
+        adbName = `emulator-${placeholderPort}`;
+      }
+      return adbName;
```
noomorph, line 26: DeviceRegistry.allocateDevice could use a sort of `.toString()` method to extract the unique id, so you don't have to run these perversions with scoped `let` variables.

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -82,13 +80,10 @@ class EmulatorDriver extends AndroidDriver {
     return this._emuVersionResolver.resolve();
   }
 
-  async _boot(avdName, adbName) {
-    const coldBoot = !!this.pendingBoots[adbName];
-
+  async _boot(avdName, adbName, bootPort) {
+    const coldBoot = !!bootPort;
```
noomorph, line 84: Please, give me an example when the boot **is not** cold.
I can't imagine this situation by looking at code, maybe you have an idea, @d4vidi ?

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -107,6 +102,11 @@ class EmulatorDriver extends AndroidDriver {
     });
```
noomorph, line 102: Regarding your `_waitForBootToComplete` implementation — how well does it play with Jest timeouts? I have a bad gut feeling here, honestly, @d4vidi . ⚠️

noomorph: Where do you handle a cleanup call, which cancels the `_waitForBootToComplete` routine, Amit?

detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
```diff
@@ -0,0 +1,72 @@
...
+    return {
+      instance,
+      isNew,
+      toString: () => `GenyCloud:${instance.name} (${instance.uuid})`,
```
noomorph, line 21: See my comment regarding `EmulatorHandle`, maybe we could use a sort of `GenycloudInstanceHandle`?

detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    const options = {
+      backoff: 'none',
+      retries: 18,
+      interval: 10000,
+    };
+
+    return await retry(options, async () => {
+      const _instance = await this.instanceLookupService.getInstance(instance.uuid);
```
noomorph, line 56: See my prior comment about cleanup() handling when Jest runner times out.

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -17,43 +17,48 @@ class GenyCloudDriver extends AndroidDriver {
...
+    this._assertRecipe(deviceQuery, recipe);
 
-    const cookie = { coldBoot: false };
-    const adbName = await this.allocateDevice({
-      recipe,
-      cookie,
-      toString: () => recipe.toString(),
-    });
+    const { instance, isNew } = await this._deviceAllocator.allocateDevice(recipe);
+    const { adbName, uuid } = instance;
```
noomorph, line 38: is `uuid` = `recipeUUID`? I'd name it so, if this is the case, otherwise I'm not sure what it means

detox/src/devices/drivers/android/genycloud/exec/GenyCloudExec.js
```diff
@@ -7,12 +6,11 @@ class GenyCloudExec {
   }
 
   getInstances() {
-    return this._exec(`instances list`);
+    return this._exec(`instances list -q`);
```
noomorph, line 9: Was there an issue with exit code ≠ 0?

detox/src/devices/drivers/android/genycloud/services/dto/GenyInstance.js
```diff
@@ -1,6 +1,16 @@
...
+const STATE_ONLINE = 'ONLINE';
+const STATE_CREATING = 'CREATING';
+const STATE_BOOTING = 'BOOTING';
+const STATE_STARTING = 'STARTING';
+const initStates = [
+  STATE_CREATING,
+  STATE_BOOTING,
+  STATE_STARTING,
+];
+
```
noomorph, line 13: 👍 Nice!

detox/src/devices/drivers/android/tools/AppInstallHelper.js
```diff
@@ -1,3 +1,5 @@
+// TODO Tweak such that if apk's already exist on the device (need to store uniquely), they will not be resent (would optimize cloud, for example)
```
noomorph, line 1: Good idea. Make sure to calculate hashes, at least.

detox/src/devices/drivers/ios/SimulatorDriver.js
```diff
@@ -215,12 +215,8 @@ class SimulatorDriver extends IosDriver {
...
   async _groupDevicesByStatus(deviceQuery) {
     const searchResults = await this._queryDevices(deviceQuery);
-
-    const { busy, free}  = _.groupBy(searchResults, device => {
-      return this.deviceRegistry.isDeviceBusy(device.udid)
-        ? 'busy'
-        : 'free';
-    });
+    const takenDevices = this.deviceRegistry.getRegisteredDevices();
+    const { busy, free }  = _.groupBy(searchResults, ({ udid }) => takenDevices.includes(udid) ? 'busy' : 'free');
```
noomorph, line 219: why not rename it to tekken and not-tekken ? 😀

d4vidi: See #2573

PR #2471, opened 2020-11-18
Genycloud teardown

d4vidi:
## Description
Another (mandatory) step towards support for Genymotion's SaaS solution (see #2429): supporting tear-down.

One of the most important aspects of using cloud emulators is the proper management of their deletion (in particular, because pricing is usage-time based). This change introduces instances deletion that in a Jest-based environment, is based on the global-teardown hook, with the API invocation looking pretty much like this:
```js
// global-teardown.js
async function globalTeardown() {
  await detox.globalCleanup();
}
module.exports = globalTeardown;
```
_Namely, such that the internal details of the shutdown are obscure from the Detox user's perspective, and are available through an API that feels natural._

> In Mocha - or any other test runner that does not support parallel-execution, this can be integrated by dispatching `detox.globalCleanup()` immediately after `detox.cleanup()`. In due time, we will add this to our Documentation and scaffolding scripts.

Underneath, what happens in a nutshell is this:
1. A new, custom device-registry is initialized (on top the existing ones), which never clears its content throughout the execution.
2. Each driver spinning up a new cloud instance in it's init, adds the associated `uuid` to the device-registry.
3. Eventually, through global-teardown (at global context), the Genymotion driver statically reads the registry, issues and executes deletion operations in a `Promise.all()` fashion.
4. The end.

> Important: This technique is quite robust and may seem elegant, but it's important to keep in mind that it is nevertheless **not ideal**. Being charged by the minute, ideally, we would like to delete each instance once its worker finishes the last bits of its work. I've experimented ways of implementing this by registering a `process.on('beforeExit', () => {...})`. While it simplified the implementation a great deal, it didn't work, for two reasons:
> 1. Jest seems to hold workers up and running right until the very bitter end (global teardown), so the callback isn't called earlier than with the current approach.
> 2. Jest doesn't approve of asynchronous work inside each worker as it waits for it to be killed: it warns the user with the famous _open handles_ log, and sometimes force-kills before the `beforeExit` callback completes.
As a bottom line, then, in order to introduce the _ideal_ optimization, we would have to push this into Jest itself (see [here](https://github.com/facebook/jest/issues/8708#issuecomment-708492340)).

Last but not least, this also adds support for `driver.shutdown()`, such that genymotion-cloud testing would respect Detox' `--cleanup` argument.

### DIFF
--- a/detox/src/DetoxExportWrapper.js
+++ b/detox/src/DetoxExportWrapper.js
@@ -1,4 +1,3 @@
-const _ = require('lodash');
 const funpermaproxy = require('funpermaproxy');
 const Detox = require('./Detox');
 const DetoxConstants = require('./DetoxConstants');
@@ -102,6 +101,16 @@ class DetoxExportWrapper {
   _suppressLoggingInitErrors() {
     this[_shouldLogInitError] = false;
     return this;
+  }
+}
+
+DetoxExportWrapper.prototype.globalCleanup = async function() {
+  try {
+    // TODO For the next consumer, need to come up with some kind of infra-code to allow for dynamic registration of cleanup-callbacks.
+    const GenyCloudDriver = require('./devices/drivers/android/genycloud/GenyCloudDriver');
+    await GenyCloudDriver.globalCleanup();
+  } catch (error) {
+    log.warn({ event: 'GLOBAL_CLEANUP' }, 'An error occurred trying to shut down Genymotion-cloud emulator instances!', error);
   }
 }
 
--- a/detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
+++ b/detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
@@ -3,15 +3,19 @@ const retry = require('../../../../utils/retry');
 const logger = require('../../../../utils/logger').child({ __filename });
 
 class GenyCloudDeviceAllocator extends AndroidDeviceAllocator {
-  constructor(deviceRegistry, instanceLookupService, instanceLifecycleService) {
+  constructor(deviceRegistry, deviceCleanupRegistry, instanceLookupService, instanceLifecycleService) {
     super(deviceRegistry, logger);
 
+    this.deviceCleanupRegistry = deviceCleanupRegistry;
     this.instanceLookupService = instanceLookupService;
     this.instanceLifecycleService = instanceLifecycleService;
   }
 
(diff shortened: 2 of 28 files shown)


detox/src/DetoxExportWrapper.js
```diff
@@ -105,4 +104,14 @@ class DetoxExportWrapper {
...
 
+DetoxExportWrapper.prototype.globalCleanup = async function() {
+  try {
+    // TODO For the next consumer, need to come up with some kind of infra-code to allow for dynamic registration of cleanup-callbacks.
+    const GenyCloudDriver = require('./devices/drivers/android/genycloud/GenyCloudDriver');
+    await GenyCloudDriver.globalCleanup();
+  } catch (error) {
+    log.warn({ event: 'GLOBAL_CLEANUP' }, 'An error occurred trying to shut down Genymotion-cloud emulator instances!', error);
+  }
+}
```
d4vidi, line 115: @noomorph if you have any better suggestions as to how to provide this in a cleaner way (leaving the TODO note aside), I'd be happy to hear them.

detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
```diff
@@ -47,7 +51,7 @@ class GenyCloudDeviceAllocator extends AndroidDeviceAllocator {
     }
 
     const options = {
-      backoff: 'none',
+      backoff: 'none', // TODO apply reverse-linear polling
```
rotemmiz, line 54: What is everse-linear polling? backoff strategy?

detox/src/DetoxExportWrapper.js
rotemmiz: Maybe we could add a link to the instance's web GUI, so it would be easy to open and examine it.
Also, it might be a good idea to add that link when a cloud emulator spins up.

d4vidi: Yes, I will do both!

detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
d4vidi: Exactly

d4vidi:
@noomorph please review post-factum, along with #2466. Although I'm still not done with all of the comments from the first PR (#2446), I will get to everything, eventually.

PR #2476, opened 2020-11-22
Improve Genymotion leakage warning by including instance names

d4vidi:
## Description

Improve Genymotion leakage warnings by including instance names, in addition to meaningless UUID's.
This incorporates 2 main things:
1. Enhancing the device registry such that it would allow for managing object, and not just strings (!)
2. Managing the device cleanup registry using objects, containing both instance `uuid`'s _and_ names (thus allowing for the instance names to be logged, eventually, in global-cleanup).

Related to #2429.

### DIFF
--- a/detox/src/devices/DeviceRegistry.js
+++ b/detox/src/devices/DeviceRegistry.js
@@ -45,7 +45,7 @@ class DeviceRegistry {
   }
 
   includes(deviceId) {
-    return this._lockfile.read().includes(deviceId);
+    return !!_.find(this._lockfile.read(), (item) => _.isEqual(item, deviceId));
   }
 
   getRegisteredDevices() {
@@ -74,7 +74,7 @@ class DeviceRegistry {
     const state = this._lockfile.read();
     const newState = busy
       ? _.concat(state, deviceId)
-      : _.without(state, deviceId);
+      : _.filter(state, (item) => !_.isEqual(item, deviceId));
     this._lockfile.write(newState);
   }
 
--- a/detox/src/devices/DeviceRegistry.js
+++ b/detox/src/devices/DeviceRegistry.js
@@ -3,6 +3,9 @@ const environment = require('../utils/environment');
 const ExclusiveLockfile = require('../utils/ExclusiveLockfile');
 const safeAsync = require('../utils/safeAsync');
 
+const getDeviceEqualsFn = (deviceHandle) => (otherDeviceHandle) => _.isEqual(otherDeviceHandle, deviceHandle);
+const getDeviceDifferFn = (deviceHandle) => (otherDeviceHandle) => !_.isEqual(otherDeviceHandle, deviceHandle);
+
 class DeviceRegistry {
   constructor({ lockfilePath }) {
     /***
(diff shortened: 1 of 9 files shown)


detox/src/devices/DeviceRegistry.js
```diff
@@ -45,7 +45,7 @@ class DeviceRegistry {
   }
 
   includes(deviceId) {
-    return this._lockfile.read().includes(deviceId);
+    return !!_.find(this._lockfile.read(), (item) => _.isEqual(item, deviceId));
```
noomorph, line 48: I would not hurry to agree with your approach here, Amit.
I'd rather use small wrapper classes where you define your own `equals(other: T): boolean` logic.
WDYT?

noomorph: See 1: https://github.com/wix/Detox/pull/2466#discussion_r529422120
See 2: https://github.com/wix/Detox/pull/2466#discussion_r529404752
See 3: https://github.com/wix/Detox/pull/2466#discussion_r529420727
See 4: https://github.com/wix/Detox/pull/2466#discussion_r529431175

detox/src/devices/drivers/android/genycloud/GenyCloudDeviceAllocator.js
```diff
@@ -13,16 +13,18 @@ class GenyCloudDeviceAllocator extends AndroidDeviceAllocator {
 
   async _doAllocateDevice(recipe) {
     let { instance, isNew } = await this._doSynchronizedAllocation(recipe);
+    const { uuid, name } = instance;
+
     if (isNew) {
-      await this.deviceCleanupRegistry.allocateDevice(instance.uuid);
+      await this.deviceCleanupRegistry.allocateDevice({ uuid, name });
```
noomorph, line 19: Here the perversions start again 😃  — if you use specific-designed data classes for EmulatorHandle, GenymotionInstanceHandle with some equality or identity method implementations, you don't have to create on-fly these small objects.

detox/src/devices/drivers/android/genycloud/services/dto/GenyRecipe.js
```diff
@@ -1,13 +1,12 @@
-const _ = require('lodash');
-
-class Recipe {
+class GenyRecipe {
   constructor(rawRecipe) {
-    Object.assign(this, _.pick(rawRecipe, 'uuid', 'name'));
+    this.uuid = rawRecipe.uuid;
```
noomorph, line 3: Looks better :)

detox/src/devices/drivers/android/genycloud/services/dto/GenyInstance.js
```diff
@@ -13,7 +12,9 @@ const initStates = [
 
 class GenyInstance {
   constructor(rawInstance) {
-    Object.assign(this, _.pick(rawInstance, 'uuid', 'name', 'state'));
+    this.uuid = rawInstance.uuid;
+    this.name = rawInstance.name;
+    this.state = rawInstance.state;
```
noomorph, line 17: Thanks!

noomorph (commented):
That's it for now.

detox/src/devices/DeviceRegistry.js
d4vidi: The funny thing is that this is how I would undoubtedly have done this in Java/Kotlin. Seems I've underestimated JS. I will definitely try to do this as you suggested 💪🏻

d4vidi: So I haven't anything idiomatic that corresponds to your suggestion (namely such that evidently, an equality operator would result in calling a custom function). Could you pinpoint me here?

PR #2480, opened 2020-11-23
Prevalidate gmsaas: binary and auth

d4vidi:
## Description

As per insightful notes by @noomorph - added a validation of that `gmsaas` is in path (so as to avoid pointless retires to take place in trying to execute initial setup command, resulting in obscure, unfriendly errors).

Actually, with genymotion's SaaS, there are two things to handle before enabling access to the cloud:
1. Setting up `gmsaas` in path
2. Logging in

This PR effectively handles both aspects.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -13,10 +13,10 @@ class Device {
   }
 
   async prepare() {
+    await this.deviceDriver.prepare();
+
     this._deviceId = await this.deviceDriver.acquireFreeDevice(this._deviceConfig.device || this._deviceConfig.name);
     this._bundleId = await this.deviceDriver.getBundleIdFromBinary(this._deviceConfig.binaryPath);
-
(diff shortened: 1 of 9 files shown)


detox/src/devices/Device.js
```diff
@@ -13,10 +13,10 @@ class Device {
   }
 
   async prepare() {
+    await this.deviceDriver.prepare();
+
     this._deviceId = await this.deviceDriver.acquireFreeDevice(this._deviceConfig.device || this._deviceConfig.name);
     this._bundleId = await this.deviceDriver.getBundleIdFromBinary(this._deviceConfig.binaryPath);
-
```
d4vidi, line 18: @LeoNatan be advised. The only current usage of this is by SimulatorDriver. I believe it would be better for it, too.

noomorph (commented):
LGTM. 👍

PR #2482, opened 2020-11-25
feat(artifacts): always loglevel=trace for log artifacts

noomorph:
- [x] This is a small change 

---

**Description:**

When something wrong happens, we usually go to the logs.
When things mostly go alright, we don't turn on verbose logging mode.
At the moment, verbosity in log artifacts is tightly coupled to the verbosity of CLI output.

This PR suggests using always `loglevel=trace` for log artifacts, presuming that a user will stick to `--record-logs failing` in the majority of cases. If it is `--record-logs all` — better to take care of CI storage. ;)


rotemmiz:
Great!

d4vidi:
LGTM 2 😄

PR #2500, opened 2020-12-02
Fix text object being of unexpected class, other than string

LeoNatan:
@valentynberehovyi


valentynberehovyi:
Thank you for the fix!

PR #2505, opened 2020-12-03
Timeline NG

d4vidi:
## Description
Significantly improve the value of the time-line artifact, by introducing a global tracing API and wiring the artifact-plugin to it -- thus allowing for events reporting from anywhere and at any time, even during the very init of Detox (i.e. before artifacts are even initialized).

> Note: I haven't added too many events to the artifact so far. Instead, I mainly focused on making it easy to apply new events from this point on.

Here's an example of what the final output might look like:

Screen Shot 2020-12-03 at 14 05 04

## How to review

Best to first review the new `trace` and `ChromeTracingExporter` **utility** modules, which together roughly replace the previous `Trace` class, associated with the artifacts plugin, specifically. Namely, instead of having a one-size-fits all `trace` class as before, we now have a global singleton class that more generically _collects_ events, and another one that simply exports them to the chrome-tracing "format".
From there, move on to the new implementation of `TimelineArtifactPlugin.onBeforeCleanup()`, which puts the trace and other things together.
Then move on to review the rest of the code - namely, which events are being reported, and where.

## Next up

1. On top of this, there's more work pending for regression-testing the artifact's integrity. For that, I've introduced a new, interesting, way of executing integration tests. See [here](https://github.com/wix/Detox/compare/timeline-ng...timeline-integ-test).
2. Even more so - I've had the new `trace` and `traceCall` methods exported via the `detox` object, so that devs could make use of it in their tests. See [here](https://github.com/wix/Detox/compare/timeline-integ-test...timeline-trace-export).


LeoNatan:
@d4vidi Why not just bump Node requirement? We already updated the docs to require Node 10 in DetxoNext.

noomorph:
@LeoNatan, flatMap appears in Node 11, not 10.

LeoNatan:
Isn't that very old too? Let's require Node 12.

d4vidi:
I'm all for it, but on a different scope. For now the solution is very simple.

LeoNatan:
In Detox 18, I meant.

noomorph:
@LeoNatan , I am afraid our CI still uses Node 10. 🤔

LeoNatan:
The CI can easily be upgraded.

d4vidi:
Guys, I'm going to move with this feature as-is. Let's reiterate until we are all satisfied with the API and implementation.

PR #2522, opened 2020-12-09
Timeline self-integration testing

d4vidi:
## Description

A follow-up on #2505, which introduces regression testing over the _timeline-artifact_ feature through the introduction of the very concept of (JS) integration testing to Detox. The general approach is to enable the generation of the artifact in a deterministic way (i.e. same result as long as nothing changes), and then snapshot-test it. This is especially challenging because the artifact's output contain timestamps...

The concept includes the following:
1. Introduction of a new kind of a driver+expectation set, which implement the various methods using stub sleeps (driver) or no-ops (expectations); Allowing its instantiation through a new type of a Detox configuration, named `stub`. _Effectively, this enables running all of the JS code, throughout Detox' lifecycle, with no devices involved (!)_
2. Generation of the time-line artifact using the associated artifact plugin in a special mode which injects fake timestamps, so that as long as nothing in the code changes - the end-result would be the same (the opposite is also true, of course).
3. Introduce integration-testing infra, based on Jest.
4. Introduce a single integration suite, which runs Detox over 3 stub e2e suites using the stub driver and snapshot tests the resulted artifact. That's right, a test suite that runs Detox (in stub mode) 😄 .


detox/src/artifacts/timeline/TimelineArtifactPlugin.js
```diff
@@ -4,20 +4,23 @@ const ArtifactPlugin = require('../templates/plugin/ArtifactPlugin');
...
   endSection: _noop,
 };
 
 class TimelineArtifactPlugin extends ArtifactPlugin {
   constructor(config) {
     super(config);
+    this._useFakeTimestamps = config.useFakeTimestamps;
 
-    this._trace = this.enabled ? trace : traceStub;
+    const threadId = process.env.JEST_WORKER_ID || process.pid;
```
d4vidi, line 19: We soon ought to normalize the concept of worker-ID, making it runner-agnostic.

detox/src/devices/DriverRegistry.test.js
```diff
@@ -54,6 +55,14 @@ describe('DriverRegistry', () => {
       expect(EmulatorDriver).toHaveBeenCalledWith(opts);
     });
 
+    it('should resolve "stub" to StubDriver', () => {
+      const StubDriver = require('./drivers/stub/StubDriver');
+      const driver = registry.resolve('stub', opts);
```
noomorph, line 60: I don't like this at all. We should not export our internal for-test code to the end users. Why can't you use a direct path to the driver file instead, and put the driver inside `detox/test/integration`? @d4vidi

d4vidi: Why - what's the concern? This isn't an external driver per se

noomorph: It is used only one place so far, to test one thing here for us. Why it should be packaged to npm registry at all, e.g.?

d4vidi: I'm actually intrigued to see whether the community could find something creative to do with this 😆

d4vidi: ok ok I've applied that in my latest commit. Please resolve :)

PR #2540, opened 2020-12-18
feat(config): launch app manually

noomorph:
**Description:**

This feature is aimed at improving usability when you need to debug native Detox errors on iOS and Android respectively.

```json
{
  "ios.manual": {
    "type": "ios.simulator",
    "binaryPath": "... path to the app ...",
    "device": "... your device config ...",
    "artifacts": false,
    "behavior": {
      "launchApp": "manual"
    },
    "session": {
      "autoStart": true,
      "server": "ws://localhost:8099",
      "sessionId": "someTestSessionId"
    }
  },
  "android.manual": {
    "type": "android.emulator",
    "binaryPath": "... path to the app ...",
    "device": "... your device config ...",
    "behavior": {
      "launchApp": "manual"
    },
    "session": {
      "autoStart": true,
      "server": "ws://localhost:8099",
      "sessionId": "someTestSessionId"
    }
}
```

So, in this example, when you run `detox test -c ios.manual` or `detox test -c android.manual` on a test you want to debug, at some point of time when you get to `device.launchApp()` call, instead of launching the app, Detox will pause its execution with a message similar to the ones shown below.

iOS:

```
detox[16804] INFO:  [SimulatorDriver.js] Waiting for you to manually launch your app in Xcode.
Make sure to pass the launch arguments listed below:
  -detoxServer ws://localhost:8099
  -detoxSessionId com.wix.detox-example

Press any key to continue...
```

Android:

```
detox[53038] INFO:  [AndroidDriver.js] Waiting for you to manually launch your app in Android Studio.

Instrumentation class: com.wix.detox.test.test/com.example.DetoxTestAppJUnitRunner
Instrumentation arguments:
------------------------------------
Key            | Value
------------------------------------
detoxServer    | ws://localhost:8099
detoxSessionId | test
------------------------------------

Press any key to continue...
```

You'll need to run your app with the said arguments from Xcode or Android Studio:

iOS:

Xcode - Edit Schema - Arguments tab

Android:

image

When you run the app, return to your running Detox test, and press any key (of course, except Ctrl+C which just terminates the process). As a result, you are expected to see a confirmation from Detox, e.g.::

```
Found the app (com.wix.detox-example) with process ID = 16854. Proceeding...
```

**Of course, this does not cancel the necessity of preparing of your Xcode or Android Studio project**: e.g. for Xcode, that at least implies unpacking `node_modules/detox/Detox-ios-src.tbz` and embedding it as a build phase:

image

Nevertheless, this feature somewhat improves the developer experience compared to `ios.none` config, especially when you have to orchestrate a sequence of mock servers before you can launch the app since this workflow does not require changes in the source code of your e2e tests.

### TODO

- [x] update the docs (config: artifacts, behavior, client/server; troubleshooting guide, debugging natively).
- [x] decide if we want to disable that monitor: https://github.com/wix/Detox/blob/master/detox/android/detox/src/main/java/com/wix/detox/reactnative/ReactNativeLoadingMonitor.kt#L56


LeoNatan:
Still not clear to me what the difference between ios.none and this is, or why this is needed.

detox/src/devices/drivers/android/AndroidDriver.js
```diff
@@ -303,6 +337,43 @@ class AndroidDriver extends DeviceDriverBase {
...
   }
+
+  async _waitForAppLaunch(adbName, bundleId, launchArgs) {
+    const instrumentationClass = await this.adb.getInstrumentationRunner(adbName, bundleId);
+    this._printInstrumentationHint({ instrumentationClass, launchArgs });
+    await pressAnyKey();
+    await this._reverseServerPort(adbName);
+
+    // TODO: think about a more elegant way to bypass instrumentation
+    this.instrumentation.waitForCrash = async () => {};
```
noomorph, line 348: @d4vidi, any idea how to improve here in a way true to your instrumentation helper object design?

noomorph:
This can work with all artifacts enabled, as usual, for both platforms.
As for the redundant arguments — you are technically right, but you can actually omit them only (!) if they match the default values, which is not always true:

E.g. the configuration below has only `behavior: { launchApp": "manual" }` override, and that's it:

```json
{
      "ios.manual": {
        "type": "ios.simulator",
        "binaryPath": "ios/build/Build/Products/Debug-iphonesimulator/example.app",
        "behavior": {
          "launchApp": "manual"
        },
        "device": {
          "type": "iPhone 11 Pro"
        }
      }
}
```

And yet it works the same as the regular `ios.simulator` version, with artifact plugins enabled as well, which is something you cannot do for `ios.none` that easily (because it is agnostic to device/simulator distinction).

And look, it also autogenerates URL and sessionId like in default settings:

```
➜  test git:(feat/launch-app-manually) ✗ npx detox test -c ios.manual e2e/01.sanity.test.js      detox[2448] INFO:  [test.js] configuration="ios.manual" useCustomLogger=true DETOX_START_TIMESTAMP=1608554308067 reportSpecs=true nyc jest --config e2e/config.js --testNamePattern '^((?!:android:).)*$' --maxWorkers 1 e2e/01.sanity.test.js
detox[2450] INFO:  [DetoxServer.js] server listening on localhost:50130...
detox[2450] INFO:  [AppleSimUtils.js] Waiting for you to manually launch your app in Xcode.
Make sure to pass the launch arguments listed below:
  -detoxServer "ws://localhost:50130"
  -detoxSessionId "845f9c9a-9a34-3c90-828a-69bf96baa754"

Press any key to continue...
```

So this is a case when you actually NEED to copy&paste the said values to your IDE.
Overall, I cannot exclude the possibility of debugging multi-worker issues as well, so I like that this solution can be very close to the real configuration on CI, for example.

LeoNatan:
Why not fix ios.none to work with artifacts?

noomorph:
Because it is a simulator-agnostic driver, and it will make me think more about how to debug it right in a complex configured suite of tests. Also, it will result in an awkward AndroidNoneDriver, if we decide to override "launchApp" implementation so it waits for you to launch the app, because contrary to IosDriver, the AndroidDriver has a full `launchApp` implementation over adb.

noomorph:
And I never said I won't fix `ios.none` artifacts issues you have reported.
I just wanted to group that together with single-timeline rewrite task, not just now.

LeoNatan:
I didn’t ask just about fixing, but perhaps extend the artifacts to support ios.none fully.

detox/src/devices/drivers/android/AndroidDriver.js
d4vidi: @noomorph if `instrumentation` is not at play here (test is instrumented manually), can the call to do [this function](https://github.com/wix/Detox/pull/2540/files#diff-1f95bce1adbbe6aae44166a89a62433fba2b9bb8d98ab439342abd509f8ff937R278) be omitted altogher, instead?

noomorph: @d4vidi , I am not sure if I am calling that function at all, in the manual flow.

I'd say that the problem stems rather from the fact that no one calls instrumentation init, while later we call this place:
```
        await Promise.race([super.waitUntilReady(), this.instrumentation.waitForCrash()]);
```

and non-initialized instrumentation throws an error when waitingForCrash - because it cannot find the process id of the instrumentation running, which is obvious because we never wanted to launch it in the first place.

LeoNatan (changes requested):
Do not touch Debugging in Xcode guide please, nor merge it with Android. 

Also, this new guide complicates things too much, unlike the previous guide of adding `ios.none`. I am sorry, bug I don't like it. This maybe cool internal guide on how to debug internal apps, but I want the Xcode debug guide to be as simple and short as possible.

LeoNatan:
Also, I will be rewriting the Xcode guide soon for XCUITest, and that too will be based on `ios.none` as I don't want any Detox complexities to mix with XCUITest.

detox/src/devices/drivers/android/AndroidDriver.js
d4vidi: Well off the top of my head I can say that waiting for crash isn't all that relevant here either. Since detox doesn't launch instrumentation, it should not account for its crashing. I think it's safe to say that in this manual mode, the user is hands on with these things.

noomorph: Okay, I see what you mean. Correct!
I think I know what to do. Thanks.

PR #2555, opened 2020-12-27
Warn about Genycloud instances leakage in case of abrupt terminations

d4vidi:
## Description

Part of the ongoing work on Genymotion-cloud support (#2429).

Warn about Genycloud instances leakage in case test suite execution is terminated prematurely (e.g. due to the user hitting a CTRL+C, or a kill signal). That is to take place on top of the standard (controlled) cleanup performed by Detox following a proper execution of the test suites down to the very end.

- This relates mostly to users cancelling their jobs manually while running test locally.
- It is not related to whether the tests in the suite have failed: A test suite failure still undergoes a proper cleanup.
- It will _not_ run in case of a CI agent's sudden shutdown.

### DIFF
--- a/detox/src/devices/DeviceRegistry.js
+++ b/detox/src/devices/DeviceRegistry.js
@@ -1,19 +1,30 @@
 const _ = require('lodash');
 const environment = require('../utils/environment');
+const fs = require('fs-extra');
 const ExclusiveLockfile = require('../utils/ExclusiveLockfile');
 const safeAsync = require('../utils/safeAsync');
 
 const getDeviceEqualsFn = (deviceHandle) => (otherDeviceHandle) => _.isEqual(otherDeviceHandle, deviceHandle);
 const getDeviceDifferFn = (deviceHandle) => (otherDeviceHandle) => !_.isEqual(otherDeviceHandle, deviceHandle);
+const readOptions = {
+  encoding: 'utf8',
+};
 
 class DeviceRegistry {
   constructor({ lockfilePath }) {
+    /***
+     * @private
+     * @type {string}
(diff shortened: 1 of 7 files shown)


detox/src/devices/DeviceRegistry.js
```diff
@@ -64,6 +75,11 @@ class DeviceRegistry {
     return result;
   }
 
+  readRegisteredDevicesUNSAFE() {
+    const contents = fs.readFileSync(this._lockfilePath, readOptions);
+    return JSON.parse(contents);
+  }
```
d4vidi, line 81: Note: This is effectively needed - not because the unsafe access is mandatory here, but really for making the _sync_ read. From within a signal-catch callback, no further queued work is honored, and hence promises (and thus async-await) cannot be used.

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -106,28 +110,34 @@ class GenyCloudDriver extends AndroidDriver {
     }
   }
 
+  static async globalInit() {
+    onSignalExit(() => {
+      const deviceCleanupRegistry = GenyDeviceRegistryFactory.forGlobalShutdown();
```
noomorph, line 115: Overall, my only question is why don't you KILL on signal exit? Why only print?

noomorph: Per discussion, the motivation is to leave the devices running if the programmer clicks Ctrl+C with an intention to look at something specific about those currently running devices.

PR #2556, opened 2020-12-27
Added back --args to xcrun simctl launch, was wrongly removed in #2540

rotemmiz:
**Description:**
Fix an issue introduced in #2540, where `--args` was dropped from `xcrun simctl launch`.
Missing: some sort of testing for this, to make sure args are being passed correctly.

### DIFF
--- a/detox/src/devices/drivers/ios/tools/AppleSimUtils.js
+++ b/detox/src/devices/drivers/ios/tools/AppleSimUtils.js
@@ -331,7 +331,7 @@ class AppleSimUtils {
 
     const cmdArgs = this._mergeLaunchArgs(launchArgs, languageAndLocale).join(' ');
     let launchBin = `SIMCTL_CHILD_DYLD_INSERT_LIBRARIES="${dylibs}" ` +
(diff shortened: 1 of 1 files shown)


d4vidi:
@rotemmiz @LeoNatan @alon-ha what use case could we apply as a test in order to cover future regressions?

LeoNatan:
This was removed on purpose I think. @noomorph ?
It was being sent to AppleSimUtils which is incorrect.

noomorph:
🤦  Leo, but you told it is not needed?

rotemmiz:
@LeoNatan it was sent to `xcrun simctl launch`, not to AppleSimUtils, or am I getting you wrong?

LeoNatan:
It’s not needed for AppleSimUtils.

noomorph:
@LeoNatan, so it looks like a reliance coming from the app engine itself?

LeoNatan:
Yes, it’s buggy app behavior. This should be reverted and app fixed.

LeoNatan:
@rotemmiz @noomorph

PR #2560, opened 2020-12-30
DeviceRegistry to support all handle types

d4vidi:
## Description

Expand `DeviceRegistry` so as to be made capable of storing device-handles of 3 possible types, rather than just ~2:
- Explicit strings
- Plain objects (functions stripped away)
- New: Class instances.

That is achieved by making changes on two levels:
- Reimplementation of `DeviceRegistry`'s modification methods (e.g. device allocation / dispose) - also `includes`, so as to sanitize (i.e. strip-off functions, any class meta data, etc.) the given handle before using it. This is a non-API-breaking change.
- Change of registry whole-state query methods, namely `getRegisteredDevices()`, `readRegisteredDevices()` and `readRegisteredDevicesUNSAFE()` such that instead of returning an array, they'd return an abstraction (`DeviceHandlesList` class) that's different in two ways:
  - It allows for a full access to the device list, but only via the `sanitizedDevices` property which makes it clear for the caller that they are actually only seeing _sanitized_ handles -- implementation constraints 🤷🏻  (to be clear, this was how things were regardless of this change - it's just been made explicit by it).
  - It provides an `includes` method (equivalent to `Array.includes()`, formerly used) that can work with all of the 3 handle types (i.e. by sanitizing before making any comparisons) - allowing the user to provide a handle in the original form used for allocation.

For the latter, associated usage of these API's have been changed to match the changes.

Last but not least, in turn -- this fixes a regression in Genymotion-cloud shutdown when using the `--cleanup` flag, introduced due to the switching to class instances as device-registry handles for Genymotion instances cleanup, in #2476. That in fact was what motivated this change in the first place.

### DIFF
--- a/detox/src/devices/DeviceRegistry.test.js
+++ b/detox/src/devices/DeviceRegistry.test.js
@@ -2,6 +2,33 @@ const fs = require('fs-extra');
 const tempfile = require('tempfile');
 const environment = require('../utils/environment');
 
+const deviceHandle = 'emulator-5554';
+
+const deviceHandleAsObjSanitized = {
+  type: 'mocked-device-type',
+  adbName: 'emulator-5556',
+};
+const deviceHandleAsObj = {
+  ...deviceHandleAsObjSanitized,
+  mockFn: () => 'mock-fn-result',
+};
+
+class DeviceHandle {
+  constructor() {
+    this.type = 'mocked-device-type';
+    this.adbName = 'localhost:11111';
+  }
+
+  mockFunc() {
+    return 'mocked-func-result';
+  }
+}
+const deviceHandleAsClassInstance = new DeviceHandle();
+const deviceHandleInstanceSanitized = {
+  type: 'mocked-device-type',
+  adbName: 'localhost:11111',
+}
+
 describe('DeviceRegistry', () => {
   let DeviceRegistry;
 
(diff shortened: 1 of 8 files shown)


detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -76,9 +77,11 @@ class GenyCloudDriver extends AndroidDriver {
   }
 
   async shutdown(instance) {
+    const instanceHandle = new GenyCloudInstanceHandle(instance);
+
     await this.emitter.emit('beforeShutdownDevice', { deviceId: instance.adbName });
     await this._instanceLifecycleService.deleteInstance(instance.uuid);
-    await this._deviceCleanupRegistry.disposeDevice(instance.uuid);
+    await this._deviceCleanupRegistry.disposeDevice(instanceHandle);
```
d4vidi, line 84: This is the bug-fix in question

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -111,18 +114,20 @@ class GenyCloudDriver extends AndroidDriver {
   }
 
   static async globalInit() {
-    onSignalExit(() => {
-      const deviceCleanupRegistry = GenyDeviceRegistryFactory.forGlobalShutdown();
-      const instanceHandles = deviceCleanupRegistry.readRegisteredDevicesUNSAFE();
-      if (instanceHandles.length) {
-        reportGlobalCleanupSummary(instanceHandles);
+    onSignalExit((code, signal) => {
+      if (signal) {
```
d4vidi, line 118: Just a fluke I spotted along the way

detox/src/devices/DeviceRegistry.test.js
```diff
@@ -2,6 +2,33 @@ const fs = require('fs-extra');
 const tempfile = require('tempfile');
 const environment = require('../utils/environment');
 
+const deviceHandle = 'emulator-5554';
+
+const deviceHandleAsObjSanitized = {
```
noomorph, line 7: Why "sanitized"? What hacker attack we are preventing?

d4vidi: Lol. I will rename across the board. What would be the proper terminology? "clean"? Stripped?!

PR #2565, opened 2021-01-01
Save frequent subprocess constant results for faster launchApp

rotemmiz:
**Description:**
This PR handles a few unrelated optimizations.
1. Save frequent subprocess constant results for faster launchApp.
2. Split `03.actions.test.js` suite into two (`03.actions-scroll.test.js`) , as it is very long, and takes significantly more time than others. By splitting it, it makes parallel test execution with many workers (5-10) faster.
3. Added timeline logging on all execWithRetriesAndLogs for better visibility.

### DIFF
--- a/detox/src/devices/drivers/ios/tools/AppleSimUtils.js
+++ b/detox/src/devices/drivers/ios/tools/AppleSimUtils.js
@@ -345,12 +345,16 @@ class AppleSimUtils {
     return result;
   }
 
-  async _printLoggingHint(udid, bundleId) {
+  async _assembleLoggingHint(udid, bundleId) {
     const appContainer = await this.getAppContainer(udid, bundleId);
     const CFBundleExecutable = await exec.execAsync(`/usr/libexec/PlistBuddy -c "Print CFBundleExecutable" "${path.join(appContainer, 'Info.plist')}"`);
     const predicate = `process == "${CFBundleExecutable}"`;
-    const command = `/usr/bin/xcrun simctl spawn ${udid} log stream --level debug --style compact --predicate '${predicate}'`;
+    this.loggingHintCommand = `/usr/bin/xcrun simctl spawn ${udid} log stream --level debug --style compact --predicate '${predicate}'`;
+    return this.loggingHintCommand;
+  }
 
+  async _printLoggingHint(udid, bundleId) {
+    const command = this.loggingHintCommand || await this._assembleLoggingHint(udid, bundleId);
     log.info(`${bundleId} launched. To watch simulator logs, run:\n        ${command}`);
   }
 
--- a/detox/src/devices/drivers/ios/tools/AppleSimUtils.js
+++ b/detox/src/devices/drivers/ios/tools/AppleSimUtils.js
@@ -345,16 +345,12 @@ class AppleSimUtils {
     return result;
   }
 
-  async _assembleLoggingHint(udid, bundleId) {
+  async _printLoggingHint(udid, bundleId) {
     const appContainer = await this.getAppContainer(udid, bundleId);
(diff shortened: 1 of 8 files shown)


detox/src/devices/drivers/ios/tools/AppleSimUtils.js
```diff
@@ -345,12 +345,16 @@ class AppleSimUtils {
...
     const appContainer = await this.getAppContainer(udid, bundleId);
     const CFBundleExecutable = await exec.execAsync(`/usr/libexec/PlistBuddy -c "Print CFBundleExecutable" "${path.join(appContainer, 'Info.plist')}"`);
     const predicate = `process == "${CFBundleExecutable}"`;
-    const command = `/usr/bin/xcrun simctl spawn ${udid} log stream --level debug --style compact --predicate '${predicate}'`;
+    this.loggingHintCommand = `/usr/bin/xcrun simctl spawn ${udid} log stream --level debug --style compact --predicate '${predicate}'`;
+    return this.loggingHintCommand;
+  }
 
+  async _printLoggingHint(udid, bundleId) {
+    const command = this.loggingHintCommand || await this._assembleLoggingHint(udid, bundleId);
```
noomorph, line 357: Overall, while I understand and support your good intention, I don't think this code is really valid.

The biggest reason is that the applesimutils class, like DeviceDriver class and unlike Device class is currently state less, and this change breaks maybe stupid but till now consistent logic. 

The smaller but a more practical reason is that this code will be invalid after first next app reinstall (weirdly enough, but app container path is changing all the time, the GUID part in the middle of it).

An even smaller reason which still is valid is that any attempt to launch another app will produce an incorrect debug logging hint. 

I don't recommend to leave it this way, it breaks many things in a subtle way.

detox/src/utils/environment.js
```diff
@@ -157,12 +158,16 @@ function getDetoxVersion() {
...
-async function getFrameworkPath() {
-  const detoxVersion = this.getDetoxVersion();
+async function setFrameworkPath() {
+  const detoxVersion = getDetoxVersion();
   const sha1 = (await exec(`(echo "${detoxVersion}" && xcodebuild -version) | shasum | awk '{print $1}'`)).stdout.trim();
   return `${DETOX_LIBRARY_ROOT_PATH}/ios/${sha1}/Detox.framework`;
 }
 
+function getFrameworkPath() {
+  return IOS_FRAMEWORK_PATH;
```
noomorph, line 168: I'd insist on lazy calling setFrameworkPath here, if the promise is not initialized.
Modules with side effects upon require are less pleasant to test. 

Also this code might break on Windows and Linux if you call it for no good reason.

detox/src/utils/exec.js
```diff
@@ -32,7 +33,7 @@ async function execWithRetriesAndLogs(bin, options = {}) {
       } else if (statusLogs.retrying) {
         _logRetrying(log, cmd, retryNumber, lastError);
       }
-      result = await exec(cmd, { timeout });
+      result = await traceCall(`cmd: ${cmd}`, () => exec(cmd, { timeout }));
```
noomorph, line 36: Good!

detox/src/utils/trace.js
```diff
@@ -1,6 +1,7 @@
 class Trace {
   constructor() {
     this.events = [];
+    this._timestampProviderFn = Date.now;
```
noomorph, line 4: Was there a specific motivation for this?

detox/test/e2e/03.actions-scroll.test.js
```diff
@@ -0,0 +1,74 @@
+const custom = require('./utils/custom-it');
+
+describe('Actions - Scroll', () => {
```
noomorph, line 3: The splitting sounds reasonable.

noomorph (changes requested):

noomorph:
TODO - extract e2e test split.

PR #2572, opened 2021-01-05
test: add <ActivityIndicator> sync test

noomorph:
- [x] This is a small change 

---

**Description:**

Detox is stuck forever when React Native's built-in `<ActivityIndicator />` is being rendered.

The issue can be reproduced on iOS since Detox 18.0.0 (the latest version is 18.0.3, at the moment, and it is affected as well).

```js
it('should show activity indicator without stucking', async () => {
  await element(by.id('showIndicator')).tap(); // here the test is stuck
  await expect(element(by.id('indicator'))).toBeVisible();
});

```

iOS simulator stuck

### DIFF
--- a/detox/test/src/Screens/SyncScreen.js
+++ b/detox/test/src/Screens/SyncScreen.js
@@ -1,0 +1,47 @@
+import React, { Component } from 'react';
+import {
+  Text,
+  TouchableOpacity,
+  ActivityIndicator,
+  Platform,
+  StyleSheet,
+  SafeAreaView,
+  View,
+} from 'react-native';
+
+const styles = StyleSheet.create({
+  screen: {
+    flex: 1,
+  },
+
+  container: {
+    justifyContent: 'flex-start',
+    paddingTop: Platform.OS === 'ios' ? 30 : 0,
+  },
+
+  textButton: {
+    color: 'blue',
+    marginVertical: 10,
+    textAlign: 'center',
+  },
+});
+
+export default class SyncScreen extends Component {
+  state = {
+    showIndicator: false,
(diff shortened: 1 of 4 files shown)


LeoNatan:
Will look, thanks

LeoNatan:
Just to understand, DetoxSync reports the system as busy when the activity indicator is animating? That's the expected result.

detox/test/src/Screens/SyncScreen.js
```diff
@@ -0,0 +1,47 @@
...
+
+  render() {
+    return (
+      <SafeAreaView style={styles.container}>
+        <View testID="SyncScreenContainer" style={styles.container}>
+          <TouchableOpacity testID="showIndicator" onPress={() => this.setState({ showIndicator: true })}>
+            <Text style={styles.textButton}>Show Activity Indicator</Text>
+          </TouchableOpacity>
+
+          { this.state.showIndicator && <ActivityIndicator testID="indicator" key="indicator" size="large" color="black" /> }
```
LeoNatan, line 42: This test is invalid and will never be idle. If an indicator appears and animates, the system is considered busy, and will never be idle.

LeoNatan (changes requested):

LeoNatan:
Just verified. Once the `stopAnimating` method of `UIActivityIndicatorView` is called, Detox correctly notices the app is idle.

LeoNatan:
In fact, `UIActivityIndicatorView` is implemented internally as an animating `UIImageView` (like a GIF or a motion image). So any of those will also make the app not idle while animating.

PR #2573, opened 2021-01-06
Rework android device allocation framework

d4vidi:
## Description

Genymotion-cloud work tech debt (#2429): Avoid returning complex values from Android device allocation helpers onto the drivers, by pushing *all* responsibility onto the helper classes (rather than have it split between the two).

This cleans up driver code, simplified contracts and increases unit test coverage.


d4vidi:
@noomorph, 'not going to wait for a review with this one - it's all based on your feedback, basically. Obviously if you're into it as a sport, I would appreciate it.

PR #2582, opened 2021-01-11
Update "does differently" to correct md file

denis-tsoi-mox:
- [x] This is a small change 

---

**Description:**
The markdown file referenced in the readme is a dead link. Updating according to existing docs.


LeoNatan:
🤦‍♂️

Thank you!

Issue #2563, opened 2020-12-31
Configuration revamp (multiple apps support)


d4vidi:
tl;dr we'd like to allow for a more flexible configuration. We're considering something like this - which motivations will be explained below (todo):

```json
{
  "detox": {
    "devices": {
      "iphone.sim": {
        "type": "ios.simulator",
        "name": "iPhone 11"
      },
      "ipad.sim": {
        "type": "ios.simulator",
        "name": "iPad Pro"
      },
      "android.emu": {
        "type": "android.emulator",
        "name": "Pixel_API_28"
      },
      "android.tablet": {
        "type": "android.emulator",
        "name": "Pixel_C_API_29"
      }
    },
    "applications": {
      "app1.release": {
        "android": {
          "binaryPath": "build/e2e.android/devRelease/app1/App.apk",
          "testBinaryPath": "./android/app/build/outputs/apk/androidTest/development/debug/app-development-debug-androidTest.apk",
          "build": "cd android ; ./gradlew assembleRelease assembleAndroidTest -DtestBuildType=debug ; cd -"
        },
        "ios": {
          "binaryPath": "build/e2e.ios/devRelease/app1/App.app",
          "build": "xcodebuild -project ios/example.xcodeproj -UseNewBuildSystem=NO -scheme example -configuration Release -sdk iphonesimulator -derivedDataPath ios/build"
        }
      },
      "app1.debug": {
        "android": {
          "binaryPath": "build/e2e.android/debug/app1/WixOneApp.apk",
          "testBinaryPath": "./android/app/build/outputs/apk/androidTest/development/debug/app-development-debug-androidTest.apk",
          "build": "cd android ; ./gradlew assembleDebug assembleAndroidTest -DtestBuildType=debug ; cd -"
        },
        "ios": {
          "binaryPath": "build/e2e.ios/debug/app1/App.app",
          "build": "xcodebuild -project ios/example.xcodeproj -UseNewBuildSystem=NO -scheme example -configuration Debug -sdk iphonesimulator -derivedDataPath ios/build"
        }
      },
      "app2.release": {
        "android": {
          "binaryPath": "build/e2e.android/devRelease/app2/App.apk",
          "testBinaryPath": "./android/app/build/outputs/apk/androidTest/development/debug/app-development-debug-androidTest.apk",
          "build": "cd android ; ./gradlew assembleRelease assembleAndroidTest -DtestBuildType=debug ; cd -"
        },
        "ios": {
[...]

PR #2587, opened 2021-01-12
Multiple applications support (Phase 1)

noomorph:
- [x] This change has been discussed in issue #2563 and the solution has been agreed upon with maintainers.

---

**Description:**

- [x] Add new definitions for the config
- [x] Implement new device config resolution
- [x] Implement new apps config resolution
- [x] Implement device.selectApp(name) method
- [x] In `composeAppsConfig.js`: reimplement `cliConfig.appLaunchArgs` override (temporary deleted at the moment)
- [x] In `composeAppsConfig.test.js`: cover unhappy scenarios for: colliding apps, property name typos, mismatches between app and device platform, restore launch args tests, etc...
- [x] implement multiple `detox build` capability and improve some warnings
- [x] In `detox init` – adapt template to the new configuration schema
- [x] Move `utilBinaryPaths` to device configuration
- [x] Expand `DetoxConfigErrorBuilder.test.js` to cover all new methods
- [x] In `DetoxConfigErrorBuilder.js`: replace stub methods with real validation error creation:
- [x] Forcibly relaunch app on `device.launchApp()` after selecting another app — overall, the best workaround at the moment is to terminate the current app when selecting another one. While theoretically it is not required, practically this is the safest choice considering that we have a single session id for all the apps. 😞 We might improve that later.
- [x] Add a small e2e test for switching between the apps
- [x] Update the docs

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -405,16 +405,6 @@ declare global {
             selectApp(app: string): Promise<void>;
 
             /**
-             * @protected
-             * Unselect the current app.
-             * After execution, all app-specific device methods will be throwing an error.
-             *
-             * @example
-             * await device.selectApp(null); // explicitly unselects any app
-             */
-             selectApp(app: null): Promise<void>;
-
-            /**
              * Launch the app.
              *
              * <p>For info regarding launch arguments, refer to the [dedicated guide](https://github.com/wix/Detox/blob/master/docs/APIRef.LaunchArgs.md).
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -18,14 +18,14 @@ declare global {
     const detoxCircus: Detox.DetoxCircus;
 
     namespace NodeJS {
-      interface Global {
-        device: Detox.DetoxExportWrapper['device'];
-        element: Detox.DetoxExportWrapper['element'];
-        waitFor: Detox.DetoxExportWrapper['waitFor'];
-        expect: Detox.DetoxExportWrapper['expect'];
-        by: Detox.DetoxExportWrapper['by'];
-        detoxCircus: Detox.DetoxCircus;
-      }
+        interface Global {
+            device: Detox.DetoxExportWrapper['device'];
+            element: Detox.DetoxExportWrapper['element'];
+            waitFor: Detox.DetoxExportWrapper['waitFor'];
+            expect: Detox.DetoxExportWrapper['expect'];
+            by: Detox.DetoxExportWrapper['by'];
+            detoxCircus: Detox.DetoxCircus;
+        }
     }
 
     namespace Detox {
@@ -48,6 +48,9 @@ declare global {
             artifacts?: DetoxArtifactsConfig;
             behavior?: DetoxBehaviorConfig;
             session?: DetoxSessionConfig;
+            apps?: Record<string, DetoxAppConfig>;
+            devices?: Record<string, DetoxDeviceConfig>;
+            selectedConfiguration?: string;
             configurations: Record<string, DetoxConfiguration>;
         }
 
@@ -98,9 +101,19 @@ declare global {
             sessionId?: string;
         }
 
+        type DetoxAppConfig = (DetoxIosAppConfig | DetoxAndroidAppConfig) & {
+            /**
+             * App name to use with device.selectApp(appName) calls.
+             * Can be omitted if you have a single app under the test.
+             *
+             * @see Device#selectApp
+             */
+            name?: string;
+        };
+
         type DetoxDeviceConfig = DetoxBuiltInDeviceConfig | DetoxCustomDriverConfig;
 
-        type DetoxConfiguration = DetoxPlainConfiguration;
+        type DetoxConfiguration = DetoxPlainConfiguration | DetoxAliasedConfiguration;
 
         interface DetoxLogArtifactsPluginConfig {
             enabled?: boolean;
@@ -145,16 +158,28 @@ declare global {
             enabled?: boolean;
         }
 
+        interface DetoxIosAppConfig extends DetoxLooseIosAppConfig {
+            type: 'ios.app';
+        }
+
         interface DetoxLooseIosAppConfig {
             binaryPath: string;
+            bundleId?: string;
             build?: string;
+            launchArgs?: Record<string, string | null | undefined>;
+        }
+
+        interface DetoxAndroidAppConfig extends DetoxLooseAndroidAppConfig {
+            type: 'android.apk';
         }
 
         interface DetoxLooseAndroidAppConfig {
             binaryPath: string;
+            bundleId?: string;
             build?: string;
             testBinaryPath?: string;
             utilBinaryPaths?: string[];
+            launchArgs?: Record<string, string | null | undefined>;
         }
 
         type DetoxBuiltInDeviceConfig =
@@ -172,6 +197,10 @@ declare global {
           | (DetoxGenymotionCloudDriverConfig & DetoxLooseAndroidAppConfig)
           | (DetoxCustomDriverConfig)
           );
+
+        type DetoxAliasedConfiguration =
+          | DetoxAliasedConfigurationSingleApp
+          | DetoxAliasedConfigurationMultiApps;
 
         interface DetoxIosSimulatorDriverConfig {
             type: 'ios.simulator';
@@ -196,11 +225,12 @@ declare global {
 
         interface DetoxGenymotionCloudDriverConfig {
             type: 'android.genycloud';
-            device: { recipeUUID: string; } | { recipeName: string; };
+            device: string | { recipeUUID: string; } | { recipeName: string; };
         }
 
         interface DetoxCustomDriverConfig {
             type: string;
+
             [prop: string]: unknown;
         }
 
@@ -218,6 +248,20 @@ declare global {
             behavior?: DetoxBehaviorConfig;
             session?: DetoxSessionConfig;
         };
+
+        interface DetoxAliasedConfigurationSingleApp {
+            type?: never;
+            device: DetoxAliasedDevice;
+            app: string | DetoxAppConfig;
+        }
+
+        interface DetoxAliasedConfigurationMultiApps {
+            type?: never;
+            device: DetoxAliasedDevice;
+            apps: string[];
+        }
+
+        type DetoxAliasedDevice = string | DetoxDeviceConfig;
 
         // endregion DetoxConfig
 
@@ -241,8 +285,11 @@ declare global {
              * });
              */
             init(configOverride?: Partial<DetoxConfig>, options?: DetoxInitOptions): Promise<void>;
+
             beforeEach(...args: any[]): Promise<void>;
+
             afterEach(...args: any[]): Promise<void>;
+
             /**
              * The cleanup phase should happen after all the tests have finished.
              * This is the phase where the Detox server shuts down.
@@ -336,6 +383,35 @@ declare global {
              * The value will be undefined until the device is properly prepared (i.e. in detox.init()).
              */
             name: string;
+
+            /**
+             * Select the current app (relevant only to multi-app configs) by its name.
+             * After execution, all app-specific device methods will target the selected app.
+             *
+             * @see DetoxAppConfig#name
+             * @example
+             * await device.selectApp('passenger');
+             * await device.launchApp(); // passenger
+             * // ... run tests for the passenger app
+             * await device.uninstallApp(); // passenger
+             * await device.selectApp('driver');
+             * await device.installApp(); // driver
+             * await device.launchApp(); // driver
+             * // ... run tests for the driver app
+             * await device.terminateApp(); // driver
+             */
+            selectApp(app: string): Promise<void>;
+
+            /**
+             * @protected
+             * Unselect the current app.
+             * After execution, all app-specific device methods will be throwing an error.
+             *
+             * @example
+             * await device.selectApp(null); // explicitly unselects any app
+             */
+             selectApp(app: null): Promise<void>;
+
             /**
              * Launch the app.
              *
@@ -358,6 +434,7 @@ declare global {
              * });
              */
             launchApp(config?: DeviceLaunchAppConfig): Promise<void>;
+
             /**
              * Access the user-defined launch-arguments predefined through static scopes such as the Detox configuration file and
              * command-line arguments. This access allows - through dedicated methods, for both value-querying and
@@ -394,6 +471,7 @@ declare global {
              * await device.terminateApp('other.bundle.id');
              */
             terminateApp(bundle?: string): Promise<void>;
+
             /**
              * Send application to background by bringing com.apple.springboard to the foreground.
              * Combining sendToHome() with launchApp({newInstance: false}) will simulate app coming back from background.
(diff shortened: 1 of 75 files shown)


detox/src/artifacts/screenshot/ScreenshotArtifactPlugin.js
```diff
@@ -17,13 +17,13 @@ class ScreenshotArtifactPlugin extends TwoSnapshotsPerTestPlugin {
     return this.api.preparePathForArtifact(`${name}.png`, testSummary);
   }
 
-  async onBeforeCleanup(e) {
+  async onBeforeCleanup() {
```
d4vidi, line 20: Hmm... okay?

detox/test/detox.config.js
```diff
@@ -0,0 +1,150 @@
+/** @type {DetoxConfig} */
+module.exports = {
```
d4vidi, line 2: Shouldn't we remove this from `package.json`, in that case?

d4vidi (approved):
See some minor comments 🚀

detox/src/artifacts/screenshot/ScreenshotArtifactPlugin.js
noomorph: Totally :) there is no argument there at all. Noticed thanks to TypeScript, actually.

detox/test/detox.config.js
noomorph: Yep, forgot. 😛

d4vidi (approved):

d4vidi:
@noomorph, approved, but note the main test suite is broken on CI.

d4vidi:
Heads up, closing and reopening to verify a subtle issue with jenkins. Not related to this PR in any way.

detox/index.d.ts
```diff
@@ -145,16 +158,28 @@ declare global {
...
+        }
+
         interface DetoxLooseIosAppConfig {
             binaryPath: string;
+            bundleId?: string;
             build?: string;
+            launchArgs?: Record<string, string | null | undefined>;
+        }
+
+        interface DetoxAndroidAppConfig extends DetoxLooseIosAppConfig {
```
d4vidi, line 172:
```suggestion
        interface DetoxAndroidAppConfig extends DetoxLooseAndroidAppConfig {
```

detox/index.d.ts
```diff
@@ -145,16 +158,28 @@ declare global {
...
+            type: 'android.apk';
         }
 
         interface DetoxLooseAndroidAppConfig {
```
d4vidi, line 176: I understand where the loose concept is coming from but since they are never directly referenced - only the concrete types do, then maybe it'd be best to simplify by unifying everything onto the concrete types and that's it?

Or alternatively, create one "loose" config that includes 'em properties that are relevant to both (i.e. all) platforms (e.g `launchArgs`)? And then reconsider the name loose, and try to go with conventions such as "abstract" or just "base"?... 🤔

detox/index.d.ts
```diff
@@ -173,6 +198,10 @@ declare global {
           | (DetoxCustomDriverConfig)
           );
 
+        type DetoxAliasedConfiguration =
```
d4vidi, line 201: (nitpicking-alert) move this lower to reside with the actual config flavors?

detox/index.d.ts
```diff
@@ -287,6 +334,35 @@ declare global {
...
+
+            /**
+             * @protected
+             * Unselect the current app.
+             * After execution, all app-specific device methods will be throwing an error.
+             *
+             * @example
+             * await device.selectApp(null); // explicitly unselects any app
+             */
+             selectApp(app: null): Promise<void>;
```
d4vidi, line 364: I'm kind-of picking my brain trying to think whether this is really needed, because none of the API's are really app-specific (or are they?)

detox/index.d.ts
noomorph: Mmm, I can see your point - let me check one more time.

detox/index.d.ts
noomorph: This is a protected method - I use it internally with multiple apps config to unselect the default app so you always will be getting an error unless you explicitly select your app after Detox init.

Do you think it is better to remove it at all? I don't have a strong opinion here, so pls let me know.

d4vidi: Oh, it _is_ `@protected` indeed :)
Your call, then!

But to be sure I understand correctly: eventually, this is so as to protect users from forgetting to select an app when multiple apps have been configured in the selected configuration?

noomorph: @d4vidi, exactly for that purpose — we can't decide what application should be default there.

detox/local-cli/build.js
```diff
@@ -17,18 +18,35 @@ module.exports.builder = {
...
+    }
 
-  if (buildScript) {
-    log.warn("\n\nImportant: 'detox build' is a convenience shortcut for calling your own build command, as provided in the config file.\nFailures in this build command are not the responsibility of Detox. You are responsible for maintaining this command.\n");
-    log.info(buildScript);
-    cp.execSync(buildScript, { stdio: 'inherit' });
-  } else {
-    throw errorBuilder.missingBuildScript();
+    if (app.binaryPath && !fs.existsSync(app.binaryPath)) {
+      log.warn("\nImportant: after running the build command, Detox could not find your app at the given binary path:\n\t" + app.binaryPath + "\nMake sure it is correct, otherwise you'll get an error on an attempt to install your app.\n");
```
d4vidi, line 49: This alone yields 💯^2 points in dev ux 🔥

detox/src/utils/environment.test.js
```diff
@@ -268,7 +268,7 @@ describe('Environment', () => {
         const executable = 'gmsaas';
         process.env.PATH = path.join(tempSdkPath, 'somewhere');
         await genExec(`somewhere/${executable}`);
-        expect(Environment.getGmsaasPath()).toMatch(new RegExp(`${tempSdkPath}.+somewhere.${executable}`));
+        expect(Environment.getGmsaasPath()).toMatch(path.join(tempSdkPath, `somewhere/${executable}`));
```
d4vidi, line 271: ❤️

detox/test/detox.config.js
```diff
@@ -32,61 +34,111 @@ const config = {
       uiHierarchy: "enabled"
     }
   },
-  configurations: {
-    "ios.sim.debug": {
+
+  apps: {
```
d4vidi, line 38: So clean... 'love it!

detox/test/detox.config.js
```diff
@@ -32,61 +34,111 @@ const config = {
...
+    "ios.debug": {
+      type: "ios.app",
```
d4vidi, line 40: can it be just `ios`, maybe? (and `android` instead of `android.apk`?)

detox/test/detox.config.js
```diff
@@ -97,57 +149,41 @@ const config = {
...
-      utilBinaryPaths: [
-        "./cache/test-butler-app.apk"
-      ],
-      build: "cd android && ./gradlew assembleFromBinRelease assembleFromBinReleaseAndroidTest -DtestBuildType=release && cd ..",
       type: "android.emulator",
       device: {
         avdName: "Pixel_API_28"
       },
+      build: "cd android && ./gradlew assembleFromBinRelease assembleFromBinReleaseAndroidTest -DtestBuildType=release && cd ..",
+      binaryPath: "android/app/build/outputs/apk/fromBin/release/app-fromBin-release.apk",
```
d4vidi, line 168: 'this here on purpose, in order to test config overrides?

detox/src/devices/LaunchArgs.js
```diff
@@ -6,14 +6,15 @@ class LaunchArgs {
...
+    for (const name of Object.keys(launchArgs)) {
+      this._setLaunchArg(name, launchArgs[name]);
+    }
   }
 
   reset() {
-    this._args = {};
+    for (const name of Object.keys(this._args)) {
+      this._setLaunchArg(name, undefined);
+    }
```
d4vidi, line 17: and also... why? 😢

detox/src/devices/LaunchArgs.js
```diff
@@ -6,14 +6,15 @@ class LaunchArgs {
   }
 
   modify(launchArgs) {
-    Object
-      .keys(launchArgs)
-      .forEach((name) =>
-        this._setLaunchArg(name, launchArgs[name]));
+    for (const name of Object.keys(launchArgs)) {
+      this._setLaunchArg(name, launchArgs[name]);
+    }
```
d4vidi, line 11: but... but why?

detox/src/devices/Device.js
```diff
@@ -5,119 +5,95 @@ const { traceCall } = require('../utils/trace');
...
-  }
-
-  get appLaunchArgs() {
-    return this._launchArgs;
-  }
 
-  get id() {
-    return this._deviceId;
+    this._currentApp = appConfig;
+    this._cachedAppLaunchArgs = null;
```
d4vidi, line 81: (Rubber-duck'ing) So this will work but would omit any previous changes made in the context of the selected app. For example:
```js
device.selectApp('app1');
device.appLaunchArgs.modify(...);
device.selectApp('app2'); // changes to app1's args are gone...
device.selectApp('app1'); // ... so here we start from scratch
```

That's ok under the assumption of that `selectApp()` is meant to be called once per test. Was that the (implicit) assumption here?

detox/src/devices/Device.js
```diff
@@ -5,119 +5,95 @@ const { traceCall } = require('../utils/trace');
...
-  get type() {
-    return this._deviceConfig.type;
+  /**
+   * @deprecated
+   */
+  async relaunchApp(params = {}, bundleId) {
+    if (params.newInstance === undefined) {
+      params['newInstance'] = true;
+    }
+    await this.launchApp(params, bundleId);
```
d4vidi, line 96: Sign this deprecation up for Detox 19 😄

d4vidi (commented):
Still a WIP, but this is what I have so far. Good work, overall! I hope to be able to share more notes soon.

detox/src/configuration/index.test.js
```diff
@@ -39,7 +39,7 @@ describe('composeDetoxConfig', () => {
         argv: {
```
d4vidi, line 39: As a general note (not directly related to the PR), this test is a bit too integrative to my taste. With the various configuration domain each broken down to its own composer (session, runner, apps, etc) along with the awesome unit testing coverage -- I feel a total mocking of each with some minimal glue-logic testing here should suffice.

I know the approach is arguable, no need to accept it. In any case, no need to address this right now.

detox/src/devices/Device.js
```diff
@@ -5,119 +5,95 @@ const { traceCall } = require('../utils/trace');
...
+    return this._deviceConfig.type;
+  }
 
-    if (params.delete) {
-      await this._terminateApp();
-      await this._reinstallApp();
-    } else if (newInstance) {
-      await this._terminateApp();
+  get appLaunchArgs() {
+    if (!this._cachedAppLaunchArgs) {
```
d4vidi, line 46: (as discussed 1x1) Consider revisiting the implicitness of this approach.
Maybe as a tradeoff, a mere rename (e.g `_currentAppArgs`) might help make it more understandable.
Or to solve altogether - what if a LaunchArgs object is created per each app, to replace the plain config? Just a suggestion.

d4vidi (commented):
Some more notes after going over the changes in Device and the various composers.
I couldn't go very deep with those I'm afraid. I trust that real-life feedback would be more beneficial, anyways!

detox/src/devices/Device.js
noomorph: > but... but why?
> and also... why? 😢

Exactly to avoid this scenario you demonstrated.

detox/src/devices/Device.js
noomorph: AFAIR, we spoke with Leo about this — and looks like we might refrain from deprecating it at all.

detox/test/detox.config.js
noomorph: Yeah — but maybe it is too weird to keep it this way.

detox/src/Detox.js
```diff
@@ -195,31 +196,34 @@ class Detox {
...
   }
 
+  async _reinstallAppsOnDevice() {
+    const appNames = Object.keys(this._appsConfig);
+
+    for (const appName of appNames) {
+      await this.device.selectApp(appName)
+      await this.device.uninstallApp();
+      await this.device.installApp();
+      await this.device.installUtilBinaries();
```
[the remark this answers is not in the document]
noomorph: Fixed according to your comments - see https://github.com/wix/Detox/pull/2587/commits/4ccc073a2302d281ddeb1788d5ef7f8281137b89

detox/src/Detox.test.js
```diff
@@ -159,13 +162,14 @@ describe('Detox', () => {
       it('should prepare the device', () =>
         expect(device().prepare).toHaveBeenCalled());
 
-      it('should reinstall the app', () => {
+      it('should select and reinstall the app', () => {
+        expect(device().selectApp).toHaveBeenCalledWith('default');
         expect(device().uninstallApp).toHaveBeenCalled();
         expect(device().installApp).toHaveBeenCalled();
```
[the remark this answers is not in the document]
noomorph: I've deleted `installUtilBinaries` assertion there, but not sure what did you mean exactly?

detox/src/Detox.js
```diff
@@ -195,31 +196,34 @@ class Detox {
     }
 
     if (handle.status === Deferred.PENDING) {
-      handle.reject(
-        new DetoxRuntimeError({
-          message: 'Aborted detox.init() execution, and now running detox.cleanup()',
-          hint: 'Most likely, your test runner is tearing down the suite due to the timeout error'
-        })
-      );
+      handle.reject(this._runtimeErrorBuilder.abortedDetoxInit());
```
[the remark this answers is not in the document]
noomorph: thanks. 🤘

detox/src/utils/environment.test.js
noomorph: Better works for Windows as well.

detox/test/detox.config.js
noomorph: Pleased to ~hear~ read 😄

detox/src/devices/LaunchArgs.js
noomorph: answered below. looks like it is going to be a temporary workaround

detox/src/devices/LaunchArgs.js
noomorph: answered below. looks like it is going to be a temporary workaround

detox/src/devices/Device.js
noomorph: Anyway, we've discussed the motivation in 1:1 mode — although, good to replace that mutation-reliant approach with some more sturdy.

detox/src/devices/Device.js
noomorph: I'll add a TODO after the merge

detox/src/configuration/index.test.js
noomorph: Let's add a TODO to discuss this later.

PR #2600, opened 2021-01-18
Drag and drop

### LINKED ISSUE #201: Add support for drag & drop
I want to move element1 to element2, is this possible ?

alon-ha:
**Description:**

Closes #201. Integrate drag and drop for iOS


detox/ios/Detox/Invocation/Element.swift
```diff
@@ -124,6 +124,10 @@ class Element : NSObject {
 		view.dtx_longPress(at: point, duration: duration)
 	}
 	
+	func longPress(at normalizedPoint: CGPoint, duration: TimeInterval, dragToElement targetElement: Element, normalizedTargetPoint: CGPoint, speed: CGFloat, holdForDuration lastHoldDuration: TimeInterval) {
+		view.dtx_longPress(at:normalizedPoint, duration:duration, target:targetElement, normalizedTargetPoint:normalizedTargetPoint, speed:speed, lastHoldDuration:lastHoldDuration)
```
LeoNatan, line 128: > target:targetElement

I don't think this is correct. You need to send `targetElement.view`.

alon-ha: Correct, you are right.

detox/ios/Detox/Actions/NSObject+DetoxActions.h
```diff
@@ -18,6 +18,8 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)dtx_longPressAtAccessibilityActivationPoint;
 - (void)dtx_longPressAtAccessibilityActivationPointForDuration:(NSTimeInterval)duration;
 - (void)dtx_longPressAtPoint:(CGPoint)point duration:(NSTimeInterval)duration NS_SWIFT_NAME(dtx_longPress(at:duration:));
+- (void)dtx_longPressAtPoint:(CGPoint)normalizedPoint duration:(NSTimeInterval)duration thenDragToElement:(NSObject*)target normalizedTargetPoint:(CGPoint)normalizedTargetPoint withSpeed:(CGFloat)speed thenHoldForDuration:(NSTimeInterval)lastHoldDuration
```
d4vidi, line 21: Is this the top-level native-side API? Can it be changed to something that conveys there will be a drag&drop done? I think the initial long-press should be a side-effect of a function with a more descriptive name, not the other way around... wdyt? @alon-ha

detox/ios/Detox/Actions/NSObject+DetoxActions.m
```diff
@@ -254,6 +309,58 @@ - (void)dtx_pinchWithScale:(CGFloat)scale velocity:(CGFloat)velocity angle:(CGFl
 	_DTXApplyPinch(window, startPoint1, endPoint1, startPoint2, endPoint2, 1.0 / velocity);
 }
 
+- (void)dtx_dragAndDrop:(NSObject *)targetElement targetPositionOffset:(CGPoint)positionOffset initialHoldDuration:(NSTimeInterval)initialHoldDuration speed:(CGFloat)speed
```
d4vidi, line 312: How come this never gets called? Is it because this is just a draft?

detox/ios/Detox/Actions/NSObject+DetoxActions.h
LeoNatan: The naming convention is similar to have Apple names their method in XCUITest.

LeoNatan: The native methods will be named in a similar fashion to XCUITest so that we can have interchangeable driver eventually.

detox/ios/Detox/Actions/NSObject+DetoxActions.m
alon-ha: Didn't remove it yet, still doing changes. Will be removed next commit

detox/ios/Detox/Actions/NSObject+DetoxActions.h
alon-ha: As @LeoNatan said we decided to go with a convention similar to Apple API
The call from the JS side will be something like this:
`await element(by.id('elementToDrag')).longPressAndDrag(.....)`

LeoNatan (changes requested):

detox/ios/Detox/Actions/NSObject+DetoxActions.m
```diff
@@ -71,7 +71,62 @@ - (void)dtx_longPressAtPoint:(CGPoint)point duration:(NSTimeInterval)duration
...
+	CGPoint calcNormalizedPoint = DTXCalcNormalizedPoint(normalizedPoint, self);
+	CGPoint calcNormalizedTargetPoint = DTXCalcNormalizedPoint(normalizedTargetPoint, target);
+	
+	[self.dtx_view dtx_assertHittableAtPoint:[self.dtx_view.coordinateSpace convertPoint:calcNormalizedPoint fromCoordinateSpace:self.dtx_view.window.screen.coordinateSpace]];
+	[target.dtx_view dtx_assertHittableAtPoint:[target.dtx_view.coordinateSpace convertPoint:calcNormalizedTargetPoint fromCoordinateSpace:target.dtx_view.window.screen.coordinateSpace]];
+	
+	// Converting end point to the window coordinate space of the view we are going to drag
+	// Setting the startPoint for better code readbility
+	CGPoint startPoint = calcNormalizedPoint;
+	CGPoint endPoint = [self.dtx_view.window.coordinateSpace convertPoint:calcNormalizedTargetPoint fromCoordinateSpace:target.dtx_view.window.screen.coordinateSpace];
```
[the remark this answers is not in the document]
alon-ha: That's exactly why I do this convention, we have the element that we going to drag and the target one. 
If those two in separate windows, I want to convert the position of the target one to the window of the element I'm going to drag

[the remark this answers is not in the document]
LeoNatan: But you are converting incorrectly. You are taking screen coordinates and converting to source window coordinates, but you should be taking destination window coordinates and converting to source window coordinates.

detox/ios/Detox/Invocation/Action.swift
```diff
@@ -213,14 +213,14 @@ class LongPressAction : Action {
 		
 		let normalizedTargetingPoint = getNormalizedPoint(xPosition: parameters[3], yPosition: parameters[4])
 		
-		var speed = CGFloat(1.0)
+		var speed = CGFloat(0.5)
```
LeoNatan, line 216: Why? It's too slow.

alon-ha: It was actually extremely fast in the simulator, with this speed it looks much better

detox/src/ios/expectTwo.js
```diff
@@ -126,6 +126,24 @@ class Element {
     return this.withAction('longPress', duration);
   }
 
+  longPressAndDrag(duration, normalizedPositionX, normalizedPositionY, targetElement,
+            normalizedTargetPositionX, normalizedTargetPositionY, speed = 'fast', holdDuration = 1000) {
```
LeoNatan, line 130: I will change this to have the `normalizedTargetPositionX, normalizedTargetPositionY` as optional, because usually people will want to have `NaN` there anyway.

LeoNatan (approved):
This looks good! Once CI has been fixed and is green, this can be merged.

detox/src/ios/expectTwo.js
alon-ha: Sure

PR #2613, opened 2021-01-24
Introduce device.setLaunchArg()

d4vidi:
## Description

I've decided to take a step forward and reduce the added complexity in e2e's of large-scale apps (as Wix' internal ones) in trying to prepare and set up suitable app launch arguments.

The typical use case is of modules settings up mock servers, in particular.
Some mock servers are provided by frameworks, while some by modules/apps, which are separate - yet all must be aggregated into the same call to `launchApp({ launchArgs: { ... } })`. This API allows the separation to be properly retained, and eliminate hacks such as `launchApp` swizzling, currently used to overcome this.

_This is in a way both a PR and a discussion, BTW. I've went ahead and implemented this ahead of time because there's very little production code involved._

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -19,6 +19,7 @@ class Device {
     this.deviceDriver = deviceDriver;
     this.deviceDriver.validateDeviceConfig(deviceConfig);
     this.debug = debug;
+    this._launchArgs = {};
   }
 
   async prepare() {
@@ -50,6 +51,7 @@ class Device {
 
     const baseLaunchArgs = {
       ...this._deviceConfig.launchArgs,
+      ...this._launchArgs,
       ...params.launchArgs,
     };
 
@@ -101,6 +103,10 @@ class Device {
     if(params.detoxUserActivityDataURL) {
       await this.deviceDriver.cleanupRandomDirectory(params.detoxUserActivityDataURL);
     }
+  }
+
+  setLaunchArg(key, value) {
+    this._launchArgs[key] = value;
   }
 
   get id() {
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -106,7 +106,15 @@ class Device {
   }
 
   setLaunchArg(key, value) {
-    this._launchArgs[key] = value;
+    if (value === undefined) {
+      this.clearLaunchArg(key);
+    } else {
+      this._launchArgs[key] = value;
+    }
+  }
+
+  clearLaunchArg(key) {
+    delete this._launchArgs[key];
   }
 
   get id() {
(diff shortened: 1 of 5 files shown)


detox/src/devices/Device.js
```diff
@@ -103,6 +105,10 @@ class Device {
     }
   }
 
+  setLaunchArg(key, value) {
+    this._launchArgs[key] = value;
```
noomorph, line 109: Can we validate against non-strings?
1. if string -> set
2. if `false | undefined | null` -> unset
3. if something else -> throw
?

So it fits with the rest of the logic....

LeoNatan:
This is good idea. Any reason why have the API in singles and not just allow the user to provide a set of arguments?

noomorph:
Also, I'd suggest renaming the method to something like `setDefaultAppLaunchArgs` (or something a bit less verbose), since it is going to affect all apps, not only a selected app, I assume.

detox/src/devices/Device.js
d4vidi: TBH, I couldn't entirely make up my mind about this. Need to reiterate, because:
1. We accept all primitives (booleans, numbers apply as well), and also objects and arrays ([ref1](https://github.com/wix/Detox/blob/37a7e038d3510dee3dbc89061e76d02357a01751/detox/test/e2e/22.launch-args.test.js#L15) [ref2](https://github.com/wix/Detox/blob/37a7e038d3510dee3dbc89061e76d02357a01751/detox/test/e2e/22.launch-args.test.js#L30))
2. I agree regarding undefined & null. However I'm generally in favour of making things explicit, so - would it better to throw in a `clearLaunchArg()`?...

d4vidi:
@LeoNatan Since per each argument a key must be provided, handling a set of arguments may end up being a bit tedious in either the API (e.g. handling an array of args) or the implementation (e.g. iterating an object's set of keys). I don't mind adding it - technically, but I'm inclined to say it will never be used... I could be wrong of course.

@noomorph I agree regarding the name - it's not very descriptive on the one hand, but other options seemed a bit verbose to me as well. Not my strong suite I suppose, more suggestions will be welcomed 😄 
As for multiple-apps -- I admit not taking that into account! I don't think it should affect all apps - on the contrary, I think we ought to throw in an additional app-name argument, for the multi-apps case!

LeoNatan:
Isn't it just a set of key/values? I think it's OK as API.

d4vidi:
UPDATE: Added `device.clearLaunghArg()`. @noomorph regarding multi-apps: I think it'd be best to associated the args with the currently "selected" app. wdyt?

d4vidi:
@noomorph can this be merged? wdyt? See my last comment

PR #2624, opened 2021-02-03
Introduce launch-args modification API

d4vidi:
## Description
A rewrite of #2618, #2613, introducing app launch-args editing according to the following syntax:
```js
device.appLaunchArgs.modify({
  newArg: 'aValue',
  existingArgToOverride: 'newValue',  
  existingArgToDelete: undefined,
});
```

and there's also:
```js
device.appLaunchArgs.reset();
device.appLaunchArgs.get();
```

In the future, we expect to support multiple apps according to the following terms:
- In a multi-apps environment, launch-args will be managed per each, separately (following the equivalent logic of the global Detox config).
- Arguments will be available via the same API, in the form of `device.appLaunchArg('app-name')`.
- In case `app-name` isn't explicitly specified -- will modify the currently "selected" app.

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -286,7 +286,7 @@ declare global {
              * Modify the launch-arguments via a modifier object, according to the following logic:
              *  - Concrete modifier properties would either set anew or override the value of existing properties with the same name, with
              *    the specific value.
-             *  - Modifier properties set to either `undefined` or `null` would have the equivalent property cleared.
+             *  - Modifier properties set to either `undefined` or `null` would have the equivalent property deleted.
              *
              * @param modifier The modifier object.
              *
@@ -296,13 +296,13 @@ declare global {
              * //   mockServerPort: 1234,
              * //   mockServerCredentials: 'user@test.com:12345678',
              * // }
-             * device.modify({
+             * device.appLaunchArgs.modify({
              *   mockServerPort: 4321,
              *   mockServerCredentials: null,
              *   mockServerToken: 'abcdef',
              * };
              * await device.launchApp();
-             * // => launch-arguments become:
+             * // ==> launch-arguments become:
              * // {
              * //   mockServerPort: 4321,
              * //   mockServerToken: 'abcdef',
@@ -374,7 +374,7 @@ declare global {
              * //   mockServerPort: 4321,
              * //   mockServerToken: 'uvwxyz',
              * // }
-             * device.appLaunchArgs().modify({
+             * device.appLaunchArgs.modify({
              *   mockServerPort: 4321,
              *   mockServerToken: 'abcdef',
              * });
@@ -382,7 +382,7 @@ declare global {
              *
              * @see AppLaunchArgs
              */
-            appLaunchArgs(): AppLaunchArgs;
+            appLaunchArgs: AppLaunchArgs;
             /**
              * Terminate the app.
              *
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -274,6 +274,55 @@ declare global {
             reuse?: boolean;
         }
 
+        /**
+         * A construct allowing for the querying and modification of user arguments passed to an app upon launch by Detox.
+         *
+         * @see AppLaunchArgs#modify
+         * @see AppLaunchArgs#reset
+         * @see AppLaunchArgs#get
+         */
+        interface AppLaunchArgs {
+            /**
+             * Modify the launch-arguments via a modifier object, according to the following logic:
+             *  - Concrete modifier properties would either set anew or override the value of existing properties with the same name, with
+             *    the specific value.
+             *  - Modifier properties set to either `undefined` or `null` would have the equivalent property cleared.
+             *
+             * @param modifier The modifier object.
+             *
+             * @example
+             * // With current launch arguments set to:
+             * // {
+             * //   mockServerPort: 1234,
+             * //   mockServerCredentials: 'user@test.com:12345678',
+             * // }
+             * device.modify({
+             *   mockServerPort: 4321,
+             *   mockServerCredentials: null,
+             *   mockServerToken: 'abcdef',
+             * };
+             * await device.launchApp();
+             * // => launch-arguments become:
+             * // {
+             * //   mockServerPort: 4321,
+             * //   mockServerToken: 'abcdef',
+             * // }
+             */
+            modify(modifier: object): void;
+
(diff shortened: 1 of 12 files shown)


detox/index.d.ts
```diff
@@ -274,6 +274,55 @@ declare global {
...
+             *
+             * @param modifier The modifier object.
+             *
+             * @example
+             * // With current launch arguments set to:
+             * // {
+             * //   mockServerPort: 1234,
+             * //   mockServerCredentials: 'user@test.com:12345678',
+             * // }
+             * device.modify({
```
noomorph, line 299: Not `device`, but `device.appLaunchArgs`

detox/index.d.ts
```diff
@@ -274,6 +274,55 @@ declare global {
...
+             *   mockServerToken: 'abcdef',
+             * };
+             * await device.launchApp();
+             * // => launch-arguments become:
+             * // {
+             * //   mockServerPort: 4321,
+             * //   mockServerToken: 'abcdef',
+             * // }
+             */
+            modify(modifier: object): void;
```
noomorph, line 311: Maybe `assign`, to be in line with `Object.assign` ?

detox/index.d.ts
d4vidi: 💪🏻

detox/index.d.ts
d4vidi: Well my line of though is that since for the undefined/null case we effectively delete properties, assigning isn't 100% descriptive. We're effectively modifying, perhaps even editing 🤔

detox/index.d.ts
```diff
@@ -785,8 +866,8 @@ declare global {
              */
             delete?: boolean;
             /**
-             * Detox can start the app with additional launch arguments
-             * The added launchArgs will be passed through the launch command to the device and be accessible via [[NSProcessInfo processInfo] arguments]
+             * Arguments to pass-through into the app.
+             * Refer to the [dedicated guide](https://github.com/wix/Detox/blob/master/docs/APIRef.LaunchArgs.md) for complete details.
              */
             launchArgs?: any;
```
noomorph, line 872: @d4vidi, don't you find this weird? I don't remember we ever had such an API... 😮 🤔

detox/src/devices/Device.js
```diff
@@ -49,7 +50,7 @@ class Device {
     }
 
     const baseLaunchArgs = {
-      ...this._deviceConfig.launchArgs,
+      ...this._launchArgs.get(),
```
noomorph, line 53: @d4vidi , I am confused — so how I am supposed to suppress base launchArgs if passing `null` or `undefined` simply removes properties inside `this._launchArgs`?

detox/src/devices/LaunchArgs.js
```diff
@@ -0,0 +1,32 @@
...
+
+  modify(launchArgs) {
+    Object
+      .keys(launchArgs)
+      .forEach((name) =>
+        this._setLaunchArg(name, launchArgs[name]));
+  }
+
+  reset() {
+    this._args = {};
```
noomorph, line 16: Wait, wait — are we resetting to an empty object or to the original `launchArgs`? This looks like a discrepancy.

detox/src/devices/Device.js
noomorph: Why do we need an extra complexity here? Can't we just have it as a blank key-value map and merge at the very last stage all these three: `this._deviceConfig.launchArgs, this._launchArgs.get(), ...params.launchArgs`?

Is the goal to make sure users can dynamically read the launch args? Do we really need this much flexibility?

noomorph: Okay, now I see. You want to be able to reset all launchArgs forcibly. Makes sense.

detox/test/detox.config.js
```diff
@@ -99,7 +99,12 @@ module.exports = {
       type: "android.emulator",
       device: {
         avdName: "Pixel_API_28"
-      }
+      },
+      launchArgs: {
```
noomorph, line 103: This is JavaScript, so I'd advise reusing a single `launchArgs` object across all the configurations.

detox/test/e2e/22.launch-args.test.js
```diff
@@ -26,25 +26,55 @@ describe(':android: Launch arguments', () => {
...
     await device.launchApp({newInstance: true, launchArgs});
-
     await element(by.text('Launch Args')).tap();
 
     await assertLaunchArg('complex', JSON.stringify(launchArgs.complex));
     await assertLaunchArg('complexlist', JSON.stringify(launchArgs.complexlist));
   });
 
+  it('should allow for arguments modification', async () => {
+    device.appLaunchArgs.modify({
```
noomorph, line 48: Could we add an assertion regarding the original appLaunchArgs values? ... i.e.:

```js
if (!_.isEqual(device.appLaunchArgs.get(), { app: 'le', goo: 'gle', micro: 'soft' }) {
  fail('Please check detox.config.js - your configuration should have custom app launchArgs');
}
```

docs/APIRef.LaunchArgs.md
```diff
@@ -0,0 +1,43 @@
+# Launch Arguments
+
+Detox allows in its [`device` API](APIRef.DeviceObjectAPI.md) for the launching the app under test in the test device via an explicit call to `device.launchApp()`. Through various means, Detox enables specifying a set of user-defined arguments (key-value pairs) to be passed on to the app when launched, so as to make them available inside the app itself at runtime (both on the native side, and - if applicable, on the Javascript side).
```
noomorph, line 3: Great effort, really, but should I mention that the article is a bit too wordy?
We can revisit it later if we have the capacity, though.

noomorph: Nevertheless, kudos for writing this! 👍

detox/index.d.ts
noomorph: Okay.

detox/index.d.ts
d4vidi: hmmm wdym?

noomorph: Check the actual implementation of `Device.js` and find it.

noomorph: Oh, Github you're drunk. :( That was just 306 / 362 line jump. No worries, resolving.

noomorph (approved):

PR #2631, opened 2021-02-09
add configuration for the build command

a-c-sreedhar-reddy:
- [x ] This is a small change 

---


LeoNatan:
I've updated the document with a more appropriate content. Thank you for your contribution!

PR #2642, opened 2021-02-15
Clean up default log printing to prevent clutter

LeoNatan:
This will be the default output:
Good

Instead of:
Bad

### DIFF
--- a/detox/src/utils/customConsoleLogger.js
+++ b/detox/src/utils/customConsoleLogger.js
@@ -46,9 +46,9 @@ function overrideConsoleMethods(console, bunyanLogger) {
   if (!console.__detox_log__) {
     const log = bunyanLogger;
 
-    override(console, 'log', log.info.bind(log));
+    override(console, 'log', log.debug.bind(log));
     override(console, 'warn', log.warn.bind(log));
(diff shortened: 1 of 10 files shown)


detox/src/utils/customConsoleLogger.js
```diff
@@ -46,9 +46,9 @@ function overrideConsoleMethods(console, bunyanLogger) {
   if (!console.__detox_log__) {
     const log = bunyanLogger;
 
-    override(console, 'log', log.info.bind(log));
+    override(console, 'log', log.debug.bind(log));
```
noomorph, line 49: Not entirely sure if this is a legit change. If users want `console.log` they should get console.log.

noomorph: It is absolutely their problem in their code if this clutters things.

LeoNatan: OK, that makes sense. Will revert this.

LeoNatan: Ah cool, you reverted.

LeoNatan:
Let me encode the gif, and then merge.

PR #2646, opened 2021-02-17
pr2pr: terser web api

noomorph:
# This is a PR to PR

**Description:**

Proposes another syntax for the web actions and expecations:

```diff
-        await web.expect(
-            web.getWebView(by.id("webview")).element(web.by.cssSelector("#cssSel"), 2)
-            ).toExist();

+        await expect(
+            web(by.id("webview")).element(by.cssSelector("#cssSel"), 2)
+        ).toExist();
```

and

```diff
-        await web.expect(web.getWebView().element(web.by.id("btnSave"), 2)).toExist();
+        await web.element(by.htmlId("btnSave")).tap();
```

Further details to be done.


d4vidi:
I think it's much better that the `expect` becomes seamless. However, I'm not fond of how two very different type of matchers (e.g. id vs htmlid, type vs classType) all reside in the same level. It'd put too much strain on the users and is rather error-prone.

To return to my suggestion from this discussion of the PR itself, I think I may have figure out a way to improve it thanks to this PR. Please speak your opinion on this suggested API:
```js
await withWebView(({ expect, element, by }) => {
   await expect(element(by.id('testingPar')).toExist();
   
   const editable = await element(by.className('public-DraftEditor-content'));
   await editable.scrollToView();
});
```
With `expect`, `element` and `by` as arguments to the user's callback, they can enforce the actual API and significantly reduce the change of errors due to type checking.
This is the kotlin-ish way to achieve this, as I mentioned privately...

noomorph:
I've given it a thought, then one more thought, and my feedback is rather vague.

1. Separate and numerous `expect` realms don't feel like the right solution provided we still want to be compatible with `chai` and `jest` expectation system in the future. The `expect` should be single for all.
2. Provided you have a complex scenario with switching back and forth between the web view and native, the syntax will look clunky:

```js
await withWebView(({ element, by }) => {
   await element(by.id('notReceivedSms')).tap();
});
await expect(element(by.id('notification')).toHaveText('Please wait until SMS');
await withWebView(({ expect, element, by }) => {
   await expect(element(by.id('code')).toExist();
   await element(by.id('code')).type('0451');
   await element(by.id('confirm')).tap();
});
await expect(element(by.id('notification')).not.toExist();
```

compared to:

```js
await web.element(by.id('notReceivedSms')).tap();
await expect(element(by.id('notification')).toHaveText('Please wait until SMS');
await expect(web.element(by.id('code')).toExist();
await web.element(by.id('code')).type('0451');
await web.element(by.id('confirm')).tap();
await expect(element(by.id('notification')).not.toExist();
```

3. > It'd put too much strain on the users and is rather error-prone.

Again, after giving it a thought, I'd rather disagree. In the scenario when you use a non-existing matcher by mistake:

```
await withWebView(({ expect, element, by }) => {
   await expect(element(by.type('input')).toExist();
});
```

what will happen? If `by` object has no `type` property, the value will be resolved to `undefined`, meaning that `element(undefined)` will throw an error, but which one? Instead of telling `Hey, you are using a wrong matcher!` we'll be saying `Dude, no idea what happened, but it seems like you call element() without arguments`, which is far from the truth.

Secondly, we can do a bit better in regards of naming:

```js
by.html.id();
by.html.class();
by.html.name();
by.html.tag();
by.html.querySelector();
by.html.xpath();
by.html.href();
by.html.hrefPartial();
```

Terse enough, and hard to confuse.

PR #2651, opened 2021-02-20
Fix currentStatus logging on Android

noomorph:
- [x] This is a small change 

---

**Description:**

It seems that there was a quick'n'dirty reimplementation, and by accident nobody remembered that Android native implementation sends legacy format of `currentStatusResult`.

Hereby I return the historical justice, although, of course, the best solution is to eventually align the JSON formats for `currentStatusResult` for both iOS and Android.

Before:

Screenshot 2021-02-20 at 12 15 06

After:

image


d4vidi:
@noomorph thanks for this ❤️ I can't help but wonder, though, if it won't be easier to switch to the new format on the native side (esp. since it's basically a stub)...

noomorph:
@d4vidi , thanks for the idea — I've reimplemented it on Android.

PR #2654, opened 2021-02-23
fix: move device.typeText() to internals

noomorph:
- [x] This is a small change 

---

**Description:**

Address the concern of publicly available half-baked semi-private API.


d4vidi (commented):
I mean, if this were Java or Kotlin, I would just settle on defining `device.typeText()` as a package-private method. I think we should follow the same idea. This feels like too much. I'd rather have `device._typeText()` used internally.

PR #2669, opened 2021-03-03
code: rename by.html to by.web

noomorph:
- [x] This is a small change 

---

**Description:**

Per request of @OrenZak, I've renamed `by.html` matchers to `by.web`, and `by.html.selector` now is `by.web.cssSelector`.

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -693,15 +693,15 @@ declare global {
             /**
              * Collection of web matchers
              */
-            readonly html: ByHtmlFacade;
+            readonly web: ByWebFacade;
         }
 
-        interface ByHtmlFacade {
+        interface ByWebFacade {
             /**
              * Find an element on the DOM tree by its id
              * @param id
              * @example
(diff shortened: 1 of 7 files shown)


detox/index.d.ts
```diff
@@ -1030,29 +1030,29 @@ declare global {
...
+             * @example await expect(webview.element(by.web.id('UniqueId205'))).not.toExist();
              */
             not: this;
 
             /**
              * Expect the element content to have the `text` supplied
              * @param text expected to be on the element content
              * @example
-             * await expect(webview.element(by.htmlId('UniqueId205'))).toHaveText('ExactText');
+             * await expect(webview.element(by.web.id('UniqueId205'))).toHaveText('ExactText');
```
OrenZak, line 1041: `webview.element` should be changed also to `web.element` right ?

noomorph: not necessarily, if you say `const webview = web(by.id('myWebView));` 
but to avoid confusion — yeah, better to do.

PR #2681, opened 2021-03-10
fix(logger): consider also -l, the alias for --loglevel

noomorph:
- [x] This is a small change 

---

**Description:**

It appears that non-`detox test` commands do not really react to `-l` shorthand alias for `--loglevel`.

For example, `detox run-server` – it is confusing that you cannot start it in a standalone mode with `--loglevel trace` when required.

It is easy to show that on the example with INFO/WARN levels.

Before (incorrect, because WARN level is higher than INFO):

image

After (correct, the log level now allows only ERROR and WARN messages):

image

### DIFF
--- a/detox/src/utils/logger.js
+++ b/detox/src/utils/logger.js
@@ -73,7 +73,7 @@ function createPlainBunyanStream({ logPath, level }) {
 }
 
 function init() {
-  const levelFromArg = argparse.getArgValue('loglevel');
+  const levelFromArg = argparse.getArgValue('loglevel', 'l');
   const level = adaptLogLevelName(levelFromArg);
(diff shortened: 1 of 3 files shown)


detox/src/utils/logger.js
```diff
@@ -73,7 +73,7 @@ function createPlainBunyanStream({ logPath, level }) {
 }
 
 function init() {
-  const levelFromArg = argparse.getArgValue('loglevel');
```
d4vidi, line 76: @noomorph how come this was OK for detox test, but not for other subcommands?

noomorph: Because detox test wraps -l into DETOX_LOGLEVEL environment variable for Jest, and the argparse logic works with env vars too.

PR #2683, opened 2021-03-12
docs: note about testNamePattern on detox test

lupenq:
- [x] This is a small change 
---

**Description:**

I came across the fact that the DetoxCLI documentation does not indicate that you can pass the --testNamePattern flag directly. I decided to fix this point in the documentation

### DIFF
--- a/detox/local-cli/utils/testCommandArgs.js
+++ b/detox/local-cli/utils/testCommandArgs.js
@@ -109,6 +109,12 @@ module.exports = {
     number: true,
     default: 1,
   },
+  t: {
+    alias: 'testNamePattern',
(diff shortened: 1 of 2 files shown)


noomorph:
I won't work this way - `detox test` CLI passes through only unknown arguments.

In your pull request, you make `--testNamePattern` a known argument to Detox, and now it stops being passed to the test runner since you are expected to write now your own logic.

I don't think that the suggested solution fits the current state of things.
Maybe you could check if there is any `postscriptum` API in `yargs` so that we just can print some explanatory message.

I'll keep this pull request closed until you find another solution.

Thanks for looking into this.

PR #2688, opened 2021-03-14
Integrate genymotion SaaS into Detox' CI

d4vidi:
## Description

Integrate Genymotion cloud into Detox' CI.

Following this change, instead of running the full demo-app test suite over 3 local emulators, we will:
- Execute the entire suite using 5 (not 3) cloud instances.
- Execute ~5 suites using 3 local (Google) emulators.

Everything else stays the same.


detox/src/utils/exec.js
```diff
@@ -33,12 +32,12 @@ async function execWithRetriesAndLogs(bin, options = {}) {
       } else if (statusLogs.retrying) {
         _logRetrying(log, cmd, retryNumber, lastError);
       }
-      result = await traceCall(`cmd: ${cmd}`, () => exec(cmd, { timeout }));
+      result = await exec(cmd, { timeout });
```
d4vidi, line 35: @noomorph I had to remove `traceCall()` because this is now called from the global Jest context - where trace is not initialized (and thus crashes), and doesn't get collected.
We should think of a way to fix this. Possibly, we could initialize it in some fake manner from the global context, just so as to avoid crashing.

Issue #2655, opened 2021-02-23
Add fully-blown support for Genycloud emulators


d4vidi:
A follow-up on the work on Genymotion-cloud integration done in the context of a grand PoC in #2429.
While #2429 sums up the work required in order to start rolling out Genymotion cloud support internally at Wix, there is more to be done in case the integration proves scalable and robust - so as to make it a polished, generally used solution provided via Detox. Here is the list:

- [ ] [wip] Better tweaking of Detox' CI script, so as to support exec on both local and cloud emulators in each build.
- [ ] Documentation + scaffolding script revisit. **Important:** must cover the global Detox init and teardown. Important 2: Must cover mocha.js.
- [ ] [wip] Revisit some technical debts (PR feedback notes, possible filenames change, internal TODO) [#2573, #2575, #2577]
- [ ] Dependency Injection - [InversifyJS](https://github.com/inversify/InversifyJS) / [di4js](https://github.com/gedbac/di4js)
- [ ] [tbd] Allow for running debug mode w/ or w/o some sort of instance reuse
- [ ] [tbd] Add support for running on a specific cloud instance (mostly for local dev usage / debugging)
- [ ] Optimize post-testing idle usage time by introducing a worker-teardown hook into Jest, then use it to kill the associated instance (note: currently, all workers -- and thus cloud instances are kept alive until the very end of the suite).
- [ ] Internal usage/error monitoring system. this is a TBD.

nhoughto:
Would be great to support a way of hooking into geny motion device naming too, we want to create a fallback device cleanup process to cleanup devices whose CI job has completed, but the `Detox-<timestamp>-<jest index>` can't be related back to a CI job, so there is no way of knowing which device is for which job.

As we cant guarantee that the detox device cleanup process will _always_ fire a fallback is required.

d4vidi:
Hey @nhoughto, quite exciting to see interest in this feature so soon (hasn't yet officially launched...)
Detox does take care of cleaning up all used Genymotion-cloud instances. You have to issue a call to `detox.globalInit()` in Jest's global init script, and an equivalent `detox.globalCleanup()` call in Jest's global teardown.

When all of this is mature enough, we will properly document it (as mentioned in the issue' description 😄)

nhoughto:
Yeah I figured it wasn’t quite finished yet, still works quite well.

[...]

PR #2691, opened 2021-03-16
Passthrough or set user-agent to gmsaas

d4vidi:
## Description

As per request from Genymobile, adding `detox` as the (default) user-agent extra-data for requests sent out via Detox.
The actual value can be overridden by a user by setting `GMSAAS_USER_AGENT_EXTRA_DATA` themselves, before invoking detox.

Relates to #2655.


noomorph:
Won't it make sense to also include Detox version number and Node.js version as well as a user-agent?

d4vidi:
No - this won't be visible for us, just Genymobile. All they want is to be able to get a sense of usage through detox. Probably not much any time soon.

PR #2702, opened 2021-03-23
Adding UserTracking permission

ANTNNJT:
## Description
iOS 14 introduced a new user tracking permission. 
AppleSimulatorUtils added support for this permission on [v0.9](https://github.com/wix/AppleSimulatorUtils/releases/tag/0.9).
I've added the missing definition to enable the use of this permission to `DevicePermissions` and the related `type`


noomorph (approved):
Thanks.

PR #2703, opened 2021-03-25
Android debug sync

### LINKED ISSUE #1788: --debug-synchronization not working on Android
I'm trying to debug some extra waiting issues I'm facing on Android by executing: `detox test --debug-synchronization` or `detox test --debug-synchronization 20` and there's no extra logging happening. I've used this command a while ago while I was in Detox version 13.4 I think and it worked, now that I'm on the latest is no longer working.

**Environment:**
 - Detox: 14.7.1 first and 14.8.5 after upgrading

d4vidi:
Fixes #1788.

## Description

- This pull request fixes the issue described here: #1788

In this pull request, I have enabled usage of `debug-synchronization` for Android.
The main issue why this was not working - in spite of most of the groundwork already laid out, was that on the Detox native end, there was only one thread ever available for handling server requests. Since requests typically require responses, they are handled synchronously, effectively forcing a multi-thread environment in order to allow things to run in parallel (e.g. invocation and status-query).
Before this change, the status query was mostly handled only after in-flight invocations have completed, and thus executed when the environment is already idle. On the node side, you would only see output saying the app is currently idle - never busy.

---

> _For API changes:_


detox/android/detox/build.gradle
```diff
@@ -131,16 +135,16 @@ if (rootProject.hasProperty('isOfficialDetoxLib') ||
             junitPlatform {
                 filters {
                     engines {
-                        include 'spek2'
+                        include 'spek2', 'junit-vintage'
```
d4vidi, line 138: Apparently, non-Spek tests were not executed via `gradle`!
https://github.com/spekframework/spek/issues/232#issuecomment-610732158

detox/android/detox/src/full/java/com/wix/detox/Detox.java
```diff
@@ -11,6 +11,7 @@
 import com.wix.detox.config.DetoxConfig;
 import com.wix.detox.espresso.UiControllerSpy;
 
+import java.util.concurrent.ThreadPoolExecutor;
```
noomorph, line 14: Probably, a redundant import.

detox/android/detox/src/full/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -101,20 +101,27 @@ class QueryStatusActionHandler(
...
+
+    private fun formatResource(resource: IdlingResource): String =
+        if (resource is DescriptiveIdlingResource) {
+            resource.getDescription()
+        } else if (resource.javaClass.name.contains("LooperIdlingResource") && resource.name.contains("mqt_js")) {
+            "Javascript code execution"
+        } else if (resource.javaClass.name.contains("LooperIdlingResource") && resource.name.contains("mqt_native")) {
+            "Javascript code execution (native)"
+        } else {
+            "Resource ${resource.name} being busy"
```
noomorph, line 123: This is definitely something we needed !! 👍

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
```diff
@@ -24,42 +26,47 @@
...
-        Bundle arguments = InstrumentationRegistry.getArguments();
-        detoxServerUrl = arguments.getString(DETOX_SERVER_ARG_KEY);
-        detoxSessionId = arguments.getString(DETOX_SESSION_ID_ARG_KEY);
+        detoxServerUrl = launchArgs.getDetoxServerUrl();
+        detoxSessionId = launchArgs.getDetoxSessionId();
 
         if (detoxServerUrl == null || detoxSessionId == null) {
-            Log.i(LOG_TAG, "Missing arguments : detoxServer and/or detoxSession. Detox quits.");
-            stop();
+            Log.i(LOG_TAG, "Missing arguments: detoxServer and/or detoxSession. Detox quits.");
```
noomorph, line 69: Wouldn't it make sense to:
1) Log it as an error
2) Move it `void start()` implementation... ?

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
```diff
@@ -69,13 +76,17 @@ public String invoke(Throwable t) {
 
     void start() {
         if (detoxServerUrl != null && detoxSessionId != null) {
-            handler.post(new Runnable() {
```
noomorph, line 72: Was it the source of the problem? So that now I see we are using a thread factory?

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
```diff
@@ -85,12 +96,12 @@ public void run() {
...
     void stop() {
-        Log.i(LOG_TAG, "Stopping Detox.");
-        handler.postAtFrontOfQueue(new Runnable() {
+        Log.i(LOG_TAG, "Stopping Detox!");
+        executor.execute(new Runnable() {
             @Override
             public void run() {
                 if (stopping) return;
-                stopping = true;
+                stopping = true; // TODO is this needed?
```
noomorph, line 104: Not sure either, but better to be safe than sorry. 😆

noomorph: Also, the question is, if both jobs are busy (you have 2 max), does that mean that `stop()` will have to wait quite a bit? Just wondering...

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
```diff
@@ -24,42 +26,47 @@
...
 class DetoxManager implements WebSocketClient.ActionHandler {
 
-    private final static String LOG_TAG =  "DetoxManager";
-
-    private final static String DETOX_SERVER_ARG_KEY = "detoxServer";
-    private final static String DETOX_SESSION_ID_ARG_KEY = "detoxSessionId";
+    private static final String LOG_TAG =  "DetoxManager";
 
-    private final static Function1<Throwable, String> errorParseFn = new Function1<Throwable, String>() {
+    private static final Function1<Throwable, String> errorParseFn = new Function1<Throwable, String>() {
```
noomorph, line 31: Do we use it anywhere? I see that you have removed it below:

```diff
-    actionHandlers.put("invoke", new InvokeActionHandler(new MethodInvocation(), wsClient, errorParseFn));
+    actionHandlers.put("invoke", new InvokeActionHandler(new MethodInvocation(), wsClient));
```

detox/android/detox/src/full/java/com/wix/detox/LaunchArgs.java
```diff
@@ -29,6 +31,14 @@ String getUrlOverride() {
         return InstrumentationRegistry.getArguments().getString(DETOX_URL_OVERRIDE_ARG);
     }
 
+    String getDetoxServerUrl() {
```
noomorph, line 34: Hehe, makes a lot of sense :) Single responsibility FTW.

detox/android/detox/src/full/java/com/wix/detox/TestEngineFacade.kt
```diff
@@ -15,7 +14,9 @@ class TestEngineFacade {
         null
     }
     fun syncIdle() = UiAutomatorHelper.espressoSync() // TODO Check whether this can be replaced with #awaitIdle()
-    fun getBusyIdlingResources() = EspressoDetox.getBusyEspressoResources() as List<IdlingResource>
+    fun getBusyIdlingResources() = IRStatusInquirer.INSTANCE.getAllBusyResources()
```
noomorph, line 17: Mmm, noice! 😄  what's the difference, btw?

noomorph: Nevermind, sorry, now I see both classes are ours.

detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/AnimatedModuleIdlingResource.java
```diff
@@ -63,6 +64,12 @@ public String getName() {
         return AnimatedModuleIdlingResource.class.getName();
     }
 
+    @NotNull
+    @Override
+    public String getDescription() {
+        return "Animations running on screen";
```
noomorph, line 70: 👏

detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/DescriptiveIdlingResource.kt
```diff
@@ -0,0 +1,5 @@
+package com.wix.detox.reactnative.idlingresources
+
+interface DescriptiveIdlingResource {
```
noomorph, line 3: Just wondering if it makes sense to extend the `IdlingResource` in order not to write that twice, e.g.:

```diff
-public abstract class DetoxBaseIdlingResource implements IdlingResource, DescriptiveIdlingResource {
+public abstract class DetoxBaseIdlingResource implements DescriptiveIdlingResource {
```

detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
```diff
@@ -64,6 +66,12 @@ public String getName() {
         return NetworkIdlingResource.class.getName();
     }
 
+    @NotNull
+    @Override
+    public String getDescription() {
+        return "In-flight network activity";
```
noomorph, line 72: So, here, I assume we can do a bit better in the future, right? E.g., collect `runningCalls()` and compose a verbose message with pending URLs, huh?

detox/android/detox/src/testFull/java/com/wix/detox/espresso/registry/IRStatusInquirerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        val resourceIdle = anIdleResource()
+        val resourceBusy = aBusyResource()
+        givenIdlingResources(resourceIdle, resourceBusy)
+
+        val result = uut.getAllBusyResources()
+        assertThat(result).isEqualTo(arrayListOf(resourceBusy))
+    }
+
+    @Test
+    fun `should execute busy-resources inquiry on main thread`() {
```
noomorph, line 60: Wow, 🤯  👍

detox/local-cli/utils/testCommandArgs.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
       return Number.isNaN(+value) ? value : +value;
     },
     describe:
-      '[iOS Only] Customize how long an action/expectation can take to complete before Detox starts querying the app why it is busy. ' +
+      'Customize how long an action/expectation can take to complete before Detox starts querying the app why it is busy. ' +
```
noomorph, line 66: ❤️

noomorph:
Overall, looks stunning! 🤗  Can't wait when it lands. 🚀

d4vidi:
@noomorph Thanks for the kind review! ❤️ I will address the comments, for sure - but first, must inspect where exactly the errors are coming from. I already know where to start - just need the time-of-day.

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
d4vidi: Yes, but all in due time

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
d4vidi: We have >1 thread, to run things (anything) concurrently

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
d4vidi: No, it should be able to close things abruptly

detox/android/detox/src/full/java/com/wix/detox/LaunchArgs.java
d4vidi: Yes yes, very old debts

detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/DescriptiveIdlingResource.kt
d4vidi: Good thinking. Really, it's not a pluggable that adds a methods to whatever, rather a very specific addition to idling-resources 👍🏻

detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
d4vidi: Exactly. I think the gap isn't that big, even.

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
noomorph: Add a TODO, maybe?

detox/android/detox/src/full/java/com/wix/detox/DetoxManager.java
noomorph: @d4vidi, it is going to close things abruptly ONLY in case if `currentStatus` query execution is rapid. Otherwise, if your `invoke` is stuck, and then the app sends `currentStatus` for the clarification, your thread pool already is full (2 out of 2).

Probably, there's no way `currentStatus` can get stuck but just saying. 🤷

d4vidi: Oh I see what you mean. Yes, `currentStatus` should finish up promptly. Are you suggesting to increase the pool to 3? That can be something to consider indeed.

noomorph: Oh, I'm not entirely sure if `pool=3` configuration is a safer thing to introduce. Let's leave it as is, but I recommend adding a TODO in the code at least so it is highlighted as a potential issue.

d4vidi: On 2nd though, is it really bad if `stop` takes a while to complete?

noomorph: Well, the Node.js side can just kill the app in that case, indeed. So you are probably right.

detox/android/detox/src/full/java/com/wix/detox/DetoxActionHandlers.kt
```diff
@@ -89,37 +89,44 @@ class CleanupActionHandler(
...
-        if (busyResources.isEmpty()) {
-            status = "The app is idle."
-        } else {
-            status = "Busy idling resources:\n"
-            for (res in busyResources) {
-                status += "\t- ${res.name}\n"
-            }
+        data["status"] = "App synchronization debug: " +
+            if (busyResources.isEmpty()) {
+                "The app appears to be idle!"
```
d4vidi, line 107: @noomorph I'm thinking about omitting the output for this use case. Makes the log dirty, and I'm not sure the reason is good enough, as the information is not very valuable. wdyt?

noomorph: Might be not the best idea, although I understand the motivation. Look, if a request does not end in 10s, should not that be suspicious? Of course, for `waitFor` case it could be nice to check whether the app was busy or not while we have been waiting. But for the other requests, the unresponsiveness can only raise suspicions.

We might although put those messages under debug loglevel on JS side, what do you think?

detox/android/detox/src/full/java/com/wix/detox/adapters/server/DetoxServerAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+        private val actionsDispatcher: DetoxActionsDispatcher,
+        private val detoxServerInfo: DetoxServerInfo,
+        private val readyActionType: String,
+        private val terminationActionType: String)
+    : WebSocketClient.WSEventsHandler, OutboundServerAdapter {
+
+    private val wsClient = WebSocketClient(this)
+
+    fun connect() {
+        wsClient.connectToServer(detoxServerInfo.serverUrl, detoxServerInfo.sessionId)
```
noomorph, line 20: Should not we log that we are attempting to connect to the server here.?

detox/android/detox/src/full/java/com/wix/detox/adapters/server/DetoxServerInfo.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.wix.detox.LaunchArgs
+import com.wix.detox.common.DetoxErrors
+
+class DetoxServerInfo internal constructor(launchArgs: LaunchArgs = LaunchArgs()) {
+    val serverUrl: String? = launchArgs.detoxServerUrl
+    val sessionId: String? = launchArgs.detoxSessionId
+
+    init {
+        if (serverUrl == null || sessionId == null) {
+            throw DetoxErrors.DetoxRuntimeException("Detox-server connection info is missing - breaking bad! url=$serverUrl, sessionId=$sessionId")
```
noomorph, line 12: While we are already rewritting things here, let's sync with ios implementation. On ios, there is a default value - localhost: 8099 and sessionId=package-name.

detox/android/detox/src/full/java/com/wix/detox/adapters/server/WebSocketClient.java
```diff
@@ -54,13 +54,13 @@ public void connectToServer(String url, String sessionId) {
...
         final Request request = new Request.Builder().url(url).build();
-        this.websocket = client.newWebSocket(request, wsEventsHandler);
+        this.websocket = client.newWebSocket(request, wsEventListener);
 
         client.dispatcher().executorService().shutdown();
     }
 
     public void sendAction(String type, Map params, Long messageId) {
-        Log.i(LOG_TAG, "At sendAction");
+        Log.i(LOG_TAG, "Sending out action '" + type + "'");
```
noomorph, line 64: Do you mind if we print message if as well?

detox/android/detox/src/full/java/com/wix/detox/adapters/server/WebSocketClient.java
```diff
@@ -69,21 +69,19 @@ public void sendAction(String type, Map params, Long messageId) {
...
     private void receiveAction(String json) {
-        Log.i(LOG_TAG, "At receiveAction");
         try {
             final JSONObject object = new JSONObject(json);
             final String type = (String) object.get("type");
             final Object params = object.get("params");
             final long messageId = object.getLong("messageId");
 
-            Log.d(LOG_TAG, "Detox Action Received: " + type);
+            Log.d(LOG_TAG, "Received action '" + type + "' (params=" + params + ")");
```
noomorph, line 81: The same comment about message id

detox/src/android/espressoapi/UiDeviceProxy.js
```diff
@@ -16,7 +15,7 @@ class UiDeviceProxy {
           return async (...params) => {
             const call = target[prop](invoke.callDirectly(uiAutomator.uiDevice()), ...params);
             const invokeResult = await this.invocationManager.execute(call);
-            if (invokeResult && invokeResult && invokeResult.result) {
+            if (invokeResult && invokeResult.result) {
```
noomorph, line 18: 😁

detox/test/e2e/24.uidevice.test.js
```diff
@@ -1,25 +1,27 @@
...
-      await expect(element(by.text(text))).toBeVisible();
-    });
+  it(`should type in an element using pressKeyCode()`, async () => {
+    const text = "a1b2c3";
+    const textAsCodes = [29, 8, 30, 9, 31, 10];
+    const uiDevice = device.getUiDevice();
+
+    await element(by.text('Actions')).tap();
+    await element(by.id('UniqueId937')).tap();
+    await forEachSeries(textAsCodes, (keyCode) => uiDevice.pressKeyCode(keyCode));
```
noomorph, line 24: Nice lib, 🤘

detox/android/detox/src/main/java/com/wix/detox/common/UIThread.kt
```diff
@@ -0,0 +1,69 @@
+package com.wix.detox.common
+
+import android.os.Handler
+import android.os.Looper
+import java.util.concurrent.Callable
+import java.util.concurrent.FutureTask
+
+object UIThread {
```
noomorph, line 8: Could you elaborate about this class? I am curious.

d4vidi: Sure! The main thread is a first-class citizen on Android, and there is special interest in enqueuing tasks to run over it (in various flavours - queue back, queue front, sync/async...). There are many ways to achieve access to the main thread, but they're not all accessible in all contexts. So, the easiest solution is to unify this functionality onto one object that enables to do that anywhere.

detox/android/detox/src/full/java/com/wix/detox/adapters/server/DetoxServerInfo.kt
d4vidi: Is that so! Sure, an easy fix. Is there logic to handle a case when only either of the two is null?

detox/android/detox/src/full/java/com/wix/detox/adapters/server/DetoxServerAdapter.kt
d4vidi: Done

detox/android/detox/src/full/java/com/wix/detox/adapters/server/WebSocketClient.java
d4vidi: Done

detox/android/detox/src/full/java/com/wix/detox/adapters/server/WebSocketClient.java
d4vidi: Done

detox/src/android/espressoapi/UiDeviceProxy.js
d4vidi: Pretty funny, indeed :)

detox/test/e2e/24.uidevice.test.js
d4vidi: 🤷🏻

detox/android/detox/src/full/java/com/wix/detox/adapters/server/DetoxServerInfo.kt
d4vidi: Done

detox/android/detox/src/full/java/com/wix/detox/DetoxActionHandlers.kt
d4vidi: Okay, left it as-is for now. Presumably, users won't see much of it, as it is mostly shown when a high-frequency debug check is employed.

Issue #2550, opened 2020-12-23
Extendable-presets support


d4vidi:
## Describe

Properly configuring Detox is a nontrivial task with many layers. Even with automated inception scripts such as `detox init`, there are still many things to go over so as to tailor to a project/environment's needs. That leaves many potential points of failure, and makes the setup complicated.

As a solution, we should  introduce the ability to use pre-baked presets, which would hold most of the necessary configurations required for various environment flavours. These presets should be extendable in the sense that they could be used as configuration bases, allowing developers to simplify their configuration work a great deal -- by focusing strictly on the changes required in tailoring things for them. Potentially, presets can be part of standalone github repos. 

To conclude, ideally, the final outcome - which is feasible, should be like so:
With an imaginary _preset_ node module called `default-aws-preset`, the default configuration would look like so:
```json
"detox": {
  "preset": "default-aws-preset"
}
```

This also paves way to at least 2 potential improvements:
1. Allowing for a base detox configuration which would be the default for all users (even presented in the example apps).
2. Allowing for an easy configuration standardization across related projects (e.g. apps of the same organization).

noomorph:
Can we close this issue now?

d4vidi:
By all means. Of course, we can now also go ahead and create actual presets, to be used as skeleton configs =>
simplify `detox init`, maybe documentation (setup), etc...

PR #2707, opened 2021-03-29
feat(config): extends

noomorph:
## Description

- This pull request addresses the issue described here: #2550

Now you can use "extends" capability, similar to `.babelrc` or `tsconfig.json`:

```json
{
  "extends": "my-base-preset"
}
```


detox/src/configuration/loadExternalConfig.test.js
```diff
@@ -56,11 +57,39 @@ describe('loadExternalConfig', () => {
...
+      artifacts: {
+        rootDir: 'someRootDir',
+        plugins: {
+          log: "all",
+          screenshot: "all",
+          video: "all",
+        },
+      },
+    });
+  });
```
d4vidi, line 75: I'm not well acquainted with this code so that may be the source of confusion, but I was sort of expecting a test that would `expect` the actual (post-merged) result... perhaps even a snapshot test. Is this not the place? (not referring to e2e's of course)

detox/src/errors/__snapshots__/DetoxConfigErrorComposer.test.js.snap
```diff
@@ -997,9 +997,18 @@ exports[`DetoxConfigErrorComposer (from configuration/index) .failedToReadConfig
...
 `;
 
+exports[`DetoxConfigErrorComposer (from configuration/index) .noConfigurationAtGivenPath should create an error with the attempted "extends" path 1`] = `
+[DetoxConfigError: Failed to find the base Detox config specified in:
+{
+  "extends": "some-detox-preset"
+}
+
+HINT: Check your Detox config at path:
+package.json]
```
d4vidi, line 1007: Very well articulated.

detox/test/package.json
```diff
@@ -53,5 +53,8 @@
     "ancestorSeparator": " > ",
     "classNameTemplate": "E2E Results.{classname}",
     "titleTemplate": "{classname} > {title}"
+  },
+  "detox": {
+    "extends": "./e2e/detox.config"
```
d4vidi, line 58: Quite legit!

PR #2716, opened 2021-04-07
fix(artifacts): more aggressive interruptProcess()

noomorph:
## Description

Internal change. Addresses rare conditions on a busy CI agent with unstable video recording commands.

1. `interruptProcess` now accepts a schedule of signals. For example:

```js
await interruptProcess(myChildProcessPromise, {
  SIGINT: 0, // first try gently, with SIGINT
  SIGTERM: 1000, // if is running after 1s, send SIGTERM
  SIGKILL: 5000, // if it is still running after 5s, 💀 
});
```

2. Simplified code in `SimulatorLogRecording` (reusing 1).
3. Added a forceful exit in `SimulatorRecordVideoPlugin` — should unblock the flow of artifacts saving.

### DIFF
--- a/detox/src/artifacts/log/ios/SimulatorLogRecording.js
+++ b/detox/src/artifacts/log/ios/SimulatorLogRecording.js
@@ -56,6 +56,7 @@ class SimulatorLogRecording extends Artifact {
 
   async _tryInterruptProcessGracefully(process) {
     await exec.interruptProcess(process, {
+      SIGINT: 0,
       SIGTERM: this._config.delayBeforeSigterm,
     });
   }
--- a/detox/src/artifacts/log/ios/SimulatorLogRecording.js
+++ b/detox/src/artifacts/log/ios/SimulatorLogRecording.js
@@ -55,14 +55,9 @@ class SimulatorLogRecording extends Artifact {
   }
 
   async _tryInterruptProcessGracefully(process) {
-    const graceful = await Promise.race([
-      sleep(this._config.delayBeforeSigterm).then(() => false),
-      exec.interruptProcess(process, 'SIGINT').then(() => true),
-    ]);
-
-    if (!graceful) {
-      await exec.interruptProcess(process, 'SIGTERM');
-    }
+    await exec.interruptProcess(process, {
+      SIGTERM: this._config.delayBeforeSigterm,
+    });
   }
 
   async doSave(artifactPath) {
--- a/detox/src/utils/exec.js
+++ b/detox/src/utils/exec.js
@@ -124,15 +124,14 @@ function _composeCommand(bin, prefix, args) {
 }
 
 function spawnAndLog(command, flags, options) {
-  const trackingId = _operationCounter++;
   const cmd = _joinCommandAndFlags(command, flags);
-  const log = execLogger.child({ fn: 'spawnAndLog', cmd, trackingId });
 
   const result = spawn(command, flags, {stdio: ['ignore', 'pipe', 'pipe'], ...options});
   const { childProcess } = result;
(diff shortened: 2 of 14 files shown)


detox/src/utils/exec.js
```diff
@@ -164,22 +163,33 @@ function _joinCommandAndFlags(command, flags) {
   return result;
 }
 
-async function interruptProcess(childProcessPromise, signal = 'SIGINT') {
-  const log = execLogger.child({ fn: 'interruptProcess' });
+const DEFAULT_KILL_SCHEDULE = {
+  SIGINT: 0,
```
noomorph, line 167: @d4vidi , I am wondering whether it makes sense to add extra signals by default here? e.g. SIGTERM after 10s or something like that... 🤔

d4vidi:
@noomorph Did this fix the issue for @doronpr ?

noomorph:
Yes.

detox/src/utils/exec.js
```diff
@@ -164,22 +163,33 @@ function _joinCommandAndFlags(command, flags) {
...
 
-  log.debug({ event: 'EXEC_KILL', signal, process_pid: pid }, `sending ${signal} to [pid = ${pid}]: ${spawnargs}`);
-
-  childProcess.kill(signal);
-  await childProcessPromise.catch(e => {
-    /* istanbul ignore if */
-    if (e.exitCode != null) {
-      throw e;
-    }
+  const handles = _.mapValues({ ...DEFAULT_KILL_SCHEDULE, ...schedule }, (ms, signal) => {
```
d4vidi, line 176: I wonder whether a provided schedule should in fact override the default... wdyt?

detox/src/utils/exec.js
d4vidi: I don't think so. It doesn't feel like the common case. However, maybe a dedicated version of `interruptProcess` that uses a more exotic schedule can be predefined here, and then reused where needed? `interruptProcessFTW()`

detox/src/artifacts/log/ios/SimulatorLogRecording.js
```diff
@@ -55,14 +55,10 @@ class SimulatorLogRecording extends Artifact {
...
-      exec.interruptProcess(process, 'SIGINT').then(() => true),
-    ]);
-
-    if (!graceful) {
-      await exec.interruptProcess(process, 'SIGTERM');
-    }
+    await exec.interruptProcess(process, {
+      SIGINT: 0,
+      SIGTERM: this._config.delayBeforeSigterm,
+    });
```
d4vidi, line 61: So much better :)

detox/src/utils/exec.js
```diff
@@ -164,22 +163,33 @@ function _joinCommandAndFlags(command, flags) {
   return result;
 }
 
-async function interruptProcess(childProcessPromise, signal = 'SIGINT') {
-  const log = execLogger.child({ fn: 'interruptProcess' });
+const DEFAULT_KILL_SCHEDULE = {
+  SIGINT: 0,
+};
 
+async function interruptProcess(childProcessPromise, schedule) {
```
d4vidi, line 170: May I suggest - replacing `schedule` with `options`, having `schedule` a specific option inside? I couldn't intuitively and simply find it clear - in the places where this API is used, what the signals map meant.

d4vidi (commented):
OK so I'm cool with the technical details of this change. However, I'm not fully settled regarding why this is needed in the first place. In particular, why various and different signals are required in order to kill the simulator's recording. Wouldn't repeating SIGERM, for example, be enough? Call me naive, but I believe things should be pretty deterministic in the simulators world...

PR #2719, opened 2021-04-08
Fixed an issue with DetoxSync which affected react native UI queue management

alon-ha:
## Description
Important fix for a leak in DetoxSync when tracking the UI queue management


noomorph (approved):
Me approves. 😍

PR #2721, opened 2021-04-09
fix: stabilize stuck cleanup

noomorph:
## Description

1. Resolves the edge case, when the app does not respond to `cleanup` action at all, hence it breaks the entire Jest + Detox cleanup phase. Technically, I am adding a 5s graceful period before we reject the pending `cleanup` request and go further in our `detox.cleanup()` routine.

2. `client.cleanup()` now theoretically cannot throw an error now. That's important if we want to finish properly the outer  `detox.cleanup()`.

3. Also, here I fix an incorrect app PID saving (it is prescribed to run too late, only after the app responds to "isReady"). After the fix, I am able to verify the existing PID (=> app is running) even in cases when the app is not eager to respond it is ready. Not entirely related to this PR (although it was relevant in the first commit with the first implementation, then I changed my mind and we have a no-app-terminate-just-reject-inflightpromise solution), but it is certainly worth fixing.

### DIFF
--- a/detox/src/client/Client.js
+++ b/detox/src/client/Client.js
@@ -23,6 +23,7 @@ class Client {
     this._onBeforeAppCrash = this._onBeforeAppCrash.bind(this);
     this._onAppDisconnected = this._onAppDisconnected.bind(this);
     this._onUnhandledServerError = this._onUnhandledServerError.bind(this);
+    this._logError = this._logError.bind(this);
 
     this._sessionId = sessionId;
     this._slowInvocationTimeout = debugSynchronization;
@@ -73,23 +74,16 @@ class Client {
 
     try {
       if (this.isConnected) {
-        this._scheduleAppTermination();
-
-        try {
-          await this.sendAction(new actions.Cleanup(this._successfulTestRun));
-        } catch (e) {
-          log.error({ event: 'ERROR' }, DetoxRuntimeError.format(e));
-        }
+        await this.sendAction(new actions.Cleanup(this._successfulTestRun)).catch(this._logError);
 
         this._whenAppIsConnected = this._invalidState('while cleaning up')
         this._whenAppIsReady = this._whenAppIsConnected;
       }
     } finally {
-      this._unscheduleAppTermination();
-      delete this.terminateApp; // property injection
-
-      await this._asyncWebSocket.close();
+      await this._asyncWebSocket.close().catch(this._logError);
     }
+
+    delete this.terminateApp; // property injection
   }
 
   setEventCallback(event, callback) {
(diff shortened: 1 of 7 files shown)


detox/src/client/Client.js
```diff
@@ -73,16 +73,23 @@ class Client {
...
+        }
 
         this._whenAppIsConnected = this._invalidState('while cleaning up')
         this._whenAppIsReady = this._whenAppIsConnected;
       }
     } finally {
+      this._unscheduleAppTermination();
+      delete this.terminateApp; // property injection
+
       await this._asyncWebSocket.close();
```
noomorph, line 91: Probably, I should guard it here as well, to ensure the entire cleanup flow can pass. 🤔

detox/src/client/Client.js
```diff
@@ -336,7 +345,9 @@ class Client {
       this._asyncWebSocket.rejectAll(this._pendingAppCrash);
       this._pendingAppCrash = null;
     } else if (this._asyncWebSocket.hasPendingActions()) {
-      this._asyncWebSocket.rejectAll(new DetoxRuntimeError('The app has unexpectedly disconnected from Detox server.'));
+      this._asyncWebSocket.rejectAll(wasAppTerminatedByUs
+        ? new DetoxRuntimeError('The app has been unresponsive and Detox had to terminate it forcibly.')
```
noomorph, line 349: TODO: check if my English wording is good here 😅

detox/src/client/Client.js
```diff
@@ -73,16 +73,23 @@ class Client {
 
     try {
       if (this.isConnected) {
-        await this.sendAction(new actions.Cleanup(this._successfulTestRun));
+        this._scheduleAppTermination();
```
noomorph, line 76: @d4vidi , actually this decision is a bit arguable — as you see, here I plan to terminate the app if it is unresponsive.

The other option would be just to cancel the cleanup request awaiting and go on. But concerns me, if the app is in broken state, why should we keep it running on the device? Better to kill it, no?

What do you think?

detox/src/client/Client.js
```diff
@@ -73,16 +73,23 @@ class Client {
...
+
+        try {
+          await this.sendAction(new actions.Cleanup(this._successfulTestRun));
+        } catch (e) {
+          log.error({ event: 'ERROR' }, DetoxRuntimeError.format(e));
```
noomorph, line 81: TODO: add some extra description of the failure, e.g.:

```js
          log.error({ event: 'ERROR' }, DetoxRuntimeError.format(new DetoxRuntimeError({
            message: 'Failed to conduct the app cleanup over web sockets due to the error.',
            debugInfo: e,
          })));
```

detox/src/client/Client.js
noomorph: Disregard, I think it is too early to app termination to here. Let it be just stuck.

detox/src/client/Client.js
noomorph: I decided to enhance `AsyncWebSocket` so that it has a `timeout=N` option you can attach to the request.

detox/src/client/Client.js
noomorph: After a night of sleep, I think it is more reasonable to let the app hang. Maybe users will find it useful in their local environment. On CI, you can run `detox test ... --cleanup` so that emulator/simulator is terminated after all the tests, so no big sense.

detox/src/client/Client.js
noomorph: Both asyncwebsocket.close and sendAction are `.catch(() => log)`ged. We are good now.

PR #2724, opened 2021-04-13
Improve mocking guide instructions

badsyntax:
## Description

In this pull request, I have improved the mocking guide slightly. Due to the way it was previously worded, I completely skipped step 1 as I assumed it was not relevant for me as I am using a modern version of RN. Had I read step 1 in more detail, I would have seen that I needed to follow this step. This change makes the instructions easier to consume to prevent confusion over the RN versions.

---

> _For API changes:_


d4vidi:
Thanks

PR #2767, opened 2021-04-29
To be focused expectation

### LINKED ISSUE #2766: Add toBeFocused expectation

alon-ha:
## Description

Closes #2766. Added to be focused functionality

### DIFF
--- a/detox/src/ios/expectTwo.js
+++ b/detox/src/ios/expectTwo.js
@@ -19,6 +19,14 @@ class Expect {
     return this.not.toBeVisible();
   }
 
+  toBeFocused() {
+    return this.expect('toBeFocused');
+  }
+
+  toBeNotFocused() {
+    return this.not.toBeFocused();
+  }
+
   toExist() {
     return this.expect('toExist');
   }
--- a/detox/src/ios/expectTwo.test.js
+++ b/detox/src/ios/expectTwo.test.js
@@ -272,6 +272,39 @@ describe('expectTwo', () => {
     expect(testCall).deepEquals(jsonOutput);
   });
 
+  it(`expect(element(by.text('Tap Working!!!'))).toBeFocused()`, () => {
+    const testCall = e.expect(e.element(e.by.text('Tap Working!!!'))).toBeFocused();
+    const jsonOutput = {
+      invocation: {
+        type: 'expectation',
+        predicate: {
+          type: 'text',
+          value: 'Tap Working!!!'
+        },
+        expectation: 'toBeFocused'
+      }
+    };
+
+    expect(testCall).deepEquals(jsonOutput);
+  });
+
+  it(`expect(element(by.text('Tap Working!!!'))).toBeNotFocused()`, () => {
+    const testCall = e.expect(e.element(e.by.text('Tap Working!!!'))).toBeNotFocused();
+    const jsonOutput = {
+      invocation: {
+        type: 'expectation',
+        predicate: {
+          type: 'text',
+          value: 'Tap Working!!!'
+        },
+        modifiers: ['not'],
+        expectation: 'toBeFocused'
+      }
+    };
+
+    expect(testCall).deepEquals(jsonOutput);
+  });
+
   it(`expect(element(by.id('UniqueId204'))).toHaveText('I contain some text')`, () => {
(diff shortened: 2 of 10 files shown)


detox/src/ios/expectTwo.js
```diff
@@ -19,6 +19,14 @@ class Expect {
     return this.not.toBeVisible();
   }
 
+  toBeFocused() {
+    return this.expect('toBeFocused');
+  }
+
+  toBeNotFocused() {
```
d4vidi, line 26: We've recently decided that these types of API's are deprecated, and that users should use `.not.toBeXYZ()` instead

detox/src/ios/expectTwo.test.js
```diff
@@ -272,6 +272,39 @@ describe('expectTwo', () => {
...
+          type: 'text',
+          value: 'Tap Working!!!'
+        },
+        modifiers: ['not'],
+        expectation: 'toBeFocused'
+      }
+    };
+
+    expect(testCall).deepEquals(jsonOutput);
+  });
```
d4vidi, line 306: Not directly related to this feature... But, really, this set of tests surely seems like it should be converted to snapshot tests. @noomorph wdyt?

detox/test/src/Screens/ExpectationsScreen.js
```diff
@@ -0,0 +1,50 @@
+import React, { Component } from 'react';
```
d4vidi, line 1: So, this is pretty great overall, yet I wonder whether it ought to be better to reuse the 3 input fields we already have in the main actions screens. Wdyt? In essence, this separation is a great initiative, but we should probably either skip it, or do it all the way (i.e. with transferring all other expectations here as well, which isn't necessarily possible)

detox/test/src/Screens/index.js
```diff
@@ -51,5 +52,6 @@ export {
   DeviceScreen,
   ElementScreenshotScreen,
   WebViewScreen,
-  VirtualizedListStressScreen,  
+  VirtualizedListStressScreen,
```
d4vidi, line 55: Isn't this out of context?

d4vidi (commented):
@alon-ha kudos for this! please see my notes 🙏🏻

detox/src/ios/expectTwo.js
alon-ha: Ohh I just follow the toBeNotVisible() function.. Maybe it's better in those instances we decide to change things to deprecate such methods or at least leave a comment in the code above such.
Should I delete it?

detox/src/ios/expectTwo.test.js
alon-ha: I also agree it would be better

detox/test/src/Screens/ExpectationsScreen.js
alon-ha: I do think it will be great to split things up.. At least from this point. To put it in the actions screen (which is already loaded with stuff) does not make sense to me.. Naming and functionality

detox/test/src/Screens/index.js
alon-ha: I just added it at the last position because it's new. I don't necessarily think it should be the same position it shows up it the actual main screen with all the titles. Doesn't really matter to me, I can change it if you want

noomorph:
Oi va voi, this pull request lacks the typings. 😄 
Nevermind, you are covered this time — I've added them: 419a451a70a10762691f860dce5831f4f1c433a9

d4vidi:
@noomorph I was actually inclined to keeping this undocumented (docs + typings), until Android support is introduced.

detox/src/ios/expectTwo.js
d4vidi: I see now that there's a `not` equivalent to each expectation we support 🙄 
So let's keep as-is for now.

detox/test/src/Screens/ExpectationsScreen.js
d4vidi: Well for now it looks that the new screen only contains 2 input fields, so I'm not sure why the actions screen is insufficient...

d4vidi:
@alon-ha let's revisit the review notes, and talk about docs.

PR #2768, opened 2021-05-04
Reorganize roles in Emu delegates to match cleanup vs. shutdown

d4vidi:
## Description

I've found this while working on redefining the concept of a `DeviceId` throughout our JS framework, as a tech debt left from the work on Genymotion (see #2655):

In the context of Android drivers, the terms 'cleanup' and 'shutdown' were (semantically) wrongly interpreted inside the related delegate (i.e. `EmulatorDeviceAllocation` class). While the implementation was correct, the associated code was simply misplaced inside it.

In this pull request, I have reorganized `EmulatorDeviceAllocation` such that the code inside `deallocateDevice()` would in fact simply have it removed from the device registry, instead of shutting it down altogether. The latter was relocated onto a new method in an existing class: `EmulatorLauncher.shutdown()`.

UPDATE: I've done the same for Genymotion.

---

> _For API changes:_

### DIFF
--- a/detox/src/devices/drivers/android/emulator/EmulatorDriver.js
+++ b/detox/src/devices/drivers/android/emulator/EmulatorDriver.js
@@ -6,7 +6,7 @@ const AndroidDriver = require('../AndroidDriver');
 const EmulatorDeviceAllocation = require('./helpers/EmulatorDeviceAllocation');
 const AVDValidator = require('./AVDValidator');
 const AVDsResolver = require('./AVDsResolver');
-const EmulatorLauncher = require('./EmulatorLauncher');
+const EmulatorLauncher = require('./helpers/EmulatorLauncher');
 const EmulatorVersionResolver = require('./EmulatorVersionResolver');
 const FreeEmulatorFinder = require('./FreeEmulatorFinder');
 const { EmulatorExec } = require('../exec/EmulatorExec');
@@ -25,14 +25,14 @@ class EmulatorDriver extends AndroidDriver {
     this._deviceRegistry = DeviceRegistry.forAndroid();
 
     const emulatorExec = new EmulatorExec();
-    const emulatorLauncher = new EmulatorLauncher(emulatorExec);
+    this._emulatorLauncher = new EmulatorLauncher(emulatorExec, this.emitter);
     this._emuVersionResolver = new EmulatorVersionResolver(emulatorExec);
 
     const avdsResolver = new AVDsResolver(emulatorExec);
     this._avdValidator = new AVDValidator(avdsResolver, this._emuVersionResolver);
 
     const freeEmulatorFinder = new FreeEmulatorFinder(this.adb, this._deviceRegistry)
-    this._deviceAllocation = new EmulatorDeviceAllocation(this._deviceRegistry, freeEmulatorFinder, emulatorLauncher, this.adb, this.emitter);
+    this._deviceAllocation = new EmulatorDeviceAllocation(this._deviceRegistry, freeEmulatorFinder, this._emulatorLauncher, this.adb, this.emitter);
   }
 
   get name() {
@@ -71,12 +71,15 @@ class EmulatorDriver extends AndroidDriver {
   }
 
   async cleanup(deviceId, bundleId) {
-    await this._deviceRegistry.disposeDevice(deviceId);
-    await super.cleanup(deviceId, bundleId);
+    try {
+      await super.cleanup(deviceId, bundleId);
+    } finally {
+      await this._deviceAllocation.deallocateDevice(deviceId)
+    }
   }
 
   async shutdown(deviceId) {
-    await this._deviceAllocation.deallocateDevice(deviceId)
+    await this._emulatorLauncher.shutdown(deviceId);
   }
 
   async _fixAvdConfigIniSkinNameIfNeeded(avdName) {
--- a/detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
+++ b/detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
@@ -1,6 +1,7 @@
 const semver = require('semver'); // eslint-disable-line node/no-extraneous-require
 const onSignalExit = require('signal-exit');
 const AndroidDriver = require('../AndroidDriver');
+const InstanceLauncher = require('./helpers/GenyCloudInstanceLauncher');
 const GenyCloudInstanceAllocation = require('./helpers/GenyCloudInstanceAllocation');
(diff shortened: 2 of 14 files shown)


detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -69,12 +71,15 @@ class GenyCloudDriver extends AndroidDriver {
   }
 
   async cleanup(instance, bundleId) {
-    await this._deviceRegistry.disposeDevice(instance.uuid);
-    await super.cleanup(instance, bundleId);
+      try {
+        await super.cleanup(instance, bundleId);
+      } finally {
+        await this._instanceAllocation.deallocateDevice(instance.uuid);
```
d4vidi, line 77: Now `GenyCloudInstanceAllocation.deallocateDevice` effectively disposes from the registry, instead of executing the shutdown sequence.

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -69,12 +71,15 @@ class GenyCloudDriver extends AndroidDriver {
...
+      }
   }
 
   async shutdown(instance) {
-    await this._instanceAllocation.deallocateDevice(instance);
+    await this._instanceLauncher.shutdown(instance);
```
d4vidi, line 82: `GenyCloudInstanceLauncher.shutdown()` now executes the logic previously implemented in `GenyCloudInstanceAllocation.deallocateDevice`

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -71,12 +71,15 @@ class EmulatorDriver extends AndroidDriver {
   }
 
   async cleanup(deviceId, bundleId) {
-    await this._deviceRegistry.disposeDevice(deviceId);
-    await super.cleanup(deviceId, bundleId);
+    try {
+      await super.cleanup(deviceId, bundleId);
+    } finally {
+      await this._deviceAllocation.deallocateDevice(deviceId)
```
d4vidi, line 77: Now `EmulatorDeviceAllocation.deallocateDevice()` effectively disposes the device from the registry.

detox/src/devices/drivers/android/emulator/EmulatorDriver.js
```diff
@@ -71,12 +71,15 @@ class EmulatorDriver extends AndroidDriver {
...
+    }
   }
 
   async shutdown(deviceId) {
-    await this._deviceAllocation.deallocateDevice(deviceId)
+    await this._emulatorLauncher.shutdown(deviceId);
```
d4vidi, line 82: The `EmulatorLauncher.shutdown()` executes the shutdown sequence (i.e. via telnet, etc.) instead of `EmulatorDeviceAllocation.deallocateDevice()`.

d4vidi:
Merging even though build is red because there was an ANR on the local emulators in 1 of the 2 main Android builds, which isn't related to cleanup etc.

detox/src/devices/drivers/android/genycloud/helpers/GenyCloudInstanceAllocation.js
```diff
@@ -1,18 +1,17 @@
...
-const GenyCloudInstanceHandle = require('../GenyCloudInstanceHandle');
 const DetoxRuntimeError = require('../../../../../errors/DetoxRuntimeError');
 const retry = require('../../../../../utils/retry');
 const logger = require('../../../../../utils/logger').child({ __filename });
 
 const { ALLOCATE_DEVICE_LOG_EVT } = AndroidDeviceAllocation;
 
 class GenyCloudInstanceAllocation extends AndroidDeviceAllocation {
-  constructor(deviceRegistry, deviceCleanupRegistry, instanceLookupService, instanceLifecycleService, eventEmitter) {
+  constructor(deviceRegistry, instanceLookupService, instanceLifecycleService, instanceLauncher, eventEmitter) {
```
noomorph, line 9: Too many args in the constructor to my taste, I'd recommend using destructuring instead.

detox/src/devices/drivers/android/genycloud/helpers/GenyCloudInstanceLauncher.test.js
```diff
@@ -0,0 +1,90 @@
...
+      expect(eventEmitter.emit).toHaveBeenCalledWith('shutdownDevice', { deviceId: instance.adbName });
+    });
+
+    it('should not emit shutdownDevice prematurely', async () => {
+      givenAnInstanceDeletionError();
+
+      try {
+        const instance = anInstance();
+        await uut.shutdown(instance);
+        fail('Expected an error');
```
noomorph, line 83: Bad practice, Jest Circus does not have `fail()` function at all.

detox/src/devices/drivers/android/genycloud/helpers/GenyCloudInstanceLauncher.js
```diff
@@ -0,0 +1,36 @@
...
+   * @param instance {GenyInstance} The freshly allocated cloud-instance.
+   * @returns {Promise<void>}
+   */
+  async launch(instance) {
+    const instanceHandle = new GenyCloudInstanceHandle(instance);
+    await this._deviceCleanupRegistry.allocateDevice(instanceHandle);
+  }
+
+  async shutdown(instance) {
+    const instanceHandle = new GenyCloudInstanceHandle(instance);
```
noomorph, line 27: Generic question — aren't you trying here to make `detox test --cleanup` runs more favorable to Genycloud instead of planning a sprint where we move the device allocation/shutdown to global Jest setup and teardown?

In my opinion, `shutdown` should SHUT DOWN the device, and not do only book-keeping. Let's check if we are on the same page here with you, @d4vidi .

detox/src/devices/drivers/android/genycloud/helpers/GenyCloudInstanceLauncher.test.js
d4vidi: Odd, I was actually working to remove all `fail()` calls. Fix in progress.

detox/src/devices/drivers/android/genycloud/helpers/GenyCloudInstanceLauncher.js
d4vidi: Yes we will sort these things out sooner than later -- let's enqueue for the next Q! :)
In the meantime, the current code simply tries to comply to the currently expected behaviour, broken as it may be.

Issue #2193, opened 2020-07-08
Make detox setup and teardown global


d4vidi:
Today, in Jest environments, we initialize Detox from scratch in each and every suite-run from within the context of our circus `Environment` class. As a rule, initialization includes launching emulators/simulators, when needed.

This approach creates some drawbacks, as things like prebooting and preconfiguring test emulators, cannot be centrally controlled. We generally want to have Detox init split into two use cases:
1. Global init, potentially integrated into Jest's global-setup phase.
2. Per-suite init, which - for example, locks-on on a specific device.
> Teardown should be addressed equivalently, of course.

Until we introduce this change, the following fixes/enhancements shall remain blocked:
- In #2188, we've created a technical debt where in some cases, util-binary APK's (e.g. test-butler's) cannot overwrite currently installed version of the same APK's, when upgrading them. The reason is that there was no single place to preinstall these APK's, and - so as to avoid reinstalling them again and again in the beginning of each suite (can be painfully time consuming), we instead chose to refrain from installing if the package already exists on the device.

- Also in #2188 (or actually #2194), we've applied a manual resolution of the device type using internal detox API (see https://github.com/wix/Detox/pull/2194/files#diff-a984eb86f26b819ce6f51b8a8fe9226bR4). A preliminary global init, would allow for the device's type to be available at a stage as early as this (e.g. using `device.getPlatform()`).

- Device cleanup using the `--cleanup` arg isn't properly supported in parallel-execution environments (#1975 refers to that): instead of tearing down the emulators once when the whole suite ends, it does this in the end of each suite. The runtime penalty here is huge. In internal projects, we've instructed devs to manually apply a `killall` in scripts, but that doesn't scale.

- For the sake of native animations' disabling on Android, we apply the necessary configuration via `adb` in each init, and thus in the beginning of each suite. This is another time-waster, as it should only be done once as part of the global setup.

- Device launch implementation is overly complex, and cannot be simplified. Since centralized prebooting is not possible, we are forced - at least on Android, at prelaunch, to allocated a unique adb port, outside the recommended range, and run emulators synthetically - preguessing their port-based name (e.g. `emulator-12345`), which is bad practice. A much more simple solution would be to just launch an emulator, having the system allocate its port, and that's that.

yungkittty:
Thanks for the tutorial btw.

[...]

PR #2778, opened 2021-05-12
Make DeviceRegistry DB-like: hold ids+data instead of objects

d4vidi:
## Description

- This pull request is associated with the issue described here: #2655.

In this pull request, I have reworked the device-registry implementation such that instead of allowing for the holding of objects (rather than plain strings as ID's), it now allows for storing ID's (strings) as DB-like keys, alongside associated data (arbitrary types).

The reason for ever allowing for objects storing was to support Genycloud, and it was a mistake. The main reason was that on top of holding an instance' UUID, we want to also have its name as meta-data, for proper logging during global tear-down.

Allowing for objects in device-registry is, by principle, wrong - because it makes the comparisons (i.e. in look-up queries) wrong in the sense that both the actual ID *AND* the meta data are forced to be identical in order to create a match. Effectively, only the ID should matter...

To put things in a more illustrative way, I've basically changed the registry's content format (in the case of genycloud) from this:

```
[
  {"uuid":"6151855a-4405-4690-8aa8-0e5524f3a4f9","name":"Detox-1620851733498.1"},
  {"uuid":"dd6b598f-a552-4393-bd96-e9923c01cccc","name":"Detox-1620851733498.2"}
]
```

to this:

```
[{
  "id":"6151855a-4405-4690-8aa8-0e5524f3a4f9",
  "data":{"name":"Detox-1620851733498.1"}
},
{
  "id":"dd6b598f-a552-4393-bd96-e9923c01cccc",
  "data":{"name":"Detox-1620851733498.2"}
}]
```
Ultimately, the resulted DeviceRegistry implementation is far simpler 👍🏻 

> ⚠️  Note: Storing meta-data in device-registry is nonetheless still somewhat of an abuse. A cleaner way would be to completely detach the locking mechanism from the actual data, and rather apply usage of an actual DB - as the likes of [Simple-JSONdb](https://github.com/nmaggioni/Simple-JSONdb) or [lowdb](https://github.com/typicode/lowdb). However, this feels like an overkill-refactor at the moment.

---

> _For API changes:_

### DIFF
--- a/detox/src/devices/DeviceRegistry.js
+++ b/detox/src/devices/DeviceRegistry.js
@@ -4,25 +4,35 @@ const fs = require('fs-extra');
 const ExclusiveLockfile = require('../utils/ExclusiveLockfile');
 const safeAsync = require('../utils/safeAsync');
 
-const getDeviceEqualsFn = (deviceHandle) => (otherDeviceHandle) => _.isEqual(otherDeviceHandle, deviceHandle);
-const getDeviceDifferFn = (deviceHandle) => (otherDeviceHandle) => !_.isEqual(otherDeviceHandle, deviceHandle);
 const readOptions = {
   encoding: 'utf8',
 };
 
-class DeviceHandlesList {
+const FIELD_NAME_ID = 'id';
+const FIELD_NAME_DATA = 'data';
+
+class DevicesList {
   constructor(devices) {
-    this.rawDevices = Object.freeze(devices);
+    this._devices = Object.freeze(devices);
   }
 
-  includes(deviceHandle) {
-    return DeviceHandlesList._includes(deviceHandle, this.rawDevices);
+  /**
+   * @returns {{id: string, data: *?}[]}
+   */
+  get rawDevices() {
+    return this._devices;
   }
 
-  static _includes(deviceHandle, devices) {
-    const rawDeviceHandle = getRawDeviceHandle(deviceHandle);
-    const deviceEqualsFn = getDeviceEqualsFn(rawDeviceHandle);
-    return !!_.find(devices, deviceEqualsFn);
+  /**
+   * @param {string} deviceId
+   * @returns {boolean}
+   */
+  includes(deviceId) {
+    return DevicesList._includes(deviceId, this._devices);
+  }
+
+  static _includes(deviceId, devices) {
+    return _.some(devices, [FIELD_NAME_ID, deviceId]);
   }
 }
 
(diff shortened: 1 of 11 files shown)


detox/src/devices/DeviceRegistry.js
```diff
@@ -120,16 +131,30 @@ class DeviceRegistry {
     return [];
   }
 
-  /***
+  /**
+   * @private
+   */
+  _registerDevice(deviceId, data) {
+    const state = this._lockfile.read();
+    let newState = _.reject(state, [FIELD_NAME_ID, deviceId]);
```
noomorph, line 139: Shouldn't we be signalizing if we are registering an already registered device?

detox/src/devices/DeviceRegistry.js
```diff
@@ -4,25 +4,35 @@ const fs = require('fs-extra');
...
-const getDeviceDifferFn = (deviceHandle) => (otherDeviceHandle) => !_.isEqual(otherDeviceHandle, deviceHandle);
 const readOptions = {
   encoding: 'utf8',
 };
 
-class DeviceHandlesList {
+const FIELD_NAME_ID = 'id';
+const FIELD_NAME_DATA = 'data';
+
+class DevicesList {
```
noomorph, line 14: Overall, why, if you are already using ID as a primary factor, why don't you just turn the list into map, i.e.:

```
[id, meta],
[id, meta],
[id, meta],
```

to 

```
id: meta,
id: meta,
id: meta,
```

?

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
```diff
@@ -165,4 +167,10 @@ function reportGlobalCleanupSummary(deletionLeaks) {
   }
 }
 
+function rawDevicesToInstanceHandles(rawDevices) {
```
noomorph, line 170: So, basically, you still have to dance with a tambourine here, to not rewrite further code in this PR?

noomorph (commented):
To be honest, I don't see any usefulness in this pull request – the things stay very complicated, and our current abstractions are falling apart. In my opinion, there is no easy way to fix this situation unless we complete a heavy rewrite.

No objections to merging, but I am not too enthusiastic either. 🤷‍♂️

detox/src/devices/DeviceRegistry.js
d4vidi: Probably so, but that's slightly out of scope, as it breaks the module's contract. Need to consider that separately.

detox/src/devices/DeviceRegistry.js
d4vidi: I was trying to avoid the place where in the common use case we'd have a file containing a JSON with many keys and empty objects as data:
```json
{
  "id1": {},
  "id2: {},
   ...
}
```

Do you think it'd be better?

detox/src/devices/drivers/android/genycloud/GenyCloudDriver.js
d4vidi: LOL it's mostly for clarity. Writing further code boils down to #2193, doesn't it?

d4vidi:
@noomorph thanks for sincere feedback, I truly appreciate it! 🙏🏻 
Note that the point here was not to iron everything through. For that, we'll have to properly introduce #2193 (and we will - in due time) - because then all global data could be kept in-memory under the global context...
The goal here was to eliminate a data management practice that was downright wrong.

PR #2798, opened 2021-05-25
lint: stricter rules

noomorph:
1. Enforces imports order.
2. Enforces single quotes or backticks (except for escaping).
3. Enforces spaces between curly brackets.
4. Enforces semicolons.
5. Prints warnings on unused variables.
6. Updated .nvmrc 
7. Prevents forgotten `it.only` in unit tests.


d4vidi:
@noomorph ok this is great and all, but please, let's put the breaks on this a bit and continue one-rule-at-time?! 😆

Issue #2757, opened 2021-04-22
Multiapps config: undescriptive error in case of missing device config


d4vidi:
## Description
In a use case where a configuration is defined with an app in the new format (i.e. with an `app` or even `apps` field), but a `device` is not specified (rather, an explicit `type` is used - e.g. `"type": "ios.simulator"`), the resulted error is this:

```
Missing "binaryPath" property in the app config.
```

Which isn't helpful. Need to consider this use case in the config validation, and output a clearer error.

noomorph:
Yeah, will do. Thanks.

noomorph:
@d4vidi, could you provide an exact Detox config where this issue reproduces?
Unfortunately, I am not able to reproduce this according to the description.
Let's make bug reports less vague, ok? What would it have costed for you to paste an excerpt of Detox config back then, right? 😊

d4vidi:
Hmm. IMO, the description is pretty clear 😄 
You can try to edit one of Detox' self configs - have the device removed, and instead add a `type` field. Does it not reproduce this way? 😢

noomorph:
Resolved in [18.16.0](https://github.com/wix/Detox/releases/tag/18.16.0).

PR #2800, opened 2021-05-25
fix(config): better detection of app/apps errors

noomorph:
## Description

- This pull request addresses the issue described here: #2757

In this pull request, I have improved error handling for the next cases:

1. Someone is using a legacy (plain) config yet wants to adds `app` or `apps` property:

```
DetoxConfigError: Your configuration "plain" appears to be in a legacy format, which can’t contain "app" or "apps".

HINT: Remove "type" property from configuration and use "device" property instead:
a) "device": { "type": "android.emulator", ... }
b) "device": "<alias-to-device>" // you should add that device configuration to "devices" with the same key

Check your Detox config at path:
/home/detox/myproject/.detoxrc.json

{
  configurations: {
    plain: {
      type: 'android.emulator',
      device: 'Pixel_3a_API_30_x86',
      binaryPath: 'path/to/apk',
      app: 'my-app'
    }
  }
}
```

2. Someone is using a legacy (plain) config and has app configuration error, e.g. missing binaryPath:

```diff
-DetoxConfigError: Missing "binaryPath" property in the app config.
+DetoxConfigError: Missing "binaryPath" property in configuration "plain".
```

In other words, I fix here the phrasing from `in the app config` to `configuration <name>` to avoid confusion.

3. Same as 2, but for another property, `launchArgs`:

```diff
-DetoxConfigError: Invalid type of "launchArgs" property in the app config.
+DetoxConfigError: Invalid type of "launchArgs" property in configuration "plain".
```

That should improve UX when transitioning between old and new format, and overall.

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -190,6 +190,8 @@ declare global {
             build?: string;
             testBinaryPath?: string;
             launchArgs?: Record<string, any>;
+            app?: never;
+            apps?: never;
         }
 
         type DetoxBuiltInDeviceConfig =
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -190,8 +190,6 @@ declare global {
             build?: string;
             testBinaryPath?: string;
             launchArgs?: Record<string, any>;
-            app?: never;
-            apps?: never;
         }
 
         type DetoxBuiltInDeviceConfig =
--- a/detox/src/configuration/composeAppsConfig.test.js
+++ b/detox/src/configuration/composeAppsConfig.test.js
@@ -237,6 +237,31 @@ describe('composeAppsConfig', () => {
   });
 
   describe('unhappy scenarios:', () => {
+    describe('plain configuration:', () => {
+      beforeEach(() => {
+        localConfig = {
+          type: 'ios.simulator',
+          device: 'Phone',
+          binaryPath: 'path/to/app',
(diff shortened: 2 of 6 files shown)


detox/src/configuration/composeAppsConfig.test.js
```diff
@@ -237,6 +237,31 @@ describe('composeAppsConfig', () => {
...
+      it('should throw if the config has .app property defined', () => {
+        localConfig.app = 'myapp';
+        expect(compose).toThrowError(errorComposer.oldSchemaHasAppAndApps());
+      });
+
+      it('should throw if the config has .apps property defined', () => {
+        localConfig.apps = ['myapp'];
+        expect(compose).toThrowError(errorComposer.oldSchemaHasAppAndApps());
+      });
+    });
```
d4vidi, line 263: Nice, I like the approach

d4vidi (approved):

detox/index.d.ts
```diff
@@ -190,8 +190,6 @@ declare global {
             build?: string;
             testBinaryPath?: string;
             launchArgs?: Record<string, any>;
-            app?: never;
-            apps?: never;
```
d4vidi: oh why?

noomorph:
TS is not very smart with type inference here. I'd love to keep it, but
apparently it does not work reliably.

PR #2822, opened 2021-06-06
Add toBeFocused on Android

jonathanmos:
#### Description
Adds toBeFocused on Android, to handle issue #2667


d4vidi:
@jonathanmos I'm generally okay with this thorough PR 👍🏻 
However there's something we need to consider before merging in: While through the newly added Detox assertion API - Espresso's `hasFocus()` matcher is used -- which in turn boils down to `view.hasFocus()`, for the focus status returned over a view's **attributes** set, the`view.isFocused()` method is used.
The two are very similar, but different. The `ViewGroup`'s override over `hasFocus()` shows how. I think we should consider which one is more appropriate, and then align both.

PR #2829, opened 2021-06-10
fix(artifacts): suppress some stderr logs from iOS video recordings

noomorph:
## Description

Suppresses redundant messages from **stderr** of `xcrun ... recordVideo` command (used by `--record-videos all` on iOS), that started appearing since Xcode 12.4 (or even earlier).

```diff
 detox[10505] INFO:  [test.js] DETOX_CONFIGURATION="ios.sim.release" DETOX_RECORD_VIDEOS="all" DETOX_REPORT_SPECS=true DETOX_START_TIMESTAMP=1623313422454 DETOX_USE_CUSTOM_LOGGER=true nyc jest --config e2e/config.js --testNamePattern '^((?!:android:).)*$' --maxWorkers 1 e2e/01.sanity.test.js
 detox[10507] INFO:  [AppleSimUtils.js] com.wix.detox-example launched. To watch simulator logs, run:
         /usr/bin/xcrun simctl spawn 6371830E-8B44-422B-A52C-DB0B62E6B6E3 log stream --level debug --style compact --predicate 'process == "example"'
 detox[10507] INFO:  Sanity is assigned to 6371830E-8B44-422B-A52C-DB0B62E6B6E3 {"type":"iPhone 12 Pro Max"}
 detox[10507] INFO:  Sanity: should have welcome screen
-detox[10507] ERROR: [SPAWN_STDERR, #10617] Recording started
-
-detox[10507] ERROR: [SPAWN_STDERR, #10617] Wrote video to: /private/var/folders/lm/thz8hdxs4v3fppjh0fjc2twhfl_3x2/T/1a19c7de-8be9-4a6b-af36-fe4f32749755.detox.mp4
 
 detox[10507] INFO:  Sanity: should have welcome screen [OK]
 detox[10507] INFO:  Sanity: should show hello screen after tap
-detox[10507] ERROR: [SPAWN_STDERR, #10631] Recording started
-
-detox[10507] ERROR: [SPAWN_STDERR, #10631] Wrote video to: /private/var/folders/lm/thz8hdxs4v3fppjh0fjc2twhfl_3x2/T/71ac0573-9d1a-4ede-8935-4a6024e109dc.detox.mp4
 
 detox[10507] INFO:  Sanity: should show hello screen after tap [OK]
 detox[10507] INFO:  Sanity: should show world screen after tap
-detox[10507] ERROR: [SPAWN_STDERR, #10644] Recording started
-
-detox[10507] ERROR: [SPAWN_STDERR, #10644] Wrote video to: /private/var/folders/lm/thz8hdxs4v3fppjh0fjc2twhfl_3x2/T/17ffa532-edb7-4620-b58e-9e78a1532a4a.detox.mp4
 
 detox[10507] INFO:  Sanity: should show world screen after tap [OK]
 
  PASS  e2e/01.sanity.test.js (9.555 s)
   Sanity
     ✓ should have welcome screen (1026 ms)
     ✓ should show hello screen after tap (1334 ms)
     ✓ should show world screen after tap (1458 ms)
```

### DIFF
--- a/detox/src/utils/exec.js
+++ b/detox/src/utils/exec.js
@@ -128,7 +128,8 @@ function _composeCommand(bin, prefix, args) {
 function spawnAndLog(command, flags, options) {
   const cmd = _joinCommandAndFlags(command, flags);
 
-  const result = spawn(command, flags, { stdio: ['ignore', 'pipe', 'pipe'], ...options });
+  const { logLevelPatterns, silent, ...spawnOptions } = { stdio: ['ignore', 'pipe', 'pipe'], ...options };
+  const result = spawn(command, flags, spawnOptions);
   const { childProcess } = result;
   const { exitCode, stdout, stderr } = childProcess;
 
@@ -139,9 +140,18 @@ function spawnAndLog(command, flags, options) {
     log.error({ event: 'SPAWN_ERROR' }, `${cmd} failed with code = ${exitCode}`);
   }
 
-  if (!options || !options.silent) {
-    stdout && stdout.on('data', (chunk) => log.trace({ stdout: true, event: 'SPAWN_STDOUT' }, chunk.toString()));
-    stderr && stderr.on('data', (chunk) => log.error({ stderr: true, event: 'SPAWN_STDERR' }, chunk.toString()));
+  if (!silent) {
+    stdout && stdout.on('data', (chunk) => {
+      const line = chunk.toString();
+      const loglevel = inferLogLevel(line, logLevelPatterns) || 'trace';
+      log[loglevel]({ stdout: true, event: 'SPAWN_STDOUT' }, line);
+    });
+
+    stderr && stderr.on('data', (chunk) => {
+      const line = chunk.toString();
+      const loglevel = inferLogLevel(line, logLevelPatterns) || 'error';
+      log[loglevel]({ stderr: true, event: 'SPAWN_STDERR' }, line);
+    });
   }
 
   function onEnd(e) {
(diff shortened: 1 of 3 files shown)


detox/src/utils/exec.js
```diff
@@ -139,9 +140,18 @@ function spawnAndLog(command, flags, options) {
...
+      const line = chunk.toString();
+      const loglevel = inferLogLevel(line, logLevelPatterns) || 'trace';
+      log[loglevel]({ stdout: true, event: 'SPAWN_STDOUT' }, line);
+    });
+
+    stderr && stderr.on('data', (chunk) => {
+      const line = chunk.toString();
+      const loglevel = inferLogLevel(line, logLevelPatterns) || 'error';
+      log[loglevel]({ stderr: true, event: 'SPAWN_STDERR' }, line);
+    });
```
d4vidi, line 154: @noomorph since this is "real-time"-ish, may I suggest to preregister different handler-callbacks that do not check for the value of `logLevelPatterns` on each `chunk`, in case it's empty.

noomorph: Checked a good couple of times with profiler, etc — no impact here on the performance. Not sure if it really makes sense 🤷‍♂️

noomorph: @d4vidi , it is exiting very fast there, see the source code: https://github.com/lodash/lodash/blob/master/isEmpty.js?source=post_page---------------------------#L47 there. Function calls themselves are quite cheap. `chunk.toString()` takes way more time.

d4vidi: Yes I understand - but still, no point in checking again and again whether it's empty because it never changes through-out the life of the callback...

Issue #2678, opened 2021-03-07
Detox Native, Next Steps


d4vidi:
## Description

A follow-up on #2672: Do all the missing tasks under 'to-do', there.

PR #2835, opened 2021-06-14
Detox native - run on CI

d4vidi:
## Description

- This pull request addresses the issue described here: #2678.

In this pull request, I have integrated a new example module, that demonstrates how Detox can be used in its pure-native form, in pure-native apps (Android). I've also updated the CI scripts such that the module's demo-tests would be executed on CI.

On a separate track, I've removed support for taking view screenshots through Detox' API. I've come to realize that it wouldn't be useful because applitools' API already provides this ability, and in an encapsulated way -- that both accepts a matcher and uploads the image directly.

---

> _For API changes:_


detox/android/detox/src/coreNative/java/com/wix/detox/actions/DetoxViewActions.kt
```diff
@@ -32,7 +29,4 @@ public object DetoxViewActions {
 
     public fun scrollRightBy(amountInDp: Double, startOffsetPercentX: Float? = null, startOffsetPercentY: Float? = null): ViewAction =
         actionWithAssertions(DetoxScrollAction(MOTION_DIR_RIGHT, amountInDp, startOffsetPercentX, startOffsetPercentY))
-
-    public fun takeScreenshot(): ViewActionWithResult<ScreenshotResult?> =
```
noomorph, line 36: What happened, @d4vidi? Why? I am curious.

d4vidi: Oh I actually described that briefly in the issue's description

PR #2841, opened 2021-06-16
Add details of network busy resources on android

### LINKED ISSUE #2886: Add details of In-flight network requests to debug-sync logs.

jonathanmos:
Add mechanism to print details for busy resources (at the moment only network) on Android. Fixes #2886.

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
@@ -72,7 +72,13 @@ public class NetworkIdlingResource extends DetoxBaseIdlingResource implements Ch
     @NotNull
     @Override
     public String getDescription() {
-        return "In-flight network activity";
+        String description = "In-flight network activity";
+
+        if (!busyResources.isEmpty()) {
+            description += "\nDetails:\n\t - " + busyResources.toString();
+        }
+
+        return description;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
```diff
@@ -86,6 +90,7 @@ protected boolean checkIdle() {
                 }
             }
             if (!idle) {
+                busyResources.add(call.request().url().toString());
                 break;
```
d4vidi, line 94: Is this `break` really needed?

d4vidi:
@jonathanmos The change in the network IR seems to be able to solve it!
Nevertheless, may I request that you'd unify the whole process onto a single call per each IR: Namely, you can have `DescriptiveIdlingResource.getDescription` changed in any way you see fit (name / signature) such that it would provide either the _whole_ in one go), or if the resource is idle nothing (e.g. `null`).

Apart for keeping the API on the more simple side (IMO), this would also eliminate the need for the questionable `busyResources` data member, which doesn't really affect the object's state...

 🙏🏻

detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
jonathanmos: it seems unnecessary - removing it

PR #2858, opened 2021-06-23
test: fix double app install issue

noomorph:
## Description

This is an idea of how to fix that double `device.installApp()` happening in our E2E self-test suite, and add a live example of advanced Jest environment use, i.e.: Detox + Jest docblockPragmas. That should cut down build time on our Jenkins OSS CI.

### DIFF
--- a/detox/test/e2e/22.launch-args.test.js
+++ b/detox/test/e2e/22.launch-args.test.js
@@ -1,3 +1,7 @@
+/**
+ * @detox-config-path ./detox-multiapps.config.json
+ */
+
 const _ = require('lodash');
 
 // Note: Android-only as, according to Leo, on iOS there's no added value here compared to
--- a/detox/test/e2e/setup.js
+++ b/detox/test/e2e/setup.js
@@ -1,6 +1,7 @@
 const { device } = require('detox');
 
 beforeAll(async () => {
-  await device.selectApp('example');
-  await device.launchApp();
+  if (device._currentApp) { // HACK: currently there's no API to tell whether we are in multi-apps mode or not
+    await device.launchApp();
+  }
 });
--- a/detox/test/e2e/detox-multiapps.config.json
+++ b/detox/test/e2e/detox-multiapps.config.json
@@ -1,0 +1,23 @@
+{
+  "extends": "./detox.config",
+  "configurations": {
+    "android.manual": {
+      "apps": ["android.debug", "android.debug.withArgs"]
+    },
+    "android.emu.debug": {
+      "apps": ["android.debug", "android.debug.withArgs"]
+    },
+    "android.emu.debug.fromSource": {
+      "apps": ["android.fromSource", "android.fromSource.withArgs"]
+    },
+    "android.emu.release": {
+      "apps": ["android.release", "android.release.withArgs"]
+    },
+    "android.genycloud.release": {
+      "apps": ["android.release", "android.release.withArgs"]
+    },
+    "android.genycloud.release2": {
+      "apps": ["android.release", "android.release.withArgs"]
+    }
+  }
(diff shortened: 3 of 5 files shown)


noomorph:
Tested on DWP deps, looks stable! ✅

detox/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "detox",
   "description": "E2E tests and automation for mobile",
-  "version": "18.18.1",
```
d4vidi, line 4: @noomorph be sure to unsmoke before merging in 😅

detox/test/e2e/22.launch-args.test.js
```diff
@@ -1,3 +1,7 @@
+/**
+ * @detox-config-path ./detox-multiapps.config.json
```
d4vidi, line 2: Albeit the coolness of this feature 😎, I'm inclined to say it even further complicates the already not-so-trivial configuration paradigm that Detox offers to users.

Also, IIUC, the end-result in the users' perspective would be that while a they might (for example) specify `android.emu.release`, they would end up getting a different configuration with the same name, which wouldn't be validated for (e.g. in terms of syntax) - or would it?
Is this really a proper dev-ex? let's discuss here...

detox/test/e2e/setup.js
```diff
@@ -1,6 +1,7 @@
 const { device } = require('detox');
 
 beforeAll(async () => {
-  await device.selectApp('example');
-  await device.launchApp();
+  if (device._currentApp) { // HACK: currently there's no API to tell whether we are in multi-apps mode or not
```
d4vidi, line 4: Need a better way to achieve that...

detox/test/e2e/22.launch-args.test.js
noomorph: First, it is not a feature — it is an advanced configuration of Jest that **WE** (and not someone else) is using internally for their self-test suites. I don't add anything new to Detox itself, so it complicates one specific thing in our e2e suite, not for **users**.

noomorph: Second, all the validations are performed as usual — Detox first merges the base config from the file and the override, and then triggers all the standard validation. It indeed will be a bit different configuration, but in fact not any different from `detox test -C e2e/detox-multiapps.config.json -c android.emu.release` launch.

detox/test/e2e/setup.js
noomorph: @d4vidi, we need to agree upon a public API then. If we have a `device.selectApp()`, then should we have `device.currentApp` public getter?

detox/package.json
noomorph: @d4vidi, do I really need to? I don't think so. According to semantic version calculation, it is totally correct. prerelease → patch.

detox/test/e2e/setup.js
noomorph: @d4vidi, here I am making sure you see my question — do you think we should add
```
get currentApp() { ... }
```
 as a public getter to Device?

detox/test/e2e/22.launch-args.test.js
d4vidi: @noomorph sorry for not getting to this sooner.
I think it's important that we would offer an optimization not only for ourselves, but for all devs.
Is it not likely that somewhere, developers have (or would have) 2 apps where `app 2` is really `app 1b` (i.e. different config, same binaries)?

noomorph:
I am abandoning this variant of solution and closing the pull request, in the reconciliation with this sentiment by @d4vidi:

> I think it's important that we would offer an optimization not only for ourselves but for all devs.

PR #2907, opened 2021-07-21
perf: skip initial multi-app reinstalls when binary paths are the same

noomorph:
## Description

This is another approach to crack the problem originally addressed by #2858.

The introduced optimization should eliminate that silly double uninstall-install process on our CI.

The implementation idea is to compare `binaryPath + testBinaryPath` for their uniqueness before calling uninstallApp/reinstallApp.


d4vidi:
This is awesome. Do we know what the impact is on our CI builds?

Issue #2904, opened 2021-07-19
App installs on Android, never launches, get ERROR: [APP_UNREACHABLE] Failed to reach the app over the web socket connection.


stmiller-shamrock:
### Describe the bug

When I run the tests on Android, I get:

ERROR: [APP_UNREACHABLE] Failed to reach the app over the web socket connection.

I see no other errors in any of the logs.

### Steps To Reproduce

To get the Android app to build I had to replace:
androidTestImplementation('com.wix:detox:+')
with
implementation('com.wix:detox:+')
in 
app/build.gradle
As explained [here](https://stackoverflow.com/questions/57723138/detox-build-fails-on-android-error-while-dexing).
Or the build would not include Detox libraries and fail. But other than that I double-checked my configuration. Even tried setting Detox up as a compiling dependency with the same results.

1. Run detox test -c android.debug --record-logs all
2. The emulator starts
3. The app gets installed
4. The app never launches and the test fails

### Expected behavior

The app launches in the emulator, and the tests run

### Detox Trace-Logs

detox[81194] TRACE: [DETOX_CREATE] created a Detox instance with config:
[...]

PR #2908, opened 2021-07-21
Reorganize troubleshooting onto a first-class guide, and ref it in code

d4vidi:
## Description

- This pull request addresses the issue described here: #2904.

In this pull request, I have collected, aggregated and reorganized troubleshooting instructions Detox has in the various troubleshooting guides.

Originally, the point was to try to better warn regarding mistakes related to `DetoxTest.java`. However, it was not easy to do + much boyscouting seemed to have been necessary, regardless.

Eventually, most work boiled down to documentation improvements. However, in addition, I've had the existing Detox-app connection error messages - in the code itself, trimmed down so as to mostly contain a ref to the associated troubleshooting guide. They were a bit overly verbose, I believe -- making developers not really wanting to read through them through, as they should.

---

> _For API changes:_

### DIFF
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -16,7 +16,6 @@
 
   For older Android gradle plugin support use `detox@6.x.x` instead ([previous setup guide here](https://github.com/wix/detox/blob/97654071573053def90e8207be8eba011408f977/docs/Introduction.Android.md)).
 
-
 **Note: As a rule of thumb, we consider all old major versions discontinued; We only support the latest Detox major version.**
 
 ## Setup :gear:
@@ -390,5 +389,5 @@ Please be aware that the `minSdkVersion` needs to be at least 18.
 
 ## Troubleshooting
 
-Please refer to our index of [troubleshooting guides](Troubleshooting.md).
+Please refer to [troubleshooting guide about build issues](Troubleshooting.BuildingTheApp.md#android) for assistance.
 
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -186,10 +186,12 @@ _Note: most guides advise of defining a global `kotlinVersion` constant - as in 
 
 ***Note that Detox has been tested for version 1.1.0 of Kotlin, and higher!***
 
-
 ### 5. Create a Detox-Test Class
 
-Add the file `android/app/src/androidTest/java/com/[your.package]/DetoxTest.java` and fill as in [the detox example app for RN](../examples/demo-react-native/android/app/src/androidTest/java/com/example/DetoxTest.java). **Don't forget to change the package name to your project's**.
+Detox requires a dummy implementation of a single Android-native test.
+
+1. Add a new file to your project, under this path and name: `android/app/src/androidTest/java/com/[your.package]/DetoxTest.java`. **Double-check check that the path is correct!**
+2. Copy & paste the content of the equivalent file from [the detox example app for RN](../examples/demo-react-native/android/app/src/androidTest/java/com/example/DetoxTest.java), into it. **Don't forget to change the package name to your project's package name!**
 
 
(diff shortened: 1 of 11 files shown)


noomorph (changes requested):
I'll review tomorrow.

docs/Introduction.Android.md
```diff
@@ -388,119 +390,5 @@ Please be aware that the `minSdkVersion` needs to be at least 18.
...
-           Constraint path 'OurApp:detox:unspecified' --> 'org.jetbrains.kotlin:kotlin-stdlib-common' strictly '1.3.0' because of the following reason: debugRuntimeClasspath uses version 1.3.0
-```
-(i.e. the project indirectly depends on different versions of `kotlin-stdlib`, such as `1.3.0`, `1.3.30`, `1.2.60`)
-
-#### Resolving for a compiling subproject
-
-Detox requires the Kotlin standard-library as it's own dependency. Due to the [many flavours](https://kotlinlang.org/docs/reference/using-gradle.html#configuring-dependencies) by which Kotlin has been released, multiple dependencies often create a conflict.
-
-For that, Detox allows for the exact specification of the standard library to use using two Gradle globals: `detoxKotlinVersion` and `detoxKotlinStdlib`. You can define both in your  root build-script file (i.e.`android/build.gradle`):
+Please refer to our index of [troubleshooting guides](Troubleshooting.md).
```
noomorph, line 393: I don't think it is wise to redirect to the general troubleshooting. Could you point straight to the relevant section in a specific file?

docs/README.md
```diff
@@ -30,10 +30,7 @@
 
 ### Troubleshooting
 
-- [Troubleshooting Detox Installation](Troubleshooting.Installation.md)
-- [Troubleshooting Failing Tests](Troubleshooting.RunningTests.md)
-- [Troubleshooting Synchronization](Troubleshooting.Synchronization.md)
-- [Dealing With Flakiness in Tests](Troubleshooting.Flakiness.md)
+- [Troubleshooting ToC](Troubleshooting.md)
```
noomorph, line 33: I don't like this idea. Could you just combine both, e.g.?

### Troubleshooting

- [Table of contents](Troubleshooting.md)
  - [Troubleshooting Detox Installation](Troubleshooting.Installation.md)
  - [Troubleshooting Failing Tests](Troubleshooting.RunningTests.md)
  - [Troubleshooting Synchronization](Troubleshooting.Synchronization.md)
  - [Dealing With Flakiness in Tests](Troubleshooting.Flakiness.md)

docs/Troubleshooting.RunningTests.md
```diff
@@ -25,6 +31,40 @@ detox test --loglevel trace
 
 See [here](https://github.com/wix/detox/blob/master/docs/Troubleshooting.Synchronization.md#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).
```
noomorph, line 32: By looking at this, I recalled that we already have it enabled by default... this section needs an extra review someday.

noomorph (changes requested):
The idea with "Table of contents" looks not bad overall, but I think you're overusing it here — it will worsen user experience if you redirect to it in the context of a specific issue. It will double the search effort by the user. We're not RDBMS, we are humans after all.

Other than that, AMAZING JOB, Amit. I like a lot the direction. ❤️

docs/Introduction.Android.md
d4vidi: I was having 2nd thoughts about this myself, TBH. Thanks for the feedback - will do!

docs/Troubleshooting.RunningTests.md
d4vidi: We need a documentation revamp. This is just one symptom.

docs/README.md
d4vidi: Done

PR #2915, opened 2021-07-25
Align device.id for Genymotion to ext-api contract

d4vidi:
## Description

In this pull request, I have aligned `device.id` for the Genymotion-cloud case, which - according to our existing contract, should return `adbName` (and instead returns a fancy, highly-detailed description).

---

> _For API changes:_

### DIFF
--- a/detox/src/devices/drivers/ios/SimulatorDriver.test.js
+++ b/detox/src/devices/drivers/ios/SimulatorDriver.test.js
@@ -1,8 +1,9 @@
 const AsyncEmitter = require('../../../utils/AsyncEmitter');
 
 describe('IOS simulator driver', () => {
-  let MockClient;
-  let uut, sim, emitter;
+  let client;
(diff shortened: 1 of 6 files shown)


detox/src/devices/drivers/ios/SimulatorDriver.test.js
```diff
@@ -15,7 +16,16 @@ describe('IOS simulator driver', () => {
...
 
-    MockClient = jest.requireMock('../../../client/Client');
+    const ClientMock = jest.requireMock('../../../client/Client');
+    client = new ClientMock();
+
+    const SimulatorDriver = require('./SimulatorDriver');
+    uut = new SimulatorDriver({ client, emitter });
+  });
+
+  it('should return the UDID as the external ID', () => {
```
d4vidi, line 26: This is the only real difference here. Everything else is pure boyscouting.

Issue #2750, opened 2021-04-19
detox-cli silently infinite spawns itself when installed locally rather than globally


cdavie-weconnect:
### Description
If you install detox-cli locally (`npm install detox-cli`) instead of globally (`npm install -g detox-cli`), running the `detox` command will spawn itself infinitely.

I suspect this is the same problem reported in #1376.

- [X] I have tested this issue on the latest Detox release and it still reproduces

#### Reproduction

1. `npm install detox detox-cli`
2. `detox help`

#### Expected behavior
`detox` commands work, or at least fail gracefully with instructions about installing globally.

#### Actual behavior
`detox` silently calls itself forever, with no output or warnings.

##### Installed locally

When **both** `detox` and `detox-cli` are installed locally (ie, in my repo's package.json), this is where the symlink points:

```
./node_modules/.bin/detox -> ../detox-cli/cli.js
```

Running `detox` causes `detox-cli/cli.js` to call itself, never actually running any detox commands.

##### Installed globally

When `detox` is locally installed, but `detox-cli` is removed from the package.json and installed globally instead, the symlinks look like this:

```
~/.nvm/versions/node/v14.6.0/bin/detox -> ../lib/node_modules/detox-cli/cli.js

./node_modules/.bin/detox -> ../detox/local-cli/cli.js
```

Running `detox` now causes `detox-cli/cli.js` to call `detox/local-cli/cli.js`, which is the real entrance point that parses arguments & runs detox commands.

#### Environment (please complete the following information):
 - Detox: 18.12.0
 - Node: 14.6.0
 - macOS: 11.2.1 (Big Sur)

noomorph:
The report might be legit — I assume this is not a blocker, but indeed is worth fixing.
I am not available this week, but I can look into it next week.
Thanks for taking the time to raise the issue!

d4vidi:
@noomorph `detox-cli` was never meant to work / ever being installed non-globally. I don't think there's much to do here.

chelovekdrakon:
Run into the same issue. Had to go inside and debug ... 
Not nice to spend time on such issues, when you are just trying to the init script 📦

noomorph:
Thanks, guys! The fix will land soon.

noomorph:
Resolved in [18.13.0](https://github.com/wix/Detox/releases/tag/18.13.0).

d4vidi:
Note: Reopened in Detox `18.20.1`. Will be address later this month.

hugows:
[...]

PR #2925, opened 2021-07-30
revert(global-cli): is installed globally

### LINKED ISSUE #2825: Error: "detox-cli" package is not meant to be installed locally, exiting...
Related issue: https://github.com/wix/Detox/issues/2789

Following the instructions to install Detox, I got the same error from the related issue. I already tried to uninstall and install again but seems that the version is still outdated.

My environment:
- macOS Big Sur (Just updated with everything)
- Node v14.17.0
- npm v6.14.13
- yarn v1.22.10

noomorph:
## Description

Resolves #2825.
Reopens #2750. 😢 

Since #2750 is not pleasant but not blocking, while #2825 is a dealbreaker, I have to revert the implementation which relies on `is-installed-globally`.

I'll need to submit fixes to that package, `is-installed-globally`, before I can return these checks back.

### DIFF
--- a/detox-cli/cli.js
+++ b/detox-cli/cli.js
@@ -3,7 +3,6 @@ const cp = require('child_process');
 const fs = require('fs');
 const path = require('path');
 const chalk = require('chalk');
-const isInstalledGlobally = require('is-installed-globally');
 
 /**
  * @param {string} msg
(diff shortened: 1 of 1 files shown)


d4vidi:
@noomorph should we not remove this from our dependencies list, as well?

noomorph:
Better if done, right

PR #2932, opened 2021-08-09
Added Typescript definitions for element.pinch() and element.takeScreenshot()

Vannevelj:
I noticed that some of the advertised APIs in the Docs weren't working with the latest version (I'm on 18.20.1). This makes two changes:

* `element().pinch()` has its `speed` and `angle` parameters marked as optional
* Added the missing API for `element().takeScreenshot()`, previously it only existed on `Device`

I happened to need both of these APIs in a new test I'm writing (iOS) and can confirm that the current functionality works just fine with this API definition.

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -1111,6 +1111,7 @@ declare global {
 
             /**
              * Pinches with the given scale, speed, and angle. (iOS only)
+             * @param speed default is `fast`
              * @param angle value in radiant, default is `0`
              * @example
              * await element(by.id('PinchableScrollView')).pinch(1.1);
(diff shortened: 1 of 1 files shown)


d4vidi (changes requested):
Thanks for this!
For `takeScreenshot`, please find the suitable place to add a reference to our element-screenshots guide: https://github.com/wix/Detox/blob/master/docs/APIRef.Screenshots.md#element-level-screenshots

Vannevelj:
@d4vidi I have added a link to that URL in the same style as I saw specified in a few other JSDocs.

d4vidi (approved):

PR #2958, opened 2021-08-27
fix(artifacts): save app artifacts before its uninstall

noomorph:
## Description

If there are pending app artifacts (iOS view hierarchies, screenshots made by app - e.g. visibility failures), and the app is being uninstalled – they just get lost in the middle. Artifacts plugin prints many warnings about missing files.

This is why here I improve the E2E test suite for this edge cases and make sure `uninstallApp()` does not affect the artifacts collection.

- In this pull request, I have implemented an artifact relocation mechanism for two plugins: UI hierarchy and screenshot plugin. If the app is about to be uninstalled, they will relocated all the pending artifacts (not saved and not deleted ones).

### DIFF
--- a/detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
+++ b/detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
@@ -1,3 +1,4 @@
+const fs = require('../../utils/fsext');
 const log = require('../../utils/logger').child({ __filename });
 const FileArtifact = require('../templates/artifact/FileArtifact');
 const temporaryPath = require('../utils/temporaryPath');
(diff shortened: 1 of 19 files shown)


detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
```diff
@@ -36,16 +37,31 @@ class SimulatorScreenshotPlugin extends ScreenshotArtifactPlugin {
...
+    const artifacts = {
+      visibilityFailingScreenshots: params.visibilityFailingScreenshotsURL,
+      visibilityFailingRects: params.visibilityFailingRectsURL,
+    };
 
-    this._registerSnapshot('visibilityFailingRects/', new FileArtifact({
-      temporaryPath: visibilityFailingRectsURL
-    }));
+    for (const [key, artifactPath] of Object.entries(artifacts)) {
+      if (!artifactPath || await fs.isDirEmpty(artifactPath)) {
```
noomorph, line 57: For some reason, I can see dirs sent, but they are empty in fact. Maybe a bit later, we should look into it on iOS.

d4vidi:
@noomorph so to recap - this happens with artifacts actively generated by the tests through various API's, in particular, right? Otherwise I'd have to naively ask whether app uninstallation and artifacts capturing are not in sync... Is that question relevant, at all?

d4vidi:
In any case, could you please briefly describe the flow that creates this problem?

noomorph:
@d4vidi 

```js
// --take-screenshots failing

test('create problem', async () => {
  await element(by.id('doesNotExist')).tap();
});

afterEach(async () => {
  await device.uninstallApp();
});
```

will log:

```
WARN:  [MOVE_FILE_MISSING] did not find temporary file: /Users/yaroslavs/Library/Developer/CoreSimulator/Devices/6371830E-8B44-422B-A52C-DB0B62E6B6E3/data/Containers/Data/Application/C0C5C5DA-72F2-4D3A-8AC5-C6D3BCF98262/tmp/A41D9E22-8D22-47AD-8B98-F96872BDB29A.png
```

because the app folder (that contains that artifact) is deleted together with the app, and it is too late.

PR #2961, opened 2021-08-31
fix: concurrency issues caused by #2958

noomorph:
## Description

This is a follow-up for https://github.com/wix/Detox/pull/2958

It fixes two race condition issues that have been noticed during testing.

### DIFF
--- a/detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
+++ b/detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
@@ -38,23 +38,26 @@ class SimulatorScreenshotPlugin extends ScreenshotArtifactPlugin {
   }
 
   async onBeforeUninstallApp(event) {
-    const snapshots = [
-      ...Object.values(this.snapshots.fromTest),
-      ...Object.values(this.snapshots.fromSession)
-    ];
+    await this.api.requestIdleCallback(async () => {
+      const snapshots = [
+        ...Object.values(this.snapshots.fromTest),
+        ...Object.values(this.snapshots.fromSession)
+      ];
 
-    await Promise.all(snapshots.map(s => s && s.relocate()));
+      await Promise.all(snapshots.map(s => s && s.relocate()));
+    });
+
     await super.onBeforeUninstallApp(event);
   }
 
-  async _onInvokeFailure({ params }) {
+  _onInvokeFailure({ params }) {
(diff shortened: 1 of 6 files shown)


detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
```diff
@@ -38,23 +38,26 @@ class SimulatorScreenshotPlugin extends ScreenshotArtifactPlugin {
...
-  async _onInvokeFailure({ params }) {
+  _onInvokeFailure({ params }) {
     const artifacts = {
       visibilityFailingScreenshots: params.visibilityFailingScreenshotsURL,
       visibilityFailingRects: params.visibilityFailingRectsURL,
     };
 
     for (const [key, artifactPath] of Object.entries(artifacts)) {
-      if (!artifactPath || await fs.isDirEmpty(artifactPath)) {
+      if (!artifactPath || fs.isDirEmptySync(artifactPath)) {
```
noomorph, line 60: Since the callback is SYNC, it is a bad idea to use async functions.

That omission caused a side-effect when artifacts of the current tests, due to prolonged asynchronous execution of `_onInvokeFailure` callback, got assigned to the next test context.

I had to switch to `fs.readdirSync` under the hood, to block the execution in a reliable way for this callback.

detox/src/artifacts/screenshot/SimulatorScreenshotPlugin.js
```diff
@@ -38,23 +38,26 @@ class SimulatorScreenshotPlugin extends ScreenshotArtifactPlugin {
   }
 
   async onBeforeUninstallApp(event) {
-    const snapshots = [
-      ...Object.values(this.snapshots.fromTest),
-      ...Object.values(this.snapshots.fromSession)
-    ];
+    await this.api.requestIdleCallback(async () => {
```
noomorph, line 41: The problem here is that all the screenshot and video artifacts are being saved in an unhurried one-by-one and very asynchronous way (historically, due to Android's `adb pull` considerations) → that would mean that at the time we get into this callback, there would be some delayed move/copy operations which cannot be easily accessed.

This is why I am forcing the plugin to block the main flow execution, so that all the pending operations are completed and then I continue blocking the flow until all the remaining artifacts get relocated to a safe location (outside of the application folder).

Only after that we resume the flow.

PR #2990, opened 2021-09-23
feat(device): add .permanent option to app launch args

noomorph:
## Description

> :shrug: ...it appears that Detox **clears `device._currentAppLaunchArgs` every time you select an app.**
> In other words, this `appLaunchArgs` object is bound to the lifecycle of the current app. No app = no extra launch args. 
> @d4vidi, I wonder if we can add eventually some, ahem…, more global app launch args that can be persisted between app relaunches.
> Otherwise, it is going to be a bit ugly and verbose, as you can see – any mock server that implants myAppMockServerPort: 24xx  to default app launch args will need the same call again and again, i.e.:

```javascript
await device.selectApp('app1');
await device.appLaunchArgs.modify({ ourMockServerPort: 4242 });
await device.selectApp('app2'); // unfortunately, it resets appLaunchArgs
await device.appLaunchArgs.modify({ ourMockServerPort: 4242 }); // so let's do it again, 🤦 
```

In this pull request, I have added an ability to persist app launch args between `device.selectApp(name)` calls, e.g.:

```javascript
device.appLaunchArgs.modify({ ourMockServerPort: 9999 }, { permanent: true });
await device.selectApp('app2');
device.appLaunchArgs.get(); // { ourMockServerPort: 9999 }
device.appLaunchArgs.reset();
device.appLaunchArgs.get(); // { ourMockServerPort: 9999 }
device.appLaunchArgs.reset({ permanent: true });
device.appLaunchArgs.get(); // { }
```

> _For API changes:_


d4vidi:
@noomorph I hate to say this, especially in retrospect, but I'm not very fond of this new form of the API. I think the effect of using the `permanent` arg across the launch-arg methods (modify, set & get) is a bit too implicit: it doesn't convey clearly enough the fact that the actions are in fact associated with a different (sub)-set of the lunch-args.

Also, I'm not sure `permanent` is the right way to describe what it does - at least from the _user_'s perspective. I think that for a user, the point is that the argument is cross-app; It being permanent is somewhat an implementation detail, if I'm not mistaken.

Could we take some time to revisit?

d4vidi:
I'm thinking here that the distinction between transient and permanent args is in fact strong enough so as to introduce something in the form of `device.baseLaunchArgs` with a similar API. WDYT? Just musing

d4vidi:
@noomorph ping 😬

PR #2991, opened 2021-09-23
code: fix review comments to #2826

noomorph:
## Description

Reacting to a batch of 4-month old code review comments from #2826.
Better late than never.

### DIFF
--- a/detox/src/devices/Device.js
+++ b/detox/src/devices/Device.js
@@ -1,9 +1,9 @@
 const _ = require('lodash');
 
 const DetoxRuntimeError = require('../errors/DetoxRuntimeError');
-const cutStackTraces = require('../utils/cutStackTraces');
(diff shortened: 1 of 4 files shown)


detox/src/devices/Device.js
```diff
@@ -17,7 +17,7 @@ class Device {
     sessionConfig,
     runtimeErrorComposer,
   }) {
-    cutStackTraces(this, [
+    wrapWithStackTraceCutter(this, [
```
d4vidi, line 20: This is great ❤️
Could we get rid of the methods list, too? :prettyplease:

noomorph: No TypeScript, no candies, sorry.

d4vidi: Lol...
A simple `filter()` lambda with a distinction between "public" and "private" (`startsWith('_')`) predicate?!

PR #2992, opened 2021-09-25
Resolve UIWindowScene _keyWindowScene unrecognized selector

calebmackdavenport:
## Description

Please feel free to edit anything you need to, just would like to get https://github.com/wix/Detox/issues/2895 resolved.

- This pull request addresses the issue described here: 
https://github.com/wix/Detox/issues/2895

In this pull request, I have
Resolved an iOS 15 issue:
image

Reference: https://stackoverflow.com/a/58447461/8665723 thanks to @hcwesson for the link
https://github.com/baronapp/Detox as well
and #2895 (comment) for initial solution

Working example shown below or [here](https://github.com/calebmackdavenport/Detox/tree/demo-react-native-nav)
Would love to add the example as well to improve ability to test similar issues in the future, but would prefer a quick resolution and a separate discussion for the example.

---

> _For features/enhancements:_
 - n/a

> _For API changes:_
 - n/a
 

https://user-images.githubusercontent.com/30021449/135011870-cfeb3ea3-f894-4425-99d0-a4aa2d1761ff.mov

Issue #2895, opened 2021-07-14
support iOS 15


alon-ha:
We need to support iOS 15, check if anything broke

d4vidi:
@alon-ha does this include Applesimutils?

alon-ha:
Yes, in case we need to update there something as well

d4vidi:
@alon-ha something to look into this week?

alon-ha:
> @alon-ha something to look into this week?

I think so, it will be good time to check it

mickmaccallum:
👋  Hello everyone! I just wanted to call attention to the fact that the Apple event where the iOS 15 release date is expected to be announced is next week. If anyone at Detox could find time to look into this it would be much appreciated.

d4vidi:
Thanks @mickmaccallum, we will do our best!

Kudo:
had some investigation for xcode 13.0 beta5 + ios 15 simulator today. there was a detox crash like this:

```
2021-09-13 18:17:46.144 Df BareExpoDetox[51267:2b20a1] (CoreFoundation) *** Terminating app due to uncaught exception 'NSUnknownKeyException', reason: '[<_NSMainThread 0x6000025c42c0> valueForUndefinedKey:]: this class is not key value coding-compliant for the key private.'
*** First throw call stack:
(
        0   CoreFoundation                      0x000000011adb88a8 __exceptionPreprocess + 242
        1   libobjc.A.dylib                     0x00000001135ecbe7 objc_exception_throw + 48
        2   CoreFoundation                      0x000000011adb8549 -[NSException init] + 0
        3   Foundation                          0x000000011614d38e -[NSObject(NSKeyValueCoding) valueForUndefinedKey:] + 225
        4   Foundation                          0x00000001161fba76 -[NSThread valueForUndefinedKey:] + 85
        5   Foundation                          0x000000011614c16e -[NSObject(NSKeyValueCoding) valueForKey:] + 307
        6   Foundation                          0x000000011614cba8 -[NSObject(NSKeyValueCoding) valueForKeyPath:] + 244
        7   Detox                               0x00000001136bfd02 __DTXHandleCrash + 130
        8<…>
```

[...]

PR #2996, opened 2021-09-28
fix(ios): select key window in ios15-compatible way

mikehardy:
## Description

- This pull request addresses the issue described here: #2895

In this pull request, I have followed the general outline from this stackoverflow https://stackoverflow.com/a/58447461/9910298 to find the key window (and key window's scene) in a way that works on ios15 now that they have removed the hidden ios internal that was previously used.

- I do not believe this requires any documentation or typescript change
- It may be correct to remove these internals lines now as well but I was less sure on that: https://github.com/wix/Detox/blob/25c240b157bc4bbb98e4da94c55f141594dceb51/detox/ios/Detox/AppleInternals/DTXAppleInternals.h#L51-L52

This PR serves as a possibly-working but minimal version of the patch I originally authored here https://github.com/wix/Detox/issues/2895#issuecomment-926181391 that @calebmackdavenport incorporated here #2992, but #2992 includes some whitespace changes which I don't understand and also includes a new example which (as a maintainer on other repos) I fear will slow down the PR incorporation, as demos are typically unrelated to bug fixes and add maintenance burden to a repo so slow down PR review etc

I just want to see ios15 support merged, so if Caleb saw success with the actual fix I made (that went in to 2992) and nothing else was required, then I'd love a clean PR to be merged ASAP :-)

### DIFF
--- a/detox/ios/Detox/Utilities/NSObject+DetoxUtils.m
+++ b/detox/ios/Detox/Utilities/NSObject+DetoxUtils.m
@@ -371,7 +371,7 @@ BOOL __DTXPointEqualToPoint(CGPoint a, CGPoint b)
 {
 	NSMutableDictionary* rv = [NSMutableDictionary new];
 	
-	rv[@"viewHierarchy"] = [[UIWindowScene _keyWindowScene] dtx_recursiveDescription];
+ 	rv[@"viewHierarchy"] = [[UIWindow dtx_keyWindowScene] dtx_recursiveDescription];
 	
 	NSMutableArray* windowDescriptions = [NSMutableArray new];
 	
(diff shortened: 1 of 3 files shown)


mikehardy:
I note that the ios Jenkins task succeeded :sparkles: , there appears to be a problem with android but this PR touched no Android files so they cannot be related

alon-ha:
@mikehardy Thank you for the PR, I chose to merge this [one](https://github.com/wix/Detox/pull/2992).
Both were basically identical but I think the first one created before. 
Anyway it will fix the issue which is the important part. Closing this one

mikehardy:
Thanks! merging either works for me, Caleb and I were collaborating on it. I will say that I have not been able to successfully run Detox on ios15 yet even with the patch. I suspect it is my local environment as Caleb (from the other PR) reports success, but just a heads up

alon-ha:
I'm sure there are others issue in iOS15 as well. We had a long holidays here.. I will catch up soon with iOS15 issues regarding detox

Issue #2601, opened 2021-01-19
Fail tests verbosely in case of concurrent invocations


d4vidi:
## Description

It's been more than once that users have approached us with "weird issues" which root cause ended up being a missing `await` over a Detox expectation or interaction.

Sometimes, such a human error can be relatively easy to detect by going over the code. However, at times, this error can appear very elusive, such as in this (highly simplistic) example:

```js
async function tapOnOldButton() {
  // ...
  console.log('Tapped on old button');
}

async function tapOnNewButton() {
  // ...
  console.log('Tapped on new button');
}

async function tapOnButton(isNew) {
  await isNew ? tapOnNewButton() : tapOnOldButton();
}

it('blah blah', async () => {
  // ...
  await tapOnButton(true);
  console.log('Test is complete');
});

```
The resulting output is likely to be:
```
Test is complete
Tapped on new button
```

Which is not what the user intended for it to be, obviously.

As a much-needed enhancement, Detox can introduce a means to detect the invocation of an expectation/interaction while another one is in-flight. Upon detection, it would fail the test with a very descriptive error, hinting that there's been an ignored interaction promise and that most likely, an `await` is missing.

With this kind of a feature, we would be:
1. Improving the experience of testers and their overall satisfaction, esp. the ones running complex suites.
2. Reducing the time needed for support.

noomorph:
I can certainly expand the `AsyncWebSocket` implementation so it has two queues: sync requests (99%) and async requests (CurrentStatus, as far as I can tell). It can throw errors if we are attempting to send something while the app has not responded to the previous something.

d4vidi:
Sounds good - best to have this solved on the JS (tester) side. However, if not feasible, I'm confident we could sort this out on the device' end.

PR #3003, opened 2021-10-03
Detect pending in-flight requests

jonathanmos:
## Description
Detects pending in-flight messages and throws an exception if they are detected: 'Attempting to send a new interaction without responding to the previous one. Previous interaction may be missing await'. 

Sync messages ('currentStatus') are specifically excluded from this check. 

This pull request addresses the issue described here: https://github.com/wix/Detox/issues/2601


detox/src/client/AsyncWebSocket.js
```diff
@@ -124,6 +129,14 @@ class AsyncWebSocket {
     return _.some(this.inFlightPromises, p => p.message.type !== 'currentStatus');
   }
 
+  hasMultiplePendingActions() {
+    const remainingAfterRemovingCheckIdleMessages = _.filter(this.inFlightPromises, p =>
```
noomorph, line 133: Not correct. You should also exclude `captureViewHierarchy`, `setRecordingState`, `takeScreenshot`.

Also, you should never block `cleanup` action, I guess. @alon-ha , could you advise too?

detox/src/client/AsyncWebSocket.test.js
```diff
@@ -296,8 +308,8 @@ describe('AsyncWebSocket', () => {
   describe('.rejectAll()', () => {
     it(`should throw error to all pending promises`, async () => {
```
noomorph, line 309: The name of the test looks outdated.

noomorph:
@jonathanmos, also please make sure that the error stack traces point exactly to the line where that invocation happens, e.g.:



The picture above is taken out of the context, from some older pull request. I am just illustrating what I mean.

detox/src/client/AsyncWebSocket.js
```diff
@@ -124,6 +129,14 @@ class AsyncWebSocket {
...
+      p.message.type !== 'currentStatus'
+    );
+
+    return remainingAfterRemovingCheckIdleMessages.length > 1;
```
d4vidi, line 137: I think you can utilize `_.some()` instead of `_.filter()`

detox/src/client/AsyncWebSocket.js
d4vidi: Naming... please use a shorter variable name here 😄

detox/src/client/AsyncWebSocket.js
```diff
@@ -95,6 +95,11 @@ class AsyncWebSocket {
 
     const messageId = message.messageId;
     const inFlight = this.inFlightPromises[messageId] = new InflightRequest(message).withTimeout(options.timeout);
+
+    if (this.hasMultiplePendingActions()) {
+      throw  new DetoxRuntimeError('Attempting to send a new interaction without responding to the previous one. Previous interaction may be missing await');
```
d4vidi, line 100: I think should be revised, to a more detox-ish jargon (is that a thing?). More in the direction of:
```
Detox has detected multiple interactions taking place simultaneously. Have you forgotten an await in your test code?
```

noomorph (approved):

detox/src/client/Client.js
```diff
@@ -131,22 +131,18 @@ class Client {
...
 
   _inferSendOptions(action) {
-    if ( action instanceof actions.CurrentStatus
-      || action instanceof actions.Login
-      || action instanceof actions.Cleanup
-    ) {
-      return { queryStatus: false, timeout: 5000 };
-    }
+    const timeout = action.timeout;
+    const shouldQueryStatus = timeout === 0;
```
d4vidi, line 143: What about plain old `null` usage instead of 0, for `timeout`? 😄

detox/src/client/actions/actions.js
```diff
@@ -14,6 +14,16 @@ class Action {
       throw new DetoxInternalError(`was expecting '${type}' , got ${JSON.stringify(response)}`);
     }
   }
+
+  /** @returns {boolean} */
+  get canBeConcurrent() {
```
d4vidi, line 19: Rename suggestion (sponsored by naming-is-hard.com):
canBeConcurrent = not overly descriptive and drives logic negation in evaluation code.
How does `isAtomic` sound? Meaning, should be executed with no interruptions? This may be the Java dev in me talking...

d4vidi (changes requested):
@jonathanmos great job!
I Have just a few, subtle yet reasonable, notes... please review. Otherwise, approved.

d4vidi (approved):

Issue #3033, opened 2021-10-24
Upgrade to RN 0.66 (Android)


d4vidi:
## Description

Add/verify support for the React Native 0.66 release:

- [x] Check (and migrate) Detox test app on developer env
- [ ] Upgrade demo apps (demo-react-native, then check demo-react-native-jest and detox-react-native-detox-instruments)
- [x] Upgrade associated CI job config to always use 0.66
- [x] Update support table in main README.md

jonathanmos:
https://github.com/facebook/react-native/issues/32594

FadiAboMsalam:
Any update on this 🤔  ?

d4vidi:
@FadiAboMsalam yes, #3035 is review-pending. Should probably land in by the end of the week 👍🏻

Issue #2899, opened 2021-07-14
Android 12 and Java 11


d4vidi:
## Description

Need to validate compliance with the upcoming release of Android 12

mikehardy:
Testing on Android 12 today and while I needed to manually set up the adb reverse to hit the bundler (unexpected, maybe not Detox' problem - `adb reverse tcp:8081 tcp:8081`) everything in the react-native-firebase e2e tests appeared to pass with Detox current-as-of-now - v18.20.4

Note that we are not doing anything fancy with regard to artifacts and dialogs and permissions and such, but it does at least run, vs ios15 which is currently a crash-on-startup per #2895

d4vidi:
Thanks @mikehardy, that's great input.

mikehardy:
If you just need to bump your app to targetSdkVersion 31, this diff will work as a patch file named `patches/detox+18.23.1.patch` in your apps for [patch-package](https://github.com/ds300/patch-package), proposed in #3055 and in use now in react-native-firebase's e2e test harness

```
diff --git a/node_modules/detox/android/detox/build.gradle b/node_modules/detox/android/detox/build.gradle
index 8bef864..3689459 100644
--- a/node_modules/detox/android/detox/build.gradle
+++ b/node_modules/detox/android/detox/build.gradle
@@ -86,12 +86,12 @@ dependencies {
     // Versions are in-sync with the 'androidx-test-1.2.0' release/tag of the android-test github repo,
     // used by the Detox generator. See https://github.com/android/android-test/releases/tag/androidx-test-1.2.0
     // Important: Should remain so when generator tag is replaced!
-    api('androidx.test.espresso:espresso-core:3.3.0') { // Needed all across Detox but also makes Espresso seamlessly provided to Detox users with hybrid apps/E2E-tests.
+    api('androidx.test.espresso:espresso-core:3.4.0') { // Needed all across Detox but also makes Espresso seamlessly provided to Detox users with hybrid apps/E2E-tests.
         exclude group: 'com.google.code.findbugs', module: 'jsr305'
     }
-    api 'androidx.test.espresso:espresso-web:3.3.0' // Web-View testing
-    api 'androidx.test:rules:1.2.0' // Needed because of ActivityTestRule. Needed by users *and* internally used by Detox.
-    api 'androidx.test.ext:junit:1.1.1' // Needed so as to seamlessly provide AndroidJUnit4 to Detox users. Depends on junit core.
[...]

PR #3055, opened 2021-11-04
fix(android, api31): use api31 compatible transitive androidx deps

mikehardy:
## Description

This PR enables apps using Detox to raise targetSdkVersion to 31

The activities exported by the libraries in the updated dependency
versions correctly specify "android:exported" attributes, so apps
that use Detox may successfully bump their targetSdkVersion to 31

Related #2899 - Detox already ran on Android12 but apps using Detox could not target API31
Related #3033 - CI failures without JDK11 and is updating lots of things at once, this is focused

#3033 is useful but looks like it will take a while, this unblocks Detox-using apps right now

### DIFF
--- a/detox/android/detox/build.gradle
+++ b/detox/android/detox/build.gradle
@@ -83,15 +83,15 @@ dependencies {
 // All are aligned with this release: https://developer.android.com/jetpack/androidx/releases/test#1.2.0
 dependencies {
 
-    // Versions are in-sync with the 'androidx-test-1.2.0' release/tag of the android-test github repo,
-    // used by the Detox generator. See https://github.com/android/android-test/releases/tag/androidx-test-1.2.0
+    // Versions are in-sync with the 'androidx-test-1.4.0' release/tag of the android-test github repo,
+    // used by the Detox generator. See https://github.com/android/android-test/releases/tag/androidx-test-1.4.0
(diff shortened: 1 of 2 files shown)


mikehardy:
Sadly, this fails react-native 0.59 detox testing as well: https://jenkins-oss.wixpress.com/job/detox-android-59-9-pr/2397/console - it may be that react-native 0.59 is an intractable problem at this point and minimum compatible react-native could be raised

mikehardy:
Not sure what is going on with RN59 Jenkins, I was able to build the app successfully locally following the CI script :shrug: 

You do need the second commit that bumps up the android gradle plugin in use, as far as I can tell, but whatever the compileSdkVersion issue is I can't figure out. Spending any time at all on react-native 0.59 feels like time wasted at this point so I'm going to leave it.

mikehardy:
For anyone that needs detox to work on an app using targetSdkVersion 31 you may use the patch-package patch from this change here: https://github.com/invertase/react-native-firebase/pull/5835/files#diff-c3c4ac892ab301a8340adc456d7acef27b5ece166b125c2acd237628f4a83539

It does work!

I expect this PR to superceded by #3033 though (which is great, that's a full upgrade of everything including example apps here) so the PR itself may not go anywhere even though the change is functional and works

PR #3063, opened 2021-11-11
SyncStatusFormatter: initial commit.

asafkorem:
Add `SyncStatusFormatter` in which will be used for formatting app status responses in the JS side, as a part of the unification of synchronization logs.

---

### Examples for _Android_ sync-status formatted messages:
```
detox[9160] INFO:  [APP_STATUS] The app seems to be idle
```
```
detox[9160] INFO:  [APP_STATUS] The app is busy with the following tasks:
• "LooperIdlingResource-2215-mqt_native_modules" (Native Modules Thread) is executing (native module calls).
• "LooperIdlingResource-2214-mqt_js" (JS Thread) is executing (JavaScript code).
• UI elements are busy:
  - Reason: UI rendering activity.
```

### Examples for _iOS_ sync-status formatted messages:

```
detox[9733] INFO:  [APP_STATUS] The app is busy with the following tasks:
• Run loop "JS Run Loop" is awake.
• There are 2 work items pending on the dispatch queue: "Main Queue (<OS_dispatch_queue_main: com.apple.main-thread>)".
• The event "Runloop Perform Block" is taking place with object: "JS Run Loop".
• The event "Runloop Perform Block" is taking place with object: "JS Run Loop".
• Run loop "Main Run Loop" is awake.
• The event "React Native (bundle load)" is taking place.
```

```
detox[9733] INFO:  [APP_STATUS] The app is busy with the following tasks:
• There are 1 work items pending on the dispatch queue: "Main Queue (<OS_dispatch_queue_main: com.apple.main-thread>)".
• Run loop "Main Run Loop" is awake.
• 1 enqueued native timers:
  - Timer #1:
    + Fire date: 2021-11-11 14:19:57 +0200.
    + Time until fire: 0.072.
    + Repeat interval: 0.
    + Is recurring: NO.
```

For a broader view, see the branch with all unification changes: [here](https://github.com/wix/Detox/pull/3065).


detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+      throw new DetoxRuntimeError(`Given sync status is invalid, app is busy but busy-resources are not defined`);
+    }
+    const resourcesDescriptions = SyncStatusFormatter.resourcesDescriptionsFromJSON(jsonStatus.busy_resources);
+    return `The app is busy with the following tasks:\n${resourcesDescriptions.join('\n')}`;
+  }
+
+  static isValidJSONStatus(jsonStatus) {
+    const ajv = new Ajv();
+    const isValidStatus = ajv.compile(SyncStatusFormatter.jsonStatusSchema());
+    return isValidStatus(jsonStatus);
```
noomorph, line 26: Why don't you compile it in advance to use just the function, `isValidStatus`?

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  static jsonStatusSchema() {
+    return {
```
noomorph, line 30: I wonder maybe we can extract it to a JSON file aside `SyncStatusFormatter.js` and `require()` it?

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: You're right, this should be done outside in advance. Nice catch!

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+          }
+        }
+      },
+      required: ['app_status'],
+      additionalProperties: false
+    };
+  }
+
+  static isAppIdle(jsonStatus) {
+    return jsonStatus.app_status === 'idle' ||
```
noomorph, line 56: Is there a reason to introduce the ambiguity?
`isAppIdle({ app_status: 'busy' })` will return true. Feels weird a bit, could you elaborate?

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+        return SyncStatusFormatter.looperResourceDescriptionFromProperties(properties);
+
+      case 'io':
+        return SyncStatusFormatter.resourceTitle(`Disk I/O activity.`);
+
+      case 'unknown':
+        return SyncStatusFormatter.unknownResourceDescriptionFromProperties(properties);
+
+      default:
+        throw new DetoxRuntimeError(`Given sync status is invalid, cannot find resource name: \`${resourceName}\``);
```
noomorph, line 109: You should not use `DetoxRuntimeError` for things that our users don't bear any responsibility. Use `DetoxInternalError`.

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+        app_status: { 'enum': ['busy', 'idle'] },
+        busy_resources: {
+          type: 'array',
+          items: {
+            type: 'object',
+            properties: {
+              name: { 'type': 'string' },
+              description: {
+                type: 'object',
+                additionalProperties: true
```
noomorph, line 42: Why won't you use JSON schema deeper for this object?
https://stackoverflow.com/questions/35450405/how-to-represent-sum-union-types-in-in-json-schema

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+    return descriptions;
+  }
+
+  static resourceDescriptionFromJSON(jsonDescription) {
+    const resourceName = jsonDescription.name;
+    const properties = jsonDescription.description;
+
+    switch (resourceName) {
+      case 'delayed_perform_selector':
+        return SyncStatusFormatter.delayedPerformSelectorResourceDescriptionFromProperties(properties);
```
noomorph, line 76: Wouldn't it be easier to have a map?

```js
const formatter = this._formatters[resourceName];
if (!formatter) {
  throw ...
}
return formatter(properties);
```

where:

```js
this._formatters = {
  delayed_perform_selector: () => {...},
  dispatch_queue: () => {...}
};
```

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,277 @@
...
+    return SyncStatusFormatter.resourceTitle(`There are ${workItemsCount} work items pending on the ` +
+      `dispatch queue: "${queue}".`);
+  }
+
+  static runLoopResourceDescriptionFromProperties(properties) {
+    const runLoopName = SyncStatusFormatter.getPropertyFromObject('name', properties);
+    return SyncStatusFormatter.resourceTitle(`Run loop "${runLoopName}" is awake.`);
+  }
+
+  static oneTimeEventDescriptionFromProperties(properties) {
```
noomorph, line 145: The entire object looks static.

Please just use `const` and `function` instead of having a class.

detox/src/client/actions/SyncStatusFormatter.test.js
```diff
@@ -0,0 +1,379 @@
+const format = require('./SyncStatusFormatter').formatJSONStatus;
+
+describe('Sync Status Formatter', () => {
```
noomorph, line 3: I'd suggest using snapshot tests in the majority of the cases here.

noomorph (changes requested):

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: great, done

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: well, perhaps it would be better to assert in case the app is `busy` and the `busy_resources` list is empty.
done

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: OK, got you :+1:

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: mmm why not.. that's indeed better, I'll do it
also thinking about this we can enforce `busy_resource` to be non-empty from the schema and not from the

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: both easy, but I agree that map will look better here :)
done

detox/src/client/actions/SyncStatusFormatter.test.js
asafkorem: done

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: sounds great. done

asafkorem:
Fixed, squashed and rebased.

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,186 @@
...
+  const executionType = properties.execution_type;
+  return resourceTitle(`${properties.thread} is executing` +
+    `${executionType !== undefined ? ` (${executionType}).` : `.`}`);
+}
+
+function unknownResourceDescriptionFromProperties(properties) {
+  return resourceTitle(`Resource "${properties.identifier}" is busy.`);
+}
+
+module.exports = { formatJSONStatus };
```
d4vidi, line 186:
```suggestion
module.exports = formatJSONStatus;
```

(this will have side-effects)

detox/src/client/actions/SyncStatusFormatter.js
```diff
@@ -0,0 +1,186 @@
...
+  for (const jsonDescription of jsonDescriptions) {
+    const description = resourceDescriptionFromJSON(jsonDescription);
+    descriptions.push(description);
+  }
+
+  return descriptions;
+}
+
+const resourceFormatters = {
+  delayed_perform_selector: delayedPerformSelectorResourceDescriptionFromProperties,
```
d4vidi, line 42: I think that maybe these should be extracted onto separate files (without necessarily breaking the unit test into several ones). I think it would help with:
- Shortening the various function names (for example, `delayedPerformSelectorResourceDescriptionFromProperties` => `propertiesFormatter.delayedPerformSelector`).
- Putting things like `uiResourceCounterNameMapping` more intimately alongside their concern.

d4vidi (changes requested):
Looks REALLY good! Some notes, nevertheless...

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: can you elaborate please?

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: Sounds great :+1:

asafkorem:
Fixed and squashed.

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: discussed :+1:

noomorph (approved):

detox/src/client/actions/SyncStatusFormatter.js
asafkorem: done, extracted to separated formatters.

asafkorem:
Fixed and rebased.

PR #3069, opened 2021-11-12
--retries: change `at least` to `at most`

arnaudambro:
According to [jest-circus documentation](https://jestjs.io/docs/jest-object#jestretrytimes)

> Runs failed tests n-times until they pass or until the max number of retries is exhausted.

## Description
Just update the readme


noomorph:
Good catch! 😉 
Thanks, I'll merge your right after this one: https://github.com/wix/Detox/pull/3068

PR #3082, opened 2021-11-16
Update demo-plugin example to fit Detox 19

d4vidi:
## Description

- This pull request addresses the issue described here: #2980

In this pull request, I have aligned the 3rd party / plugin driver example to match the changes in Detox 19.

---

> _For API changes:_


detox/src/client/Client.js
```diff
@@ -63,8 +63,12 @@ class Client {
     return this._serverUrl;
   }
 
+  async open() {
+    return this._asyncWebSocket.open();
+  }
```
d4vidi, line 68: @noomorph please let me know if you are comfortable with this change.
Apparently, when introducing the pluggable drivers, they made a hidden assumption that the web socket is available to the outside world. I'm highly against keeping it that way, yet suppose we ought to make the Client useable from within the driver, nonetheless.

examples/demo-plugin/driver.js
```diff
@@ -1,7 +1,10 @@
-const DeviceDriverBase = require('detox/src/devices/drivers/DeviceDriverBase');
+const DeviceDriverBase = require('detox/src/devices/runtime/drivers/DeviceDriverBase');
```
noomorph, line 1: Just so you're aware, it is a bad practice to allow plugins to go deep inside the module: either Detox should export some endpoint like `require('detox/plugin-api')`, or export that stuff at the top level `const { plugin } = require('detox')`.

noomorph (approved):
Go for it!

PR #3097, opened 2021-11-22
docs: rewrite the mocking guide

noomorph:
## Description

1. I guess react-native-repackager mention is blatantly outdated.
2. Then, it is not clear what patterns are for the following scenarios:

* **Scenario 1**: You have two self-contained implementations: `module.js`  and `module.e2e.js`.
* **Scenario 2**: You need to replace a tiny function in `module.js` – how to reuse the 99% of the shared code.

Related discussions: https://github.com/wix/Detox/discussions/2935

### DIFF
--- a/docs/Guide.Mocking.md
+++ b/docs/Guide.Mocking.md
@@ -7,94 +7,10 @@ sidebar_label: Mocking
 
 ## Mocking
 
-> This guide is a bit out of date. We hope to have it updated soon.
-
-Mocking is an important part of testing. You may want to alter some behavior of your app during test and replace it with a mock. Here are some example reasons why this could be useful:
-
-- Change server endpoints to point to a mock/staging server instead of the regular production server
-- Stub a feature the simulator doesn’t support
-- Prepare mock environment data like GPS position, Contacts/Photos found on the device, etc
-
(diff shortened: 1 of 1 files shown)


ruben00james:
This is an interesting article i found about mocking launchImageLibrary from react-native-image-picker when testing with Detox
https://medium.com/geekculture/how-to-mock-react-native-image-picker-using-detox-9f7ced41b67f

noomorph:
I'm likely to take it next week for as a part of the guild week.

d4vidi:
not stale

noomorph:
yeah, just a guild week arriving late :)

d4vidi:
not stale

noomorph:
Godspeed, today I'm going to do it. 🙏

d4vidi:
@noomorph this work is 🤯<sup>^10</sup>

d4vidi:
I do feel nevertheless that at least on a fundamental level, there should be at least some motivational description of why and when mocking can be helpful. It may probably overlap what we already have under [LaunchArgs](https://wix.github.io/Detox/docs/api/launch-args#motivation) but it would probably be worth it. WDYT?

Issue #3107, opened 2021-11-25
docs: improve getting started


noomorph:
**Is your feature request related to a problem? Please describe.**

See the current page: https://wix.github.io/Detox/docs/introduction/ios

![easy-to-say](https://user-images.githubusercontent.com/1962469/143428060-5ffbe7f9-76f1-4bf2-b568-42c64c6f1c3c.png)

**Describe the solution you'd like**

Describe somehow less vaguely that the user needs to write a command similar to this one:

```
xcodebuild -workspace ios/MyProject.xcworkspace -configuration Debug -scheme MyProject -sdk iphonesimulator -derivedDataPath ios/build
```

asafkorem:
@noomorph in the introduction doc, we refer the user to the configuration API guide:
![image](https://user-images.githubusercontent.com/55082339/143442516-48e641c6-7d35-44d8-9880-7eb2fa002971.png)

https://wix.github.io/Detox/docs/api/configuration#detox-cli:

> - Building with `xcodebuild`:
> 
>   ```sh
>   xcodebuild -project ios/YourProject.xcodeproj -scheme YourProject -sdk iphonesimulator -derivedDataPath ios/build
>   ```

However, I agree that we can (and should) be clearer there.

PR #3109, opened 2021-11-25
docs(Introduction.iOS): add `build` and `binaryPath` to the example.

asafkorem:
Add a `xcodebuild` command example and `binaryPath` example to the provided configuration example. No reason to have a placeholder if that's an example (and not a template), and the configuration keys are explain immediately after.

In response to this issue: https://github.com/wix/Detox/issues/3107.

### DIFF
--- a/docs/Introduction.iOS.md
+++ b/docs/Introduction.iOS.md
@@ -27,9 +27,10 @@ Whether you’ve selected to apply the configuration in a  `.detoxrc.json` or bu
   },
   "apps": {
     "ios.release": {
+      "name": "YourProject",
       "type": "ios.app",
-      "binaryPath": <path to .app bundle>,
-      "build": <xcodebuild command>
+      "binaryPath": "ios/build/Build/Products/Release-iphonesimulator/YourProject.app",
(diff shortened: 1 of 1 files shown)


docs/Introduction.iOS.md
```diff
@@ -41,10 +42,14 @@ Whether you’ve selected to apply the configuration in a  `.detoxrc.json` or bu
 }
 ```
 
+:::info
+
 For a comprehensive explanation of Detox configuration, see our [dedicated API-reference guide](APIRef.Configuration.md).
 
+:::
+
 In the above configuration example, make sure to provide the correct information for your project/app. Under the key `"binaryPath"`, you should provide the path of the .app bundle to use. Normally, this is the path where the `"build”` command would output this bundle. Under the key `"build"`, specify the `xcodebuild` command for your project.
```
noomorph, line 51: @asafkorem , I wonder if you can take a screenshot of Xcode starting the build of the project and imply that a user can copy the command from there? Not critical, but still I couldn't help but ask.

asafkorem: I can't, the `xcodebuild` command is not invoked by Xcode when building the app from there.

noomorph (approved):
Thanks, @asafkorem

PR #3111, opened 2021-11-25
Validate proper app vs. test APK config

### LINKED ISSUE #2521: Better ux for wrong usage of binaryPath
A follow-up on #2436. We already handle the case where the target APK isn't available - with a suitable, descriptive error, but not where it's the _wrong_ APK.

We can try to see whether `aapt` generates output that we can use in order to assert the integrity of the APK. Alternatively, we can warn if the package to install ends with a `test` suffix compared to the core package name.

**Additional context**
N/A

cc @noomorph

d4vidi:
## Description

- This pull request resolves #2521.

In this pull request, I have incorporated new APK type validations:
- The file configured as the app APK (i.e. under `binaryPath`) should NOT in fact accidentally be the instrumentation test APK.
- The file either configured as the test APK (i.e. under `testBinaryPath`) or automatically resolved as such, should in fact be an instrumentation testing APK.

Given validation errors, descriptive warnings are issued:

image

---

> _For API changes:_


detox/src/devices/common/drivers/android/tools/apk.js
```diff
@@ -0,0 +1,29 @@
+const path = require('path');
```
d4vidi, line 1: This was actually moved (and renamed) from APKPath

detox/src/devices/common/drivers/android/tools/apk.test.js
```diff
@@ -0,0 +1,30 @@
+const path = require('path'); // Using path methods will normalize slashes to backslashes on win32, so tests must match.
```
d4vidi, line 1: This was actually a rename of APKPath.test.js

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -233,20 +233,45 @@ class AndroidDriver extends DeviceDriverBase {
...
   _getTestApkPath(originalApkPath) {
-    const testApkPath = APKPath.getTestApkPath(originalApkPath);
+    const testApkPath = apkUtils.getTestApkPath(originalApkPath);
 
     if (!fs.existsSync(testApkPath)) {
-      throw new DetoxRuntimeError(`'${testApkPath}' could not be found, did you run './gradlew assembleAndroidTest'?`);
+      throw new DetoxRuntimeError({
+        message: `The test APK could not be found at path: '${testApkPath}'`,
+        hint: 'Try running the detox build command, and make sure it was configured to execute a build command (e.g. \'./gradlew assembleAndroidTest\')\n' +
+          'For further assistance, visit the Android setup guide: https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md',
```
asafkorem, line 273: It may be better to refer to the website doc page instead of the GH one:
https://wix.github.io/Detox/docs/introduction/android

detox/src/devices/common/drivers/android/exec/AAPT.test.js
```diff
@@ -1,70 +1,89 @@
...
+    jest.mock('../../../../../utils/exec');
+    exec = require('../../../../../utils/exec').execWithRetriesAndLogs;
+    givenAAPTResult('');
+
+    jest.mock('../../../../../utils/pipeCommands');
+    escape = require('../../../../../utils/pipeCommands').escape.inQuotedString;
+    escape.mockImplementation(mockEscapeFunc);
 
-    AAPT = require('./AAPT');
+    const AAPT = require('./AAPT');
```
asafkorem, line 32: I think this should be extracted from the `beforeEach`.

detox/src/devices/runtime/drivers/android/AndroidDriver.js
d4vidi: I agree but I didn't want to שעטנז it :-)
we should probably change things across the board, if we ever.

detox/src/devices/common/drivers/android/exec/AAPT.test.js
```diff
@@ -1,70 +1,89 @@
...
+    jest.mock('../../../../../utils/environment');
+    environment = require('../../../../../utils/environment');
+    environment.getAaptPath.mockResolvedValue(mockAAPTPath);
+
+    jest.mock('../../../../../utils/exec');
+    exec = require('../../../../../utils/exec').execWithRetriesAndLogs;
+    givenAAPTResult('');
+
+    jest.mock('../../../../../utils/pipeCommands');
+    escape = require('../../../../../utils/pipeCommands').escape.inQuotedString;
```
asafkorem, line 29: Same goes for the other requires. Why is it inside the `beforeEach`?

asafkorem (changes requested):

detox/src/devices/common/drivers/android/tools/apk.js
```diff
@@ -0,0 +1,29 @@
...
+
+function getTestApkPath(originalApkPath) {
+  const originalApkPathObj = path.parse(originalApkPath);
+  let tempPath = originalApkPathObj.dir.split(path.sep);
+  const splitFileName = originalApkPathObj.name.split('-');
+
+  const buildType = _.last(splitFileName);
+  const flavorDimensions = _.slice(splitFileName, 1, splitFileName.length - 1);
+
+  tempPath = _.dropRight(tempPath, 1); // buildType
```
asafkorem, line 15: the `flavourDimensions` statement should move closer to its usage

detox/src/devices/common/drivers/android/tools/apk.js
```diff
@@ -0,0 +1,29 @@
...
+
+  let flavorDimensionsPath = '';
+  if (flavorDimensions.length > 0) {
+    flavorDimensionsPath = string.lowerCamelCaseJoin(flavorDimensions);
+    tempPath = _.dropRight(tempPath, 1); // flavorDimensions
+  }
```
asafkorem, line 21: can be extracted to a nullable method IMO `getFlavorDimensionsPath(splitFileName)`, then:
```js
const flavorDimensionsPath = getFlavorDimensionsPath(..);
tempPath = flavorDimensionsPath ? _.dropRight(tempPath) : tempPath;

const finalFileName = ...

const pathComponents = [path.join(tempPath), 'androidTest', flavorDimensionsPath, buildType, finalFileName].filter(Boolean);
return path.join(pathComponents);
```

asafkorem: WDYT?

asafkorem (commented):
good stuff btw!

a few more suggestions, might be a little picky

detox/package.json
```diff
@@ -119,7 +119,6 @@
       "src/devices/allocation/factories",
       "src/devices/allocation/factories/drivers",
       "src/devices/cookies",
-      "src/devices/common/drivers/android/exec/AAPT.js",
```
noomorph, line 122: This is nice!

detox/src/devices/runtime/drivers/android/AndroidDriver.js
noomorph: I tend to agree with @asafkorem. Amit, you're adding here a *new* link to the error description, so he's right.

d4vidi: As a Detox user, would feel odd if Detox were to refer me to the github docs and the website intermittently. Nevertheless, my opinion here is not super strong, so I will update this if you insist on it :-)

detox/src/devices/common/drivers/android/exec/AAPT.test.js
d4vidi: Discussed f2f

detox/src/devices/common/drivers/android/exec/AAPT.test.js
d4vidi: Discussed f2f

detox/src/devices/common/drivers/android/tools/apk.js
d4vidi: I agree!

detox/src/devices/common/drivers/android/tools/apk.js
```diff
@@ -0,0 +1,29 @@
+const path = require('path');
+
+const _ = require('lodash');
+
+const string = require('../../../../../utils/string');
+
+function getTestApkPath(originalApkPath) {
+  const originalApkPathObj = path.parse(originalApkPath);
```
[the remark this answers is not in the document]
d4vidi: Well, technically, `path.parse` does [return an object](https://nodejs.org/api/path.html#pathparsepath). But I can relate to what you're saying here. Obj is too obscure, even so.

d4vidi:
@asafkorem I decided not to boy-scout with respect to your comments on the apk utils file -- at least not this time around. Apart from the rename, the file itself has no true meaning in this change...

d4vidi:
@asafkorem @noomorph would appreciate another round of reviews, as things have somewhat changed fundamentally in the last commit.

detox/src/devices/common/drivers/android/tools/ApkValidator.js
```diff
@@ -0,0 +1,52 @@
+const DetoxRuntimeError = require('../../../../../errors/DetoxRuntimeError');
+
+const setupGuideHint = 'For further assistance, visit the Android setup guide: https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md';
```
asafkorem, line 3: I really think that we should change all of our links to the website. However, you can keep it as you want for now and we can change it in a separated PR with all other links. I opened an issue (https://github.com/wix/Detox/issues/3117) for that, but it should be an easy change.

Note that there are already several links to the web so the experience you described is already broken :)

asafkorem: If you think otherwise let's discuss it on this in the issue, I don't think it should block us with this PR.

detox/src/devices/common/drivers/android/tools/ApkValidator.test.js
```diff
@@ -0,0 +1,54 @@
...
+    it('should validate the APK is not a test APK', async () => {
+      givenAAPTResult(false);
+      await uut.validateAppApk(binaryPath);
+      expect(aapt.isTestAPK).toHaveBeenCalledWith(binaryPath);
+    });
+
+    it('should throw a descriptive error if app APK happens to be a test APK', async () => {
+      givenAAPTResult(true);
+      await expect(uut.validateAppApk(binaryPath)).rejects.toThrowErrorMatchingSnapshot();
+    });
```
asafkorem, line 29: I'd add expectation in all tests (`expect(aapt.isTestAPK).toHaveBeenCalledWith(binaryPath);`)

detox/src/devices/common/drivers/android/tools/ApkValidator.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  async validateAppApk(binaryPath) {
+    let isAppApk;
+
+    try {
+      isAppApk = !await this._aapt.isTestAPK(binaryPath);
+    } catch (e) {
+      throw this._composeAaptBasedError(e, binaryPath, 'binary');
+    }
```
asafkorem, line 17: not tested

asafkorem (approved):
Minor comments. Besides that, looks very good! 🚀 🎸

detox/src/devices/common/drivers/android/tools/ApkValidator.test.js
d4vidi: Good catch!

d4vidi: Ohh, I though you meant something else. Simply adding as-is would be redundant :-/

detox/src/devices/common/drivers/android/tools/ApkValidator.js
d4vidi: It's covered by "should throw a specific error if AAPT throws" (https://github.com/wix/Detox/pull/3111/files/558eba81d47d302aa4e258910f267343497f8711..42c7aed553a47cff000f05f6bf1093735fb51d9c#diff-e41e8194f7736d596a564bc5396f164b42ff84d5c6e7893f9071042733b5c7f4R33)

Also note that we force 100% coverage so without coverage here we'd have red builds

detox/src/devices/common/drivers/android/tools/ApkValidator.js
d4vidi: Let's do it in the scope of #3117 (great that it's there, btw)

detox/src/devices/common/drivers/android/tools/ApkValidator.js
asafkorem: Cool, I missed this one 👍🏼

PR #3112, opened 2021-11-26
build: update linting

noomorph:
## Description

This pull request is purely auxiliary, without any breaking changes. 🍸 

1. Updates `eslint` to `^8` in `detox` project.
2. Adds `eslint@7` to `detox/test` project (cannot use `^8` due to: https://github.com/facebook/react-native/issues/32528). 😱 It turns out, we had no linting in the test project at all – it made me very frustrated because it just looks dirty and inconsistent.
3. **The huge change**. Switched ESLint parser to TypeScript for both the projects. Consider it to be a baby step towards TypeScript adoption across the project.
4. Adds [eslint-plugin-unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn). The primary reason is because I want to see warnings for `TODO: comments` in the code (btw, it supports expiring TODOs), but it also contains some nice to have things.

It consists of two commits:
1. The core commit – it changes only some test files and obviously ESLint/Typescript config files (don't worry, TypeScript files are `noEmit` ones, it physically can't impact anything). There are new devDependencies in `package.json`s.
2. The huge technical commit: it adds to each file that does not pass TypeScript checks this kind of comment:
    ```diff
    +// @ts-nocheck
    const path = require('path');
    ...
    ```
    
I have validated 2 manually two times, it has no surprises.

## Factual changes

We'll get a bit more warnings:

* around 40 warnings in `detox` project. Mostly, about `unused` variables.
* around 6 warnings in `detox/test` project. Mostly, about skipped tests.

Hence, again, if CI checks pass, there's nothing scary or wrong about this PR. I worked very hard to make it as less intrusive as possible.


noomorph:
Looks like there's some Node.js version conflict on Jenkins, I'll look later. `TypeError: Module.createRequire is not a function`

PR #3114, opened 2021-11-29
fix: prevent a retry loop from hanging up the test runners

### LINKED ISSUE #2481: Stabilize exiting from Jest when device allocation takes forever
At the moment, on the JS side we have an over-simplified logic for booting Android devices:

```js
  async _waitForBootToComplete(deviceId) {
    await retry({ retries: 240, interval: 2500 }, async () => {
      const isBootComplete = await this.adb.isBootComplete(deviceId);
      if (!isBootComplete) {
        throw new DetoxRuntimeError({
          message: `Android device ${deviceId} has not completed its boot yet.`,
        });
      }
     });
```

I'm not sure what's about iOS device booting, but this comment might relevant there as well.

The fact that there is no built-in reaction to `device.cleanup()` calls, as I am afraid, will result in uncontrolled `jest-circus` timeouts because nothing can abort that retry loop. I don't think we want to have leaky exits with that `--detectOpenHandles` yellow message, which worsens the developer experience by a mile.

Personally, I witnessed 5-8 times when an Android emulator was stuck forever when requested to boot, so I am confident I'm talking about realistic use cases. Let's find some time eventually to think how to fix this.

/cc @d4vidi

noomorph:
## Description
- This pull request resolves #2481

In this pull request, I have:

* added `shouldUnref` option to `sleep` utility
* added `shouldUnref ` option to `retry` utility
* added `shouldUnref: true` to the `retry()` call for Android emulator
* updated unit tests
* checked manually the end result.

To simulate the hanging emulator I monkey-patched `detox/src/devices/common/drivers/android/exec/BinaryExec.js` and forced it to use this script:

```bash title='emulator'
#!/usr/bin/env bash

while read -r line;
do
   echo "$line" ;
done
```

### Before

Screenshot 2021-11-29 at 17 39 43

### After

Screenshot 2021-11-29 at 17 40 09

### DIFF
--- a/detox/src/utils/retry.js
+++ b/detox/src/utils/retry.js
@@ -1,5 +1,6 @@
 const sleep = require('./sleep');
 
+const DEFAULT_UNREF = false;
 const DEFAULT_INITIAL_SLEEP = 0;
 const DEFAULT_RETRIES = 9;
 const DEFAULT_INTERVAL = 500;
@@ -23,12 +24,13 @@ async function retry(optionsOrFunc, func) {
     backoff = DEFAULT_BACKOFF_MODE,
     conditionFn = DEFAULT_CONDITION_FN,
     initialSleep = DEFAULT_INITIAL_SLEEP,
+    unref = DEFAULT_UNREF,
   } = options;
 
   const backoffFn = backoffModes[backoff]();
 
   if (initialSleep) {
-    await sleep(initialSleep);
+    await sleep(initialSleep, { unref });
   }
 
   // eslint-disable-next-line no-constant-condition
@@ -41,7 +43,7 @@ async function retry(optionsOrFunc, func) {
       if (!conditionFn(e) || (totalTries > retries)) {
         throw e;
       }
-      await sleep(backoffFn({ interval, totalTries }));
+      await sleep(backoffFn({ interval, totalTries }), { unref });
     }
   }
 }
--- a/detox/src/utils/retry.js
+++ b/detox/src/utils/retry.js
@@ -1,6 +1,5 @@
 const sleep = require('./sleep');
 
-const DEFAULT_UNREF = false;
 const DEFAULT_INITIAL_SLEEP = 0;
 const DEFAULT_RETRIES = 9;
 const DEFAULT_INTERVAL = 500;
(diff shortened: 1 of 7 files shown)


detox/src/utils/retry.js
```diff
@@ -1,5 +1,6 @@
 const sleep = require('./sleep');
 
+const DEFAULT_UNREF = false;
```
asafkorem, line 3: you already have a default value for this method in its signature.. why to define it in another place? 
let's have a single source of truth

detox/src/utils/sleep.js
```diff
@@ -1,5 +1,12 @@
-async function sleep(ms) {
-  return new Promise(resolve => setTimeout(resolve, ms));
+const defaultOptions = { unref: false };
+
+async function sleep(ms, options = defaultOptions) {
```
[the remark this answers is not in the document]
asafkorem: also, I'd call this argument `shouldUnref` instead of `unref`

detox/test/package.json
```diff
@@ -39,7 +39,7 @@
   "devDependencies": {
     "@babel/core": "^7.12.9",
     "@babel/runtime": "^7.12.5",
-    "@types/node": "^14.14.20",
+    "@types/node": "^12.20.37",
```
asafkorem, line 42: why do we need to downgrade the requirement?

detox/src/utils/retry.js
asafkorem: In a second thought, it has some sense to separate between the configurations (I mean that `DEFAULT_UNREF ` is the specific default value of `retry.js`), but since you already have in tests the check that we don't use `unref`, it isn't necessary to explicitly define it in the implementation.

detox/test/package.json
noomorph: Because Detox supports Node.js 12 minimally. That was erroneous to orientate ourselves on types that might be not available in minimally claimed version to be supported by Detox.

detox/src/utils/sleep.js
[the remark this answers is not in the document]
noomorph: Because positional arguments tend to grow. I prefer to have them named instead of checking every time what is `fn(value, true, false, 20, 'ads')`.

detox/src/utils/retry.js
noomorph: Passing now `undefined`.

detox/src/utils/sleep.test.js
```diff
@@ -0,0 +1,24 @@
...
+  it('should not call .unref() by default', async () => {
+    await sleep(0);
+    expect(setTimeoutMock.mock.results[0].value.unref).not.toHaveBeenCalled();
+  });
+
+  it('should call .unref() when given options = { shouldUnref: true }', async () => {
+    await sleep(0, { shouldUnref: true });
+    expect(setTimeoutMock.mock.results[0].value.unref).toHaveBeenCalled();
+  });
+});
```
d4vidi, line 24: Nicely done

Issue #2834, opened 2021-06-14
Interactive touch outside React-Native-Navigation Overlay when detox run


hqdai:
### Description
I'm using React-native-navigation (RNN) with an overlay is showing as alert, my issue is when manual test, I can't touch outside the alert message, that is correct, but when detox runs for automation test, it can touch and scroll outside the alert area. 

So, the problem is why Detox can interact outside the overlay area ...

- [x] I have tested this issue on the latest Detox release and it still reproduces.

#### Reproduction

To reproduce, please check out my pet project and just do the test run. You can see detox can scroll even though the alert message is shown on top. Please be noted that I've turned option: interceptTouchOutside: true already. 

https://user-images.githubusercontent.com/19548213/121865961-b1c81d80-cd28-11eb-9f0f-019226435704.mov

[Interactive 2.zip](https://github.com/wix/Detox/files/6646930/Interactive.2.zip)

#### Environment:
 - Detox: 18.16.0
 - React Native: 0.64.1
 - Node: 14.10.1
 - Device: iPhone 12 (Simulator)
 - Xcode: 12
 - iOS: 14.5
 - macOS: 11.3

noomorph:
@hqdai, sorry for replying late.
Thanks for providing the example project.
Hopefully we can check it next week.

hqdai:
> @hqdai, sorry for replying late.
> Thanks for providing the example project.
> Hopefully we can check it next week.

Thank you @noomorph :D, it seems the issue still happens. Please reopen the issue.

d4vidi:
@hqdai Detox has super powers 😄
I wonder whether it is plausible that touch events get somehow associated with specific views (based on interactability & visibility). @alon-ha would you mind taking a look into this sometime next week? There's an example project here ready to go.

alon-ha:
Will check it this week and update

phuongwd:
Not stale,

phuongwd:
Hello @alon-ha,
Could you please give me the an update?

Many thanks

d4vidi:
No updates so far, we apologize.

phuongwd:
> No updates so far, we apologize.

No problems, just checking to update our progress 😊

phuongwd:
Not stale

asafkorem:
Thanks @hqdai for reaching out with this bug report! 
[This PR](https://github.com/wix/Detox/pull/3115) will solve the bug you described.

PR #3115, opened 2021-11-29
fix(scrollTo): raise if scroll view element is not hittable.

asafkorem:
## Description

This pull request addresses the issue described here: #2834.

In this pull request, I have reproduced the bug from the issue, by adding more tests to the test-app. Apparently, only the `scrollTo` didn't work as expected because we don't assert if the called element is hidden by another view or window.

This fixes the bug as we did in all other UI interaction actions (taps, long-presses, date-picking), by raising an assertion when the view is not hittable.

Similar failed tests could be achieved by hiding the scroll view with another view, but I still decided to implement them with `UIWindow` ("overlay") since it is not tested anywhere in our test suite, and that's an iOS-specific behaviour (in Android we touch the visible screen, and not the window behind the element).

Also, in addition to this fix, I rewrote the mechanism of assertions before interacting with a view (for any UI gesture).

### DIFF
--- a/detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
+++ b/detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
@@ -290,9 +290,9 @@ if(isnan(normalizedStartingPoint.main) || normalizedStartingPoint.main < 0 || no
 	
 	CGPoint startPoint = CGPointMake(safeAreaToScroll.origin.x + safeAreaToScroll.size.width * normalizedStartingPoint.x, safeAreaToScroll.origin.y + safeAreaToScroll.size.height * normalizedStartingPoint.y);
 
+	CGPoint viewPoint = [self convertPoint:startPoint fromView:self.window];
+	[self dtx_assertHittableAtPoint:viewPoint];
 
-	[self dtx_assertHittableAtPoint:startPoint];
-	
 	NSUInteger successfullyAppliedScrolls = 0;
 	while (offset.x != 0.0 || offset.y != 0.0)
 	{
--- a/detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
+++ b/detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
@@ -289,6 +289,9 @@ if(isnan(normalizedStartingPoint.main) || normalizedStartingPoint.main < 0 || no
 	}
 	
 	CGPoint startPoint = CGPointMake(safeAreaToScroll.origin.x + safeAreaToScroll.size.width * normalizedStartingPoint.x, safeAreaToScroll.origin.y + safeAreaToScroll.size.height * normalizedStartingPoint.y);
+
+
+	[self dtx_assertHittableAtPoint:startPoint];
 	
 	NSUInteger successfullyAppliedScrolls = 0;
 	while (offset.x != 0.0 || offset.y != 0.0)
--- a/detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
+++ b/detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
@@ -291,7 +291,7 @@ if(isnan(normalizedStartingPoint.main) || normalizedStartingPoint.main < 0 || no
 	CGPoint startPoint = CGPointMake(safeAreaToScroll.origin.x + safeAreaToScroll.size.width * normalizedStartingPoint.x, safeAreaToScroll.origin.y + safeAreaToScroll.size.height * normalizedStartingPoint.y);
 
 	CGPoint viewPoint = [self convertPoint:startPoint fromView:self.window];
-	[self dtx_assertHittableAtPoint:viewPoint];
+	[self dtx_assertScrollableAtPoint:viewPoint];
 
 	NSUInteger successfullyAppliedScrolls = 0;
 	while (offset.x != 0.0 || offset.y != 0.0)
@@ -308,7 +308,16 @@ if(isnan(normalizedStartingPoint.main) || normalizedStartingPoint.main < 0 || no
 	DTXViewAssert(strict == NO || successfullyAppliedScrolls > 0, self.dtx_elementDebugAttributes, @"Unable to scroll %@ in “%@”", _DTXScrollDirectionDescriptionWithOffset(offset), self.dtx_shortDescription);
 	
 	self.dtx_disableDecelerationForScroll = NO;
-//	self.bounces = oldBounces;
+}
+
+- (void)dtx_assertScrollableAtPoint:(CGPoint)startPoint {
+  NSError *error;
+  DTXAssert([self dtx_isHittableAtPoint:startPoint error:&error],
+			@"View is not scrollable at the given start point. Start point (view coordinates): %@" \
+			"\n- Full error: %@\n----\nMore about scroll action API here: " \
+			"https://wix.github.io/Detox/docs/api/actions-on-element/" \
+			"#scrolloffset-direction-startpositionx-startpositiony",
+			NSStringFromCGPoint(startPoint), error.localizedDescription);
 }
 
 @end
--- a/detox/ios/Detox/Utilities/UIResponder+First.m
+++ b/detox/ios/Detox/Utilities/UIResponder+First.m
@@ -1,0 +1,26 @@
+//
+//  UIResponder+First.m
+//  Detox
+//
+//  Created by Asaf Korem on 02/12/2021.
+//  Copyright © 2021 Wix. All rights reserved.
+//
+
+#import "UIResponder+First.h"
+
+static __weak UIResponder *_dtx_first;
+
+@implementation UIResponder (First)
+
++ (instancetype)dtx_first {
+  _dtx_first = nil;
+  [[UIApplication sharedApplication] sendAction:@selector(responderAction:) to:nil from:nil
+									   forEvent:nil];
+  return _dtx_first;
+}
+
+- (void)responderAction:(id)sender {
+  _dtx_first = self;
+}
+
+@end
--- a/detox/ios/Detox/Utilities/UIView+DetoxUtils.m
+++ b/detox/ios/Detox/Utilities/UIView+DetoxUtils.m
@@ -28,29 +28,11 @@
 DTX_DIRECT_MEMBERS
 @implementation UIView (DetoxUtils)
 
-- (void)dtx_assertHittable
-{
-	[self _dtx_assertHittableAtPoint:self.dtx_accessibilityActivationPointInViewCoordinateSpace isAtActivationPoint:YES];
-}
-
 - (void)dtx_assertVisibleAtRect:(CGRect)rect percent:(nullable NSNumber *)percent {
 	NSError* error;
 	BOOL assert = [self dtx_isVisibleAtRect:rect percent:percent error:&error];
-	
+
 	DTXViewAssert(assert, self.dtx_elementDebugAttributes, @"%@", error.localizedDescription);
-}
-
-- (void)dtx_assertHittableAtPoint:(CGPoint)point
-{
-	[self _dtx_assertHittableAtPoint:point isAtActivationPoint:NO];
-}
-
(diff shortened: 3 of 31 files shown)


detox/test/e2e/35.overlay.test.js
```diff
@@ -0,0 +1,46 @@
+const { expectToThrow } = require('./utils/custom-expects');
+
+describe(':ios: Overlay', () => {
```
noomorph, line 3: I'll run tomorrow, thanks!

noomorph:
@asafkorem maybe I'll mention the obvious, but this change should pass pre-release lifecycle on our Wix internal projects, I guess. Nightly RCs for all the projects.

detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
```diff
@@ -266,6 +266,8 @@ - (void)dtx_scrollWithOffset:(CGPoint)offset normalizedStartingPoint:(CGPoint)no
 
 - (void)_dtx_scrollWithOffset:(CGPoint)offset normalizedStartingPoint:(CGPoint)normalizedStartingPoint strict:(BOOL)strict
 {
+	[self dtx_assertVisibleWithPercent:@1];
+
```
asafkorem, line 270: I don't like this.
However, I don't like the current approach also (>=75% visible), what we use in other views actions ([see for example](https://github.com/wix/Detox/blob/260e32a9b50f1a38f21bfbce7461147da8733625/detox/ios/Detox/Actions/UIPickerView%2BDetoxActions.m#L17)), since the view can be half visible and still accessible..

Also, the `normalizedStartingPoint` is undefined in many cases and then we use some almost-arbitrary point (from one of the edges). Theoretically I could use this point to check whether it's hittable on that point, but then it might result in breakage of tests (where this points are hidden but the view is partially visible, like in `03.actions-scroll.test.js`).

asafkorem: It might be the best we can offer for this bug fix, of-course I'd love to hear more thoughts 😄

asafkorem:
@wix/detox-team I'll appreciate if one of you can review this change :)

detox/ios/Detox/Utilities/UIView+DetoxUtils.m
```diff
@@ -216,7 +216,7 @@ - (BOOL)_dtx_isTestedRegionObscuredWithVisiblePixels:(NSUInteger)visible
 	
 	if (isRegionObscured) {
 		*explanation = [NSString stringWithFormat:@"View does not pass visibility percent "
-						"threshold (%@)", [DetoxPolicy percentDescriptionForValue:percent]];
+						"threshold (%lu)", (unsigned long)percent];
```
d4vidi, line 219: 👍🏻

detox/ios/Detox/Utilities/UIView+DetoxUtils.m
```diff
@@ -277,14 +277,11 @@ - (BOOL)_dtx_testVisibilityInRect:(CGRect)rect percent:(NSUInteger)percent
 	CGRect testedRegionInWindowCoords = [windowToUse convertRect:rect fromView:self];
 	
 	CGRect visibleBounds = self.dtx_visibleBounds;
-	
-	if (CGRectIsNull(visibleBounds) || [self _dtx_isRegionObscured:visibleBounds
-		 										  fromTestedRegion:self.dtx_visibleBounds
-														   percent:percent]) {
+	if (CGRectIsNull(visibleBounds) || CGRectIsEmpty(visibleBounds) ||
```
d4vidi, line 280: Apple say [`GCRecIsEmpty`](https://developer.apple.com/documentation/coregraphics/1454917-cgrectisempty) checks for nullability. Does that render `GCRectIsNull` redundant?

detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
d4vidi: I agree that adding the ≥75 assertion offhand would be reckless, and also will not make sense as long as the percentage cannot be specified by the user.

As for the idea you suggested - checking hittability, I wouldn't get it off the table altogether. In any case, on Android we essentially do something quite similar with respect to visibility -- asserting the view is non-empty.

detox/test/src/Screens/OverlayScreen.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default class OverlayScreen extends Component {
+  render() {
+    return (
+      <ScrollView style={styles.container} testID='VerticalScrollView'>
+        <TouchableOpacity onPress={() => NativeModule.presentOverlayWindow()} style={styles.button} testID='ShowOverlayButton'>
+          <Text style={styles.text}>Show Overlay</Text>
+        </TouchableOpacity>
+
+        <View style={styles.item}><Text style={styles.itemText}>Text1</Text></View>
```
d4vidi, line 14: Are the `View` wrappers a must?

detox/test/e2e/35.overlay.test.js
```diff
@@ -0,0 +1,46 @@
...
+      await showOverlayButton.tap();
+      await expectToThrow(() => showOverlayButton.tap());
+    });
+  });
+
+  describe('scroll view', () => {
+    it('should be intractable when overlay is not shown', async () => {
+      await verticalScrollView.scrollTo('bottom');
+      await expect(showOverlayButton).not.toBeVisible();
+    });
```
d4vidi, line 39: I think is not necessary because it is tested in the actions suite

detox/test/e2e/35.overlay.test.js
```diff
@@ -0,0 +1,46 @@
...
+
+    it('should not be intractable when overlay is shown', async () => {
+      await showOverlayButton.tap();
+      await expectToThrow(() => verticalScrollView.scrollTo('bottom'));
+    });
```
d4vidi, line 44: Somehow this test failed on my computer (scrolling worked)

detox/test/ios/example/NativeModule.m
```diff
@@ -87,4 +87,19 @@ @implementation NativeModule
...
 
+RCT_EXPORT_METHOD(presentOverlayWindow) {
+    static UIWindow *overlayWindow;
+
+    dispatch_async(dispatch_get_main_queue(), ^{
+        CGRect screenBounds = UIScreen.mainScreen.bounds;
+        overlayWindow = [[UIWindow alloc] initWithFrame:screenBounds];
+
+        [overlayWindow setWindowLevel:UIWindowLevelStatusBar];
+        [overlayWindow setBackgroundColor:[UIColor.grayColor colorWithAlphaComponent:0.5]];
```
d4vidi, line 98: Would it make sense to challenge further and use a 0 alpha?

detox/test/e2e/35.overlay.test.js
asafkorem: Basically you are right but not really, since I'm testing the whole screen in this test I made sure that we start from the expected state (where the button disappears after scrolling). This covers the case where the scroll view might be not-scrollable (for example, too few items on some screen).

detox/test/e2e/35.overlay.test.js
asafkorem: maybe `detox clean-framework-cache && detox build-framework-cache` is required?
I introduced a new native method (for presenting the overlay) in this change.

detox/test/ios/example/NativeModule.m
asafkorem: Holy s**t, that's an excellent question!
If the overlying view has a transparent background (or, `clearColor`), we consider the underlying view as visible. So in this case the views are considered as "visible" but they are not tappable / scrollable. 

So it makes a lot of sense, but the tests will break, but we need to cover these cases also since they shouldn't (touch gestures should not pass).

We must have some better mechanism here than just asserting on visibility.

detox/ios/Detox/Utilities/UIView+DetoxUtils.m
asafkorem: definitely right.

detox/test/e2e/35.overlay.test.js
d4vidi: It appear so 💪🏻

detox/test/e2e/35.overlay.test.js
d4vidi: Still kind of sounds like this: https://github.com/wix/Detox/blob/2c6478d6137bd0f3b3d3e92ee30fa82ad6f1d365/detox/test/e2e/03.actions-scroll.test.js#L33

asafkorem: I disagree, I'm not testing the functionality of a scroll view, but the expectation from the initial state of this screen. Let's discuss about this f2f?

asafkorem:
@d4vidi fixed and squashed. I rewrote the mechanism of is-hittable check on iOS. 

The guiding logic is:
- If not point was given, finds a visible point of the view (otherwise the view is not hittable).
- Finds the topmost window, tries to tap on the visible point, on the topmost window using `hitTest`.
- Checks whether the hitted view is the view itself / descendent of the tested view.

detox/ios/Detox/Actions/UIScrollView+DetoxActions.m
asafkorem: Done ✔️ we will now check hittability instead of visibility.
There's a lot of changes in the new implementation, so this may be effective to review them face to face.

asafkorem: of course, we should also test it with the latest changes as broadly as possible using the internal wix projects. especially when it adds more assertions to touch actions.

detox/ios/Detox/Utilities/UIResponder+First.m
```diff
@@ -0,0 +1,26 @@
+//
+//  UIResponder+First.m
+//  Detox
+//
+//  Created by Asaf Korem on 02/12/2021.
+//  Copyright © 2021 Wix. All rights reserved.
+//
+
+#import "UIResponder+First.h"
```
d4vidi, line 9: Ref stackoverflow

d4vidi:
- Descriptive error for when default start point is not hittable (i.e. use offset arg + ref the docs)
- Add missing assertions in [scrolling test with offsets](https://github.com/wix/Detox/blob/7d5a2bb9e3b75da0752740192a66b033c33af29a/detox/test/e2e/03.actions.test.js#L164)
- Add expect-to-fail scroll test -- when overlay is shown, in `03.actions.test.js`
- UIScreen etc.

detox/ios/Detox/Utilities/UIResponder+First.m
asafkorem: v

asafkorem:
Changes pushed

> Descriptive error for when default start point is not hittable (i.e. use offset arg + ref the docs)

Done ✔️ 

> Add missing assertions in scrolling test with offsets

Done ✔️ 

> Add expect-to-fail scroll test -- when overlay is shown, in 03.actions.test.js

Done ✔️

detox/test/e2e/03.actions.test.js
```diff
@@ -167,7 +167,10 @@ describe('Actions', () => {
     await element(by.id('toggleScrollOverlays')).tap();
 
     await element(by.id('ScrollView161')).swipe('up', 'slow', NaN, 0.9, 0.95);
-    await expect(element(by.text('Text1'))).not.toBeVisible(1);
+    if (device.getPlatform() === 'ios') {
+      // TODO: investigate why this assertion fails on Android
+      await expect(element(by.text('Text1'))).not.toBeVisible(1);
+    }
```
asafkorem, line 173: This was already TODOed here:
https://github.com/wix/Detox/blob/d95b5e7d35e8e587aa1faa27031a49be5c649a74/detox/test/e2e/03.actions.test.js#L147-L150

@jonathanmos / @d4vidi FYI

asafkorem: Actually, it's related to a very recently reported issue we have, and Amit already answered it:
https://github.com/wix/Detox/issues/3105#issuecomment-986229563

Issue #2534, opened 2020-12-14
Improve app install time by reinstalling on-device APK


d4vidi:
On Android, we've integrated an alternative to the somewhat unstable `adb install`, by manually executing each of the subsequent commands - namely, APK's push and on-device installation.

This opens way to cutting down app installation time even further. Instead of going through the _entire_ process again and again in each suite (where the app is typically reinstalled from scratch), we could skip the APK's push part, and simply install the APK's that are already available on the device itself. Of course, this would have to come with a cost of validating the APK's integrity (e.g. by comparing checksums).

amq:
Why is reinstalling needed at all between test suites?

PR #3118, opened 2021-11-30
Improve Android apk install time

jonathanmos:
Improves the installation time for android. Addresses the following issue: #2534 

Works by testing whether the version that you want to install already exists on the device. If it exists, then it just zeroes the user data. If it exists but in a different version then it uninstalls and then installs, and if it doesn't exist at all then it just installs.


noomorph (changes requested):
While technically the PR idea is great, I strongly dislike the API and the fact it is made public, user-facing. Function calls like `runtimeDevice.existsOnDevice` sound very puzzling.

I'd think in two directions:
1) less likely. We can add a helper getter, `device.app` (or `device.activeApp`, or `device.selectedApp`). There we can put app-specific methods. I bet @d4vidi will say it is too early and out of scope of this change.
2) most likely. We can just fix a bit naming and make all those methods private, at least by underscore convention. 

Last but not least, let's think about OneApp Engine which uses only direct signatures like `device.installApp(path1, path2)` or `device.uninstallApp(bundleId)` – they don't rely on the apps configuration at all. Will your change work for their scenario, can you check?

And one more thing, if I'm building the same version of the same app, just with some patched code, will your `isInstalledWithSameVersion` detect that case?

PR #3124, opened 2021-12-01
refactor: throw DetoxRuntimeError if an invocation fails

noomorph:
## Description

This should allow us to easily distinguish between Detox-generated errors vs something else.

At the moment, we don't – `invoke` errors get wrapped into regular `Error` objects for no good reason.


d4vidi (commented):
Approved :-)

Issue #3084, opened 2021-11-18
Add timestamps for Detox logging


shamilovtim:
**Is your feature request related to a problem? Please describe.**
Timestamps would help a great deal in logging when debugging failures but we currently don't have them:
<img width="371" alt="Screen Shot 2021-11-18 at 12 43 32 PM" src="https://user-images.githubusercontent.com/3495974/142468514-a1bb4404-f4ff-4a96-b87d-ffb65761b011.png">

**Describe the solution you'd like**
The ability for all of the logging to be timestamped with the date/time

**Describe alternatives you’ve considered**
I have considered adding this ourselves with some kind of helper but I think it would be a messier solution than having it part of the CLI/API

noomorph:
@asafkorem, please write down in advance how do you plan to change the configuration options to implement this.

asafkorem:
@noomorph it seems to me that we should add some session configuration flag param: `showDate`. 
With this, we should configure the `bunyanDebugStream` to show dates (see [bunyan-debug-stream#showdate](https://github.com/benbria/bunyan-debug-stream#showdate)), so I'm not sure where do we need that custom formatted function. Let's discuss this.

d4vidi:
I tend to agree with @asafkorem; I'm not convinced as to why this flexibility is required...
That said, if we do pursue this path then a better option would be an explicit _logging_ configuration, IMO.

noomorph:
Implemented in Detox v20 logger. Removing `stale: label`.

PR #3131, opened 2021-12-09
feat(logger): add configuration options from CLI.

asafkorem:
This addresses this issue: https://github.com/wix/Detox/issues/3084, and adds an optional CLI options to configure the logger (whether to show the date, logger name, PID and metadata). 

With this change, dates will be shown by default.

For example, a standard log with the default configurations (after this change):
```
Dec 9 08:56:29 detox[1556] INFO:  Actions: should expect text fields to be focused after tap but not before [OK]
```

With `--logger-show-date false` configured:
```
detox[1556] INFO:  Actions: should expect text fields to be focused after tap but not before [OK]
```


noomorph (changes requested):
It's not that I strongly insist, but nevertheless, I'd recommend taking it into a bit another direction:

```
DETOX_LOGGER_SHOW_DATE=true
DETOX_LOGGER_SHOW_LOGGER_NAME=true
DETOX_LOGGER_SHOW_PID=false
DETOX_LOGGER_SHOW_METADATA=false
```

That makes more sense extensibility-wise.

asafkorem:
@noomorph you suggest also to set `DETOX_LOGGER_SHOW_PID` to be false by default?

noomorph:
> > DETOX_LOGGER_SHOW_PID=false
> 
> @noomorph do you suggest also to set `DETOX_LOGGER_SHOW_PID` to be false by default?

I'm afraid that would be justified only in case when we are positively sure that there's only 1 worker running.

asafkorem:
@noomorph SGTM, done.

asafkorem:
It seems that we are not going to add this change, so I'm closing this PR.
@d4vidi @noomorph

PR #3132, opened 2021-12-10
Update Guide.RunningOnCI for React Native usage with Release configuration

david50407:
## Description

- This pull request is for document enhancement.

In this pull request, I added a tip for CI testing guide.

When using Detox with React Native, DetoxSync dynamic loads React Native symbols to hook it.
But in Release configuration, xcode strips symbols by default, this will cause dlsym() get null pointer and when DetoxSync executes the code, boooom, say hi to SIGSEGV. [code](https://github.com/wix/DetoxSync/blob/master/DetoxSync/DetoxSync/ReactNativeSupport/DTXReactNativeSupport.m#L117)

To solve this problem in easy (but not suitable for production environment) way is to add build config `ENABLE_TESTABILITY=YES` to `xcodebuild` to keep symbols not being striped when using React Native.

### DIFF
--- a/docs/Guide.RunningOnCI.md
+++ b/docs/Guide.RunningOnCI.md
@@ -48,6 +48,10 @@ We will need to create a [release device configuration for Detox](APIRef.Configu
 
 > TIP: Notice that the name `example` above should be replaced with your actual project name.
 
+> **Tip:** You may need to add build config `ENABLE_TESTABILITY=YES` to `xcodebuild` to keep symbols not being striped when you are using React Native.
+> Detox tests app with DetoxSync which dynamic loads React Native symbols to hook RN, so we need to keep these symbols visible.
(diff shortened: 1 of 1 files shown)


asafkorem (changes requested):
Thanks a lot @david50407!
However, I'm not convinced that we need to add this config flag. Can you describe how can I reproduce this SIGSEGV error?

Also, I think it should be under the [configuration options](https://wix.github.io/Detox/docs/api/configuration#detox-cli) doc (detox-cli -> building with xcodebuild), if at all.

david50407:
You can just build the React Native app with Release Configuration with defaulting strip symbols, then check via `nm -m` command that the output binary is not containing the React Native symbol `_RCTGetUIManagerQueue` and then you will get SIGSEGV (when the [code](https://github.com/wix/DetoxSync/blob/master/DetoxSync/DetoxSync/ReactNativeSupport/DTXReactNativeSupport.m#L117) I mentioned got exectued)

PR #3140, opened 2021-12-21
Update Introduction.iOS.md

GregoryNative:
## Description

---

> _For API changes:_

### DIFF
--- a/docs/Introduction.iOS.md
+++ b/docs/Introduction.iOS.md
@@ -54,7 +54,7 @@ In the above configuration example, make sure to provide the correct information
 
 If your app uses Cocoapods, pass `-workspace` in your `xcodebuild` command, for example:
 
-```
+```json
 "build": "xcodebuild -workspace ios/YourProject.xcworkspace -scheme YourProject -sdk iphonesimulator -derivedDataPath ios/build"
 ```
 
--- a/docs/Introduction.iOS.md
+++ b/docs/Introduction.iOS.md
@@ -42,6 +42,11 @@ Whether you’ve selected to apply the configuration in a  `.detoxrc.json` or bu
 }
 ```
 
+If your app uses Cocoapods, pass -workspace to xcodebuild command:
+```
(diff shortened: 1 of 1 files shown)


docs/Introduction.iOS.md
```diff
@@ -42,6 +42,11 @@ Whether you’ve selected to apply the configuration in a  `.detoxrc.json` or bu
 }
 ```
 
+If your app uses Cocoapods, pass -workspace to xcodebuild command:
+```
+"build": "xcodebuild -workspace ios/YourProject.xcworkspace -scheme YourProject -sdk iphonesimulator -derivedDataPath ios/build"
+```
+
```
asafkorem, line 49: Thanks for the tip!

However, there are many more options for building the app with xcodebuild command (like debug / release, passing a config file, etc.), so I'm not very comfortable with a specific elaboration.

But I do agree that most developers are not familiar with this command, so I think it would make more sense to refer the reader to Apple's tech notes regarding xcodebuild:
https://developer.apple.com/library/archive/technotes/tn2339/_index.html#//apple_ref/doc/uid/DTS40014588-CH1-HOW_DO_I_BUILD_MY_PROJECTS_FROM_THE_COMMAND_LINE_

Also, since we are using docusaurus for our docs, it would be nice to wrap this with a pretty [info block](https://docusaurus.io/docs/next/markdown-features/admonitions):

```
:::info
For more information regarding xcodebuild command, visit [Apple Docs](https://developer.apple.com/library/archive/technotes/tn2339/_index.html#//apple_ref/doc/uid/DTS40014588-CH1-HOW_DO_I_BUILD_MY_PROJECTS_FROM_THE_COMMAND_LINE_).
:::

```

asafkorem: Also, I think we should move this reference to come after the sentence "Under the key `"build"`, specify the `xcodebuild` command for your project."

asafkorem (changes requested):
Huge thanks for the clarification!

PR #3142, opened 2021-12-22
Fix RN66 workaround for screens with multiple switches

jonathanmos:
Screens with multiple RnSwitches still hang with RN66 - this pr fixes the issue.


jonathanmos:
retest this please

PR #3146, opened 2021-12-23
fix(ios): remove Xcode 12.5 + M1 workaround call.

asafkorem:
We are not using M1s, so I believe that we don't need the workaround (?)


asafkorem:
@igorgn FYI

asafkorem:
retest this please

asafkorem:
Removing this workaround call fails Jenkins? 🤔 
But we are not using M1 (AFAIK) or Xcode 12.5, so it's very weird.

PR #3145, opened 2021-12-23
fix(ios, UIView+DetoxUtils): locate top-most window at point.

asafkorem:

Issue #3130, opened 2021-12-08
Detox 19.3.0 causes "View is not hittable at its visible point" failures


Vannevelj:
### Description
I'm on Detox 19.1.0. Upgrading to 19.2.0 works just fine and I've got a nice green run of my 120 tests. If I further bump to 19.3.0 then I get a large amount of failures (20+) all with the same error:

```
DetoxRuntimeError: Test Failed: View is not hittable at its visible point. Error: Failed to hit view at point {992, 197} with `hitTest`.
- Origin view: <UIView: 0x7fa90e910b90; frame = (0 0; 1024 1366); autoresize = W+H; layer = <CALayer: 0x600000d31320>>
- Absolute origin: {0, 0}
- Hitten: <UIView: 0x7fa90e910b90; frame = (0 0; 1024 1366); autoresize = W+H; layer = <CALayer: 0x600000d31320>>
- Target view: <RCTImageView: 0x7fa91cfccb30; reactTag: 923; frame = (290 16; 32 32); clipsToBounds = YES; layer = <CALayer: 0x600000d2f220>>
- Relative point: {992, 197}
```

- [x] I have tested this issue on the latest Detox release and it still reproduces

#### Reproduction

Provide the steps necessary to reproduce the issue. If you are seeing a regression, try to provide the last known version where the issue did not reproduce.

1. Upgrade to Detox 19.3.0

#### Environment

 - Detox: 19.3.0
 - React Native: 0.65.1
 - Node: 12.13.0
 - Device: iPad Mini 6th Generation (15.0) & iPad Pro 12.9-inch (5th Generation)
 - Xcode: 13.0
 - iOS: 11.0
 - macOS: 11.6 Big Sur
 - Test-runner (select one): `jest-circus`

#### Device and verbose Detox logs

- [x] I have run my tests using the `--loglevel trace` argument and am providing the verbose log below:

https://gist.github.com/Vannevelj/ee918001bffc5ae4c794cd76f398cff8

asafkorem:
Hi @Vannevelj, we indeed changed the implementation of UI gestures in the latest version by adding an assertion for hittability.

Can you please provide here the test's code and the tested view's code?

asafkorem:
Also, are you able to execute the expected flow manually?

Vannevelj:
The code behind this is quite complicated -- not something I can easily pull out. This video shows what's happening:

[...]

PR #3150, opened 2021-12-27
Revert: is-hittable changes.

asafkorem:
Revert the latest is-hittable changes, since they caused some internal and external projects tests to fail. 
For example, https://github.com/wix/Detox/issues/3130.

Reverted changes:
- https://github.com/wix/Detox/pull/3149
- https://github.com/wix/Detox/pull/3145
- https://github.com/wix/Detox/pull/3115

Those changes were meant to solve this issue https://github.com/wix/Detox/issues/2834, but caused other issues that must be resolved before pushing this fix again.


asafkorem:
retest this please

PR #3153, opened 2021-12-28
fix(iOS,Podfile): fix workaround condition.

asafkorem:
If the React Native version is not defined using the env-arg, we are not overriding the version (currently, we use v0.66).


asafkorem:
CC @igorgn @jonathanmos

Issue #3123, opened 2021-12-01
tests crash when updated from v17 to v19. Signal 11 was raised


fatlinesofcode:
### Description
updated project to xcode 13.1, updated detox to v19. tests crash on startup frequently.
always crashes with maxWorkers set to 2, sometimes passes with maxWorkers 1. We are using react-native-firebase too

- [x ] I have tested this issue on the latest Detox release and it still reproduces

#### Screenshots / Video

```
 FAIL  e2e/000-screens/ShopScreen.e2e.js (11.085 s)
  ShopScreen
    ✕ can tap on ShopPromo and see ShopPromotionsScreen (7425 ms)

  ● ShopScreen › can tap on ShopPromo and see ShopPromotionsScreen

    DetoxRuntimeError: The pending request #-1000 ("isReady") has been rejected due to the following error:

    The app has crashed, see the details below:

    Signal 11 was raised
    (
    	0   Detox                               0x0000000108ffa6c5 +[NSThread(DetoxUtils) dtx_demangledCallStackSymbols] + 37
    	1   Detox                               0x0000000108ffd250 __DTXHandleCrash + 464
    	2   Detox                               0x0000000108ffd991 __DTXHandleSignal + 59
    	3   libsystem_platform.dylib            0x00007fff6bfe7e2d _sigtramp + 29
    	4   ???                                 0x0000000300584d80 0x0 + 12890688896
    	5   CoreFoundation                      0x00007fff2046ccdf -[__NSDictionaryM dealloc] + 128
    	6   libobjc.A.dylib                     0x00007fff2019c9f7 objc_object::sidetable_release(bool, bool) + 177
    	7   libobjc.A.dylib                     0x00007fff20198cb3 _object_remove_assocations + 578
    	8   libobjc.A.dylib                     0x00007fff20196207 objc_destructInstance + 84
    	9   libobjc.A.dylib                     0x00007fff2019c187 -[NSObject dealloc] + 21
    	10  Xxxxx Xxxxxx                       0x0000000104bd71fa -[FBSDKGraphRequestConnection dealloc] + 106
    	11  libobjc.A.dylib                     0x00007fff2019c9f7 objc_object::sidetable_release(bool, bool) + 177
    	12  Xxxxx Xxxxxx                        0x0000000104872d1d __destroy_helper_block_e8_32s + 29
    	13  libsystem_blocks.dylib              0x00007fff2008a434 _Block_release + 130
[...]

PR #3167, opened 2022-01-06
AppleSimUtils: replace zombie-enabled workaround with designated flag.

asafkorem:
Enable `GULGeneratedClassDisposeDisabled` environment flag to disable GoogleUtilities' Swizzler from disposing generated classes when executing Detox tests. This flag works with [GoogleUtilities v7.7](https://github.com/google/GoogleUtilities/blob/main/CHANGELOG.md#770).

This replaces the `NSZombieEnabled` flag, which was used as a workaround.

See: https://github.com/firebase/firebase-ios-sdk/issues/9083.


fatlinesofcode:
@asafkorem issue from #3123 has reappeared with this change

asafkorem:
@fatlinesofcode I see. I will take a look at this today. I guess we need a similar fix with Expo..
Are you using firebase in your project? If so, you need to upgrade your GoogleUtilities to v7.7 for this fix to work.

PR #3174, opened 2022-01-12
fix: port allocation race

noomorph:
## Description

Theoretically, it should fix sporadic issues on CI when there's a high number of workers (currently we've got around 6 for iOS).

screenshot

In this pull request, I have brought almost together two events: when the port number is allocated and when the web socket server starts.


d4vidi:
@noomorph thanks 🚀  I saw the change - could you please explain in a few words what that solves the problem?

Issue #3179, opened 2022-01-19
Property 'text' does not exist in getAttributes


dbilgin:
### Description

When I'm trying to use the example below in a project with Typescript I am getting an error:
```
  const attributes = await element(by.text('Tap Me')).getAttributes();
  const text = attributes.text;
```

`Property 'text' does not exist on type 'IosElementAttributes | AndroidElementAttributes | { elements: IosElementAttributes[]; }'`
The error goes away when I try to force the type to be only `IosElementAttributes | AndroidElementAttributes`. Is there a reason for this? And how should I be using it then?

### Environment

 - Detox: 19.4.2
 - React Native: 0.66.3
 - Node: 14
 - Device: iPhone|Android
 - OS: iOS|Android
 - Test-runner: jest-circus

d4vidi:
@dbilgin that makes sense, because in the case where `getAttributes()` returns an elements array, which is applicable, the `text` property doesn't exist, indeed.

The options I see are for you to either force-cast it as you suggested, or introduce a runtime check:
```ts
if (!('elements' in attributes)) {
    const text = attributes.text;
    // ...
}
```

dbilgin:
@d4vidi wow thank you for such fast response and the PR for the Android part! this seems to work for me, so I'll close the issue.

dbilgin:
Just one question though, out of curiosity @d4vidi, why would there be multiple elements returned when we request the attributes of an element?

d4vidi:
Easiest to explain through examples -
Example 1: A `by.text('submit')` can match multiple submission buttons
Example 2: Usage of [`by.type()`](https://wix.github.io/Detox/docs/api/matchers#bytypeclassname), which is discourage, can easily match all cards in a list.

etc...

dbilgin:
Ah okay, got it, we are passing the result of the query as the "element(s)" after all. Thanks!

PR #3180, opened 2022-01-19
fix(types): Apply Android type to multi-element case of getAttributes() return value

d4vidi:
## Description

- This pull request is associated with #3179

In this pull request, I have applied the `AndroidElementAttributes` type to `getAttributes()`' return value for the `{ element[] }` case.

---

> _For API changes:_


d4vidi:
@noomorph please have a quick glance

Issue #3175, opened 2022-01-17
A quicker way to clean app data than reinstalling


noomorph:
**Is your feature request related to a problem? Please describe.**

Currently, it takes too long to clean the app data the way we usually do it via uninstalling and installing it again.

**Describe the solution you'd like**

Add some way to tell:

```js
await device.cleanAppData();
```

**Describe alternatives you’ve considered**

`device.installApp({ delete: true })`, but it's too slow.

**Additional context**

iOS: https://stackoverflow.com/questions/56228470/ios-simulator-how-to-clear-cache-of-an-app-instead-of-entire-phone
Android: ??

d4vidi:
@noomorph please merge this with https://github.com/wix/Detox/issues/2537 🙏🏻

d4vidi:
@jonathanmos does this in fact align with https://github.com/wix/Detox/issues/2534?

jonathanmos:
@d4vidi Not exactly. We aren't providing an api to clear app data in #2534, rather we're clearing app data behind the scenes where possible to avoid reinstallation.

d4vidi:
But technically `resetAppState()` is going to be a public API, wouldn't it be equivalent?

Issue #2537, opened 2020-12-15
App data clear API


d4vidi:
Add support for Application clear-data, so that apps wont have to be reinstalled on Android (clearing the data is much faster).

LeoNatan:
Isn't this supported by ADB? I thought we are already using it.

d4vidi:
Sort of, but it's not as simple as running a single command.
In any case, we might want to consider an API that would do this seamlessly on Android and iOS (e.g. a `resetApp()` call that would clear data on android, and reinstall on iOS)... does that make sense?

LeoNatan:
Reinstall in iOS doesn’t guarantee data clearance in some cases. I prefer not to offer API that cannot do what the users explicitly is asking for.

d4vidi:
Yes you are right. My aim was to try to avoid spreading `if (isIos) device.launchApp({delete: true}) else device.clearData()` being implemented over and over out there.

d4vidi:
This will be introduced indirectly via #2534

noomorph:
The #3497 did not enter the master, so I think it is totally fine to keep this issue open for a while.

PR #3195, opened 2022-01-30
feat(Android): optimize app installation.

### LINKED ISSUE #2534: Improve app install time by reinstalling on-device APK

jonathanmos:
## Description

Resolves https://github.com/wix/Detox/issues/3175
Resolves https://github.com/wix/Detox/issues/2537

**Implemented only for Android.**

1. Improve APK install time by:
    * taking a hash of the binary that is being installed and writing it to the emulator as file whose filename is the binary name.
    * on the next install, checks whether the hash exists and if it does then clear user data without performing uninstall/install.
2. Adds a new method of `resetAppState` to decide between clearing user data and reinstallation.
3. Adds a boolean option to Detox `behavior` config, `optimizeAppInstall` with a corresponding CLI flag: `--optimize-app-install`.

In terms of performance: uninstall/install takes 6-7s, whereas clearing app state takes 400-600ms.

This pull request fixes #2534 .

> _For API changes:_
 - [X ] I have made the necessary changes in the [types index](https://github.com/wix/Detox/blob/master/detox/index.d.ts) file.


detox/src/devices/common/drivers/android/tools/FileXfer.js
```diff
@@ -6,6 +6,23 @@ class FileXfer {
...
+  }
+
+  async deleteByExtension(deviceId, extension) {
+    await this._adb.deleteByExtension(deviceId, this._dir, extension);
+  }
+
+  async checkFileExists(deviceId, filename) {
+    return await this._adb.checkFileExists(deviceId, this._dir, filename);
+  }
+
```
d4vidi, line 25: This shouldn't be a part of FileXfer, because it's not under it's scope. You must find a different design that would better organize this.

detox/src/devices/common/drivers/android/tools/FileXfer.js
```diff
@@ -6,6 +6,23 @@ class FileXfer {
     this._dir = destinationDir;
   }
 
+  async getFileHash(binary) {
+    const crypto = require('./CryptoUtils');
+    return await crypto.getMd5(`${this._dir}/${binary}`);
+  }
+
+  async createEmptyFile(deviceId, filename) {
```
d4vidi, line 14: Why is this needed? Isn't it enough to just create a (local) hash and push it to the device?

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -92,6 +95,34 @@ class AndroidDriver extends DeviceDriverBase {
     }
   }
 
+  async _getLocalBinaryHash(binaryFile) {
```
d4vidi, line 98: Please move private methods to the bottom of the class 🙏🏻

detox/test/integration/stub/StubRuntimeDriver.js
```diff
@@ -91,6 +91,10 @@ class StubRuntimeDriver extends DeviceDriverBase {
   async resetContentAndSettings() {
     await sleepALittle();
   }
+
+  async resetAppState() {
+    await sleepALittle();
```
d4vidi, line 96: more like `sleepSomeTime()` 😬

detox/src/devices/runtime/drivers/ios/SimulatorDriver.js
```diff
@@ -63,11 +63,16 @@ class SimulatorDriver extends IosDriver {
...
+    await this._uninstallApp(bundleId);
+    await this._installApp(binaryPath);
+  }
+
+  async _installApp(binaryPath) {
     await this._applesimutils.install(this.udid, getAbsoluteBinaryPath(binaryPath));
   }
 
-  async uninstallApp(bundleId) {
+  async _uninstallApp(bundleId) {
```
d4vidi, line 75: You can't really do that. `installApp()` and `uninstallApp()` are public methods the driver must override...

detox/src/devices/runtime/drivers/ios/SimulatorDriver.test.js
```diff
@@ -137,4 +137,27 @@ describe('IOS simulator driver', () => {
...
+      expect(uninstallSpy).toBeCalledTimes(1);
+      expect(uninstallSpy).toBeCalledWith(bundleId);
+
+      expect(applesimutils.install).toBeCalledTimes(1);
+      expect(applesimutils.install).toBeCalledWith(udid, binaryPath);
+      expect(applesimutils.uninstall).toBeCalledTimes(1);
+      expect(applesimutils.uninstall).toBeCalledWith(udid, bundleId);
+
+      expect(emitSpy).toBeCalledTimes(1);
+      expect(emitSpy).toBeCalledWith('beforeUninstallApp', {deviceId: udid, bundleId});
```
d4vidi, line 160: Please don't use spies for this, as the approach enforces a specific implementation. Better test explicitly what the outcome should be (i.e. use the fundamentals from the tests of `installApp` and `uninstallApp`...)

detox/src/devices/runtime/RuntimeDevice.js
```diff
@@ -419,6 +420,12 @@ class RuntimeDevice {
       this._currentApp.bundleId = await this.deviceDriver.getBundleIdFromBinary(binaryPath);
     }
   }
+
+  async resetAppState() {
+    await traceCall('resetAppState', () => {
+      return this.deviceDriver.resetAppState(this._currentApp.binaryPath, this._currentApp.bundleId);
+    });
+  }
```
d4vidi, line 428: Please only place private methods in the end. Put this adjacent to something more associated with the same context.

detox/src/devices/common/drivers/android/tools/AppInstallHelper.js
```diff
@@ -18,6 +18,28 @@ class AppInstallHelper {
     const binaryPathOnTarget = await this._fileXfer.send(deviceId, binaryPath, binaryFilenameOnTarget);
     await this._adb.remoteInstall(deviceId, binaryPathOnTarget);
   }
+
+  async clearUserData(deviceId, bundleId) {
```
d4vidi, line 22: should this be `clearAppData`?

detox/src/devices/common/drivers/android/tools/AppInstallHelper.js
```diff
@@ -18,6 +18,28 @@ class AppInstallHelper {
...
+  }
+
+  async _removeExistingFileHashes(deviceId) {
+    return await this._fileXfer.deleteByExtension(deviceId, 'hash');
+  }
+
+  async recordHash(deviceId, filehash) {
+    await this._removeExistingFileHashes(deviceId);
+    await this._fileXfer.createEmptyFile(deviceId,`${filehash}.hash`);
+  }
```
d4vidi, line 42: This doesn't seem associated with what the class is meant to do (namely, install APK's). You need to come up with a better design where this would all make sense. To be honest, I'm not sure this is needed at all...

detox/src/devices/common/drivers/android/tools/CryptoUtils.test.js
```diff
@@ -0,0 +1,13 @@
...
+// jest.mock('crypto-js', () => ({
+//   md5: () => mockMd5(),
+// }));
+
+const { getMd5 } = require('./CryptoUtils');
+
+describe('CryptoUtils', () => {
+  it('should call cryto-js md5 function', async () => {
+    await getMd5('test');
+    // expect(mockMd5).toHaveBeenCalledTimes(1);
```
d4vidi, line 11: todo...

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -92,6 +95,34 @@ class AndroidDriver extends DeviceDriverBase {
...
+    return await this.fileXfer.getFileHash(bundleId);
+  }
+
+  async _clearUserData(bundleId) {
+    await this.appInstallHelper.clearUserData(this.adbName, bundleId);
+  }
+
+  async _isAlreadyInstalled(hash) {
+    return await this.appInstallHelper.isAlreadyInstalled(this.adbName, hash);
+  }
```
d4vidi, line 112: These private methods should reside somewhere after all public methods 🙏🏻

detox/src/devices/common/drivers/android/tools/AppInstallHelper.js
```diff
@@ -18,6 +18,28 @@ class AppInstallHelper {
...
+    return await this._adb.clearUserData(deviceId, bundleId);
+  }
+
+  async getLocalBinaryHash(binary) {
+    const { getMd5 } = require('./CryptoUtils');
+    await getMd5(binary);
+  }
+
+  async isAlreadyInstalled(deviceId, filehash) {
+    return await this._fileXfer.checkFileExists(deviceId,`${filehash}.hash`);
```
d4vidi, line 32: This doesn't mean the app is installed. It just means that sometime in the past, something has been installed with `forceAdbInstall` turned off and that the APK was left there...

d4vidi (changes requested):
Hey hey,
this is going in a good direction but requires more work, still. We will run over this through several iterations.

Besides my comments, please be sure to also provide details regarding the positive impact this has over the overall execution time. I urge you to utilize the [timeline artifact](https://wix.github.io/Detox/docs/api/artifacts/#timeline-plugin) in order to be able to do that...

🙏🏻

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,33 @@
...
+const HASH_PATH = '/data/local/tmp/detox';
+
+class HashHelper {
+  constructor(adb, hashXfer) {
+    this._adb = adb;
+    this._hashXfer  = hashXfer;
+  }
+
+  generateHash(path, hashType = SUPPORTED_HASH_TYPES.MD5) {
+    const crypto = require('crypto');
```
d4vidi, line 14: Is this lazy on purpose?

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,33 @@
...
+    switch (hashType) {
+      case SUPPORTED_HASH_TYPES.MD5:
+      default:
```
d4vidi, line 17: I'm not a fan of this approach, because `generateHash(path, 'fubar')` would in fact result in a valid MD5 hash as well (i.e. no indication there was an invalid usage of the helper).

detox/src/devices/common/drivers/android/tools/HashFileXfer.js
```diff
@@ -0,0 +1,20 @@
...
+class HashFileXfer extends FileXfer {
+  /**
+   * @param adb
+   */
+  constructor(adb) {
+    super(adb, HASH_PATH);
+  }
+
+  async readHashFile(deviceId, bundleId) {
+    const filePath = path.posix.join(this._dir, `${bundleId}.hash`);
```
d4vidi, line 15: Isn't `path.join()` sufficient here? 🤔

detox/src/devices/common/drivers/android/exec/ADB.js
```diff
@@ -141,6 +138,19 @@ class ADB {
...
+  async clearAppData(deviceId, packageId) {
+    await this.shell(deviceId, `pm clear ${packageId}`);
+  }
+
+  async createFileWithContent(deviceId, filepath, filename, content) {
+    await this.shell(deviceId, `echo ${content} > ${filepath}/${filename}`);
+  }
+
+  async readFile(deviceId, filepath) {
+    return await this.shell(deviceId, `cat ${filepath}`, {silent: true}).catch(() => '');
```
d4vidi, line 151: Noice!

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +407,30 @@ class AndroidDriver extends DeviceDriverBase {
...
+  async _isAlreadyInstalled(bundleId, hash) {
+    return hash ? this.hashHelper.isRemoteHashEqualToLocal(this.adbName, bundleId, hash) : false;
+  }
+
+  _getLocalBinaryHash(binaryFile) {
+    return this.hashHelper.generateHash(binaryFile);
+  }
+
+  async _saveHashToRemote(hash, bundleId) {
+    if (hash) {
```
d4vidi, line 424: Why would hash be empty? Is that an actual use case?

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +407,30 @@ class AndroidDriver extends DeviceDriverBase {
       'Press any key to continue...'
     );
   }
+
+  async _clearAppData(bundleId) {
+    await this.adb.clearAppData(this.adbName, bundleId);
+  }
+
+  async _isAlreadyInstalled(bundleId, hash) {
+    return hash ? this.hashHelper.isRemoteHashEqualToLocal(this.adbName, bundleId, hash) : false;
```
d4vidi, line 416: Why would hash be empty? Is that an actual use case?

d4vidi, line 416: Technically, this logic is not enough: `hash_exists_and_equal <=> app installed`. For example, a user can manually uninstall the app, and still this would return `true`.
What's missing here, I suppose, is the preliminary check of whether the package is in fact installed on the device (a la `adb.isPackageInstalled()`...

detox/src/devices/runtime/drivers/android/AndroidDriver.test.js
```diff
@@ -579,11 +598,56 @@ describe('Android driver', () => {
...
+      hashHelper.generateHash.mockImplementation(() => mockHash);
+      hashHelper.isRemoteHashEqualToLocal.mockImplementation(() => true);
+      await uut.resetAppState(binaryPath, bundleId);
+      expect(hashHelper.generateHash).toHaveBeenCalledTimes(1);
+      expect(hashHelper.generateHash).toHaveBeenCalledWith(binaryPath);
+      expect(hashHelper.isRemoteHashEqualToLocal).toHaveBeenCalledTimes(1);
+      expect(hashHelper.isRemoteHashEqualToLocal).toHaveBeenCalledWith(adbName, bundleId, mockHash);
+      expect(adb.clearAppData).toHaveBeenCalledTimes(1);
+      expect(adb.clearAppData).toHaveBeenCalledWith(adbName, bundleId);
+      expect(appUninstallHelper.uninstall).not.toHaveBeenCalled();
```
d4vidi, line 631: Must split into several use cases...

detox/src/devices/runtime/drivers/android/AndroidDriver.test.js
```diff
@@ -579,11 +598,56 @@ describe('Android driver', () => {
...
+      expect(adb.clearAppData).not.toHaveBeenCalled();
+      expect(appUninstallHelper.uninstall).toHaveBeenCalledTimes(1);
+      expect(appUninstallHelper.uninstall).toHaveBeenCalledWith(adbName, bundleId);
+      expect(apkValidator.validateAppApk).toHaveBeenCalledTimes(1);
+      expect(apkValidator.validateAppApk).toHaveBeenCalledWith(mockBinaryPath);
+      expect(adb.install).toHaveBeenCalledTimes(2);
+      expect(adb.install).toHaveBeenNthCalledWith(1, adbName, mockBinaryPath);
+      expect(adb.install).toHaveBeenNthCalledWith(2, adbName, mockAPKPathGetTestApkPathImpl(mockBinaryPath));
+      expect(hashHelper.saveHashToRemote).toHaveBeenCalledTimes(1);
+      expect(hashHelper.saveHashToRemote).toHaveBeenCalledWith(adbName, bundleId, mockHash);
```
d4vidi, line 650: Must split into several use cases

detox/index.d.ts
```diff
@@ -1091,6 +1091,11 @@ declare global {
              */
             typeText(text: string): Promise<void>;
 
+            /**
+             * reset the state of the app (e.g. by reinstalling on the device)
+             */
+            resetAppState(): Promise<void>;
```
d4vidi, line 1097: On 2nd thought, I don't think that exposing this as an external API is the right way at all.
From the users' perspective, all they normally do (in order to relaunch the app from scratch) is to invoke `device.launchApp({delete: true})`. Whether the app gets reinstalled or has its state cleaned up should be agnostically decided [here](https://github.com/wix/Detox/blob/35b5bbc912e05dec9a8f5f1a6f8c03ec0290d883/detox/src/devices/runtime/RuntimeDevice.js#L318). This would make this whole optimization applied automatically and seamlessly into existing tests out there.

@noomorph I would highly appreciate you voicing your opinion on this as well 🙏🏻

detox/src/devices/common/drivers/android/tools/AppInstallHelper.test.js
```diff
@@ -1,7 +1,6 @@
 const deviceId = 'mock-device-id';
 const appBinaryPath = '/mock-app-binary-path/binary.apk';
 const testBinaryPath = '/mock-test-binary-path/test/binary.apk';
-
```
d4vidi, line 4: But why? 😆

detox/src/devices/common/drivers/android/tools/FileXfer.js
```diff
@@ -5,7 +5,6 @@ class FileXfer {
     this._adb = adb;
     this._dir = destinationDir;
   }
-
```
d4vidi, line 8: But why? 😆

detox/src/devices/common/drivers/android/tools/FileXfer.test.js
```diff
@@ -1,6 +1,5 @@
 const deviceId = 'mock-device-id';
 const deviceDestinationDir = '/mock-tmp-dir';
-
```
d4vidi, line 3: But why? 😆

detox/src/devices/common/drivers/android/tools/HashFileXfer.test.js
```diff
@@ -0,0 +1,19 @@
+let uut;
+let adb;
+
+describe('hash file transfer', () => {
+  it('should pass correct params to read hash file', () => {
+    const HashFileXfer = require('./HashFileXfer');
+    const ADB = jest.genMockFromModule('../exec/ADB');
+    adb = new ADB();
+    adb.readFile.mockImplementation(() => jest.fn());
+    uut = new HashFileXfer(adb);
```
d4vidi, line 10: `beforeEach` it, please

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,33 @@
...
+        return crypto.createHash('md5').update(path).digest("hex");
+    }
+  }
+
+  async saveHashToRemote(deviceId, bundleId, hash) {
+    const hashFileName = `${bundleId}.hash`;
+    await this._adb.createFileWithContent(deviceId, HASH_PATH, hashFileName, hash);
+  }
+
+  async isRemoteHashEqualToLocal(deviceId, bundleId, localHash) {
```
d4vidi, line 27: Naming: What about `compareXYZ()`?

d4vidi (changes requested):
Good progress but it's still not ready for merge. Please see my notes, especially the one regarding the external API changes.

detox/src/devices/common/drivers/android/tools/HashHelper.test.js
```diff
@@ -0,0 +1,74 @@
+const HashHelper = require('./HashHelper');
```
d4vidi, line 1: Hmm that can't be right...

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,33 @@
+const SUPPORTED_HASH_TYPES = {
+  MD5: 'MD5'
+}
+
+const HASH_PATH = '/data/local/tmp/detox';
```
d4vidi, line 5: This feels like something that should be a c'tor arg.

detox/src/devices/common/drivers/android/tools/HashHelper.test.js
```diff
@@ -0,0 +1,74 @@
...
+    await uut.saveHashToRemote(mockDeviceId, mockBundleId, mockHash);
+    expect(adb.createFileWithContent).toHaveBeenCalledTimes(1);
+    expect(adb.createFileWithContent).toHaveBeenLastCalledWith(mockDeviceId, HASH_PATH, `${mockBundleId}.hash`, mockHash);
+  });
+
+  it('should return true when remoteHash and localHash match', async () => {
+    const HashXferMock = jest.genMockFromModule('./HashFileXfer');
+    hashxfer = new HashXferMock();
+    hashxfer.readHashFile.mockImplementation(() => mockHash);
+    uut = new HashHelper(adb, hashxfer);
```
d4vidi, line 33: Is there a good reason to do this? `uut` already gets initialized in `beforeEach`.

detox/src/devices/runtime/drivers/android/AndroidDriver.js
d4vidi: @jonathanmos This is still applicable, it seems

d4vidi: @jonathanmos still applicable

d4vidi:
@jonathanmos what's the status here?

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,46 @@
...
+      throw new DetoxRuntimeError({
+        message: `Path must be provided for hash generation`,
+      });
+    }
+
+    switch (hashType) {
+      case SUPPORTED_HASH_TYPES.MD5:
+        return crypto.createHash('md5').update(path).digest("hex");
+      default:
+        throw new DetoxRuntimeError({
```
noomorph, line 28: Use `DetoxInternalError`, please.

detox/src/devices/common/drivers/android/exec/ADB.js
```diff
@@ -135,6 +132,19 @@ class ADB {
     await this.shell(deviceId, `am force-stop ${appId}`);
   }
 
+
+  async clearAppData(deviceId, packageId) {
+    return await this.shell(deviceId, `pm clear ${packageId}`, {verbosity: 'low', silent: true});
+  }
+
+  async createFileWithContent(deviceId, filepath, filename, content) {
```
noomorph, line 140: Why not just a complete `filepath`, why join `filepath` and `filename` ?

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,46 @@
...
+  generateHash(path, hashType = SUPPORTED_HASH_TYPES.MD5) {
+    if (!path) {
+      throw new DetoxRuntimeError({
+        message: `Path must be provided for hash generation`,
+      });
+    }
+
+    switch (hashType) {
+      case SUPPORTED_HASH_TYPES.MD5:
+        return crypto.createHash('md5').update(path).digest("hex");
```
noomorph, line 26: Why do you calculate an MD5 hash of a *path*, not the file contents themselves? 🤔

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  async saveHashToRemote(deviceId, bundleId, hash) {
+    const hashFileName = `${bundleId}.hash`;
+    await this._adb.createFileWithContent(deviceId, this._hashPath, hashFileName, hash);
+  }
+
+  async compareRemoteToLocal(deviceId, bundleId, localHash) {
+    const remoteHash = await this._hashXfer.readHashFile(deviceId, bundleId);
```
noomorph, line 41: What if the user:
1. runs Detox tests
2. re-installs a different version of the app on the device using React Native CLI or some other manual way, outside of Detox flows
3. returns to the initial project and runs Detox tests again?

Won't Detox treat it as if the hash did not change (hence, no need to reinstall), while it should have been more paranoid?

noomorph: Actually won't it be wiser to check last install timestamp or maybe use those hashes from dumpsys:

```
adb shell dumpsys package com.example.testapp.first

Packages:
  Package [com.example.testapp.first] (2036fd1):
    userId=10225
    pkg=Package{42e7a36 com.example.testapp.first}
    codePath=/data/app/com.example.testapp.first-1
    resourcePath=/data/app/com.example.testapp.first-1
    legacyNativeLibraryDir=/data/app/com.example.testapp.first-1/lib
    primaryCpuAbi=null
    secondaryCpuAbi=null
    versionCode=1 minSdk=21 targetSdk=24
    versionName=1.0
    splits=[base]
    apkSigningVersion=1
    applicationInfo=ApplicationInfo{29cb2a4 com.example.testapp.first}
    flags=[ HAS_CODE ALLOW_CLEAR_USER_DATA ALLOW_BACKUP ]
    privateFlags=[ RESIZEABLE_ACTIVITIES ]
    dataDir=/data/user/0/com.example.testapp.first
    supportsScreens=[small, medium, large, xlarge, resizeable, anyDensity]
    timeStamp=2016-11-03 01:12:08
    firstInstallTime=2016-11-03 01:12:09
    lastUpdateTime=2016-11-03 01:12:09
    signatures=PackageSignatures{9fe380d [53ea108d]}
    installPermissionsFixed=true installStatus=1
    pkgFlags=[ HAS_CODE ALLOW_CLEAR_USER_DATA ALLOW_BACKUP ]
    User 0: ceDataInode=474317 installed=true hidden=false suspended=false stopped=true notLaunched=true enabled=0
      runtime permissions:

``` 




https://stackoverflow.com/questions/40543815/to-what-do-the-android-adb-shell-dumpsys-package-hashes-refer-to

detox/src/devices/common/drivers/android/exec/ADB.js
```diff
@@ -135,6 +132,19 @@ class ADB {
     await this.shell(deviceId, `am force-stop ${appId}`);
   }
 
+
+  async clearAppData(deviceId, packageId) {
+    return await this.shell(deviceId, `pm clear ${packageId}`, {verbosity: 'low', silent: true});
+  }
+
+  async createFileWithContent(deviceId, filepath, content) {
+    await this.shell(deviceId, `echo ${content} > ${filepath}`);
```
noomorph, line 141: Looks unsafe. Maybe you could have created a temporary file on the local filesystem and `adb push` it to the location?

detox/src/devices/common/drivers/android/exec/ADB.js
```diff
@@ -135,6 +132,19 @@ class ADB {
...
+  }
+
+  async readFile(deviceId, filepath) {
+    return await this.shell(deviceId, `cat ${filepath}`, {verbosity: 'low', silent: true}).catch(() => '');
```
noomorph, line 145: What is the reason to suppress errors here?

detox/src/devices/common/drivers/android/tools/HashHelper.js
noomorph: So, just to summarize for the sake of history - Jonathan and I discussed this and reached the conclusion that the current implementation indeed is not totally safe, that's why we add the `optimizeAppInstall` flag to enable this optimization in verified contexts (such as CI environment, where the devices are ephemeral and the flow is deterministic).

noomorph (changes requested):
@jonathanmos, I think we should also improve the logic of `device.launchApp({ delete: true })`, so it leverages the current optimization for Android.

detox/src/devices/common/drivers/android/exec/ADB.js
jonathanmos: This method is only used to read the hash file, and if the hash file doesn't exist I don't want that error to be shown to the user, since it just means that the apk needs to be reinstalled.

docs/APIRef.DetoxCLI.md
```diff
@@ -102,6 +102,7 @@ Initiating your test suite[^1].
...
 | --force-adb-install                           | Due to problems with the `adb install` command on Android, Detox resorts to a different scheme for installing APKs. Setting true will disable that and force usage of `adb install`, instead.<br/>This flag is temporary until the Detox way proves stable.<br/>_Default: false_                          |
 | --inspect-brk                                 | Uses [node’s --inspect-brk](https://nodejs.org/en/docs/guides/debugging-getting-started/#enable-inspector) flag to let users debug the jest/mocha test runner <br />_Default: false_                                                                                                                      |
+| --optimize-apk-install                        | Uses an optimized method for installation where a hash is stored on the device allowing future app installation to be skipped when possible.<br />_Default: false_                                                                                                                                        |
```
noomorph, line 105: @jonathanmos, wait, something's off here. Why **APK**?

detox/index.d.ts
```diff
@@ -576,6 +583,12 @@ declare global {
              */
             uninstallApp(bundle?: string): Promise<void>;
 
+            /**
+             * Reset the state of the app (e.g. by reinstalling on the device)
+             * or deleting user data.
+             */
+            resetAppState(): Promise<void>;
```
noomorph, line 590: @jonathanmos @d4vidi @asafkorem 

I suggest renaming to `clearAppData`. It is more coherent with the UI terminology in iOS and Android:

image

image

and Google also suggests it to be less popular 

image

image

docs/APIRef.DetoxCLI.md
noomorph: Ok, this is a typo. Fixed.

detox/index.d.ts
d4vidi: I think that resetting app state is _what_ this method does, and clearing data is _how_ it does it (when optimized on Android, at least). So, I vote for leaving as-is 😅

detox/src/Detox.js
```diff
@@ -248,10 +248,17 @@ class Detox {
...
       await this.device.selectApp(appName);
-      await this.device.uninstallApp();
-      await this.device.installApp();
+
+      if (optimizeAppInstall) {
+        await this.device.resetAppState();
+      } else {
+        await this.device.uninstallApp();
+        await this.device.installApp();
+      }
```
d4vidi, line 261: I think it's wrong to perform this check in multiple places ([example](https://github.com/wix/Detox/pull/3195/files#diff-61fe553dde3e9ac2a314b747cf90bcadbdc0466427bced4f164bf4a924042664R322)).

If I understand correctly, the very point behind `resetAppState` was to either clear data if the `optimizeAppInstall` flag is set, or reinstall otherwise. Therefore, this code should be inside `resetAppState()` itself, shouldn't it?

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -52,10 +53,12 @@ class AndroidDriver extends DeviceDriverBase {
     this.apkValidator = deps.apkValidator;
     this.invocationManager = deps.invocationManager;
     this.fileXfer = deps.fileXfer;
+    this.hashXfer = deps.hashXfer;
```
d4vidi, line 56: This is never used, in fact, and can be removed

detox/src/devices/common/drivers/android/tools/HashFileXfer.js
```diff
@@ -0,0 +1,20 @@
+const FileXfer = require('./FileXfer');
+const path = require('path');
+
+const HASH_PATH = '/data/local/tmp/detox';
+
+class HashFileXfer extends FileXfer {
```
d4vidi, line 6: I don't understand this class, and why it inherits `FileXfer`:
1. `FileXfer`'s concern is very specific: Sending files into the device. Extending it in this way doesn't extend that concern, so it's shouldn't be done this way...
1. `FileXfer`'s methods `send` and `prepare` methods are never used in the context of hash storage. That is in fact also a testament to the first point.
1. Reading the hash file isn't about send a file, nor about pulling it - hence not about transferring file. The `readHashFile` method doesn't belong here.

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,64 @@
...
+    this._hashXfer  = hashXfer;
+    this._hashPath = hashPath;
+  }
+
+  async generateHash(path) {
+    if (!path) {
+      throw new DetoxInternalError(`Path must be provided for hash generation`);
+    }
+
+    return new Promise((resolve, reject) => {
```
d4vidi, line 21: As an `async` function - it might be better, semantically, to `await` the promise rather than returning it

detox/src/devices/common/drivers/android/tools/HashHelper.js
```diff
@@ -0,0 +1,64 @@
+const crypto = require('crypto');
+const fs = require('fs');
+
+const { DetoxInternalError } = require('../../../../../errors');
+const path = require('path');
+
+const HASH_PATH = '/data/local/tmp/detox';
+
+class HashHelper {
```
d4vidi, line 9: I'm not entirely against classes like these, but I really think it needs to be broken down into at least two separate pieces:
1. Calculating a hash: This is a concern that can be introduced as a **top-level** (non-Android specific) **util**. It could offer methods such as calculating hashes async'ly over strings, or - as a convenience, based on in-streams (as in this case...).
1. Saving the hash to the device: This is something that can indeed be under a helper as such. However, **it should utilize `TempFileXfer`** in order to send the file away: That would allow a reuse of:
    1. The target path, which is unnecessarily redefined here, at the moment.
    1. Creating the dest dir on the device, which isn't called here - but should be.

Reading the hash back from the device + comparing it can be done in the driver itself, I suppose.

d4vidi (changes requested):
Hey I left some important comments. They need to be addressed before merging 🙏🏻

detox/index.d.ts
noomorph: It is not app state. Reset app state is more like something that happens in runtime, via Redux or other state management library.

detox/src/Detox.js
noomorph: Good catch, but now I have an opposite question. Why does `resetAppState` install the app as a side effect at all? 🤔

noomorph: I think this place has a logic error. Something is wrong with handling `--reuse` flag and `--optimize-app-install` together, it is itching in the back of my head. 🤔

asafkorem:
Starting a review.

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +410,43 @@ class AndroidDriver extends DeviceDriverBase {
...
+  }
+
+  async _shouldClearData(bundleId, hash, isPkgInstalled) {
+    const isSameVersionInstalled = this._compareRemoteToLocal(bundleId, hash);
+    return isPkgInstalled && isSameVersionInstalled;
+  }
+
+  async _compareRemoteToLocal(bundleId, localHash) {
+    const hashFilename = `${bundleId}.hash`;
+    const destinationPath = path.posix.join(this.fileXfer.getFilePath(), hashFilename);
```
asafkorem, line 429: This doesn't make any sense (`this.fileXfer.getFilePath`). `fileXfer` is from type `FileXfer` in your code.

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -32,6 +32,8 @@ const log = logger.child({ __filename });
  * @property aapt { AAPT }
  * @property apkValidator { ApkValidator }
  * @property fileXfer { FileXfer }
```
asafkorem, line 34: According to the usage of this object, the type should be `TempFileXfer`

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +410,43 @@ class AndroidDriver extends DeviceDriverBase {
...
+    await this.hashHelper.saveHashToRemote(this.adbName, bundleId, hash);
+  }
+
+  async _shouldClearData(bundleId, hash, isPkgInstalled) {
+    const isSameVersionInstalled = this._compareRemoteToLocal(bundleId, hash);
+    return isPkgInstalled && isSameVersionInstalled;
+  }
+
+  async _compareRemoteToLocal(bundleId, localHash) {
+    const hashFilename = `${bundleId}.hash`;
```
asafkorem, line 428: The "`HashHelper`" class also generates the file name (with `${bundleId}.hash`). Let's avoid this duplication

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +410,43 @@ class AndroidDriver extends DeviceDriverBase {
       'Press any key to continue...'
     );
   }
+
+  _getLocalBinaryHash(binaryFile) {
+    return generateHash(binaryFile);
+  }
+
+  async _saveHashToRemote(hash, bundleId) {
+    await this.hashHelper.saveHashToRemote(this.adbName, bundleId, hash);
```
asafkorem, line 419: this should get the hash file name as a parameter (and not to generate it by itself)

detox/src/devices/runtime/drivers/android/AndroidDriver.js
asafkorem, line 428: this should get the hash file name as a parameter (and not to generate it by itself)

asafkorem: let's extract this generation to a method (maybe on the hash-helper file)

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +410,43 @@ class AndroidDriver extends DeviceDriverBase {
...
+    const remoteHash = await this.adb.readFile(this.adbName, destinationPath, true);
+    return localHash === remoteHash;
+  }
+
+  async _clearData(params) {
+    try {
+      return await this.adb.clearAppData(this.adbName, params.bundleId);
+    } catch (e) {
+      await this._reinstallApp(params);
+    }
```
asafkorem, line 439: why return on try and not to return on catch?

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -394,6 +410,43 @@ class AndroidDriver extends DeviceDriverBase {
...
+  }
+
+  async _reinstallApp(params) {
+    if (params.isPkgInstalled) {
+      await this.uninstallApp(params.bundleId);
+    }
+
+    await this.installApp(params.binaryPath, params.testBinaryPath);
+    await this._saveHashToRemote(params.hash, params.bundleId);
```
asafkorem, line 448: why do you need the uninstall if the install already uninstalls if needed?

detox/src/devices/runtime/drivers/ios/SimulatorDriver.js
```diff
@@ -63,6 +63,11 @@ class SimulatorDriver extends IosDriver {
     }
   }
 
+  async resetAppState(bundleId, binaryPath) {
+    await this.uninstallApp(bundleId);
+    await this.installApp(binaryPath);
```
asafkorem, line 68: I think we should throw exception here (not implemented) and not to fallback. this will avoid usage of the optimized-app-install feature on iOS (not exists).

detox/src/devices/runtime/drivers/ios/SimulatorDriver.test.js
```diff
@@ -137,4 +137,16 @@ describe('IOS simulator driver', () => {
...
+  describe('resetAppState', () => {
+    it('should call uninstall and install for resetAppState', async () => {
+      const binaryPath = '/tmp';
+      await uut.resetAppState(bundleId, binaryPath);
+
+      expect(applesimutils.install).toBeCalledTimes(1);
+      expect(applesimutils.install).toBeCalledWith(udid, binaryPath);
+      expect(applesimutils.uninstall).toBeCalledTimes(1);
+      expect(applesimutils.uninstall).toBeCalledWith(udid, bundleId);
+    });
```
asafkorem, line 150: check that an assertion is raised

detox/src/devices/runtime/factories/android.js
```diff
@@ -8,18 +8,21 @@ class RuntimeDriverFactoryAndroid extends RuntimeDeviceFactory {
...
 
+
     return {
       ...commonDeps,
       adb,
       aapt,
       apkValidator,
       fileXfer,
       devicePathBuilder,
+      hashHelper,
```
asafkorem, line 25: do you use it from here?
it's just a static method, so I dont think you need a service

detox/src/servicelocator/android/index.js
```diff
@@ -21,5 +22,8 @@ AndroidServiceLocator.apkValidator = new ApkValidator(AndroidServiceLocator.aapt
 AndroidServiceLocator.fileXfer = new TempFileXfer(AndroidServiceLocator.adb);
 AndroidServiceLocator.deviceRegistry = DeviceRegistry.forAndroid();
 AndroidServiceLocator.devicePathBuilder = new AndroidDevicePathBuilder();
+AndroidServiceLocator.tempFileXfer = new TempFileXfer(AndroidServiceLocator.adb);
```
asafkorem, line 25: I think you dont use it (check)

detox/src/servicelocator/android/index.js
```diff
@@ -21,5 +22,8 @@ AndroidServiceLocator.apkValidator = new ApkValidator(AndroidServiceLocator.aapt
 AndroidServiceLocator.fileXfer = new TempFileXfer(AndroidServiceLocator.adb);
```
asafkorem, line 22: you might want to rename this one to tempFileXfer since we are using it as the real temp file xfer

detox/src/servicelocator/android/index.js
```diff
@@ -21,5 +22,8 @@ AndroidServiceLocator.apkValidator = new ApkValidator(AndroidServiceLocator.aapt
...
 AndroidServiceLocator.deviceRegistry = DeviceRegistry.forAndroid();
 AndroidServiceLocator.devicePathBuilder = new AndroidDevicePathBuilder();
+AndroidServiceLocator.tempFileXfer = new TempFileXfer(AndroidServiceLocator.adb);
+AndroidServiceLocator.hashHelper = new HashHelper(AndroidServiceLocator.adb, AndroidServiceLocator.tempFileXfer);
```
asafkorem, line 26: no need I guess

detox/test/e2e/detox.config.js
```diff
@@ -27,6 +27,7 @@ const config = {
     init: {
       exposeGlobals: true
     },
+    optimizeAppInstall: Boolean(process.env.CI),
```
asafkorem, line 30: should be true on Android and false on iOS..

asafkorem: (`process.env.PLATFORM === 'android' && Boolean(process.env.CI) `)

detox/test/types/detox-global-tests.ts
```diff
@@ -28,6 +28,7 @@ describe("Test", () => {
 
     beforeAll(async () => {
         await device.reloadReactNative();
+        await device.resetAppState();
```
asafkorem, line 31: why here? @noomorph / @jonathanmos

docs/APIRef.DetoxCLI.md
```diff
@@ -102,6 +102,7 @@ Initiating your test suite[^1].
 | --use-custom-logger                           | Use Detox' custom console-logging implementation, for logging Detox (non-device) logs. Disabling will fallback to node.js / test-runner’s implementation (e.g. Jest / Mocha).<br />_Default: true_ |
 | --force-adb-install                           | Due to problems with the `adb install` command on Android, Detox resorts to a different scheme for installing APKs. Setting true will disable that and force usage of `adb install`, instead.<br/>This flag is temporary until the Detox way proves stable.<br/>_Default: false_ |
 | --inspect-brk                                 | Uses [node’s --inspect-brk](https://nodejs.org/en/docs/guides/debugging-getting-started/#enable-inspector) flag to let users debug the jest/mocha test runner <br />_Default: false_ |
+| --optimize-app-install                        | Uses an optimized method for installation where a hash is stored on the device allowing future app installation to be skipped when possible.<br />_Default: false_                                                                                                                                        |
```
asafkorem, line 105: Only Android

asafkorem: Also I wouldn't elaborate on implementation (stored hash etc..), only saying that it depends whether the APK has changed.

detox/test/e2e/detox.config.js
noomorph: Why? If it were implemented on iOS, why wouldn't we want it to work?

detox/local-cli/test.js
```diff
@@ -149,6 +149,7 @@ function prepareMochaArgs({ cliConfig, runnerArgs, runnerConfig, platform }) {
       'debug-synchronization': isFinite(cliConfig.debugSynchronization) ? cliConfig.debugSynchronization : undefined,
       'device-name': cliConfig.deviceName || undefined,
       'force-adb-install': platform === 'android' && cliConfig.forceAdbInstall || undefined,
+      'optimize-app-install': cliConfig.optimizeAppInstall || undefined,
```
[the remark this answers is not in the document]
noomorph: Yes. `false` values should be omitted.

detox/index.d.ts
```diff
@@ -578,6 +585,12 @@ declare global {
...
```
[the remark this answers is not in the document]
noomorph: @asafkorem what about other names like `wipeAppFiles` or `clearAppData`, etc?

detox/src/servicelocator/android/index.js
noomorph: wth is `xfer`?? can't we use normal english? 😬

detox/test/e2e/detox.config.js
asafkorem: But it isn't implemented in iOS...

detox/index.d.ts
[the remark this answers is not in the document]
asafkorem: I don't think these are better here @noomorph 
This method resets the state of the app to the initial state, by reinstalling or clearing the data (clearing / wiping data is only one option)

detox/src/servicelocator/android/index.js
asafkorem: Thinking about that "XFER" is a pretty common synonym for "File Transfer" 🙂 I guess we can call it `FileTransfer` or to leave it as is.. and to postpone this decision to another PR

noomorph: At least could you add JSDoc comment explaining the name... ?

detox/test/types/detox-global-tests.ts
noomorph: @asafkorem  it is not a test per se, it is a typing/compiler error check thing.

detox/src/Detox.js
```diff
@@ -250,12 +250,7 @@ class Detox {
 
     for (const appName of appNames) {
       await this.device.selectApp(appName);
-      await this.device.uninstallApp();
-    }
-
-    for (const appName of appNames) {
-      await this.device.selectApp(appName);
-      await this.device.installApp();
+      await this.device.resetAppState();
```
d4vidi, line 253: @noomorph wouldn't this create a regression of #3435 ?

detox/src/devices/common/drivers/android/exec/ADB.js
```diff
@@ -135,6 +132,18 @@ class ADB {
     await this.shell(deviceId, `am force-stop ${appId}`);
   }
 
+  async clearAppData(deviceId, packageId) {
+    return await this.shell(deviceId, `pm clear ${packageId}`, { verbosity: 'low', silent: true });
+  }
+
+  async readFile(deviceId, filepath, silent = false) {
+    if (silent) {
```
d4vidi, line 140: This never seems to be called with `silent=false`. Is this really necessary?
@jonathanmos

detox/src/devices/common/drivers/android/tools/AppInstallHelper.test.js
```diff
@@ -10,23 +10,23 @@ describe('Android app installation helper', () => {
     adb = new ADBMock();
   });
 
-  let fileXfer;
+  let tempFileTransfer;
```
d4vidi, line 13: Same comment... please revert

detox/src/devices/common/drivers/android/tools/FileTransfer.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 
-class FileXfer {
+class FileTransfer {
```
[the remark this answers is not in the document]
noomorph: @d4vidi I asked for this. Less hax0r l33t slang, please :) I know it is a legit word but `Transfer` is not that long still https://en.wiktionary.org/wiki/xfer

detox/src/devices/common/drivers/android/tools/AppInstallHelper.js
```diff
@@ -1,21 +1,21 @@
 // TODO Tweak such that if apk's already exist on the device (need to store uniquely), they will not be resent (would optimize cloud, for example)
 
 class AppInstallHelper {
-  constructor(adb, fileXfer) {
+  constructor(adb, tempFileTransfer) {
```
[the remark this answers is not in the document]
asafkorem: the mistake is in this commit https://github.com/wix/Detox/pull/3195/commits/b14825760a316970899da641cecb11bbbcbb4c94

(fileXFER should've been replaced with fileTransfer)

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -51,7 +54,7 @@ class AndroidDriver extends DeviceDriverBase {
     this.aapt = deps.aapt;
     this.apkValidator = deps.apkValidator;
     this.invocationManager = deps.invocationManager;
-    this.fileXfer = deps.fileXfer;
+    this.tempFileTransfer = deps.tempFileTransfer;
```
d4vidi, line 57: Please revert 🙏🏻

detox/src/devices/common/drivers/android/tools/ApkHashUtils.js
```diff
@@ -0,0 +1,36 @@
...
+
+const saveHashToDevice = async ({ tempFileTransfer, bundleId, deviceId, binaryPath }) => {
+  const hashFilename = _getHashFilename(bundleId);
+  await _createLocalHashFile(hashFilename, binaryPath);
+  await tempFileTransfer.prepareDestinationDir(deviceId);
+  await tempFileTransfer.send(deviceId, hashFilename, hashFilename);
+  _deleteLocalHashFile(hashFilename);
+};
+
+const isHashUpdated = async ({ adb, deviceId, bundleId, binaryPath }) => {
```
d4vidi, line 16: This just doesn't make sense to me... - to have a _utility_ get one of our core dependency objects as an argument.
It's a smell saying this should be a helper that gets injected like all other deps.

The smell itself comes from the fact the giving `adb` as an arg hints a great deal of the internal implementation - hence doesn't encapsulate it all the way. Having a Class that just gets that in the c'tor, limits that to just one function (the c'tor).

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -92,12 +95,24 @@ class AndroidDriver extends DeviceDriverBase {
     }
   }
 
+  async optimizedInstallApp(bundleId, binaryPath, testBinaryPath) {
+    const isPkgInstalled = await this.adb.isPackageInstalled(this.adbName, bundleId);
```
d4vidi, line 99: Shouldn't this take place inside `_shouldClearData`?

detox/src/devices/common/drivers/android/tools/ApkHashUtils.js
d4vidi: Also, naming-wise: Would `isHashUpToDate` make more sense?

detox/src/devices/common/drivers/android/exec/ADB.js
jonathanmos: It's to allow for calling not silently in future. Atm only the optimized flow uses this method

jonathanmos: It's to allow for calling nonsilently in future. Atm it's called only from the optimized flow

PR #3204, opened 2022-02-02
Use root locale when changing case in java

halaei:
## Description

Not setting locale for language/country neutral operation may cause bug depending on the default locale.
See https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#ROOT

Note: I am just searching for `toLowerCase()` and `toUppercase()` in my project's dependencies and send the same PR, in order to potentially raise the awareness. Although I've seen the lack of explicit locale has caused issues for us, I am not sure if wix/Detox is actually affected. 

Example related issue: https://github.com/joltup/rn-fetch-blob/issues/573

> _For API changes:_


noomorph:
@jonathanmos any chance you can look soon?

jonathanmos:
I'm merging this although I'm not sure it would affect us

PR #3210, opened 2022-02-08
fix(jest-circus): proper handling of init errors

noomorph:
## Description

Previously, if there was an error in the Detox setup phase (e.g. port allocation collision or device not booting), the entire suite with all the tests was marked as skipped. The motivation was to reduce the output for such tests.

But as it seems, it doe not play well with Jest reporters — such tests are considered skipped indeed, nevertheless, and if a suite fails on such an early error — it is not going to be retried via Detox CLI mechanism.

So, the suggested solution is to duplicate the unhandled error to each test and to noop the registered hooks (to prevent their accidental execution — yes, unfortunately, that happens).

### Before

Screen Shot 2022-02-08 at 14 10 08

### After

Screen Shot 2022-02-08 at 19 09 30

### DIFF
--- a/detox/runners/jest-circus/listeners/DetoxInitErrorListener.js
+++ b/detox/runners/jest-circus/listeners/DetoxInitErrorListener.js
@@ -1,19 +1,23 @@
 const _ = require('lodash');
 
 class DetoxInitErrorListener {
-  setup(event, { unhandledErrors, rootDescribeBlock }) {
-    if (unhandledErrors.length > 0) {
-      rootDescribeBlock.mode = 'skip';
+  add_hook(event, { unhandledErrors, currentDescribeBlock, rootDescribeBlock }) {
+    if (_.isEmpty(unhandledErrors)) {
+      return;
+    }
+
+    if (currentDescribeBlock !== rootDescribeBlock || currentDescribeBlock.hooks.length > 1) {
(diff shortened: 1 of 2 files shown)


detox/runners/jest-circus/listeners/DetoxInitErrorListener.js
```diff
@@ -1,19 +1,20 @@
...
+    }
+
+    if (currentDescribeBlock !== rootDescribeBlock || currentDescribeBlock.hooks.length > 1) {
+      _.last(currentDescribeBlock.hooks).fn = async () => {};
     }
   }
 
-  add_test(event, { currentDescribeBlock, rootDescribeBlock }) {
-    if (currentDescribeBlock === rootDescribeBlock && rootDescribeBlock.mode === 'skip') {
+  add_test(event, { unhandledErrors, currentDescribeBlock }) {
```
d4vidi, line 14: Just making sure: Do we know for a fact that `unhandledErrors` is synonymous with errors happening prior to a describe-suite execution? Could they, for example, be a result of failures of previous _tests_? (in which case the logic here renders inappropriate)

noomorph: No, they are root level, and we put them on our behalf.

PR #3216, opened 2022-02-10
fix: race condition in DetoxServer class

noomorph:
## Description

Related to https://github.com/wix/Detox/pull/3174

Hopefully, this is an ultimate fix that will never ever cause the further reproduction of that bug.

Solution: using `port: 0` to make sure `wss` server allocates the port dynamically and setting the resulting value post-factum to the client config.

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -62,7 +62,6 @@
     "find-up": "^4.1.0",
     "fs-extra": "^4.0.2",
     "funpermaproxy": "^1.0.1",
-    "get-port": "^2.1.0",
     "ini": "^1.3.4",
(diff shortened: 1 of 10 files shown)


noomorph:
Guys, since it is a fix without any API change per se, I'll merge it now, and we can discuss postfactum.

d4vidi:
>Solution: using port: 0 to make sure wss server allocates the port dynamically and setting the resulting value post-factum to the client config.

This is an AMAZING find! Could you explain what WSS' port allocation mechanism looks like (i.e. how do we know for sure it would not use the same port in two concurrent workers?)

PR #3222, opened 2022-02-15
Buildkite Migration

igorgn:
## Description
Buildkite migration changes

---

> _For API changes:_


noomorph:
I see no artifacts:

image

image

d4vidi:
Thanks for this, dear @igorgn 😍
I suspect something is a bit off here, with the sub-jobs failing but reported as "green". Could you review?
Also, the 1MB log limit is very harsh... What can be done in regards? We'd like to be able to review the logs in the browser

igorgn:
@noomorph totally forgot about artifacts.. on it
@d4vidi can you specify where's green failing part? Regarding logs, it seems that 1MB is hardcoded and can't be increased -> [docs](https://buildkite.com/docs/pipelines/managing-log-output#log-output-limits).

noomorph:
@d4vidi  so let's cut the logs' crap, shall we? @igorgn, can you check what build steps generate the most wadawadawadawadawadawadawada ?

d4vidi:
> @d4vidi so let's cut the logs' crap, shall we? @igorgn, can you check what build steps generate the most wadawadawadawadawadawadawada ?

???
We have styled logs and the only options to see the fully generated log does so in _raw_ mode which is unreadable.. :-(

igorgn:
@noomorph @d4vidi I added artifacts archiving, including what was uploading to s3 previously. Regarding logs, I think we need to make a leap and just see how (un)-comfortable it will be.

detox/src/devices/common/drivers/android/emulator/exec/EmulatorExec.js
```diff
@@ -74,7 +74,7 @@ class LaunchCommand extends ExecCommand {
     if (this._options.headless) {
       switch (os.platform()) {
         case 'darwin':
-          return 'host';
+          return 'off';
```
noomorph, line 77: @d4vidi just to make sure everyone's okay with it.

noomorph: Igor thinks it is an error on our side to have `host` default. Maybe it is, maybe not, let's think.

d4vidi: This isn't good enough because it would affect all detox users out there (running macs). We should apply `--gpu` explicitly as a command-line argument in the scripts, or conditionally (local vs. CI) in the Detox configuration scripts.

detox/src/configuration/composeDeviceConfig.test.js
```diff
@@ -81,7 +81,7 @@ describe('composeDeviceConfig', () => {
       'android.emulator': {
         type: 'android.emulator',
         device: { avdName: 'Pixel_API_28' },
-        gpu: 'auto',
+        gpu: 'off',
```
d4vidi, line 84: revert

PR #3224, opened 2022-02-15
Introduce spawnWithRetriesAndLogs and use in ADB

d4vidi:
## Description

- This pull request addresses the issue described here: https://github.com/wix-private/mobile-guild/issues/192

In this pull request, I have introduced a new child-process spawning (not exec) function that interactively logs the child-process' outputs (stdout and stderr), which also supports retries.

In the process, I've refactored `exec.js`, which has become bloated, onto `exec` and `spawn` under a new subdirectory - `childProcess/`.

> (EDITED) Note: The work on this is technically far from over, as far as code quality is concerned. Ideally, we should "merge" `exec` and `spawn` into the same skeletal implementation (perhaps even use inheritance) so as to have both well aligned in terms of supported features. As mentioned in the comments - that is not mandatory for the time being, rather, more of a JFYI note for the "next round" of maintenance work.

TODO:
- [x] Unit testing over `spawn`
- [x] Fix unit tests over ADB.
- [x] Unit test that verifies that tracking ID is coherent among usage of `exec` and `spawn`.
- [x] Inline todo's

---

> _For API changes:_


d4vidi:
@noomorph would love to get your early-bird insights on this.

detox/src/devices/common/drivers/android/exec/ADB.js
```diff
@@ -100,20 +100,15 @@ class ADB {
...
-  async install(deviceId, apkPath) {
-    apkPath = `"${escape.inQuotedString(apkPath)}"`;
-
+  async install(deviceId, _apkPath) {
+    const apkPath = escape.inQuotedString(_apkPath);
     const apiLvl = await this.apiLevel(deviceId);
+    const command = (apiLvl >= 23)
+      ? `install -r -g -t ${apkPath}`
+      : `install -rg ${apkPath}`;
+    const result = await this.adbCmdSpawned(deviceId, command, { timeout: 45000, retries: 3});
```
noomorph, line 109: Is `45000` a good magic number? Not that I want to make it configurable, but I would like to be sure it is reliable... what's the average duration of installation?

detox/src/utils/childProcess/exec.test.js
```diff
@@ -94,11 +94,14 @@ describe('exec', () => {
         retrying: true,
       },
     };
+
+    logger.debug.mockClear();
+
     await exec.execWithRetriesAndLogs('bin', options);
 
     expect(cpp.exec).toHaveBeenCalledWith(`bin --argument 123`, { timeout: 0 });
-    expect(logger.debug).toHaveBeenCalledWith({ event: 'EXEC_RETRY', retryNumber: 2 }, '(Retry #1)', 'bin --argument 123');
```
noomorph, line 100: Why did you decide to remove `retryNumber` metadata field?

noomorph (approved):
Overall, looks OK, although the code cries for more refactoring. 😭  🙂

detox/src/devices/common/drivers/android/exec/ADB.js
d4vidi: A proper installation of a large app takes no more than a few seconds, actually. Factoring in slow networks and such, 45 is quite a decent timeout. You think it should be defined more explicitly?

detox/src/utils/childProcess/exec.test.js
d4vidi: I figured that it's not very productive since the retry count appears in the log itself. WDYT?

d4vidi:
> Overall, looks OK, although the code cries for more refactoring. 😭 🙂

I was originally aiming for a single function that unifies the various `spawn()`'s and `exec()'`s, but merging both in amounted into a level of complexity I didn't think was actually worth it.

What did you have in mind?

detox/src/devices/common/drivers/android/exec/ADB.js
noomorph: Let's add a TODO pls.

detox/src/utils/childProcess/exec.test.js
noomorph: Ok.

PR #3231, opened 2022-02-22
Enable Genymotion debug

jonathanmos:
## Description
Enable running debug tests on Genycloud (discussed in [this](https://github.com/wix/Detox/issues/2655) issue)

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/Detox.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/Detox.java
@@ -4,7 +4,9 @@ import android.app.Activity;
 import android.app.Instrumentation;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 
 import com.wix.detox.config.DetoxConfig;
 import com.wix.detox.espresso.UiControllerSpy;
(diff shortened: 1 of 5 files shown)


detox/test/e2e/setup.js
```diff
@@ -1,6 +1,7 @@
 const { device } = require('detox');
 
 beforeAll(async () => {
+  await device.reverseTcpPort(8081);
```
noomorph, line 4: Is that port of the Metro bundler?

detox/android/detox/src/full/java/com/wix/detox/Detox.java
```diff
@@ -72,6 +74,7 @@
 public final class Detox {
     private static final String INTENT_LAUNCH_ARGS_KEY = "launchArgs";
     private static final long ACTIVITY_LAUNCH_TIMEOUT = 10000L;
+    private static final String DEBUG_HOST_KEY = "debug_http_host";
```
d4vidi, line 77:
```suggestion
    private static final String RN_DEBUG_HOST_KEY = "debug_http_host";
```

d4vidi (changes requested):
I'm not mentally ready to do this world-wide, as implicated by this PR.
I think we should first introduce a more local solution:
1. Specify the preferences tweak in our internal project's `DetoxTest.java`
2. Call `device.reverseTcpPort(8081)` in the single test suite where it really matters, in the associated project.

d4vidi:
@jonathanmos I think we should reiterate this, and decide on one of these 2 options:
1. Avoid integrating into Detox at this point, and solve this locally (as I've previously suggested).
2. Introduce this into Detox in a more generic and holistic way. For example, instead of `isGenymotion`, we should allow for accepting an argument that holds the URL to override with (e.g. `localhost:8081`). Also, if we do this, then we'd also have to introduce a test scheme that would run in Debug and verify that this thing works continuously.

jonathanmos:
Will be solved locally

Issue #207, opened 2017-07-20
Detox as a platform


talkol:
There has been some discussion about third party tooling to be built on top of Detox. Core Detox features like device control, commands via websocket and native synchronization are useful for a large family of tools and isn't limited only for E2E tests.

## Some examples for third-party tools

* [Fructose](https://github.com/rjanjua/fructose) is a tool by @rjanjua for functional testing of components on device
* @panarasi from Microsoft suggested creating an integration testing framework on top of Detox
* @gongreg is thinking about snapshot testing with screenshots for components on device

## Current Detox architecture and limitations

When using Detox, we have 2 different machines running side by side: the *tester* (node.js running mocha with your test scripts) and the *testee* (the app running on a device). Communication between the 2 machines is done via websocket.

JS code running on the *tester* (node.js) can have different behaviors across test scenarios. For example, if I have a mock server running on node.js, which is supposed to return A in one scenario and B in another scenario, I have no problem with configuring the server on the beginning of each scenario for the behavior I want.

JS code running on the *testee* (app) **can't** currently have different behaviors across scenarios. Since we compile and bundle the app once in the beginning of the entire suite, we can't change code between scenarios. If we want the app to behave differently in different scenarios, we can only control it differently with Detox and tap on different buttons. This is a limitation. It makes sense with E2E tests, but it doesn't make sense for the other tools people are thinking about.

## Mechanism for dynamic JS evaluation on device

One way to overcome the above limitation is to introduce API to Detox that will allow changing the JS code running on the *testee* (app) **dynamically** during a test scenario. This API will basically allow the *tester* (node.js) to evaluate JS code on the *testee* (app).

Let's assume the API will look like `device.evaluateJS('../device/file.js')`. It will run on the *tester* (node.js) and take a local JS file found on the tester, read it into a string and send the string via the Detox websocket to the *testee* (app). On the app, the Detox native framework will take the JS string and evaluate it on the current React Native JS context.

What about `require()`? If this code needs to rely on other libraries currently loaded in the React Native JS context, like `react-native` itself, we're going to have problems requiring them due to name changes in the bundling process. We can overcome this difficulty by running react native bundler in the `unbundle` mode.

Third-party tools can rely on this mechanism to manipulate the behavior of the app running on device with Detox during runtime. For example, inject the various scenarios to run on device and get integration tests instead of E2E tests.

## Communication between the app and outside the app

[...]

Issue #3208, opened 2022-02-07
iOS: XCUITest integration


asafkorem:
> [!IMPORTANT]  
> This is no longer an epic, since we decided to do the integration in smaller steps. Initial XCUITest integration is now part of [iOS: Support System Dialogs](https://github.com/wix/Detox/issues/4432) epic.

## Motivation

Currently, there are several gaps in Detox iOS we would like to eliminate, especially the non-support of these two: 
- **Multiple apps flows** (tests that involves multiple applications).
- **Real devices** (running tests on a physical device).

There are other features we would like to implement and support but we are limited to doing so, infrastructure-wise:
- Web-view interactions.
- Unmocked push notifications flows.
- System dialog interactions.

Currently, in Detox, the entire interaction with the app under test (or, **AUT**), is done from the JS side, communicating with the Detox framework, in which resides inside the app, aware of all the resources ([DetoxSync](https://github.com/wix/DetoxSync) for instance), and triggers UI events from within. This way, Detox is limited to interact only with the AUT. 

Note that, running apps on physical devices, and specifically UI tests, will require [code-signing](https://developer.apple.com/support/code-signing/) the application with signing certificates issued by Apple. 

Also, on iOS Simulators, Detox framework is injected dynamically into the application (using the `DYLD_INSERT_LIBRARIES` environment variable). For physical devices, inserting dynamic code to the application will require to code-sign the dylib with the same signing certificate of the running application ([Apple Code Signing](https://developer.apple.com/support/code-signing/)).

Since in many cases our users do not use code-signing on their applications on their local environment, we would like to keep this step optional (unless running on real devices).

## XCUITest

XCUITest is a user interface testing framework launched by Apple in 2015, the XCUITest framework rests upon two core technologies: the [XCTest framework](https://developer.apple.com/documentation/xctest) and [Accessibility](https://developer.apple.com/accessibility/ios/).

The UI testing in XCUITest is based on the semantic data about the UI that is exposed as "Accessibility", it works by locating the UI elements using queries, sending UI events to these objects and providing a rich API enabling us to examine the elements' attributes and state, and comparing them against an expected state.

The transition to accessibility-based operations, will allow Detox to interact with multiple apps, iOS SpringBoard, push notifications, system dialogs and other system elements all managed from the test target process, without having to inject any dynamic code injected into the AUT. We should also be able to operate on apps without the additional phase of inserting dynamic code.

<p align="center">
[...]

PR #3245, opened 2022-03-01
Implement backdoor feature

codebutler:
## Description

- This pull request addresses the issue described here: #207

This makes it possible for tests to send arbitrary messages to the app being tested for the purpose of configuring state or running any other special actions.

Usage inside a test looks like:

```tsx
await device.backdoor({ action: "do-something" });
```

Then receive the event in a React Native app:

```tsx
const emitter = Platform.OS === "ios" ? NativeAppEventEmitter : DeviceEventEmitter;
emitter.addListener("detoxBackdoor", ({ action }) => {
  // do something based on action
});
```

Inspired by a similar feature in the Xamarin test framework:
https://docs.microsoft.com/en-us/appcenter/test-cloud/frameworks/uitest/features/backdoors

---

 - [X] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files.

> _For API changes:_


cjshearer:
I was just looking for this functionality a week ago! I'm surprised it's not already a feature, but glad to see I'm not the only one who wants this.

My workaround for this is a hacky mess involving a mock entry file that listens for a launch argument that injects some data into the app, and even this only works in debug mode.

asafkorem:
Hey @codebutler! 
First of all - thank you for the pull request!

I went through the changes you implemented in this PR, I must say that this is definitely an interesting idea. However, I'm still not convinced that Detox (or generally, E2E testing framework) should have it, can you please provide me with some use cases (or even better, your use case) where you think this feature is necessary?

We'll have to consult about this change internally, probably next week (with @d4vidi, which is OOO this week), and to understand whether it is desirable. This change might also interfere with our current plans and work ([reimplementing Detox-iOS with XCUITest framework](https://github.com/wix/Detox/issues/3208)).

asafkorem:
@cjshearer can you please elaborate about the kind of data you'd like to pass the app on run-time, and what flows are you trying to test?

noomorph:
Looks good to me, I'll vote up. The discussion on this PR is ongoing internally.

cjshearer:
> @cjshearer can you please elaborate about the kind of data you'd like to pass the app on run-time, and what flows are you trying to test?

Our app stores data recorded from a proprietary BLE device as JSON in AsyncStorage. These recordings are displayed in a list that the user can select from, allowing them to view a graph of the recording data, and some metadata for these recordings.

cjshearer:
> @cjshearer can you please elaborate about the kind of data you'd like to pass the app on run-time, and what flows are you trying to test?

Our app stores data recorded from a proprietary BLE device as JSON in AsyncStorage. These recordings are displayed in a list that the user can select from, allowing them to view a graph of the recording data and some metadata for these recordings. We want these automated tests to run on iOS and Android simulators, so we cannot have it interact with any Bluetooth simulator. We do have plans to mock the BLE device, but we would still like to have a quick way to inject these recordings without actually making them at test-time (some of the recordings we would like to inject would be hours long). Our app supports importing recordings shared to it, so it would be convenient to simply hook into that functionality with the proposed backdoor feature.

Let me know if you need anymore info!

codebutler:
> Hey @codebutler! First of all - thank you for the pull request!
> 
> I went through the changes you implemented in this PR, I must say that this is definitely an interesting idea. However, I'm still not convinced that Detox (or generally, E2E testing framework) should have it, can you please provide me with some use cases (or even better, your use case) where you think this feature is necessary?

Thanks for taking a look! My thinking was that this is an appropriate feature for a "gray box" testing framework.

We have "developer" options in our app to tweak internal state for flows that are difficult to test. This makes it easy to hook into that existing functionality.

> We'll have to consult about this change internally, probably next week (with @d4vidi, which is OOO this week), and to understand whether it is desirable. This change might also interfere with our current plans and work ([reimplementing Detox-iOS with XCUITest framework](https://github.com/wix/Detox/issues/3208)).

Can I make any changes now to better prepare for that work?

noomorph:
@codebutler I guess what you also need is to implement some simple handler in the `detox/example` e2e test app and at least 1 E2E test.

noomorph:
@d4vidi @jonathanmos I love this feature and I'd like to see it available in some time.

noomorph:
@d4vidi @jonathanmos I love this feature and I'd like to see it available in some time.

It should serve a good purpose in lightweight integration tests, maybe Storybook-ish stuff, screenshot testing, etc.
It might help get rid of some developer-only screens we use for some E2Es.
It seems truly useful.

If you approve the idea, I can take charge of this PR and help @codebutler to get it ready for merge.

d4vidi:
Being a UI testing tool, normally we would advise doing this through a custom test screen with custom mock options. ATM I don't understand what in the the BLE use case makes it special. That said, I _would_ consider that in terms of convenience, although, the API (_backdoor_) must go out under a different name. Also, I would try to change the impl. such that it would be performed truly synchronously (namely, send `backdoorDone` message only when handling completes).

I think the right course of action in developing this idea would be to set up a sync discussion on it among ourselves (aka a meeting).

d4vidi:
BTW @codebutler as we've already merged the unit tests change, we would appreciate you rebasing over origin/master 🙏🏻

asafkorem:
@wix/detox-team guys, I don't want this PR to stay open forever. Are we going to close it or to discuss on it?

wouterh-dev:
For what it's worth I would be very interested in this functionality 🙏

It would make implementing custom mocks much easier.

noomorph:
I still think this PR is awesome and it is a shame that we didn't finish it and adopt to Detox.

noomorph:
My technical concern so far is that it looks like adding asynchronous callbacks is not possible (addEventListener).

A somewhat acceptable workaround would be to "wait until the app is ready", but I'd prefer to verify it wouldn't be flaky due to possible race conditions.

abajajct:
Would this work allow us to change the date during runtime of test?

noomorph:
Technically yes, but that would be achievable with an advanced mock server as well (so nothing actually unique to the backdoor feature) - just less legwork, yet still some good old mocking.

codebutler:
Hi, I'm no longer working with this project so I'm closing this. PR I agree that most (all?) of these uses cases can be handled by a custom mock server instead of this approach. Thanks.

noomorph:
@codebutler thanks for your hard work. I still have a hope to bring this feature to Detox in one of major updates and I'll give you all the proper credit if/when that happens.

PR #3247, opened 2022-03-03
website: add versioning feature.

asafkorem:
Adds an option for documentation version release when publishing a new version from Buildkite, to create a new version or to replace an older one. See Docusaurus [versioning doc](https://docusaurus.io/docs/next/versioning).

**Demo:**
image

PR #3250, opened 2022-03-06
website: add Twitter follow button, replace cover image format.

asafkorem:
### Twitter follow button 🐦 
image
### Replace cover image with progressive JPEG 🖼️ 
The new cover is encoded in such a way that enables browsers to display a progressively more precise image as it loads. This means that at first, the image appears blurry (or incomplete) but with full-dimensions, then it progressively gets sharper and better (improves loading performance).


asafkorem:
Extracted from https://github.com/wix/Detox/pull/3247 to minimize the amount of changes in one PR.

PR #3254, opened 2022-03-08
Fix release scripts and package

igorgn:
## Description
This is a fix for buildkite CI

incorrect setting manual tag
missing ios/android in final package

---

> _For API changes:_


asafkorem:
Let's add the tag to the version name + tag counter (if it isn't `latest`), i.e. releasing version `20.0.0` with tag `smoke` should be `20.0.0-smoke.0`, next `20.0.0`+`smoke` will be `20.0.0-smoke.1` and so on..

igorgn:
> Let's add the [prerelease identifier](http://semver.org/#spec-item-9): That is, releasing version `20.0.0` with tag `smoke` should be `20.0.0-smoke.0`, next `20.0.0`+`smoke` will be `20.0.0-smoke.1` and so on.. (same as we had with Jenkins)
> 
> **Edit:** using this: https://github.com/lerna/lerna/blob/main/commands/publish/README.md#--preid

while doing Release, need to select 'prerelease, premajor or preminor'. This will produce 'pre-' version.

asafkorem:
> > Let's add the [prerelease identifier](http://semver.org/#spec-item-9): That is, releasing version `20.0.0` with tag `smoke` should be `20.0.0-smoke.0`, next `20.0.0`+`smoke` will be `20.0.0-smoke.1` and so on.. (same as we had with Jenkins)
> > **Edit:** using this: https://github.com/lerna/lerna/blob/main/commands/publish/README.md#--preid
> 
> while doing Release, need to select 'prerelease, premajor or preminor'. This will produce 'pre-' version.

Ohh right, so we already have it.. nice catch 🙂 
That's really confusing (as we have already experienced 😅 ), let's add a note about this (maybe add some `hint` the release yaml?)

asafkorem:
mmm, so after going over this line again: https://github.com/wix/Detox/blob/95700e061cbc73a8bdcdc9e731ecebbc74e8b5e2/scripts/ci.publish.js#L37
I'm thinking maybe it will be better to replace the `prerelease`, `premajor` and `preminor` options with a flag (checkbox) that indicates whether this is a pre-version (for instance, `minor`+`isPreVersion=true` => `preminor`).

Even better, to have all pre-versions with the **same tag** ("smoke" or "alpha" or something), and to deduce it from the pre-version indicator? 
I mean, the `isPreVersion`/`isSmokeVersion` flag will determine whether this is a pre-version, and whether to add the `smoke` tag.. This way, we'll have only one source of truth.

What do you think @igorgn? 
Is there a possible scenario where we're having a pre-version with `latest` tag or a regular version with not-`latest` tag?

scripts/ci.common.js
```diff
@@ -10,29 +10,30 @@ const logSection = (message) => console.log(chalk.blue(`[RELEASE] ${message}`));
...
 
-function getIsRelease() {
-  return isRelease;
+function isRelease() {
+  return RELEASE;
 }
 
 function getReleaseVersionType() {
-  return RELEASE_VERSION_TYPE;
+  return PRE_RELEASE === 'true' ? 'pre' + RELEASE_VERSION_TYPE : RELEASE_VERSION_TYPE;
```
asafkorem, line 28: I love it! 💯

.buildkite/pipeline.release.yml
```diff
@@ -3,6 +3,16 @@ steps:
     prompt: "Fill out the details for release"
     if: 'build.message =~ /^release\$/i'
     fields:
+      - select: "PRE-RELEASE?"
+        key: "pre-release"
+        default: true
+        hint: "A corresponding prefix will be added to published version ('alpha' or the one specified in RELEASE_NPM_TAG)"
```
asafkorem, line 9: So, I'm supposed to get `alpha` npm-tag when `pre-release == true && release-npm-tag == null`?

igorgn: by default (`RELEASE_NPM_TAG === 'null'`) `master` will return 'prepatch, preminor or premajor'.
branch will return branch name, until `RELEASE_NPM !== 'null'`

so on master we could have` 1.2.3-preminor.0` OR `1.2.3-customTag.[0..n]`
branch release could produce` 1.2.3-branchName.0 `OR `1.2.3-customTag.[0..n]`

This way it is easier to distinguish master version from branch one
@d4vidi @asafkorem

asafkorem (approved):
w00t!
Looks good to me, pair-reviewed with @igorgn.

PR #3287, opened 2022-04-01
Add discord server invite links.

asafkorem:
Preview:
image


noomorph (approved):
Looks good. Btw, how are you going to publish the website without actually releasing Detox binary?

asafkorem:
> Looks good. Btw, how are you going to publish the website without actually releasing Detox binary?

Thanks. The [website GH action](https://github.com/wix/Detox/actions/workflows/website.yml) runs a test on every pull-request and publish the website with its changes on every push into `master`.

noomorph:
@asafkorem which actually means, that after we merge the `versioning` PR, we'll have two different entry points that might update the website? Two sources of truth? 🤔

asafkorem:
> @asafkorem which actually means, that after we merge the `versioning` PR, we'll have two different entry points that might update the website? Two sources of truth? 🤔

@noomorph sources of which truth?
The GitHub workflow (`website`) is responsible to update our GH pages (the website) once the codebase (on `master`) is changed.
The documentation phase in the CI release script changes the codebase, by pushing a commit to `master` with the new version. The deployment of this change is handled only by the `website` workflow (since we push it into `master`).

This is reasonable to couple them together, true, though if we prefer the idea of having all website automated stuff in one place, this means we'll have to decouple the docs versioning phase and the npm release phase, which I believe we should keep together for the reasons of version management.

noomorph:
@asafkorem, ah, now it makes more sense. I'm not sure if this won't cause merge conflicts from time to time, actually... but nevertheless, I understand your idea now.

d4vidi:
I think we should upgrade the button a bit, as it is something that can be easily done. For example, to something like this - http://www.ageofcivilizationsgame.com/uploads/monthly_2019_06/discord_button.png.3c4e31057287000ade2e552fdc414e79.png

This might be helpful - https://blog.discord.com/add-the-discord-widget-to-your-site-d45ffcd718c6

asafkorem:
> This might be helpful - https://blog.discord.com/add-the-discord-widget-to-your-site-d45ffcd718c6

This blog post refers to this kind of widget: https://user-images.githubusercontent.com/55082339/161419617-540f7d36-554d-4523-a2b2-0a46ef307c7a.png. 

However the button you sent is doable, I can use some image for that.

d4vidi:
Also, I think we should add a shields.io to the main README, while we're at it

asafkorem:
already done..

asafkorem:
image

PR #3290, opened 2022-04-03
website: replace discord link with image button.

asafkorem:
Before:
image

After:
image


d4vidi:
I hate to be nit-picky about this but kind of feels off, visually
needs to align to the other components height-wise and be a bit less crunched in internally

asafkorem:
OK, changed:
image

d4vidi (approved):

PR #3299, opened 2022-04-06
Build also for arm64-v8a

jonathanmos:
## Description
Build local test apk also for arm64-v8a to allow running on M1


d4vidi:
@jonathanmos what about the example projects, while we're at it?

d4vidi:
@igorgn would it be possible to integrated M1 agents on our OSS?

d4vidi (approved):

PR #3313, opened 2022-04-10
build(releaseDocumentation): push version changes to master.

asafkorem:
Fix the docs version release (`releaseDocumentation.js`) by pushing the version changes to `wix/Detox:master` branch.
Before that, the changes were committed, but [only locally](https://buildkite.com/wix-mobile-oss/detox/builds/497#bd4beb87-5312-42f4-a6c7-535f6a474b40/117-965) (on the local clone made during the release process).


scripts/utils/releaseDocumentation.test.js
```diff
@@ -19,13 +19,15 @@ describe('Documentation script', () => {
...
+
+    expect(exec.execSync).toHaveBeenCalledTimes(5);
     expect(process.chdir).toHaveBeenCalledWith(docsPath());
 
     expect(exec.execSync).toHaveBeenCalledWith(`npm install`);
     expect(exec.execSync).toHaveBeenCalledWith(`npm run docusaurus docs:version ${version}`);
     expect(exec.execSync).toHaveBeenCalledWith(`git add .`);
     expect(exec.execSync).toHaveBeenCalledWith(`git commit -m "Publish docs version ${version}"`);
+    expect(exec.execSync).toHaveBeenCalledWith(`git push origin master`);
     expect(process.chdir).toHaveBeenCalledWith(process.cwd());
```
d4vidi, line 31: Not that I mind all that much, but technically I don't find this type of tests very useful.

asafkorem: I definitely agree. It was forked from navigation. So you would not mind if I delete it? 😄

d4vidi: BTW, mocking `fs` and `shell-utils` (i.e. node's native modules) is discouraged.

scripts/utils/releaseDocumentation.js
```diff
@@ -16,6 +16,7 @@ function buildDocsForVersion(version) {
     exec.execSync(`npm run docusaurus docs:version ${version}`);
     exec.execSync(`git add .`);
     exec.execSync(`git commit -m "Publish docs version ${version}"`);
+    exec.execSync(`git push origin master"`);
```
d4vidi, line 19: @asafkorem @igorgn anything to worry about in terms of release script? Is it cool to hardcode master this way?

scripts/utils/releaseDocumentation.test.js
asafkorem: Removed 🙂

scripts/utils/releaseDocumentation.js
d4vidi: IIUC, technically, the only use cases where this would be invalid is where -
1. We would somehow want to preview docs of a pre-release
2. We would rename master
Both unlikely...
Or have I overlooked something?

asafkorem: mmm good point, I think we can take the upstream branch from the `process.env` args..

asafkorem: [Done](https://github.com/wix/Detox/pull/3313/commits/1dc4c24521867f215e4feb121e266fcab07e536e).

d4vidi: Oh cool. Good to be on the safe side.

d4vidi (approved):

PR #3326, opened 2022-04-12
docs: add native debugging guide

noomorph:
## Description

Covers the new way to debug native iOS code via `launchApp: 'manual'` in Detox.


asafkorem:
Good stuff!
BTW, in addition to what we've already talked about in the channel, don't forget the meta data for docusaurus ([like this](https://github.com/wix/Detox/blob/docs/native-debugging-guide/docs/APIRef.ActionsOnElement.md?plain=1#L1)) 🙂

PR #3333, opened 2022-04-14
[Detox 20] feat: new global lifecycle for Detox

noomorph:
Resolves https://github.com/wix/Detox/issues/3354

#### Why did I touch logs?

Because the global context now allows merging the logs in a cleaner way after the test session ends. 

#### Why did I touch trace/traceCall?

Not sure, maybe just for fun. I consider it to be smart to use a logging mechanism to produce timelines. That might be optional – maybe I'll revert most of the change.


d4vidi:
@noomorph can this be changed to _draft_?

noomorph:
@d4vidi , no, GitHub does not allow changing back to drafts.

PR #3359, opened 2022-04-27
build: limit peerDependencies to jest@27 max

noomorph:
## Description

It turns out that in the recent `npm` package manager versions (8.x) the fact of having `peerDependencies` can dramatically change the installation outcome. Hence, I am considering limiting the supported Jest versions to 26..27.

Detox 20 will have support for 27..28, respectively.

@d4vidi WDYT?

P. S. In fact, I myself am not 100% sure what the most correct solution is.

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -1,7 +1,7 @@
 {
   "name": "detox",
   "description": "E2E tests and automation for mobile",
-  "version": "19.6.6",
+  "version": "19.6.7-build.limit-jest-27.0",
   "bin": {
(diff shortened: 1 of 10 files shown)


d4vidi:
I'm generally in favor of locking until the skies clear out.

Other than that, the rules seem to be ok:

image


image

d4vidi (approved):

d4vidi:
What the heck has happened with Buildkite, though?

noomorph:
Superseded by #3377 – the latter provides the complete solution for the issue.

PR #3371, opened 2022-05-03
feat(config): remove legacy dash props

noomorph:
Resolves https://github.com/wix/Detox/issues/3355

## Description

Removes old and **undocumented** global configuration properties: `test-runner` and `runner-config`.

Instead, their camelCase alternatives should be used: `testRunner` and `runnerConfig`.


d4vidi:
Wow, this this is what it boils down to? No kebabs anywhere else at all? ♨️
Honestly, I wasn't even aware we ever supported that

PR #3372, opened 2022-05-03
docs: test runners clarification

noomorph:
## Description

Replaces `test-runner` property notice from YAML templates and Markdown docs with an up-to-date information:

1. Jest test runner is the default and it is implied.
2. Instead of undocumented `test-runner` syntax, we now suggest the correct and documented `testRunner` property as a source of truth.

Last but not least, `testRunner` property still remains supported by Detox for custom commands (e.g. `nyc jest`) and (possibly) other runners.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,6 @@ The most difficult part of automated testing on mobile is the tip of the testing
 - **Runs on Devices** (not yet supported on iOS): Gain confidence to ship by testing your app on a device/simulator just like a real user.
 - **Automatically Synchronized:** Stops flakiness at the core by monitoring asynchronous operations in your app.
 - **Made For CI:** Execute your E2E tests on CI platforms like Travis without grief.
-- **Test Runner Independent:** Use Jest, Mocha, AVA, or any other JavaScript test runner you like (spoiler: we have our favorite).
 - **Debuggable:** Modern `async`-`await` API allows breakpoints in asynchronous tests to work as expected.
 
 ## Supported Versions
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT_ANDROID.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT_ANDROID.yml
@@ -42,7 +42,7 @@ body:
     id: environment
     attributes:
       label: In what environment did this happen?
-      description: "Note: the test-runner is set in your detox configuration file (e.g. package.json, detox.config)."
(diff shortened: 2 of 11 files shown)


d4vidi:
Oh man

noomorph:
@d4vidi refresh.

README.md
```diff
@@ -49,7 +49,6 @@ The most difficult part of automated testing on mobile is the tip of the testing
 - **Runs on Devices** (not yet supported on iOS): Gain confidence to ship by testing your app on a device/simulator just like a real user.
 - **Automatically Synchronized:** Stops flakiness at the core by monitoring asynchronous operations in your app.
 - **Made For CI:** Execute your E2E tests on CI platforms like Travis without grief.
-- **Test Runner Independent:** Use Jest, Mocha, AVA, or any other JavaScript test runner you like (spoiler: we have our favorite).
```
noomorph, line 52: @d4vidi  I had to do it. 🤷 
@asafkorem , @jonathanmos 

If you can come up with another selling statement... like "using the most natural test runner for React Native which is Jest, mu-ha-ha" 😈

d4vidi: This is quite alright, as it aligns with what we've agreed upon. We will be reiterating this, of course 😅 
Nevertheless, there's still one more place to clean up: https://github.com/wix/Detox/blob/next/website/src/components/HomepageFeatures.js#L11

image

@asafkorem this is the place jump in on the discussion 😄

d4vidi: This is quite alright, as it aligns with what we've agreed upon. We will be reiterating this, of course 😅 
Nevertheless, there's still one more place to clean up: https://github.com/wix/Detox/blob/next/website/src/components/HomepageFeatures.js#L11

image

@asafkorem this is the place jump in on the discussion 😄

asafkorem: @noomorph I suggest to replace it with _"Supports Multiple Workers" / "Supports Parallel Testing"_ (description can be smth like: _"Run parallel tests with multiple workers."_)
We can keep the current icon (it fits), or use https://uxwing.com/engineer-workers-icon/. WDYT?

asafkorem: I mean, to replace the outdated homepage section. To the readme file we can add both bullets (parallel testing & using Jest)

noomorph: Brilliant idea! Thanks for your $0.02, Asaf - indeed, it fits!

PR #3377, opened 2022-05-05
fix: jest@28 compatibility

noomorph:
## Description

Fixes two issues with the new Jest 28:

1. Handling a possible `default` export from `jest-environment-node`.
2. Handling a possible new config format: now it is not `projectConfig`, rather `{ projectConfig, globalConfig }`.3.

### DIFF
--- a/detox/runners/jest-circus/environment.js
+++ b/detox/runners/jest-circus/environment.js
@@ -1,5 +1,6 @@
 // @ts-nocheck
-const NodeEnvironment = require('jest-environment-node');
+const maybeNodeEnvironment = require('jest-environment-node');
+const NodeEnvironment = maybeNodeEnvironment.default || maybeNodeEnvironment;
 
 const DetoxError = require('../../src/errors/DetoxError');
 const Timer = require('../../src/utils/Timer');
--- a/detox/runners/jest-circus/utils/assertJestCircus26.js
+++ b/detox/runners/jest-circus/utils/assertJestCircus26.js
@@ -3,12 +3,14 @@ const path = require('path');
 
 const DetoxRuntimeError = require('../../../src/errors/DetoxRuntimeError');
 
-function assertJestCircus26(config) {
-  if (!/jest-circus/.test(config.testRunner)) {
+function assertJestCircus26(maybeProjectConfig) {
+  const projectConfig = maybeProjectConfig.projectConfig || maybeProjectConfig;
+
+  if (!/jest-circus/.test(projectConfig.testRunner)) {
     throw new DetoxRuntimeError('Cannot run tests without "jest-circus" npm package, exiting.');
   }
 
-  const circusPackageJson = path.join(path.dirname(config.testRunner), 'package.json');
+  const circusPackageJson = path.join(path.dirname(projectConfig.testRunner), 'package.json');
   if (!fs.existsSync(circusPackageJson)) {
     throw new DetoxRuntimeError('Check that you have an installed copy of "jest-circus" npm package, exiting.');
   }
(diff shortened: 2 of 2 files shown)


detox/package.json
```diff
@@ -81,10 +81,17 @@
     "yargs-unparser": "^2.0.0"
   },
   "peerDependencies": {
-    "jest-circus": "26.0.x - 26.4.x || >=26.5.2",
-    "jest-environment-node": ">=25.0.0",
+    "jest": "26.0.x - 26.4.x || ^26.5.2 || 27.x.x || 28.x.x",
     "mocha": ">=6.0.0"
   },
+  "peerDependenciesMeta": {
```
d4vidi, line 87: Nice!!!

detox/package.json
```diff
@@ -81,10 +81,17 @@
...
+    "jest": {
+      "optional": true
+    },
+    "mocha": {
+      "optional": true
+    }
+  },
```
d4vidi, line 94: Doesn't that effectively allows both to be absent from `node_modules/`?

detox/runners/jest-circus/environment.js
```diff
@@ -1,5 +1,6 @@
 // @ts-nocheck
-const NodeEnvironment = require('jest-environment-node');
+const maybeNodeEnvironment = require('jest-environment-node'); // eslint-disable-line node/no-unpublished-require
+const NodeEnvironment = maybeNodeEnvironment.default || maybeNodeEnvironment;
```
d4vidi, line 3: Lol... But I get it 💯

detox/package.json
noomorph: Yes, but technically we allow to use Detox with other runners as well, so... 🤷‍♂️

detox/runners/jest/JestCircusEnvironment.js
```diff
@@ -1,4 +1,6 @@
-const NodeEnvironment = require('jest-environment-node');
+const maybeNodeEnvironment = require('jest-environment-node'); // eslint-disable-line node/no-unpublished-require
+// @ts-ignore
+const NodeEnvironment = maybeNodeEnvironment.default || maybeNodeEnvironment;
```
d4vidi, line 3: Though a very subtle one - still a dup. Would it make sense to import through a custom file (e.g. `detox/somewhere/NodeEnvironment.js`, instead?

noomorph: No sense. This integration is dropped in Detox 20

docs/Guide.Jest.md
```diff
@@ -30,12 +30,16 @@ For its part, Detox supports only one Jest’s concrete runner, which is [`jest-
...
+Afterward, install the respective npm package:
 
 ```sh
-npm install -D jest
+npm install -D "jest@>=27.2.5"
 ```
 
+**NOTE:** The command will install the _latest Jest version_. However, `@>=27.2.5` addendum is recommended just to be on the safe side in a common scenario, when a `package-lock.json`
+generated by an official React Native project template limits Jest version to a very old `26.x`, maybe due to some optimization mechanism.
+Generally we recommend not to stay on outdated Jest versions for too long, e.g. `jest@27.2.5` will be the minimal version supported by Detox 20.
```
d4vidi, line 41: 💪🏻

detox/runners/jest/JestCircusEnvironment.js
d4vidi: Because of this, you mean? -
>Generally we recommend not to stay on outdated Jest versions for too long, e.g. `jest@27.2.5` will be the minimal version supported by Detox 20.

noomorph: No, this integration is very old and will be dropped regardless. The next minimal version requirement itself is motivated by the streamline reporter monkey-patch.

The maybe/default trick will remain because we'll be supporting both Jest 27/28

PR #3398, opened 2022-05-16
ci: upgrade to iPhone 12 Pro

noomorph:
## Description

Reaction to: https://buildkite.com/wix-mobile-oss/detox/builds/661#baf766f9-31bc-4ede-9901-eb758aeae137/1078-1143

Hopefully, it should fix the issue.


noomorph:
@asafkorem  I wonder maybe it still makes sense to align the simulator version across test and example projects?

d4vidi:
It was aligned on iPhone 11 Pro, was it not?

d4vidi:
Sorry guys, is there a real reason for this, having already realized that the device type doesn't affect the build's stability?

asafkorem:
Guys, I fail to reproduce the original error locally :/
I think we should close this pull request for now

PR #3400, opened 2022-05-16
build: move jest-circus/ facade to jest/

noomorph:
## Description

In this PR, since there is only single integration with `jest` and `jest-circus`, and the latter is a default choice for Jest 27, and the only choice for Jest 28 (unless `jest-jasmine2` is installed and configured explicitly), I consider it justified to shorten the paths to the user-facing files.

### Required changes from the users

In the environment class file:

```diff
 const {
   DetoxCircusEnvironment,
   SpecReporter,
   WorkerAssignReporter,
-} = require('detox/runners/jest-circus');
+} = require('detox/runners/jest');
```

In Jest config:

```diff
   "testTimeout": 120000,
-  "reporters": ["detox/runners/jest/streamlineReporter"],
+  "reporters": ["detox/runners/jest/reporter"],
   "verbose": true
```

### Migration flow

As for the attempts to import the files at their old locations, the users will be seeing warnings:

```
[DEPRECATION] Detox detected an attempt to require a module from an outdated location, please change in your project:
- detox/runners/jest-circus/environment
+ detox/runners/jest/environment
```

That should enact the migration process smoothly enough.


d4vidi:
@noomorph kudos for finding those building blocks that allow moving forward with this in a step-by-step fashion ❤️

detox/runners/migration.js
```diff
@@ -0,0 +1,36 @@
+const path = require('path');
```
d4vidi, line 1: @noomorph this is quite thoughtful of you 😄
Are we going to keep this for good or could we get rid of that at some point?

noomorph: Detox 21 probably

d4vidi: OK: https://github.com/wix/Detox/discussions/3406

PR #3412, opened 2022-05-22
Add appcompat dependency to docs

jonathanmos:
## Description
When a project is setup through react-native init, the following appcompat dependency is not added. 

    implementation 'androidx.appcompat:appcompat:1.1.0'

This later causes the following error when building Detox in release mode:

    Execution failed for task ':app:processReleaseAndroidTestResources'.
    > A failure occurred while executing com.android.build.gradle.internal.res.LinkApplicationAndroidResourcesTask$TaskAction
       > Android resource linking failed
         ERROR:: AAPT: error: resource style/Widget.AppCompat.TextView ...

In this pull request I have updated the docs with the solution for this issue.

### DIFF
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -160,7 +160,7 @@ also ensure that this line appears in `dependencies`:
 ```groovy
 dependencies {
     // ...
-    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.appcompat:appcompat:1.1.0' // (check what the latest version is!)
 }
 ```
 
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -152,6 +152,14 @@ In your app’s `buildscript` (i.e. `android/app/build.gradle`) add this in `dep
 dependencies {
     // ...
     androidTestImplementation('com.wix:detox:+')
+}
+```
+
+also ensure that this line appears in `dependencies`:
+
+```groovy
+dependencies {
+    // ...
     implementation 'androidx.appcompat:appcompat:1.1.0'
 }
 ```
--- a/docs/Introduction.Android.md
+++ b/docs/Introduction.Android.md
@@ -152,6 +152,7 @@ In your app’s `buildscript` (i.e. `android/app/build.gradle`) add this in `dep
 dependencies {
     // ...
     androidTestImplementation('com.wix:detox:+')
+    implementation 'androidx.appcompat:appcompat:1.1.0'
 }
 ```
 
--- a/docs/Troubleshooting.BuildingTheApp.md
+++ b/docs/Troubleshooting.BuildingTheApp.md
@@ -13,9 +13,9 @@ For troubleshooting of other issue, refer to our [troubleshooting index](Trouble
 
 ## Android
 
-### Problem: AAPT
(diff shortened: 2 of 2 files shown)


noomorph (approved):

docs/Troubleshooting.BuildingTheApp.md
```diff
@@ -13,7 +13,27 @@ For troubleshooting of other issue, refer to our [troubleshooting index](Trouble
 
 ## Android
 
-### Problem: minSdkVersion mistmatch
+### Problem: AAPT
```
d4vidi, line 16: Lovin' this troubleshooting section. The heading, on the other hand, isn't very descriptive. `AAPT` can mean so many things...

docs/Introduction.Android.md
```diff
@@ -152,6 +152,7 @@ In your app’s `buildscript` (i.e. `android/app/build.gradle`) add this in `dep
 dependencies {
     // ...
     androidTestImplementation('com.wix:detox:+')
+    implementation 'androidx.appcompat:appcompat:1.1.0'
```
d4vidi, line 155: This is definitely the right page for this but I don't like it under this section (named, "Add the native detox dependency"). 'Need to put things under the right context...

d4vidi (changes requested):
This is great but needs polishing.

d4vidi:
@igorgn on a different matter: any ideas why buildkite appears twice (regular and pr)?

image

igorgn:
@d4vidi It is because branch build was triggered manually and not via '_git push_'. "_buildkite/detox/pr_" is there because this check is set as "_Required_", thus it will be present in **every** PR.  But manual build is considered as a regular "**buildkite/detox**", making it a separate check and thus we have 2 checks. 

In buildkite we can disable this behaviour, leaving only "**buildkite/detox**" check, i.e. PR checks also would be marked as "**buildkite/detox**"

docs/Troubleshooting.BuildingTheApp.md
```diff
@@ -13,7 +13,27 @@ For troubleshooting of other issue, refer to our [troubleshooting index](Trouble
...
+   > Android resource linking failed
+     ERROR:: AAPT: error: resource style/Widget.AppCompat.TextView ...
+```
+
+Ensure that the following line appears in your app's `buildscript` (`android/app/build.gradle`) in the `dependencies` section:
+
+```groovy
+dependencies {
+    // ...
+    implementation 'androidx.appcompat:appcompat:1.1.0'
```
d4vidi, line 32: @jonathanmos Sorry I haven't mentioned this before (although it _did_ come to mind): Using the static `1.1.0` version is tricky here. I think we should add some kind of an indication saying that `1.1.0` may not be the latest/most suitable version for the user's project, etc.

docs/Introduction.Android.md
```diff
@@ -155,6 +155,15 @@ dependencies {
 }
 ```
 
+also ensure that this line appears in `dependencies`:
+
+```groovy
+dependencies {
+    // ...
+    implementation 'androidx.appcompat:appcompat:1.1.0'
```
d4vidi, line 163: @jonathanmos version notice here too

d4vidi (approved):

PR #3432, opened 2022-05-30
fix(android): record device log during the test even if the app is not running

noomorph:
## Description

Fixes the logical error when an artifact gets **SEEMINGLY** started, stopped, saved or discarded, yet inside there's **NO-OP**.

That breaks assumptions in some third-party plugins (e.g. our alpha version of Allure integration) that _if Artifact exists it means there's some file in the end_. Let's keep it simple and enforce that convention: **an artifact means at least 1 file or directory.**

- This pull request addresses the issue reported by Vilius Mileris privately in our Detox channel.


detox/src/artifacts/log/android/ADBLogcatRecording.js
```diff
@@ -61,20 +52,12 @@ class ADBLogcatRecording extends Artifact {
...
   async doSave(artifactPath) {
-    if (this.hasRecordedFile) {
-      await this.adb.pull(this.deviceId, this.pathToLogOnDevice, artifactPath);
-      await this.adb.rm(this.deviceId, this.pathToLogOnDevice);
-    } else {
-      log.debug(`Skipping saving artifact because the recording has not started: ${artifactPath}`);
-    }
+    await this.adb.pull(this.deviceId, this.pathToLogOnDevice, artifactPath);
+    await this.adb.rm(this.deviceId, this.pathToLogOnDevice);
   }
```
asafkorem, line 57: @noomorph can we get here before the log-file is created?

asafkorem: Oh, I get this, we throw an error on `doStart` if the log-file wasn't created.

detox/src/artifacts/log/android/ADBLogcatRecording.js
```diff
@@ -26,26 +25,18 @@ class ADBLogcatRecording extends Artifact {
...
 
-  get hasRecordedFile() {
-    return !!this._waitUntilLogFileIsCreated;
-  }
-
   async doStart() {
     const pid = this.pid.get();
 
-    if (pid > 0) {
```
d4vidi, line 36: @noomorph I wonder - shouldn't we be asking ourselves how come `pid` is unavailable, instead of trying to fix?

detox/src/artifacts/log/android/ADBLogcatRecording.js
```diff
@@ -26,26 +25,18 @@ class ADBLogcatRecording extends Artifact {
...
+      pid: pid > 0 ? pid : 0,
+    });
 
-      this._waitUntilLogFileIsCreated = sleep(300).then(() => {
-        return retry(() => this._assertLogIsCreated());
-      });
-    } else {
-      log.debug('Ignoring a command to start recording, because PID of the app is missing');
-    }
+    this._waitUntilLogFileIsCreated = sleep(300).then(() => {
```
asafkorem, line 37: Can't we observe this event without sleeping (or shorter retry intervals with timeout)?

noomorph: No.

Issue #3440, opened 2022-06-07
Crash - NoSuchMethodError: No static method isBlank


ngandhy:
### What happened?

App crashes with error 

```
AndroidRuntime: java.lang.NoSuchMethodError: No static method isBlank(Ljava/lang/CharSequence;)Z in class Lj/a/a/a/b; or its super classes (declaration of 'j.a.a.a.b' appears in /data/app/~~Y3v30P120k32Qg==/com.MY.APP-p1c2xMu-ymxs_w==/base.apk!classes2.dex)
AndroidRuntime: 	at com.wix.invoke.MethodInvocation.invoke(MethodInvocation.java:30)
AndroidRuntime: 	at com.wix.invoke.MethodInvocation.invoke(MethodInvocation.java:26)
AndroidRuntime: 	at com.wix.invoke.MethodInvocation.invoke(MethodInvocation.java:20)
AndroidRuntime: 	at com.wix.detox.adapters.server.InvokeActionHandler.handle(DetoxActionHandlers.kt:54)
AndroidRuntime: 	at com.wix.detox.adapters.server.ActionsExecutor$executeAction$$inlined$let$lambda$1.run(DetoxActionsDispatcher.kt:64)
AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:938)
AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:99)
AndroidRuntime: 	at android.os.Looper.loopOnce(Looper.java:201)
AndroidRuntime: 	at android.os.Looper.loop(Looper.java:288)
AndroidRuntime: 	at com.wix.detox.adapters.server.ActionsExecutor$1.run(DetoxActionsDispatcher.kt:50)
AndroidRuntime: 	at java.lang.Thread.run(Thread.java:920)
```

### What was the expected behaviour?

Test should start and app should not crash

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [ ] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### Help us reproduce this issue!

I'm currently reproducing by:
1) detox build --configuration android.release
2) detox test --configuration android.release

### In what environment did this happen?

Detox version: 19.7.1
React Native version: 0.66.4
Node version: 
Device model: 
[...]

PR #3441, opened 2022-06-07
fix (android): for #3440 - Missing Proguard Entry for org.apache.commons.lang3.**

ngandhy:
Missing proguard entry in Detox for lang3.stringutils break method invocation. As multiple lang3 classes are are in detox, I used **

## Description

- This pull request addresses the issue described here: #3440

In this pull request, I have added the missing proguard entry to get Detox working again.

---

 - [N/A] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files.

> _For API changes:_
 - [N/A] I have made the necessary changes in the [types index](https://github.com/wix/Detox/blob/master/detox/index.d.ts) file.

### DIFF
--- a/detox/android/detox/proguard-rules-app.pro
+++ b/detox/android/detox/proguard-rules-app.pro
@@ -13,3 +13,4 @@
 -keep class kotlin.text.** { *; }
 -keep class kotlin.io.** { *; }
 -keep class okhttp3.** { *; }
+-keep class org.apache.commons.lang3.** { *; }


noomorph:
@jonathanmos can you take it?

d4vidi (approved):

d4vidi:
I got this, thanks

d4vidi (changes requested):
@ngandhy Technically, this should reside under `detox/android/detox/proguard-rules.pro`, because *Detox* uses `org.apache.**` and defines that as a dependency directly.

Could you try to check whether that solves your issue as well?

Also, while you're at it, we would appreciate adding `org.apache.commons.io.**` to the list.

ngandhy:
@d4vidi - Sorry - I'm not following on the first part. I believe I did make the change to: detox/android/detox/proguard-rules-app.pro. Currently my app won't build without the addition of -keep class org.apache.commons.lang3.** { *; }, so it doesn't seem to be automatically added if that's what you mean?

In the meantime I'll get org.apache.commons.io.** added and do a few test builds :)

d4vidi:
You added that to `proguard-rules-app.pro`, I was asking whether you could switch it to `proguard-rules.pro`. Subtle, I know, but there is a difference 😄

ngandhy:
Got it, not sure how I missed that! I've made the change locally and will update once I've confirmed everything is building correctly

d4vidi:
Not stale. Planning on reiterating that.

d4vidi (approved):

Issue #3196, opened 2022-01-30
[Android] Support multiple elements in getAttributes()


d4vidi:
**Is your feature request related to a problem? Please describe.**

As a follow-up action on #3180, we need to not only apply the type fixes for Android (i.e. for `getAttributes()`), but in fact introduce the functionality altogether.

**Describe the solution you'd like**
`getAttributes()` should allow for returning an array of the attributes of all associated elements, given an ambiguous matcher.

jaheikki:
I have tested that the following works for iOS but not for Android: "const multipleMatchedElements = await element(by.text('Multiple')).getAttributes();" (=clip from Detox documentation https://wix.github.io/Detox/docs/api/actions-on-element#getattributes. And Btw the documentation does not say this is iOS only...). 

I have detox version 19.5.7. 

Problem with Android is: "matches multiple views in the hierarchy".

So will #3196 solve this problem I have?

d4vidi:
@jaheikki That sounds exactly right. Keep track of this for resolution.

pradeipp:
Hi @d4vidi, what's the priority of this item? or any eta on this.. would be nice to see it implemented sooner rather than later

d4vidi:
It's very hard to provide ETA's on things that are not in our center-focus. Nevertheless, I estimate that within the upcoming weeks.

mkumail1:
This is a very important issue for some people, kindly look forward to this or at-least update the documentation accordingly since this only works for IOS not **Android**. I have wasted an entire day trying to fix this.

jaheikki:
Hoping this issue will be solved soon, I've been waiting for this since April. I have several Detox tests that are currently "iOS only" because of this issue. Thank you for your co-operation!

micahdasMA:
Has there been any movement on this issue? I am hitting an issue on iOS where it gives the error
```
Property 'elements' does not exist on type 'IosElementAttributes | AndroidElementAttributes | { elements: IosElementAttributes[]; }'.
  Property 'elements' does not exist on type 'IosElementAttributes'.ts(2339)
```  

when following the exact example laid out in the [documentation](https://wix.github.io/Detox/docs/api/actions/#getattributes)

d4vidi:
The issue is back on our table 🙏🏻
@asafkorem we will soon align typescript between platform (as much as possible) alongside the technical fix.

asafkorem:
[...]

PR #3458, opened 2022-06-27
Add Android support for getAttributes with multiple elements

jonathanmos:
## Description
Support retrieving attributes for multiple elements on Android. Fixes issue #3196

### DIFF
--- a/detox/src/android/interactions/native.js
+++ b/detox/src/android/interactions/native.js
@@ -1,8 +1,14 @@
+const _ = require('lodash');
+
 const DetoxRuntimeError = require('../../errors/DetoxRuntimeError');
 const { ScrollAmountStopAtEdgeAction } = require('../actions/native');
 const { NativeMatcher } = require('../core/NativeMatcher');
 const DetoxAssertionApi = require('../espressoapi/DetoxAssertion');
 const EspressoDetoxApi = require('../espressoapi/EspressoDetox');
+
+const SUPPORTED_MULTIVIEW_ACTIONS = new Set([
+  'getAttributes'
+]);
 
 function call(maybeAFunction) {
   return maybeAFunction instanceof Function ? maybeAFunction() : maybeAFunction;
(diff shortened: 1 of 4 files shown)


detox/src/android/interactions/native.js
```diff
@@ -1,20 +1,57 @@
...
 class Interaction {
   constructor(invocationManager) {
     this._call = undefined;
+    this._element = undefined;
+    this._action = undefined;
     this._invocationManager = invocationManager;
   }
 
   async execute() {
+    const actionMethod = _.get(this._action, '_call.value.method', '');
```
noomorph, line 26: I don't like this code, it looks like something anti-pattern'ish.

Could you try another approach:

```js
class MultiElementActionInteraction extends Interaction {
  constructor(invocationManager, element, action) {
    super(invocationManager);
    this._call = EspressoDetoxApi.perform(call(element._call), action._call);
    this._element = element;
    this._action = action;
  }

  async execute() {
    const resultObjects = [];
    let index = 0;
    let err;
    while (!err) {
      try {
        const invocationResult = await this._invocationManager.execute(
          EspressoDetoxApi.perform(call(_.cloneDeep(this._element).atIndex(index)._call), this._action._call)
        );

        resultObjects.push(JSON.parse(invocationResult.result));

        index++;
      } catch (e) {
        err = e;
      }
    }

    if (index === 0) {
      throw err;
    }

    return resultObjects.length === 1 ?
      resultObjects[0] :
      resultObjects;
  }
}
```


and inside `NativeElement.js` do it like this?

```js
  async getAttributes() {
    const result = await new MultiElementActionInteraction(this._invocationManager, this, new actions.GetAttributes()).execute();
    return JSON.parse(result);
  }
```

detox/test/e2e/33.attributes.test.js
```diff
@@ -221,14 +222,33 @@ describe('Attributes', () => {
...
+
+        const result = await
           element(
             by.type('com.facebook.react.views.view.ReactViewGroup')
               .withAncestor(by.id('attrScrollView'))
-          ).getAttributes()
-        ).rejects.toThrowError(/Problem views are marked with '.{4}MATCHES.{4}' below/m);
+          ).getAttributes();
+
+        const innerViews = result.filter(a => a.identifier);
```
noomorph, line 248: This signature is very problematic. I suggest to align iOS and Android results and also update TypeScript typings accordingly.

In iOS it is either Attributes or { elements: Attributes[] }.
In Android, as I see it, it is either Attributes or Attributes[].

It can't enter `master` unresolved.

noomorph (changes requested):

PR #3497, opened 2022-07-20
(WIP) feat(Android): optimize app installation.

### LINKED ISSUE #2534: Improve app install time by reinstalling on-device APK

jonathanmos:
## Description

Additional fixes on top of the [original pr](https://github.com/wix/Detox/pull/3195)

Resolves https://github.com/wix/Detox/issues/3175
Resolves https://github.com/wix/Detox/issues/2537

**Implemented only for Android.**

1. Improve APK install time by:
    * taking a hash of the binary that is being installed and writing it to the emulator as file whose filename is the binary name.
    * on the next install, checks whether the hash exists and if it does then clear user data without performing uninstall/install.
2. Adds a new method of `resetAppState` to decide between clearing user data and reinstallation.
3. Adds a boolean option to Detox `behavior` config, `optimizeAppInstall` with a corresponding CLI flag: `--optimize-app-install`.

In terms of performance: uninstall/install takes 6-7s, whereas clearing app state takes 400-600ms.

This pull request fixes #2534 .

> _For API changes:_

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -100,6 +100,14 @@ declare global {
                  */
                 keepLockFile?: boolean;
             };
+            /**
+             * Avoid installing the app on the device if it has been already installed there.
+             * Instead of reinstalling, it calls Device#resetAppState().
+             * By default, Detox will perform app reinstallation.
+             *
+             * Experimental and available for Android only.
+             */
+            optimizeAppInstall?: boolean;
             launchApp?: 'auto' | 'manual';
             cleanup?: {
                 shutdownDevice?: boolean;
@@ -753,6 +761,12 @@ declare global {
              * @example await device.installApp('other.bundle.id');
              */
             uninstallApp(bundle?: string): Promise<void>;
+
+            /**
+             * Reinstall the application if the APK has changed, or reset the user data otherwise (without reinstalling).
+             * Available for Android only.
+             */
+            resetAppState(): Promise<void>;
 
             /**
              * Mock opening the app from URL. sourceApp is an optional parameter to specify source application bundle id.
@@ -1612,7 +1626,7 @@ declare global {
              */
             userActivity?: any;
             /**
-             * Launch into a fresh installation
+             * Launch into a fresh installation.
              * A flag that enables relaunching into a fresh installation of the app (it will uninstall and install the binary again), default is false.
              */
             delete?: boolean;
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -101,13 +101,10 @@ declare global {
                 keepLockFile?: boolean;
             };
             /**
-             * Avoid installing the app on the device if it has been already installed there.
-             * Instead of reinstalling, it calls Device#resetAppState().
-             * By default, Detox will perform app reinstallation.
-             *
-             * Experimental and available for Android only.
+             * Force using the legacy flow for launchApp - uninstalling and reinstalling the app - instead of
+             * the optimized flow of deleting app data. The optimized flow exists at the moment only on Android.
              */
-            optimizeAppInstall?: boolean;
+            useLegacyLaunchApp?: boolean;
             launchApp?: 'auto' | 'manual';
             cleanup?: {
                 shutdownDevice?: boolean;
--- a/detox/src/DetoxWorker.js
+++ b/detox/src/DetoxWorker.js
@@ -7,6 +7,7 @@ const { DetoxRuntimeErrorComposer } = require('./errors');
 const { InvocationManager } = require('./invoke');
 const symbols = require('./symbols');
 const AsyncEmitter = require('./utils/AsyncEmitter');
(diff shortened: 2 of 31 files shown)


RohovDmytro:
Any chance to see it landed in v19?

d4vidi:
@RohovDmytro Detox 20 will definitely be released, beforehand.

detox/src/devices/runtime/RuntimeDevice.js
```diff
@@ -119,8 +120,13 @@ class RuntimeDevice {
 
     if (params.delete) {
       await this.terminateApp(bundleId);
-      await this.uninstallApp();
-      await this.installApp();
+
+      if (this._behaviorConfig.optimizeAppInstall) {
```
noomorph, line 124: @jonathanmos @asafkorem @d4vidi 

Do we want `delete` to behave ambiguously?

jonathanmos: This is a good point, and I would say that the user's intent when launching with 'delete: true' is to launch the app with a clean state, so I don't necessarily see resetting the cache as conflicting with that.

detox/index.d.ts
```diff
@@ -100,6 +100,11 @@ declare global {
                  */
                 keepLockFile?: boolean;
             };
+            /**
+             * Force using the legacy flow for launchApp - uninstalling and reinstalling the app - instead of
+             * the optimized flow of deleting app data. The optimized flow exists at the moment only on Android.
+             */
+            useLegacyLaunchApp?: boolean;
```
asafkorem, line 107: That's a better naming IMO.

detox/index.d.ts
```diff
@@ -754,6 +759,12 @@ declare global {
              */
             uninstallApp(bundle?: string): Promise<void>;
 
+            /**
+             * Reinstall the application if the APK has changed, or reset the user data otherwise (without reinstalling).
+             * Available for Android only.
+             */
+            resetAppState(): Promise<void>;
```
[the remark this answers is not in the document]
noomorph: Actually, it makes more sense to `Reinstall the application if the APK has changed` inside the `installApp` logic, @jonathanmos . I strongly advise to consider that.

detox/src/DetoxWorker.js
```diff
@@ -260,14 +261,24 @@ class DetoxWorker {
...
-      yield this.device.selectApp(appName);
-      yield this.device.uninstallApp();
-    }
+    const platform = this._config.device.type.split('.')[0];
+
+    if (shouldUseOptimizedInstall(platform, this._behaviorConfig.useLegacyLaunchApp)) {
+      for (const appName of appNames) {
+        yield this.device.selectApp(appName);
+        // @ts-ignore
+        yield this.device.resetAppState();
```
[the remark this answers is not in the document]
noomorph: I see the solution in a different way:

```js
const shouldUseOptimizedInstall = /* ... */;

for (const appName of appNames) {
  yield this.device.selectApp(appName);

  if (shouldUseOptimizedInstall) {
    yield this.device.resetAppState();
  } else {
    yield this.device.uninstallApp();
  }
}

for (const appName of appNames) {
  yield this.device.selectApp(appName);
  yield this.device.installApp(); // your logic should be smart enough to see when you can skip installing app if it is the same
}
```

noomorph (changes requested):
I think that overall the accent is put in a wrong place: `resetAppState` should be a rather dumb method, while `installApp` should be more economic and be able to skip installation if the hashes match.

**Last, but not least:** I'd add information about **lastUpdateTime**, not only the **hashsum**:

```
adb shell dumpsys package <app.bundle.id> | grep lastUpdateTime=
```

That way you can be sure there's no chance somebody would install the app outside of Detox – installing anything different would affect output of the command above, so if it does not match while the hash matches – you need to reinstall the app anyway.

I think that this extra caution can spare us from adding experimental modes, etc. – it would be safe enough to start using just after release.

noomorph:
@d4vidi @jonathanmos @asafkorem

Yesterday I suggested approaching this issue from a different side and formulated my vision. Does it sound logical enough?

In light of @jonathanmos' leave, what action items on this PR do we have? Will @jonathanmos proceed with it, or will it go to our team's backlog?

d4vidi:
Back to the backlog, for now. But you can definitely share you approach!

noomorph:
TODO: add this to Detox 22 epic.

PR #3513, opened 2022-07-28
ci: Introduce an additional job for RN 66 compat on Android

d4vidi:
## Description

- This pull request is associated with all post-factum work done in the context of #3492.

In this pull request, I have added a new (Android) CI job in order to regularly verify compatibility with older RN versions, of the `demo` project. An equivalent job already exists for the _main_ `test` app, although - #3492, which had to be reopened due to issues in production, has proven that this isn't enough.
There's a core difference between the `demo` and `test` projects: While the `test` app compiles Detox itself, bundled into the test .apk, the `demo` project precompiles Detox as an `.aar` and uses it. With that, the `demo` is (intentionally) much more similar to real-world users (in and out of Wix), and that can sometimes introduce specific edge-cases.

---

> _For API changes:_

### DIFF
--- a/.buildkite/jobs/pipeline.android_demo_app_rn_66.yml
+++ b/.buildkite/jobs/pipeline.android_demo_app_rn_66.yml
@@ -1,0 +1,12 @@
+  - label: ":android::mushroom:-RN_66_demo_project"
+    command:
+      - "nvm install"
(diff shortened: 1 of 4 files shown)


d4vidi:
@igorgn please approve

igorgn (approved):

Issue #3548, opened 2022-08-10
Missing configuration for react-native-gradle-plugin in 0.68


jonathanmos:
### Describe your idea

RN upgrade helper recommends adding this to settings.gradle. Not having it appears to cause an issue when dependencies migrate to the new architecture:

```
includeBuild('../node_modules/react-native-gradle-plugin')
```

causing errors that look like `could not find react-native-gradle-plugin` when building the project.

[Link to helper](https://react-native-community.github.io/upgrade-helper/?from=0.66.4&to=0.68.2)

To reproduce include @react-native-community/slider dependency above 4.3.0

PR #3549, opened 2022-08-10
fix(Android): Add react-native-gradle-plugin to settings.gradle

jonathanmos:
## Description
There is some missing configuration in settings.gradle according to the RN upgrade helper, and this leads to an error about `missing react-native-gradle-plugin` when dependencies upgrade to the new architecture. This PR adds the missing line.

Fixes issue #3548

### DIFF
--- a/detox/test/android/settings.gradle
+++ b/detox/test/android/settings.gradle
@@ -1,6 +1,8 @@
 rootProject.name = 'DetoxTest'
 include ':app'
 
+includeBuild('../node_modules/react-native-gradle-plugin')
+
 include ':detox'
 project(':detox').projectDir = new File(rootProject.projectDir, '../../android/detox')
 


d4vidi:
Wait, didn't we say we are going to lock the slider module?

d4vidi:
IDK, this specific patch sounds related more to supporting the new-architecture rather than just RN 68. The upgrade helper include both things, intermingled, and it's hard to say just based on that.

d4vidi (changes requested):
See previous questions

jonathanmos:
yes, I've already locked the slider module. However, without this change it will happen again in the future with some other dependency.

d4vidi:
I see what you mean. But doesn't that mean we can _unlock_ the slider module?

jonathanmos:
If we want to, but I'm not sure that leaving it unlocked wasn't a mistake to begin with

d4vidi:
OK, your call.

d4vidi (approved):

PR #3587, opened 2022-09-07
fix(DetoxCoreListener.js): remove duplicated `status` value.

asafkorem:
This change has no affect on functionality, as well as this field value, we override this field with the real test status anyway.


noomorph:
sure!

asafkorem:
ahhh damn 🤦🏼‍♂️  apparently this change affected the order of the fields on the snapshot so I have to update it the intg. tests with the new snapshot..

PR #3589, opened 2022-09-07
feat(trace): add support for events recorder delegate.

asafkorem:


asafkorem:
As discussed, no need for this recorder delegate, we can extract the events from `trace` directly.

PR #3594, opened 2022-09-15
Update NetworkFormatter.js

mikasius:
In logs network address we have this: 
```
• 1 network requests with URLs:
  - URL #1: https://google.com/api.
```

And this `.` at the end is very confusing when try to click on this link.
Suggest to remove this `.`


mikasius:
There are a snapshot errors.
Snapshots need to be updated.

noomorph (approved):

PR #3608, opened 2022-09-23
fix: the occasional ConcurrentModificationException error

zhouzh1:
## Description

When I executed my E2E testing on Android emulator, I found that sometimes there was a error of `ConcurrentModificationException`, it's occasional, doesn't 
appear every time, I reviewed the relevant code based on the error stack, found that the root cause is that the `ArrayList` and `HashSet` are used here, after I replaced the `ArrayList` with `CopyOnWriteArrayList` and `HashSet` with `CopyOnWriteArraySet`, and rebuilt the Detox, changed to reference the rebuilt package in my `build.gradle` file, then the issue went away and everything goes well.

Sorry for forgetting to take a screenshot of this error, but it indeed exists.
image

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java
@@ -8,7 +8,6 @@ import com.facebook.react.bridge.ReactContext;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


asafkorem (changes requested):
Thanks @zhouzh1 for investigating this error and your contribution! 🙌🏼 

I wonder if we better synchronize the set operations instead, if this isn't a common exception (concurrent modification).

Also, from a brief look, it makes sense to me that you are getting this error, my guess is that it happen when we call `isIdleNow()` on the [doFrame()](https://github.com/wix/Detox/blob/084c3805bb0e4d68718909c6c338aee134bff47f/detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/NetworkIdlingResource.java#L117) callback from different Looper threads.

zhouzh1:
@asafkorem Thanks for your review, maybe there is a better fix solution to this issue, because I don't have much experience in Java development. :)

d4vidi:
@zhouzh1 thank you, contributions are much appreciated!
@asafkorem you are likely correct - the thread management in that area is not trivial and it is likely that there's more than one thread involved. It is likely that the concurrent modification takes place while Detox's debug-sync mechanism is operational.

Typically, `ConcurrentHashMap` can be used in order to accommodate for such races. However, since is this case a certain type of transaction-like logic is performed, it would likely be a better idea to use `synchronized`. I will post an equivalent PR soon.

d4vidi:
Closing in favor of #3616 (credits apply)

zhouzh1:
Looking forward to the fix version so that we don't need to patch and rebuild. :)

asafkorem:
Detox [v19.12.3](https://github.com/wix/Detox/releases/tag/19.12.3) was released with this fix.
Thanks @zhouzh1 and @d4vidi 🙌🏼

PR #3609, opened 2022-09-26
fix(iOS 16): avoid SpringBoard crash trying to foreground the app.

asafkorem:
This change fixes the [`sendToHome`](https://wix.github.io/Detox/docs/api/device-object-api#devicesendtohome) method for iOS 16.0 and above, which were introduced along with Xcode 14, as addressed in https://github.com/wix/Detox/issues/3531.

The original bug was reproduced locally with Xcode 14 and Xcode 14.1 beta 2 (I checked if anything has changed between those versions) along with iOS 16 and iOS 16.1. Also, the fix was tested on the same Xcode versions and iOS Simulators.

SpringBoard is not directly accessible by Simctl on iOS 16.0 and above, therefore we launch and terminate the Settings app instead. This sends the currently open app to the background and brings the home screen to the foreground.

In spite of the fact that this isn't an ideal solution, I haven't found another direct method of doing it, so this is probably the only practical option.

### DIFF
--- a/detox/src/devices/common/drivers/ios/tools/AppleSimUtils.js
+++ b/detox/src/devices/common/drivers/ios/tools/AppleSimUtils.js
@@ -132,6 +132,30 @@ class AppleSimUtils {
   }
 
   async sendToHome(udid) {
+    if (await this._isSpringBoardInaccessible(udid)) {
+      // SpringBoard is not directly accessible by Simctl on iOS 16.0 and above, therefore we launch and terminate the
+      // Settings app instead. This sends the currently open app to the background and brings the home screen to the
+      // foreground.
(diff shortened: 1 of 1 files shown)


umutcanbostancii:
@asafkorem 
Hello sir. I'm not a pro ios developer. I'm only junior and trying learn the swift language for 6 months. Anyway, I upgraded my xcode the last version and now I'm seeing "  the operation couldn't be completed. xpc error " and simulator is down. I don't know this is the same problem that you wrote here but I just want to help you with this bad error. Thanks



Ekran Resmi 2022-11-09 22 14 07
Ekran Resmi 2022-11-09 22 14 02

asafkorem:
@umutcanbostancii what is your Detox version?

Issue #3507, opened 2022-07-27
[RN postmortem] Add a bloated JS bundle stress test.


asafkorem:
Pull request: https://github.com/wix/Detox/pull/3619

### Describe your idea

Add a (stress) test to Detox’s suite that runs the test app based on a bloated JS bundle that takes a long time to load; Release and debug.

This came up after facing with some back-compat problems detected only from our demo module (on _DWP_) using genycloud.

PR #3504, opened 2022-07-26
fix(Android): Make RN instance listener work in RN≤68

d4vidi:
## Description

- This pull request addresses the issue described here: #XXX

In this pull request, I have fixed a crash in registering an RN instance loading listener, in RN≤68 (following the release of the Detox `.aar` based on .68). See this [diff](https://github.com/facebook/react-native/compare/v0.67.4..v0.68.0#diff-2f01f0cd7ff8c9ea58f12ef0eff5fa8250cad144dd5490598d80d8e9e743458aR1009) --

`ReactInstanceManager.java`
image

The root of the problem is that at runtime, this error is thrown:

```
@Thread detox.primary(242):
java.lang.NoClassDefFoundError: Failed resolution of: Lcom\/facebook\/react\/ReactInstanceEventListener;
	at com.wix.detox.reactnative.ReactNativeLoadingMonitor$subscribeToNewRNContextUpdates$1.run(ReactNativeLoadingMonitor.kt:38)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:458)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at android.app.Instrumentation$SyncRunnable.run(Instrumentation.java:2163)
	at android.os.Handler.handleCallback(Handler.java:873)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loop(Looper.java:193)
	at android.app.ActivityThread.main(ActivityThread.java:6669)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

[...]

PR #3619, opened 2022-10-02
ci(demo-react-native): add tests for bloated JS bundle.

asafkorem:
Resolves https://github.com/wix/Detox/issues/3507


scripts/demo-rn-bloat-bundle-test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+
+echo "Bloating the demo project bundle"
+run_f "npm run bloat-bundle"
+
+run_f "npm run build:ios-release"
```
jonathanmos, line 6: It looks like it's specifically run on ios - what about android? Since the script is being triggered from android/ios scripts they could pass in the platform as an argument

scripts/demo-rn-bloat-bundle-test.sh
```diff
@@ -0,0 +1,10 @@
...
+run_f "npm run test:ios-release-ci"
+
+echo "Unbloating the demo project bundle"
+run_f "git restore ."
```
jonathanmos, line 10: this could be a problematic way to remove the bloat because the scope is not limited. Perhaps we can write the bloat comments with a unique prefix and then at the end reedit the app.js to remove those lines specifically, or alternatively, make a copy of the app.js and overwrite the bloated bundle at the end with the backup copy?

examples/demo-react-native/scripts/bloatBundle.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs');
+
+function bloatBundle() {
+  const bundlePath = './app.js';
+
+  try {
+    const debugBundle = fs.openSync(bundlePath, 'a');
+    try {
+      /// Bloating the bundle with a lot of data.
+      for (let i = 0; i < 25000; i++) {
```
jonathanmos, line 10: Some suggestions: 

* Extract to constants the number of lines to write and the bloat message
* Enable input args to allow overriding from outside vars here - number of lines to write, bundlepath

examples/demo-react-native/scripts/bloatBundle.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs');
+
```
jonathanmos, line 2: Perhaps we should add a comment to this file with a link to [the gh issue](https://github.com/wix/Detox/issues/3507) to provide future context of why we have a bloat test

scripts/demo-projects.android.sh
```diff
@@ -20,8 +20,7 @@ popd
 # as it runs tests in parallel. Also, it installs test-butler on them -
 # even for forward usage.
 pushd examples/demo-react-native-jest
-run_f "npm run test:android-release-ci"
-DETOX_EXPOSE_GLOBALS=0 run_f "npm run test:android-release-ci"
```
jonathanmos, line 24: What is DETOX_EXPOSE_GLOBALS and why is it removed here?

scripts/demo-projects.android.sh
```diff
@@ -31,22 +30,24 @@ if [ "$REACT_NATIVE_COMPAT_TEST" = "true" ]; then
 fi
 
 pushd examples/demo-react-native
-run_f "npm run test:android-release-ci"
-DETOX_EXPOSE_GLOBALS=0 run_f "npm run test:android-release-ci"
+  run_f "npm run test:android-release-ci"
+
+  # Run tests with bloated JS bundle:
+  source $(dirname "$0")/demo-rn-bloat-bundle-test.sh
```
jonathanmos, line 36: $(dirname "$0") occurs a couple of times in this file - should probably be assigned to a descriptive var name to make it clearer

scripts/demo-projects.ios.sh
```diff
@@ -7,17 +7,21 @@ source $(dirname "$0")/demo-projects.sh
...
-run_f "npm run test:ios-release-ci"
+  run_f "npm run test:ios-release-ci"
 popd
 
 pushd examples/demo-react-native
-run_f "npm run test:ios-release-ci"
+  run_f "npm run test:ios-release-ci"
+
+  # Run tests with bloated JS bundle:
+  source $(dirname "$0")/demo-rn-bloat-bundle-test.sh
```
jonathanmos, line 25: Same comment as for android regarding dirname $0

jonathanmos:
I suggest adding a link [to the issue](https://github.com/wix/Detox/issues/3507) to the pr description

scripts/demo-projects.android.sh
asafkorem: Good question, initially I thought we are running the same tests twice without any good reason. However, while revisiting this, looks like the original intention was to test this API: https://wix.github.io/Detox/docs/next/config/behavior/#behaviorinitexposeglobals-boolean

asafkorem: I'll revert this and add a comment, great catch.

scripts/demo-rn-bloat-bundle-test.sh
asafkorem: correct! great catch

examples/demo-react-native/scripts/bloatBundle.js
asafkorem: Done.

asafkorem:
Done @jonathanmos

.buildkite/jobs/pipeline.ios_demo_app_rn_66.yml
```diff
@@ -0,0 +1,9 @@
+  - label: ":ios::react: RN.66 + iOS: Demo app"
+    command:
+      - "nvm install"
+      - "JAVA_HOME=/usr/local/opt/openjdk@11/ ./scripts/demo-projects.ios.sh"
```
noomorph, line 4: Just wondering, why do we need that `JAVA_HOME` override for iOS?

examples/demo-react-native/scripts/bloatBundle.mjs
```diff
@@ -0,0 +1,21 @@
+import { appendFile } from 'fs/promises';
+const bundlePath = `${process.argv[2]}`;
+
+async function bloatBundle() {
+  const garbageToAppend = `/* ${ 'junk '.repeat(50000) } */`
```
noomorph, line 5: Is it effective enough? Maybe we could generate some scalable `if () {} else {}` labyrinth to make sure that the AST parser chokes?

noomorph (approved):
Looks good! 👍  See a couple of comments, though. Maybe they make sense.

.buildkite/jobs/pipeline.ios_demo_app_rn_66.yml
asafkorem: You're right, we don't. Missed it in the copy-paste from the android script 🤭

examples/demo-react-native/scripts/bloatBundle.mjs
asafkorem: Yes it is, it reproduced the error we had before [Amit's fix](https://github.com/wix/Detox/pull/3504) for the RN .68 upgrade. 
Although, adding more coverage by choking the AST won't hurt this test.

.buildkite/jobs/pipeline.ios_demo_app_rn_66.yml
asafkorem, line 4:
```suggestion
      - "./scripts/demo-projects.ios.sh"
```

.buildkite/jobs/pipeline.ios_rn_66.yml
```diff
@@ -1,4 +1,4 @@
-  - label: ":ios:-:six::six: RN_66_iOS"
+  - label: ":ios::detox: RN.66 + iOS: Tests app"
     command:
       - "nvm install"
       - "JAVA_HOME=/usr/local/opt/openjdk@11/ ./scripts/ci.ios.sh"
```
asafkorem, line 4:
```suggestion
      - "./scripts/ci.ios.sh"
```

.buildkite/jobs/pipeline.ios_rn_69.yml
```diff
@@ -1,4 +1,4 @@
-  - label: ":ios:-:six::nine: RN_69_iOS"
+  - label: ":ios::detox: RN.69 + iOS: Tests app"
     command:
       - "nvm install"
       - "JAVA_HOME=/usr/local/opt/openjdk@11/ ./scripts/ci.ios.sh"
```
asafkorem, line 4:
```suggestion
      - "./scripts/ci.ios.sh"
```

jonathanmos (approved):

examples/demo-react-native/scripts/bloatBundle.mjs
asafkorem: mmm, @noomorph maybe you have an idea what kind of labyrinth would choke the AST parser? Also, how would you test that the AST is choked?

asafkorem:
That's weird. 
According to my local machine, the snapshots tests are up-to-date, however our CI keeps failing on them (b/c of some missing line break). However, it doesn't seems to be related to this pull request because the only difference between the snapshots is some line-break.

@noomorph last sync you told something about Jest upgrade that might have cause that. Are there any steps required (locally) so the local snapshot will match the CI's snapshot?

PR #3661, opened 2022-11-02
website(showcase): make container left padding same as the right.

asafkorem:
This change affects small screens, where the showcase cards weren't centered in their container.


asafkorem:
FYI @jasonbariwix 🙂

PR #3680, opened 2022-11-10
test: add tests for IPCClient & IPCServer integration.

asafkorem:
This PR adds integration tests for `IPCServer` and `IPCClient`, along with unit tests for the `SessionState`, which is used by them.

@noomorph

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -163,7 +163,6 @@
       "runners/jest/testEnvironment",
       "src/DetoxWorker.js",
       "src/logger/utils/streamUtils.js",
-      "src/ipc",
       "src/realms"
     ],
     "resetMocks": true,
--- a/detox/src/ipc/IPCClient.js
+++ b/detox/src/ipc/IPCClient.js
@@ -34,7 +34,15 @@ class IPCClient {
     this._serverConnection = null;
 
     if (this._ipc) {
-      this._ipc.disconnect(this.serverId);
+      await new Promise((resolve, reject) => {
+        this._ipc.of[this.serverId]
+          // @ts-ignore
+          .once('disconnect', resolve)
+          .once('error', reject);
+
+        this._ipc.disconnect(this.serverId);
+      });
+
       this._ipc = null;
     }
   }
@@ -68,9 +76,9 @@ class IPCClient {
     this._serverConnection = await new Promise((resolve, reject) => {
       this._ipc.connectTo(serverId, (client) => {
         client.of[serverId]
-          .on('error', reject)
-          .on('disconnect', () => reject(new DetoxInternalError('IPC server has unexpectedly disconnected.')))
-          .on('connect', () => resolve(client.of[serverId]));
+          .once('error', reject)
+          .once('disconnect', () => reject(new DetoxInternalError('IPC server has unexpectedly disconnected.')))
+          .once('connect', () => resolve(client.of[serverId]));
       });
     });
 
@@ -90,6 +98,7 @@ class IPCClient {
     return new Promise((resolve, reject) => {
       const server = this._serverConnection;
 
+      /* istanbul ignore next */
       function onError(err) {
         server.off('error', onError);
         server.off(`${event}Done`, onDone);
@@ -110,7 +119,7 @@ class IPCClient {
   }
 
   _onSessionStateUpdate = (payload) => {
-    this._state.patch(payload);
+    this._sessionState.patch(payload);
   };
 }
 
--- a/detox/src/ipc/IPCClient.js
+++ b/detox/src/ipc/IPCClient.js
@@ -109,8 +109,8 @@ class IPCClient {
     });
   }
 
-  _onSessionStateUpdate = (payload) => {
-    this._state.patch(payload);
+  _onSessionStateUpdate = ({ testResults, workersCount }) => {
+    this._state.patch({ testResults, workersCount });
   };
 }
 
--- a/detox/src/ipc/IPCClient.js
+++ b/detox/src/ipc/IPCClient.js
@@ -11,14 +11,14 @@ class IPCClient {
     /** @type {import('./SessionState')} */
     this._state = state;
 
-    this._client = null;
+    this._ipc = null;
     this._serverConnection = null;
   }
 
   async init() {
-    this._client = new IPC();
+    this._ipc = new IPC();
 
-    Object.assign(this._client.config, {
+    Object.assign(this._ipc.config, {
       id: this._id,
       appspace: 'detox.',
       logger: (msg) => this._logger.trace(msg),
@@ -33,9 +33,9 @@ class IPCClient {
   async dispose() {
     this._serverConnection = null;
 
-    if (this._client) {
-      this._client.disconnect(this.serverId);
-      this._client = null;
+    if (this._ipc) {
+      this._ipc.disconnect(this.serverId);
+      this._ipc = null;
     }
   }
 
@@ -66,7 +66,7 @@ class IPCClient {
     const serverId = this.serverId;
 
     this._serverConnection = await new Promise((resolve, reject) => {
-      this._client.connectTo(serverId, (client) => {
+      this._ipc.connectTo(serverId, (client) => {
         client.of[serverId]
           .on('error', reject)
           .on('disconnect', () => reject(new DetoxInternalError('IPC server has unexpectedly disconnected.')))
--- a/detox/src/ipc/IPCClient.js
+++ b/detox/src/ipc/IPCClient.js
@@ -109,8 +109,8 @@ class IPCClient {
     });
   }
 
-  _onSessionStateUpdate = ({ testResults, workersCount }) => {
-    this._sessionState.patch({ testResults, workersCount });
+  _onSessionStateUpdate = (payload) => {
+    this._state.patch(payload);
   };
 }
 
--- a/detox/src/ipc/IPCClient.js
+++ b/detox/src/ipc/IPCClient.js
@@ -4,12 +4,12 @@ const { DetoxInternalError } = require('../errors');
 const { serializeObjectWithError } = require('../utils/errorUtils');
 
 class IPCClient {
-  constructor({ id, logger, state }) {
+  constructor({ id, logger, sessionState }) {
     this._id = id;
     /** @type {import('../logger/DetoxLogger')} logger */
     this._logger = logger.child({ cat: 'ipc' });
     /** @type {import('./SessionState')} */
-    this._state = state;
+    this._sessionState = sessionState;
 
     this._ipc = null;
     this._serverConnection = null;
@@ -40,7 +40,7 @@ class IPCClient {
   }
 
   get sessionState() {
-    return this._state;
+    return this._sessionState;
   }
 
   get serverId() {
@@ -49,7 +49,7 @@ class IPCClient {
 
   async registerWorker(workerId) {
     const sessionState = await this._emit('registerWorker', { workerId });
-    this._state.patch(sessionState);
+    this._sessionState.patch(sessionState);
   }
 
   /**
@@ -59,7 +59,7 @@ class IPCClient {
     const sessionState = await this._emit('reportTestResults', {
       testResults: testResults.map(r => serializeObjectWithError(r, 'testExecError')),
     });
-    this._state.patch(sessionState);
+    this._sessionState.patch(sessionState);
   }
 
   async _connectToServer() {
@@ -79,7 +79,7 @@ class IPCClient {
 
   async _registerContext() {
     const sessionState = await this._emit('registerContext', { id: this._id });
-    this._state.patch(sessionState);
+    this._sessionState.patch(sessionState);
   }
 
   async _emit(event, payload) {
@@ -110,7 +110,7 @@ class IPCClient {
   }
 
   _onSessionStateUpdate = ({ testResults, workersCount }) => {
-    this._state.patch({ testResults, workersCount });
+    this._sessionState.patch({ testResults, workersCount });
   };
 }
 
--- a/detox/src/ipc/IPCServer.js
+++ b/detox/src/ipc/IPCServer.js
@@ -14,10 +14,15 @@ class IPCServer {
     this._logger = logger.child({ cat: 'ipc,ipc-server' });
     this._ipc = null;
     this._workers = new Set();
+    this._contexts = new Set();
   }
 
   get id() {
     return this._sessionState.detoxIPCServer;
+  }
+
+  get contexts() {
+    return [...this._contexts];
   }
 
   get sessionState() {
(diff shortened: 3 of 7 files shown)


detox/src/ipc/IPCClient.js
```diff
@@ -59,27 +59,27 @@ class IPCClient {
...
+      this._ipc.connectTo(serverId, (client) => {
         client.of[serverId]
           .on('error', reject)
           .on('disconnect', () => reject(new DetoxInternalError('IPC server has unexpectedly disconnected.')))
           .on('connect', () => resolve(client.of[serverId]));
       });
     });
 
-    this._serverConnection.on('sessionStateUpdate', this._onSessionStateUpdate);
+    this._serverConnection.on('', this._onSessionStateUpdate);
```
noomorph, line 77: Why did you decide to remove the event name?

asafkorem: Sorry, ignore that 😓  that's an error. Already force pushed with the fix

detox/src/ipc/IPCClient.js
```diff
@@ -109,8 +109,8 @@ class IPCClient {
     });
   }
 
-  _onSessionStateUpdate = (payload) => {
-    this._state.patch(payload);
+  _onSessionStateUpdate = ({ testResults, workersCount }) => {
```
noomorph, line 112: Wait, no. It was meant to be generic.
Can we return that back, please?

detox/src/ipc/SessionState.js
```diff
@@ -25,12 +25,16 @@ class SessionState {
     this.workersCount = workersCount;
   }
 
-  patch(state) {
-    Object.assign(this, state);
+  patch({ testResults = null, testSessionIndex = null, workersCount = null }) {
```
noomorph, line 28: Please, let's keep it generic.

detox/src/ipc/IPCClient.js
asafkorem: I can. why do you want that to be generic if we're using only the `testResults` and the `workersCount`?

asafkorem: I think it's better to limit only for the props that you want to be able to change from this place.

noomorph: It is going to turn into a source of errors. The contract is that the IPC server itself knows what properties it should update/broadcast to all the clients. The client should be obedient in that regard.

asafkorem: How the state's props `id`, `detoxConfig` or `detoxIPCServer` are relevant here? I really see no reason to allow their change from here.

asafkorem: > It is going to turn into a source of errors. The contract is that the IPC server itself knows what properties it should update/broadcast to all the clients. The client should be obedient in that regard.

OK, I get that.

noomorph: Yes, but I see no reason to update them from the server. It does not send the entire session object – only the changed parts of it.

detox/src/ipc/SessionState.test.js
```diff
@@ -0,0 +1,69 @@
+const SessionState = require('./SessionState');
+
+describe('SessionState', () => {
+  /** @type {SessionState} */
+  let sessionState;
+
+  beforeEach(() => {
+    sessionState = new SessionState({
```
noomorph, line 8: Just to make sure, is there a function reviving test?

asafkorem: Nope, I'll add it

asafkorem: Thinking about this, do we have any use case where one of the stringified object props is a function?

detox/src/ipc/IPCClient.js
asafkorem: I can understand what you're saying from this perspective, but I still think the API from both the client and the server should be restricted only for what the server can or should update. 

I'll do the revert because I don't think this discussion worth the bikeshedding 😄

detox/src/ipc/SessionState.test.js
noomorph: Yes, unfortunately. Many Bunyan formatting options are passed as functions, e.g. `showDate: (date) => ...` and all prefixers/stringifiers too.

detox/src/ipc/IPCClient.js
asafkorem: done

detox/src/ipc/SessionState.test.js
asafkorem: done

detox/src/ipc/ipc.test.js
```diff
@@ -17,97 +20,164 @@ describe('IPC', () => {
...
-      await ipcClient1.dispose();
-      await ipcClient2.dispose();
+      it('should wait for connected clients to disconnect', async () => {
+        await ipcServer.init();
+        await ipcClient1.init();
 
-      setTimeout(() => {
-        throw new Error('dispose() should have resolved by now');
-      }, 3000);
+        const disposePromise = ipcServer.dispose();
```
noomorph, line 126: I decided to make this test complete.
First, we wait to make sure that there is no magic, and that the IPC server doesn't actually close.
Then we close the client and make sure that the server closes.

detox/src/ipc/ipc.test.js
```diff
@@ -17,97 +20,164 @@ describe('IPC', () => {
...
 
     it('should return the correct IPC server id', () => {
       expect(ipcClient1.serverId).toEqual(sessionState.detoxIPCServer);
     });
 
-    it('should throw if test results are reported before init', async () => {
-      const expected = 'IPC server bar has unexpectedly disconnected';
-      await expect(ipcClient1.reportTestResults([])).rejects.toThrow(expected);
-    });
+    describe('before init', () => {
```
noomorph, line 161: Cosmetic change. I added a describe suite here.

detox/src/ipc/ipc.test.js
```diff
@@ -17,97 +20,164 @@ describe('IPC', () => {
...
+        await expect(ipcClient1.registerWorker('foo')).rejects.toThrow(expected);
+      });
     });
 
-    it('should register context upon init', async () => {
-      await ipcClient1.init();
-
-      let expected = ['foo'];
-      expect(ipcServer.sessionState.contexts).toEqual(expected);
-      expect(ipcClient1.sessionState.contexts).toEqual(expected);
```
noomorph: This was the test that should not have passed. Using the same instance of session state led to a false positive.

detox/src/ipc/ipc.test.js
```diff
@@ -116,69 +186,108 @@ describe('IPC', () => {
...
+              testFilePath: 'file1',
+              success: true,
+            },
+          ];
+
+          await ipcClient1.reportTestResults(testResults);
+          expect(ipcServer.sessionState.testResults).toEqual(testResults);
+          expect(ipcClient1.sessionState.testResults).toEqual(testResults);
+
+          await sleep(10); // broadcasting happens with a delay
```
noomorph, line 202: Adding a "broadcasting" sleep because it happens asynchronously, later in the macro-tasks queue.

detox/src/ipc/ipc.test.js
```diff
@@ -116,69 +186,108 @@ describe('IPC', () => {
...
+          expect(ipcServer.sessionState.testResults).toEqual(expected);
+          expect(ipcClient1.sessionState.testResults).toEqual(expected);
+          expect(ipcClient2.sessionState.testResults).toEqual(expected);
+        });
 
-        expected = [...testResults2, ...testResults1];
-        expect(ipcServer.sessionState.testResults).toEqual(expected);
-        expect(ipcClient1.sessionState.testResults).toEqual(expected);
-        expect(ipcClient2.sessionState.testResults).toEqual(expected);
+        it('should update test results with retried tests', async () => {
```
noomorph, line 231: Added a case when an IPC client sends data about the same test, e.g. when retrying to run a failed test file.

detox/src/ipc/ipc.test.js
```diff
@@ -116,69 +186,108 @@ describe('IPC', () => {
...
-        await ipcClient2.registerWorker('baz');
+          await ipcClient2.registerWorker('bar');
+          await ipcClient2.registerWorker('baz');
 
-        expect(ipcServer.sessionState.workersCount).toEqual(3);
-        expect(ipcClient1.sessionState.workersCount).toEqual(3);
-        expect(ipcClient2.sessionState.workersCount).toEqual(3);
+          expect(ipcServer.sessionState.workersCount).toEqual(3);
+          expect(ipcClient2.sessionState.workersCount).toEqual(3);
+          await sleep(10); // broadcasting might happen with a delay
```
noomorph, line 274: Same broadcasting delay issue. Nothing to do. 🤷‍♂️

detox/package.json
```diff
@@ -163,7 +163,6 @@
       "runners/jest/testEnvironment",
       "src/DetoxWorker.js",
       "src/logger/utils/streamUtils.js",
-      "src/ipc",
```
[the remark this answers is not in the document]
asafkorem: make sense, considering that causing some errors might be too complicated here

detox/src/ipc/IPCClient.js
```diff
@@ -34,7 +34,15 @@ class IPCClient {
     this._serverConnection = null;
 
     if (this._ipc) {
-      this._ipc.disconnect(this.serverId);
+      await new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
asafkorem: LGTM

detox/src/ipc/IPCClient.js
```diff
@@ -68,9 +76,9 @@ class IPCClient {
     this._serverConnection = await new Promise((resolve, reject) => {
       this._ipc.connectTo(serverId, (client) => {
         client.of[serverId]
-          .on('error', reject)
-          .on('disconnect', () => reject(new DetoxInternalError('IPC server has unexpectedly disconnected.')))
-          .on('connect', () => resolve(client.of[serverId]));
+          .once('error', reject)
```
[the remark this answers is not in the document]
asafkorem: :+1: nice

detox/src/ipc/IPCClient.js
```diff
@@ -90,6 +98,7 @@ class IPCClient {
     return new Promise((resolve, reject) => {
       const server = this._serverConnection;
 
+      /* istanbul ignore next */
```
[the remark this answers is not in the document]
asafkorem: :/

detox/src/ipc/IPCClient.js
```diff
@@ -110,7 +119,7 @@ class IPCClient {
   }
 
   _onSessionStateUpdate = (payload) => {
-    this._state.patch(payload);
+    this._sessionState.patch(payload);
```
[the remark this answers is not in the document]
asafkorem: ah, ops. I see the commit history. 
that's because I reverted the change of passing the params explicitly and it was together with the renaming 😅

detox/src/ipc/IPCServer.js
```diff
@@ -14,12 +14,17 @@ class IPCServer {
     this._logger = logger.child({ cat: 'ipc,ipc-server' });
     this._ipc = null;
     this._workers = new Set();
+    this._contexts = new Set();
```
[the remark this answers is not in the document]
asafkorem: nice catch, great!

detox/src/ipc/IPCServer.js
```diff
@@ -45,15 +50,18 @@ class IPCServer {
...
-    return new Promise((resolve, reject) =>{
+    await new Promise((resolve, reject) =>{
       // @ts-ignore
-      this._ipc.server.server.close(e => e ? reject(e) : resolve());
+      this._ipc.server.server.close(e => /* istanbul ignore next */
+        e ? reject(e) : resolve());
       this._ipc.server.stop();
     });
+
+    this._ipc = null;
```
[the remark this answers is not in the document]
asafkorem: I see, looks ok

detox/src/ipc/SessionState.test.js
```diff
@@ -23,9 +22,21 @@ describe('SessionState', () => {
     });
   });
 
+  describe('when created without arguments', () => {
+    it('should have default values', () => {
+      expect(new SessionState({})).toEqual({
```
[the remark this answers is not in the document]
asafkorem: great!

detox/src/ipc/ipc.test.js
```diff
@@ -1,11 +1,14 @@
+jest.setTimeout(1000);
```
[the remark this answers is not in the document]
asafkorem: good point :+1:

detox/src/ipc/ipc.test.js
```diff
@@ -1,11 +1,14 @@
...
 jest.mock('../utils/logger');
 
 const IPCClient = require('./IPCClient');
 const IPCServer = require('./IPCServer');
 const SessionState = require('./SessionState');
 
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```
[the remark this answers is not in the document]
asafkorem: true. good call

detox/src/ipc/ipc.test.js
```diff
@@ -1,11 +1,14 @@
...
+
 describe('IPC', () => {
-  /** @type {SessionState} */
+  /** @type {*} */
```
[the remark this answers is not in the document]
asafkorem: OK :+1:

detox/src/ipc/ipc.test.js
```diff
@@ -17,97 +20,164 @@ describe('IPC', () => {
...
+    const logger = require('../utils/logger');
+
+    sessionState = {
+      id: 'session-1',
+      detoxIPCServer: 'foo',
+    };
+
+    ipcServer = new IPCServer({
+      logger,
+      sessionState: new SessionState(sessionState),
```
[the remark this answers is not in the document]
asafkorem: True.
Also this raises a question - how does a real execution go, are there distinct "SessionState" objects?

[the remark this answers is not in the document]
asafkorem: If not, I think the client & server need to clone their own copy of the `sessionState` on their constructor, to avoid mutating the same object. WDYT?

detox/src/ipc/ipc.test.js
asafkorem: :+1:

detox/src/ipc/ipc.test.js
asafkorem: :+1:

detox/src/ipc/ipc.test.js
asafkorem: 🙌🏼

asafkorem:
Can't approve my own pr 🙈, but anyway great changes and thanks for the explanations. See https://github.com/wix/Detox/pull/3680#discussion_r1024172192

detox/src/ipc/ipc.test.js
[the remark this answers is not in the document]
noomorph: Yes, they are distinct because they are running in different processes.

noomorph (approved):

detox/src/ipc/ipc.test.js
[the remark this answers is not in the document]
asafkorem: Right, right 👍🏼

PR #3686, opened 2022-11-10
Add unit-tests for the primary-secondary realms

### LINKED ISSUE #3643: [Detox 20] Unit tests: Detox contexts: primary and secondary

d4vidi:
## Description

- This pull request fixes #3643

In this pull request, I have add some unit-test coverage over the primary-context element. More to come*.

_* secondary context, among other things. Also, logger initialization, worker initialization._

---

> _For API changes:_


detox/.npmignore
```diff
@@ -15,9 +15,9 @@ wallaby.js
 __tests__
 __mocks__
 __snapshots__
-src/**/*.test.js
-src/**/*.mock.js
-src/**/*.mock.json
+**/*.test.js
```
noomorph, line 18: This is boy scouting.

When I checked whether `__tests__/helpers.js` gets into the final npm package, I found that we were not excluding some test files from the published package, e.g.: `local-cli/build.test.js`, etc.

detox/__tests__/helpers.js
```diff
@@ -0,0 +1,47 @@
+const path = require('path');
```
noomorph, line 1: See my comment on the next file, `detox/__tests__/setupJest.js`.

detox/__tests__/helpers.js
```diff
@@ -0,0 +1,47 @@
...
+      .exitProcess(false)
+      .fail((msg, err) => reject(err || msg))
+      .parse(cmd, (err) => err && reject(err));
+  });
+}
+
+exports.callCli = callCli;
+exports.latestInstanceOf = (clazz) => _.last(clazz.mock.instances);
+exports.lastCallTo = (mocked) => _.last(mocked.mock.calls);
+exports.backupProcessEnv = () => {
```
noomorph, line 41: It seems that backing up the `process.env` is quite a common scenario, so it deserves to be a common helper.

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
 jest.mock('proper-lockfile');
 jest.mock('signal-exit');
 jest.mock('../src/logger/DetoxLogger');
-
-const path = require('path');
-
-const _ = require('lodash');
-const yargs = require('yargs');
-
-function callCli(modulePath, cmd) {
```
noomorph, line 9: I tried to understand why I had been using `global.callCli` instead of importing it from a helper file, and to be frank, I have no idea. Therefore, I moved it to `detox/__tests__/helpers.js` as an exported function, and it worked flawlessly in the unit tests.

That was a huge inspiration to repeat this success for the other `global.*` helpers you introduced in this PR.

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
...
- * @example
- * await withSuspendingMock(innerMockObj, 'mockedMethodName', async ({ callSuspended }) => {
- *   await callSuspended(uut.methodUnderTest(), async () => {
- *     expect(uut.nextSolarEclipse()).toEqual( Date.parse('20 Apr 2023') );
- *
- *     return 'result-of-suspended-method'; // (Optional)
- *   }); // Resolve suspended method
- * }); // Await methodUnderTest's promise
- */
-async function withSuspendingMock(obj, func, callback) {
```
noomorph: 🙏  I'm always sorry to remove such well-worked code parts, but there's little sense in reinventing the wheel.

What you tried to achieve was trivial enough if you would have re-formulated it in Jest DSL:

```js
it('should change intermediate status to "cleanup"', async () => {
  // making sure that the test is not marked as PASSED without running the assertion below
  expect.assertions(1);
  await facade.init();

  ipcClient().dispose.mockImplementation(async () => {
    // running an expectation while `facade.cleanup()` is running
    expect(facade.getStatus()).toBe('cleanup');
  });

  await facade.cleanup(); // launching the sequence
});
```

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
...
-    resolve(result);
-
-    await promise;
-  };
-
-  await callback({ callSuspended });
-}
-
-// @ts-ignore
-global.callCli = callCli;
```
noomorph, line 42: Overall, I'm against pollution of global scope. All the methods used can be safely imported instead, including my mistake of youth, `callCli`.

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
...
-// @ts-ignore
-global.withSuspendingMock = withSuspendingMock;
-
-// @ts-ignore
-global.latestInstanceOf = (clazz) => _.last(clazz.mock.instances);
```
noomorph: This is actually a good helper. I moved it to the helpers module.

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
...
-// @ts-ignore
-global.lastCallTo = (mocked) => _.last(mocked.mock.calls);
```
noomorph: This is actually a good helper. I moved it to the helpers module.

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
...
-
-// @ts-ignore
-global.throwErrorImpl = (message = '') => { throw new Error(message); };
```
noomorph: This is a rather pointless helper.
I looked at places where you had been using it, and I can tell you that you could have used this instead:

```js
mockFn.mockRejectedValue(new Error(''));
mockFn.mockRejectedValue(new Error('init error'));
```

detox/__tests__/setupJest.js
```diff
@@ -1,94 +1,3 @@
...
-
-const _uuidRegexp = '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}';
```
noomorph: Indeed, we use UUID regexp in a few test files, so it may make sense to have a shared one **eventually**.
On the other hand, `tempFileRegexp` sounds too specific – only a handful of modules could make sense of it.

So, I think it is better to move those regular expressions to `DetoxPrimaryContext.test.js`, the only place that uses them.

detox/local-cli/build.test.js
```diff
@@ -1,11 +1,11 @@
 const tempfile = require('tempfile');
 
+const { callCli } = require('../__tests__/helpers');
```
noomorph, line 3: Fixing my "mistakes of youth."
`callCli` is now an exported helper.

detox/local-cli/run-server.test.js
```diff
@@ -1,6 +1,7 @@
-// @ts-nocheck
 jest.mock('../src/utils/logger');
 
+const { callCli } = require('../__tests__/helpers');
```
noomorph, line 3: Fixing my "mistakes of youth."
`callCli` is now an exported helper.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -1,62 +1,73 @@
...
-  uuidRegexp,
-  tempFileRegexp,
-} = global;
-
-const workerId = 19;
+} = require('../../__tests__/helpers');
 
 describe('DetoxPrimaryContext', () => {
+  //#region Fixtures and constants
+  const UUID_REGEXP_STR = '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}';
```
noomorph, line 11: I prefer to use UPPER_CASE for constant variables which are not being reset in `beforeEach()` and/or `afterEach()` hooks.

noomorph: Also, maybe it is a superstition, but I try not to declare variables out of the root describe.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -1,62 +1,73 @@
-// @ts-nocheck
```
noomorph: I removed ignoring the types and fixed TypeScript errors instead.

detox/src/realms/DetoxSecondaryContext.test.js
```diff
@@ -1,40 +1,35 @@
-// @ts-nocheck
```
noomorph: I removed ignoring the types and fixed TypeScript errors instead.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -1,62 +1,73 @@
...
+      autoStart: false,
+    },
+  });
+  const FIRST_ARGUMENT = 0;
+  //#endregion
+
   let detoxConfigDriver;
 
+  //#region Mocks
+  /** @type {jest.Mocked<import('fs-extra')>} */
```
noomorph, line 35: I fixed the types and extracted all mocked classes and instances to the region with those `let`, `let` ... `let` variables.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -1,62 +1,73 @@
...
   let facade;
 
-  let env;
-
   const detoxServer = () => latestInstanceOf(DetoxServer);
-  const detoxWorker = () => {
-    const DetoxWorker = require('../DetoxWorker');
-    return latestInstanceOf(DetoxWorker);
-  };
+  const ipcServer = () => latestInstanceOf(IPCServer);
```
noomorph, line 65: After extracting all the types, these instance extractors turned into sleek one-liners.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -1,62 +1,73 @@
...
-  const getSignalHandler = () => {
-    const onSignalExit = require('signal-exit');
-    const [signalHandler] = lastCallTo(onSignalExit);
-    return signalHandler;
-  };
+  const getSignalHandler = () => lastCallTo(signalExit)[FIRST_ARGUMENT];
 
-  beforeEach(() => env = process.env);
-  afterEach(() => process.env = { ...env });
+  backupProcessEnv();
```
noomorph, line 70: Looks like we could use this helper in much more suites, and eventually, we might do that.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -186,23 +197,23 @@ describe('DetoxPrimaryContext', () => {
...
-      await facade.init({ workerId });
-      expect(ipcServer.onRegisterWorker).toHaveBeenCalledWith({ workerId });
+      await facade.init({ workerId: WORKER_ID });
+      expect(ipcServer().onRegisterWorker).toHaveBeenCalledWith({ workerId: WORKER_ID });
     });
 
     describe('given an initialization failure', () => {
       it('should report status as "init"', async () => {
-        ipcServer.init.mockImplementation(() => throwErrorImpl('init error'));
+        IPCServer.prototype.init = jest.fn().mockRejectedValue(new Error('init failed'));
```
noomorph, line 216: I checked what you tried to achieve with that complex mocking (4 lines) for that IPCServer, and realized you could just mock the prototype, which is pretty idiomatic for JS developers.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -345,96 +357,73 @@ describe('DetoxPrimaryContext', () => {
...
+    configuration = jest.requireMock('../configuration');
     configuration.composeDetoxConfig.mockResolvedValue(detoxConfigDriver.detoxConfig);
   }
 
   function _initExternalMocks() {
     jest.mock('signal-exit');
+    signalExit = jest.requireMock('signal-exit');
     jest.mock('fs-extra');
-    fs = require('fs-extra');
+    fs = jest.requireMock('fs-extra');
```
noomorph, line 378: I'm using `jest.requireMock` a lot to make sure that the typings match.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -345,96 +357,73 @@ describe('DetoxPrimaryContext', () => {
...
     jest.mock('../DetoxWorker');
+    DetoxWorker = jest.requireMock('../DetoxWorker');
   }
-});
 
-const detoxConfigBase = Object.freeze({
-  behavior: {
-    init: {
-      keepLockFile: false,
+  class DetoxConfigDriver {
```
noomorph, line 418: It is an interesting approach with drivers. Maybe we should use it even more, at least, in some test suites.

detox/src/realms/DetoxSecondaryContext.test.js
```diff
@@ -1,40 +1,35 @@
...
-  latestInstanceOf,
-  throwErrorImpl,
-  withSuspendingMock,
-} = global;
-
-const DETOX_CONFIG_SNAPSHOT_PATH = 'mocked/detox.json';
-const sessionState = {
-  id: 'mocked-session-state',
-};
+describe('DetoxSecondaryContext', () => {
```
noomorph, line 3: There is nothing new here regarding my review comments. Please check out `DetoxPrimaryContext.test.js` and `detox/__tests__/helper.js` file to see the motivation of changes.

noomorph (approved):

detox/__tests__/setupJest.js
d4vidi: nice!

detox/__tests__/setupJest.js
d4vidi: nice! don't worry about it! makes perfect sense

detox/__tests__/setupJest.js
d4vidi: np, I'm not strongly opinionated on those things

detox/src/realms/DetoxPrimaryContext.test.js
d4vidi: Might not be a bad idea to register global before-after callbacks that do this for all tests

detox/src/realms/DetoxPrimaryContext.test.js
d4vidi: Could you elaborate?

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -29,4 +97,333 @@ describe('DetoxPrimaryContext', () => {
...
+    DeviceRegistry.forAndroid.mockReturnValue(deviceRegistryAndroid);
+
+    jest.mock('../devices/allocation/drivers/android/genycloud/GenyDeviceRegistryFactory');
+    const genycloudDeviceRegistryFactory = jest.requireMock('../devices/allocation/drivers/android/genycloud/GenyDeviceRegistryFactory');
+    deviceRegistryGenyCloud = new DeviceRegistry();
+    genycloudDeviceRegistryFactory.forGlobalShutdown.mockReturnValue(deviceRegistryGenyCloud);
+
+    // The mocking complexity here is higher than the norm so as to allow for interacting
+    // with both the class and the generated instance as mocks; With the latter - even
+    // before its creation by the tested-unit (i.e. in its init()).
```
d4vidi, line 400: Can be deleted

d4vidi:
So looks like we're just 1 subtle iteration away from merging. Will you able to get on it tomorrow?

detox/__tests__/setupJest.js
d4vidi: Oh BTW, be sure to also `expect()` that `ipcClient().dispose` has indeed been called......

d4vidi: Oh that's `expect.assertions(1)`... got it

detox/src/realms/DetoxPrimaryContext.test.js
noomorph: Well, it all was just about uniformity and readability – it's easier for the brain to scan over the boring same-pattern code than to read and try to understand why this place differs from the others.

detox/src/realms/DetoxPrimaryContext.test.js
noomorph: Well, maybe, but it's rather a red flag for me. It will worsen our self-tracking of code that depends on `process.env`. It took me a few years to get rid of `process.env` dependencies here and there (`argparse` util counts in too, because It does the same under the hood).

It would be convenient to have it explicit so we can search in the project and know exactly how many modules-under-the-test have this nasty tendency to poke in `process.env`. I'd rather had this as an exception to the rule than make every module/unit test so careless.

detox/src/realms/DetoxPrimaryContext.test.js
noomorph: In TypeScript, `jest.requireMock` returns `Mock<T>` which plays well with the JSDoc `/** @type {} */` declarations I define above. Altogether, it gives us better type-checking without the need to `@ts-ignore`.

detox/src/realms/DetoxPrimaryContext.test.js
noomorph: Oh, correct.

detox/src/realms/DetoxPrimaryContext.test.js
d4vidi: OK but all of those `#region` comments are beginning to mess up our code (e.g. if you have Webstorm as an IDE, or on Github). I'd rather hold that back a bit.

detox/src/realms/DetoxPrimaryContext.test.js
d4vidi: Right, thanks!

detox/src/realms/DetoxPrimaryContext.test.js
d4vidi: Awesome 😎

detox/src/realms/DetoxPrimaryContext.test.js
d4vidi: I agree

Issue #3692, opened 2022-11-11
🚨 [Detox 20] Platform-specific: Log artifacts are not created on Windows


noomorph:
### What happened?

Our integration tests running on Windows agents are currently showing that there are some issues with creating JSONL event streams.

This is likely to cause missing log artifacts on Windows.
This issue has a high priority and is expected to be solved today.

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### In what environment did this happen?

Detox version: 20.0.1
Node version: 14.x
OS: Windows 10
Test-runner (select one): jest

noomorph:
Closed in Detox 20.0.2.

PR #3694, opened 2022-11-11
fix: logger errors on Windows

noomorph:
## Description

- This pull request addresses the issue described here: #3692

### DetoxLogger#close()

In this pull request, I have added an internal method, `close()`, to the `DetoxLogger` class.

It is called by DetoxPrimaryContext inside the `_finalize()` method.
Closing flushes the write streams. Ignoring this step provokes issues on Windows when Detox tries to move logs – they are often empty or incomplete. Also, omitting closing makes our integration tests flaky.

### Implications of closed logger

I had to tweak the final (try-catch) stage in `local-cli/cli.js`, because the logic there tries to use an already closed logger (i.e., stdout-only state). This is not fatal, but the logged messages won't get into the log files, which is sad, especially when we can easily overcome this in the most common case.

So, now, instead of plain logging, `Command exited with code = ...`, I log duration events (`log.info.begin() → log.error.end() OR log.trace.end()`) at every launch of a test runner. This can be considered an improvement to the rich timelines we create.

### Glob package and Windows OS

Another fixed issue was caused by the `glob` package. Turns out, it accepts forward slashes (POSIX) only. I had to correct my functions using it.

### Duplexify package and Windows OS

I didn't succeed in digging to the bottom of it, but I found out that the `duplexify` package doesn't play well with Windows. Cryptic errors like "being corked" and "this._writableState is undefined" appear, while the behavior in POSIX environments seems to be fine.

As a result, I use a clumsier approach – instead of `duplexify(writable, readable)` I return an object with both the streams, and their consumers configure all the further piping.

### Unit and integration tests

While I tried to figure out what was wrong, I had to refactor that big `streamUtils` file into several smaller classes and modules and added more tests.


detox/src/logger/DetoxLogger.js
```diff
@@ -119,6 +119,25 @@ class DetoxLogger {
     this.overrideConsole();
   }
 
+  /**
+   * Closes the file descriptors to make sure that the temporary
+   * JSONL files can be moved safely to the directory with the artifacts.
```
noomorph, line 124:
```suggestion
   * JSONL files are flushed and contain the last error messages.
```

detox/src/logger/DetoxLogger.js
```diff
@@ -119,6 +119,25 @@ class DetoxLogger {
...
+   * This safety measure is especially important for Windows OS.
+   *
+   * @async
+   * @internal
+   */
+  async close() {
```
noomorph, line 130: Hm, overall, I wonder if we could try just flushing the logger instead of closing the file streams.

noomorph: Seems not really viable. We could try to subscribe to the `drain` event or something, and try to wait, but closing looks more solid.

noomorph:
@jonathanmos , could you please review this PR? Since you've been commenting last time on the logger and you're better in the context.

Maybe you could also check if there are any other review comments about DetoxLogger (from Detox 20 PR) that I haven't addressed yet.

detox/src/logger/utils/streams/DetoxJSONLParser.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  createTransformer() {
+    const log = this._log;
+    const readable = through();
+    const writable = JsonlParser.make({ checkErrors: true })
+      .on('error', (err) => {
+        /* istanbul ignore else */
+        if (err instanceof SyntaxError) {
+          log.debug({ err });
```
d4vidi, line 17: So this will not be parsed and transformed, obviously. But will the logging be successful? Have you managed to test that out?

detox/src/logger/utils/streams/DetoxJSONLParser.js
```diff
@@ -0,0 +1,31 @@
...
+          readable.end();
```
d4vidi, line 18: I'm not sure what `SyntaxError` stands for in this context, but assuming (guessing) that it is about a specific json line then wouldn't it be better to skip (and provide some other "wrapper" log), instead? Something along the lines of:
`<Failed to parse line: ${err}>`

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -210,12 +212,17 @@ class DetoxPrimaryContext extends DetoxContext {
...
       this[$logFinalizer].finalizeSync();
     } catch (err) {
-      this[symbols.logger].error({ err }, 'Encountered an error while merging the process logs:');
+      this[_logFinalError](err);
     }
   };
 
+  [_logFinalError] = (err) => {
+    this[_lifecycleLogger].error(err, 'Encountered an error while merging the process logs:');
+  };
```
d4vidi, line 224: I don't fully understand what's happening here - need some help... It looks like a ["hail mary"](https://www.dictionary.com/browse/hail-mary), and if so, would it be helpful at all? Assuming the streams finalization didn't end well - what's going to happen when trying to write a new log? 🤔

d4vidi (commented):
Hey!
I tried my best to go over this thoroughly, but it wasn't easy because it seems to have piled up into something a bit massive. So, I left some stuff overlooked.
In any case, please check out the notes on what could get into 🙏🏻

detox/src/logger/utils/streams/DetoxJSONLParser.js
noomorph: Yes, that can potentially happen on abruptly terminated log files. I tested this case manually and in integration tests.

detox/src/logger/utils/streams/DetoxJSONLParser.js
noomorph: It stands for the unfinished JSON object on the last line of JSONL file.

noomorph: The idea with "Failed to parse line" seems nice. 👍 TODO

detox/src/logger/DetoxLogger.test.js
```diff
@@ -70,8 +70,12 @@ describe('DetoxLogger', () => {
     safetyTimeout = null;
   }
 
-  function teardownSuite() {
+  async function teardownSuite() {
     process.env = _env;
+
+    await Promise.all(_loggerInstances.map((logger) => logger.close()));
```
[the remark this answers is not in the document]
noomorph: Hm, but I close the logger in the real world too, when Detox finalizes the logs. It is not only for tests. I noticed that on Windows the log files are not always complete and that made me wonder how to fix it. That's how eventually I came to close() method.

detox/src/artifacts/utils/temporaryPath.js
```diff
@@ -1,14 +1,42 @@
...
 const tempfile = require('tempfile');
 
+const { useForwardSlashes } = require('../../utils/shellUtils');
+
+const globSync = glob.sync;
+const globAsync = promisify(glob);
+
+function getRoot() {
+  return path.dirname(tempfile());
+}
```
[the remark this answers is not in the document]
noomorph: Okay, TODO.

detox/local-cli/testCommand/TestRunnerError.test.js
```diff
@@ -0,0 +1,25 @@
...
+    expect(`${error}`).toMatch(/Command failed with exit code = 1:\nfoo/);
+  });
+
+  it('should assign properties (code, signal)', () => {
+    expect(error.code).toBe(1);
+    expect(error.signal).toBe(null);
+  });
+
+  it('should be formatted to an empty string', () => {
+    // So that the try-catch block in cli.js will not print it for the second time
```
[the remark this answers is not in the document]
noomorph: Yes, I have already another idea how to fix it.
TODO.

detox/src/logger/utils/BunyanLogger.js
```diff
@@ -71,6 +71,45 @@ class BunyanLogger {
...
+          : bunyanStream.stream;
+
+        /* istanbul ignore next */
+        return stream.fd !== undefined && !stream.closed;
+    }
+  };
+
+  /** @private */
+  static _closeStream(bunyanStream) {
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
noomorph: TODO

detox/src/logger/utils/BunyanLogger.js
```diff
@@ -71,6 +71,45 @@ class BunyanLogger {
 
     return this;
   }
+
+  async closeFileStreams() {
```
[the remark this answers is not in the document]
noomorph: 🤷‍♂️

detox/src/logger/DetoxLogger.test.js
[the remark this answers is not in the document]
d4vidi: What I meant was that we're both testing the loggers' `close()` method and relying here on the fact that it works, in terms of after-test cleanup...

detox/src/realms/DetoxPrimaryContext.js
d4vidi: See where - in the console? I wonder whether that's true, seeing that all/some streams have been closed...

noomorph: Yes, in console. Only file streams are closed.

detox/src/logger/DetoxLogger.test.js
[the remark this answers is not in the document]
noomorph: Well, the entire suite is rather an integration than a unit test. I see what you mean, but in the end, I'm making assertions on real text files. Maybe we could add a dedicated unit test file to the logger if we have time, but I'd leave the integration test as it allows me to peek and see how the real logs (seen by the end-user) will look.

PR #3695, opened 2022-11-13
website(showcase): add new projects (Expo, Currenxie).

asafkorem:
The PR adds two new showcased projects (Expo, Currenxie) and replaces the fixed arbitrary order of apps with a shuffled order on every load.


website/src/components/Showcase/Showcase.js
```diff
@@ -5,8 +5,19 @@ import cardList from '@site/showcase.json';
...
 
 function Showcase() {
-  const cards = cardList.map((props) => <Card key={props.title} {...props} />);
+  const cards = _makeShuffledCardList().map(_makeCard)
+  return _makeCardsContainer(cards);
+}
+
+const _makeShuffledCardList = () => {
+  return cardList.sort(() => 0.5 - Math.random());
+}
```
d4vidi, line 14:
```suggestion
const _makeShuffledCardList = () => (cardList.sort(() => 0.5 - Math.random()));
```

website/src/components/Showcase/Showcase.js
```diff
@@ -5,8 +5,19 @@ import cardList from '@site/showcase.json';
...
+
+const _makeCard = (props) => {
+  return <Card key={props.title} {...props} />;
+}
```
d4vidi, line 18:
```suggestion
const _makeCard = (props) => <Card key={props.title} {...props} />;
```

website/src/components/Showcase/Showcase.js
d4vidi: I think we should keep the Wix ones at the top:
1. Would be weird to have it somewhere in the middle
2. Need to stay adjacent, as a group

asafkorem: OK, I agree on both points.. Done

website/src/components/Showcase/Showcase.js
asafkorem: Done :+1:

website/src/components/Showcase/Showcase.js
```diff
@@ -10,19 +10,18 @@ function Showcase() {
 }
 
 const _makeShuffledCardList = () => {
-  return cardList.sort(() => 0.5 - Math.random());
+  const alwaysOnTop = cardList.filter(card => card.shouldStayOnTop);
+  const shuffled = cardList.filter(card => !card.shouldStayOnTop).sort(() => Math.random() - 0.5);
+  return alwaysOnTop.concat(shuffled);
```
d4vidi, line 15: [Lodash partition[(https://lodash.com/docs/4.17.15#partition) 🤔

asafkorem: I thought about that but didn't care enough 😄

asafkorem: anyway https://github.com/wix/Detox/pull/3696

PR #3696, opened 2022-11-13
website(showcase): refactor with lodash methods.

asafkorem:


noomorph:
Asaf, regardless of Lodash or vanilla JS, my concern is that you are breaking hydration stage when rendering items in a non-deterministic order.

I'd advise to move shuffling into useEffect hook.

asafkorem:
@noomorph I had no idea this might happen. I now see that the images are for some reason cached and stays on their position although other parts of the card replaces the position. Locally I had no such problem. 

I'm reverting this change

PR #3713, opened 2022-11-17
fix(logger): save files on Ctrl+C

### LINKED ISSUE #3679: [Detox 20] Improve Ctrl+C behavior
### Describe your idea

Reimplement DetoxReporter logic:

* report file after first failing test
* if the test file, after all, is marked as successful (e.g., after a Jest Circus retry), "un-report" that failing test back
* report the full summary only at the end

The early failure reporting will enable keeping log artifacts on emergency teardown when the collection is configured only for the failing tests.

On second thought, it is too complicated. I found a simpler solution and posted it in the first comment.

noomorph:
## Description

Resolves #3679

Currently, when using `--record-logs failed` and interrupting Detox in the middle, it does not save the logs because it has not received a signalization about failing tests, so the logic is way too straightforward – discard the logs.

I suggest treating Ctrl+C (SIGINT, SIGTERM, etc) as an unsuccessful run so the logs are saved regardless.

That should reduce frustration when you are running into some timeout issues, and you want to interrupt Detox to read some saved logs, but oops – "no error, no logs". The latter rule is too strict, to my taste, in this case.

### DIFF
--- a/detox/src/logger/utils/DetoxLogFinalizer.test.js
+++ b/detox/src/logger/utils/DetoxLogFinalizer.test.js
@@ -172,20 +172,9 @@ describe('DetoxLogFinalizer', () => {
       expect(fs.existsSync(temporaryFiles[2])).toBe(false);
     });
 
-    it('should not create logs for successful run when configured', async () => {
+    it('should create logs even for successful run when configured to keep failing logs', async () => {
       session.detoxConfig.artifacts.plugins.log.keepOnlyFailedTestsArtifacts = true;
       session.testResults = [{ success: true }];
-
-      await createLogFiles();
-      finalizer.finalizeSync();
-      expect(fs.existsSync(artifactsDir())).toBe(false);
-      expect(fs.existsSync(temporaryFiles[1])).toBe(false);
-      expect(fs.existsSync(temporaryFiles[2])).toBe(false);
-    });
-
-    it('should create logs for failing run when configured', async () => {
-      session.detoxConfig.artifacts.plugins.log.keepOnlyFailedTestsArtifacts = true;
-      session.testResults = [{ success: false }];
 
       await createLogFiles();
       finalizer.finalizeSync();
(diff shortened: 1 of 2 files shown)


detox/src/logger/utils/DetoxLogFinalizer.test.js
```diff
@@ -172,21 +172,10 @@ describe('DetoxLogFinalizer', () => {
       expect(fs.existsSync(temporaryFiles[2])).toBe(false);
     });
 
-    it('should not create logs for successful run when configured', async () => {
+    it('should create logs even for successful run when configured to keep failing logs', async () => {
```
noomorph, line 175: Please mind that this scenario is being checked for "finalizeSync" (which is an emergency substitute for the "finalize").

detox/src/logger/utils/DetoxLogFinalizer.test.js
```diff
@@ -172,21 +172,10 @@ describe('DetoxLogFinalizer', () => {
...
       session.testResults = [{ success: true }];
 
-      await createLogFiles();
-      finalizer.finalizeSync();
-      expect(fs.existsSync(artifactsDir())).toBe(false);
-      expect(fs.existsSync(temporaryFiles[1])).toBe(false);
-      expect(fs.existsSync(temporaryFiles[2])).toBe(false);
-    });
-
-    it('should create logs for failing run when configured', async () => {
```
d4vidi, line 186: Don't you want to keep this test case? It's not a given IMO, even though it's all the same as `{success: true}` implementation-wise.

d4vidi (approved):
I'm generally ok with this, although -
1. See 1 comment that I had
2. Does this effectively keep artifacts with Ctrl+C? *What about an abrupt CI job terminate?*

noomorph:
Sigkill is sigkill, it's hammer time. Less violent signals should end with artifacts.

d4vidi:
> Sigkill is sigkill, it's hammer time. Less violent signals should end with artifacts.

I see

d4vidi:
@noomorph any work left to do here? Mind that I've already approved it

detox/src/logger/utils/DetoxLogFinalizer.test.js
noomorph: Well, since the coverage is 100% there, adding another if-else branch will reduce it, forcing you to add a new test, so hopefully this is not very critical.

asafkorem:
Detox version [20.1.0](https://github.com/wix/Detox/releases/tag/20.1.0) has been released with this fix.

PR #3715, opened 2022-11-17
refactor(trace): simpler primary PID detection

noomorph:
## Description

In this PR, I make it simpler to infer which PID is primary (when creating a Detox trace file).

Instead of making this logic tightly coupled with `process.pid`, I check which PID appears first.

That simplifies unit tests and makes the code more generic and reusable.


d4vidi (approved):
I'm basically okay with approving this, although I'm not 100% convinced why this is a better approach, TBH. Is taking the assumption that the primary context always goes first better than taking the assumption that it is the only one ever finalizing the logs? 🤷🏻 IDK, but anyways - your call.

Issue #3292, opened 2022-04-03
Clean up lint warnings


d4vidi:
### Describe your idea

Simply put, I think we've had enough of seeing this in every PR:

![image](https://user-images.githubusercontent.com/9818880/161432472-153bb69e-a538-40d5-949f-a05750b7645f.png)

noomorph:
Well, part of them will go away with Detox 20, but every item needs some groundwork to be resolved.

d4vidi:
Great. I think lifting the restrictions over ToDo's would make a huge difference, for starters.

noomorph:
I might take it tomorrow for 1-2 hours, maybe partially I can remove around 20 warnings.

d4vidi:
Great if you could get to doing this, sometime soon. Will make our PR's much cleaner.

noomorph:
@d4vidi oh damn it :) I forgot to push the branch 
Here it is
https://github.com/wix/Detox/pull/3719
You're welcome to see it.

PR #3719, opened 2022-11-20
lint: fix "unused" warnings

noomorph:
## Description

- This pull request addresses the issue described here: #3292

Mostly, I fix this by relaxing ESLint rule for unused function **args** only.

Unfortunately, I can't make it smarter (it does not discern concrete and virtual methods), so this is the most I can come up with.

On the other side, allowing unused variables is a road to hell, so I'm better off with special ESLint comments than with disabling it at all.

### DIFF
--- a/detox/.eslintrc.js
+++ b/detox/.eslintrc.js
@@ -54,12 +54,7 @@ module.exports = {
       }
     ],
     'no-prototype-builtins': 'off',
-    'no-unused-vars': [
-      'error',
-      {
-        'argsIgnorePattern': '^_'
-      }
-    ],
+    'no-unused-vars': 'off',
     'node/no-unpublished-require': 'warn',
     'object-curly-spacing': [
       'error',
--- a/detox/src/artifacts/templates/plugin/ArtifactPlugin.js
+++ b/detox/src/artifacts/templates/plugin/ArtifactPlugin.js
@@ -1,5 +1,5 @@
+/* eslint @typescript-eslint/no-unused-vars: ["error", { "args": "none" }] */
 // @ts-nocheck
-/* eslint-disable no-unused-vars */
 
 const _ = require('lodash');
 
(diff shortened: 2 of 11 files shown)


noomorph:
Also I extracted one more pull request out of this one: #3778. 

Please make sure to review it too.

detox/src/artifacts/templates/plugin/ArtifactPlugin.js
```diff
@@ -44,13 +44,13 @@ class ArtifactPlugin {
...
-   * @param {string} event.bundleId - Current bundleId
-   * @param {Object} event.launchArgs - Mutable key-value pairs of args before the launch
+   * @param {Object} _event - Launch app event object
+   * @param {string} _event.deviceId - Current deviceId
+   * @param {string} _event.bundleId - Current bundleId
+   * @param {Object} _event.launchArgs - Mutable key-value pairs of args before the launch
    * @return {Promise<void>} - when done
    */
-  async onBeforeLaunchApp(event) {
+  async onBeforeLaunchApp(_event) {
```
d4vidi, line 53: I'm not very enthusiastic about this approach. It makes the argument's crippled name the official one, for the API. It makes it weird esp. in these cases where classes inherit these methods -- both in calling `super()`, and when called as an external API.

d4vidi (changes requested):
See my single note on this. I don't like this solution...

detox/.eslintrc.js
```diff
@@ -54,12 +54,7 @@ module.exports = {
       }
     ],
     'no-prototype-builtins': 'off',
-    'no-unused-vars': [
-      'error',
-      {
-        'argsIgnorePattern': '^_'
-      }
-    ],
+    'no-unused-vars': 'off',
```
noomorph, line 57: We rely on `@typescript-eslint/no-unused-vars` for that instead.

noomorph:
Changed the solution to a less intrusive one.

PR #3731, opened 2022-11-23
docs: add web views API documentation.

asafkorem:


asafkorem:
Preview:
image

d4vidi:
Wow! YES! At a glance - seems very sexy... We will review ASAP

noomorph (changes requested):
Hi @asafkorem ! You have my review here, in your own repository as a pull request:

https://github.com/asafkorem/Detox/pull/9

docs/api/webviews.md
```diff
@@ -0,0 +1,355 @@
...
+
+Get the title of the web view.
+
+:::note
+Although this action returns the title of the presented web document, it can be called from an inner element only (for example, an iframe id or the HTML) and not from the root native web view element itself.
+:::
+
+```js
+const title = await web.element(by.web.id('identifier')).getTitle();
+```
```
asafkorem, line 287: I really don't like these APIs. It can be useful when testing an inner **iframe** URL or title, but not convenient when trying to test the root document. It will probably work when targeting the HTML tag element using `by.web.tag("html")` but I haven't tried that yet.

We should expose the same APIs for the web-view element (the root native element), for running the same operations on the web-view element (the main frame).

The Espresso code we're running with this APIs is:
```java
return getWebViewInteraction().withElement(get()).perform(Atoms.getCurrentUrl()).get();
```
https://github.com/wix/Detox/blob/master/detox/android/detox/src/full/java/com/wix/detox/espresso/web/WebElement.java#L65-L79 (`WebElement`)

While we should enable operating on the root frame **without specifying the inner element** (from `WebViewElement` object) is something like:
```java
return getWebViewInteraction().reset().perform(Atoms.getCurrentUrl()).get();
```

Should be in the same API level with `web.element(..)` https://github.com/wix/Detox/blob/master/detox/android/detox/src/full/java/com/wix/detox/espresso/web/WebViewElement.java

asafkorem: I've opened a new issue for this: https://github.com/wix/Detox/issues/3752

noomorph (approved):

noomorph:
:shipit:

Issue #3716, opened 2022-11-18
[Detox 20] (WIP) Finish Logger review


noomorph:
### Describe your idea

See #3749 which finishes the logger review.

d4vidi:
@noomorph reminder 😄 What should the actual task look like?

noomorph:
It is about this code review and your comments on it:

https://github.com/wix/Detox/pull/3694

d4vidi:
> It is about this code review and your comments on it:
> 
> #3694

OK I see now that's both about #3694 _and_ #3749. My task here is to review the latter, then.

PR #3749, opened 2022-12-09
refactor: follow-up to the logger review

noomorph:
## Description

Follow-up of #3694 .

* https://github.com/wix/Detox/pull/3694#discussion_r1023980628
* https://github.com/wix/Detox/pull/3694#discussion_r1026505297
* https://github.com/wix/Detox/pull/3694#discussion_r1026505098
* https://github.com/wix/Detox/pull/3694#discussion_r1024001526
* https://github.com/wix/Detox/pull/3694#discussion_r1026506372
* https://github.com/wix/Detox/pull/3694#discussion_r1026499800

### DIFF
--- a/detox/local-cli/utils/cliErrorHandling.js
+++ b/detox/local-cli/utils/cliErrorHandling.js
@@ -1,0 +1,15 @@
+const loggedErrors = new WeakSet();
+
+function markErrorAsLogged(error) {
+  loggedErrors.add(error);
+  return error;
+}
+
+function isErrorAlreadyLogged(error) {
+  return loggedErrors.has(error);
+}
+
+module.exports = {
+  markErrorAsLogged,
+  isErrorAlreadyLogged,
+};
--- a/detox/src/artifacts/utils/temporaryPath.js
+++ b/detox/src/artifacts/utils/temporaryPath.js
@@ -2,16 +2,14 @@ const path = require('path');
 const { promisify } = require('util');
 
 const glob = require('glob');
+const _ = require('lodash');
 const tempfile = require('tempfile');
 
 const { useForwardSlashes } = require('../../utils/shellUtils');
 
 const globSync = glob.sync;
 const globAsync = promisify(glob);
-
-function getRoot() {
-  return path.dirname(tempfile());
-}
+const getRoot = _.once(() => path.dirname(tempfile()));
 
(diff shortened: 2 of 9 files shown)


detox/src/artifacts/utils/temporaryPath.js
```diff
@@ -2,16 +2,14 @@ const path = require('path');
...
 
 const { useForwardSlashes } = require('../../utils/shellUtils');
 
 const globSync = glob.sync;
 const globAsync = promisify(glob);
-
-function getRoot() {
-  return path.dirname(tempfile());
-}
+const getRoot = _.once(() => path.dirname(tempfile()));
```
d4vidi, line 12: 🔥

detox/local-cli/utils/cliErrorHandling.js
```diff
@@ -0,0 +1,15 @@
...
+}
+
+function isErrorAlreadyLogged(error) {
+  return loggedErrors.has(error);
+}
+
+module.exports = {
+  markErrorAsLogged,
+  isErrorAlreadyLogged,
+};
```
d4vidi, line 15: So do I totally dislike this solution? Of course not! But do I _like_ this solution? Hmm........ 🤔
It _is_ kind of elegant... and as for concerns about garbage collection not kicking in on time - experimentation shows we're on the good side of things (yes, I checked). So I guess the real question is, does the error really have to be reported via [`logger.error.end()`](https://github.com/wix/Detox/pull/3749/files#diff-548b574af3df69470e3a3a4c901c845b4180a406c29b1d621ec1eaa25ec93255R145) in TestRunnerCommand? Meaning, to some degree, this duplication situation is self-inflicted, isn't it?

d4vidi (approved):
Approved, but please review my comment and try to give it some extra thought

detox/local-cli/utils/cliErrorHandling.js
noomorph: I can't come up with a more creative solution right now. It makes sense to call `log.*end()` there because there might be a few times we call an underlying test runner command and we want to log each command separately. Maybe we'll revisit this later with some additions/rewrites to the `local-cli/*`.

d4vidi:
Wait, so can this be closed? - #3716

PR #3756, opened 2022-12-13
fix: immutable matchers

### LINKED ISSUE #3675: withAncestor modifies original matcher
### What happened?

Calling `withAncestor` (and I suspect, `withDescendant`, and `and`), modifies the original selector it was called upon

I'd expect it to return a new instance of the matcher with the additional filtering applied, without overwriting the original. Overriding the original means that in a test suite, you can't easily share/reuse base matchers.

Create some simple test with a reusable component, that is then used in two different ancestors, and write a test against that. The below will then show this problem:

```
const childComponent = by.id('child');
const parentComponentA = element(sharedComponent.withAncestor(by.id('parent-a')));

// Expected: MATCHER((id == “child” && ANCESTOR(id == “parent-a”))
// Actual: MATCHER((id == “child” && ANCESTOR(id == “parent-a”)) && ANCESTOR(id == “parent-b”))
const parentComponentB = element(childComponent.withAncestor(by.id('parentB')));
```

Detox version: 19.3.0
React Native version: 0.69.6
Node version: 16.14.2
Test-runner (select one): jest-circus

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
paste logs here!
```

</details>

### More data, please!

_No response_

noomorph:
## Description

Resolves #3675

### DIFF
--- a/detox/src/android/AndroidExpect.test.js
+++ b/detox/src/android/AndroidExpect.test.js
@@ -93,6 +93,19 @@ describe('AndroidExpect', () => {
       await e.expect(e.element(e.by.id('test').or(e.by.id('test2')))).toBeVisible();
     });
 
+    test.each([
+      ['withAncestor'],
+      ['withDescendant'],
+      ['and'],
+    ])(`matcher immutability: %s`, async (combineMethodName) => {
+      const base = e.by.id('test');
+      const modifier = e.by.id('ancestor');
+
+      expect(base[combineMethodName](modifier)).not.toBe(base);
+      expect(base).toEqual(e.by.id('test'));
+      expect(modifier).toEqual(e.by.id('ancestor'));
+    });
+
     it(`should throw for invalid toBeVisible parameters`, async () => {
       await expectToThrow(() =>e.expect(e.element(e.by.label('test'))).toBeVisible(0));
       await expectToThrow(() =>e.expect(e.element(e.by.label('test'))).toBeVisible(120));
--- a/detox/src/ios/expectTwo.test.js
+++ b/detox/src/ios/expectTwo.test.js
@@ -115,6 +115,19 @@ describe('expectTwo', () => {
     expect(testCall).toDeepEqual(jsonOutput);
   });
 
+  it.each([
+    ['withAncestor'],
+    ['withDescendant'],
+    ['and'],
+  ])(`should produce immutable objects when combining matchers: %s`, async (combineMethodName) => {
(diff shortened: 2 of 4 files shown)


detox/src/android/AndroidExpect.test.js
```diff
@@ -93,6 +93,15 @@ describe('AndroidExpect', () => {
       await e.expect(e.element(e.by.id('test').or(e.by.id('test2')))).toBeVisible();
     });
 
+    it(`matcher immutability`, async () => {
+      const base = e.by.id('test');
+      const modifier = e.by.id('ancestor');
+
+      expect(base.withAncestor(modifier)).not.toBe(base);
+      expect(base.withDescendant(modifier)).not.toBe(base);
+      expect(base.and(modifier)).not.toBe(base);
```
asafkorem, line 102: More specifically, you can test that `base` and `modifier` haven't mutated, for example:
```js
const base = e.by.id('foo');
const modifier = e.by.id('bar');

const _ = base.withAncestor(modifier);
const __ = base.withDescendant(modifier);
const ___ = base.and(modifier);

expect(base).toEqual(e.by.id('foo'));
expect(modifier).toEqual(e.by.id('bar'));
```

detox/src/ios/expectTwo.test.js
```diff
@@ -115,6 +115,17 @@ describe('expectTwo', () => {
...
 
+  it.each([
+    ['withAncestor'],
+    ['withDescendant'],
+    ['and'],
+  ])(`should produce immutable objects when combining matchers: %s`, async (combineFn) => {
+    const some = e.by.id('abc');
+    const other = e.by.id('def');
+
+    expect(some[combineFn](other)).not.toBe(some);
```
asafkorem, line 126: same, I'd check if `some` and `other` still equals to their original form

asafkorem (approved):
w00t! 🛠️  some small comment regarding the tests, except for that, LGTM

detox/src/android/AndroidExpect.test.js
asafkorem: also you can separate this to different tests like you did with the `it.each` in `expectTwo.test.js`

noomorph: Thanks, fixed.

asafkorem:
🥇

PR #3757, opened 2022-12-13
Website (homepage/showcase): Update background colors

moutro:
Update Dark mode and light mode to match original designs for:

- UA banner
- homepage subheader
- section containers
- hero banner

Add border-bottom to hero banner, update footer border-top.

cc: @asafkorem

### DIFF
--- a/website/src/components/CustomBanner/StandWithUkraine.module.css
+++ b/website/src/components/CustomBanner/StandWithUkraine.module.css
@@ -2,7 +2,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  background-color: #1f243c;
+  background-color: #1f253c;
   padding: 0.5em;
 }
 
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -45,10 +45,6 @@ html[data-theme='dark'] .header-github-link:before {
   border-top: 1px solid #f2eeeee8;
 }
 
-/* html[data-theme='dark']:root {
-  --ifm-background-color: #12161b;
-} */
-
 [data-theme='dark'] .footer {
   background-color: #12161b;
   border-top: 1px solid #202429;
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -44,13 +44,13 @@ html[data-theme='dark'] .header-github-link:before {
   border-top: 1px solid #f2eeeee8;
 }
 
-html[data-theme='dark'] {
-  --ifm-background-color: #202429;
+html[data-theme='dark']:root {
+  --ifm-background-color: #12161b;
 }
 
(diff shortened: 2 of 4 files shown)


asafkorem:
Thanks @moutro, that's amazing!! I will review it soon 🤩

website/src/components/CustomBanner/StandWithUkraine.module.css
```diff
@@ -2,7 +2,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  background-color: #1f243c;
+  background-color: #1f253c;
```
asafkorem, line 5: Did they also ask to change this color?

website/src/css/custom.css
```diff
@@ -44,13 +44,13 @@ html[data-theme='dark'] .header-github-link:before {
   border-top: 1px solid #f2eeeee8;
 }
 
-html[data-theme='dark'] {
-  --ifm-background-color: #202429;
+html[data-theme='dark']:root {
+  --ifm-background-color: #12161b;
```
asafkorem, line 48: Why do we need the `:root` selector here?

asafkorem (approved):
Looks good to me! 🙌🏼 
@moutro please answer these two questions before I merge the PR

website/src/css/custom.css
noomorph: @asafkorem is right. Just use either html or :root

noomorph: Btw, you can merge with line 17.

website/src/components/CustomBanner/StandWithUkraine.module.css
moutro: They included it in the figma image for style matching. When inspecting, this is the hex showing - very minor but it does seem to be just a bit darker IMO. I can revert it if you prefer the previous. Please let me know!

asafkorem: Dont revert, that's great :+1:

website/src/css/custom.css
asafkorem: `:root` has higher specification level than `html`, so there might be something that overrides the `html` version.. 🤔

asafkorem: Actually maybe `html` selector doesn't work with JSX, because JSX is JavaScript **XML**..

moutro: I'm not 100% sure but I think an `html` selector is not directly relevant to JSX and agree with what you said above.

moutro: My apologies for the extra commit - had a technical issue and pushed 2nd commit without removing commented code. Fixed in latest commit.

Issue #3745, opened 2022-12-05
Detox: Add packager command support


noomorph:
### Describe your idea

Currently, you cannot have all the necessary information inside the Detox config to be able to run both the debug and release configurations.

For example, to be able to run tests on debug builds of React Native apps, you have to run the packager command together with the `detox test` command, i.e.:

```
npx react-native start ... & detox test ...
```

My suggestion is to support a new property for any app config called `packager`, `start`, `background` or something like that:

```json
{
  "type": "android.apk",
  "binaryPath": "...",
  "build": "./gradlew ...",
  "start": [
    "react-native start ...",
    "node myMockServer.js -p 3000",
  ]
}
```

That will allow us to keep the entire config as a part of the Detox preset and remove this knowledge from the wrapper tools such as Shono.

The suggested commands would be:

```bash
detox test -c ... # starts the commands by default
detox test -c ... --start=force # starts the tests even if some of the start commands fail
detox test -c ... --no-start # omits starting the commands
detox start -c ... # only starts the commands 
detox start -c ... --force # starts even if some of the commands fail
```

professorkolik:
That's supported, whenever you ran build, packager will start automatically, I'm confident about iOS though.
you can create shell script. and define all your commands there

noomorph:
@professorkolik this feature starts making more sense when you're using enterprise presets and you wish to keep "start" and "build" scripts together for better maintainability. Also, this proposal concerns "detox test" execution – the test+autostart feature should help users to run debug builds on CI as easily as the release ones.

PR #3766, opened 2022-12-19
feat(cli): detox start, detox test --start

noomorph:
## Description

- This pull request addresses the issue described here: #3745

In this pull request, I have implemented the following:

1. A new app config property:
```diff
 {
   type: '...',
   binaryPath: '...',
   build: '...',
+  start: 'react-native start', // ← for example
 }
```
2. A new CLI command: `detox start`:
    ```sh
    detox start -c ios.sim.debug # runs RN packager
    ```
3. A new modifier for the existing CLI command: `detox test ... --start=???`:
    ```sh
    detox test -c ios.sim.debug # runs start command if exists
    detox test -c ios.sim.debug --no-start # disables start command
    detox test -c ios.sim.debug --start=force # runs tests even if the start command fails
    ```

---

 - [x] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files. (TODO)

> _For API changes:_


detox/__tests__/setupJest.js
```diff
@@ -1,3 +1,3 @@
 jest.mock('proper-lockfile');
-jest.mock('signal-exit');
+jest.mock('signal-exit', () => jest.fn(() => () => {}));
```
noomorph, line 2: `execa` uses `signal-exit` inside, so I am providing a correct signature

detox/local-cli/__mocks__/executable
```diff
@@ -4,22 +4,32 @@ const fs = require('fs');
 
 const DETOX_REGEXP = /^DETOX_(?!DISABLE_POD_INSTALL|DISABLE_POSTINSTALL)/;
 
-const env = Object.keys(process.env)
-  .filter(k => DETOX_REGEXP.test(k.toUpperCase()))
-  .reduce((acc, key) => {
-    acc[key] = process.env[key];
-    return acc;
-  }, {});
+function main() {
```
noomorph, line 7: Modernized this mock executable so it can go to "sleep" – that helps to test a few extra cases.

detox/local-cli/cli.js
```diff
@@ -10,7 +10,7 @@ const DetoxError = require('../src/errors/DetoxError');
 yargs
   .scriptName('detox')
   .parserConfiguration({
-    'boolean-negation': false,
+    'boolean-negation': true,
```
noomorph, line 13: Makes it possible to pass `--no-start` CLI option.

detox/local-cli/testCommand/TestRunnerCommand.js
```diff
@@ -37,6 +42,15 @@ class TestRunnerCommand {
     let runsLeft = 1 + this._retries;
     let launchError = null;
 
+    if (this._startCommands.length > 0) {
+      try {
+        await Promise.race([sleep(1000), ...this._startCommands.map(cmd => cmd.execute())]);
```
noomorph, line 47: Debatable place. I decide here to wait for an extra 1 second to make sure that the "start" process does not exit early. WDYT? Any better ideas?

detox/src/configuration/collectCliConfig.js
```diff
@@ -55,6 +55,7 @@ function collectCliConfig({ argv }) {
     useCustomLogger: asBoolean(get('use-custom-logger')),
     retries: asNumber(get('retries')),
     inspectBrk: asBoolean(get('inspect-brk')),
+    start: get('start'),
```
noomorph, line 58: I was actually wondering if it makes sense to add this setting to the behavior config. So that users can disable start commands.


But I decided not to. If this turns out to be a requested feature, we can add it later.

detox/__tests__/helpers.js
```diff
@@ -25,7 +25,7 @@ function callCli(modulePath, cmd) {
     return yargs
       .scriptName('detox')
       .parserConfiguration({
-        'boolean-negation': false,
+        'boolean-negation': true,
```
noomorph, line 28: Needed to support `--no-start` variant of CLI command in `detox test`.

detox/local-cli/testCommand/builder.js
```diff
@@ -28,6 +28,11 @@ module.exports = {
     describe: 'Reuse existing installed app (do not delete + reinstall) for a faster run.',
     boolean: true,
   },
+  start: {
+    group: 'Execution:',
+    describe: 'Run app "start" scripts before running the tests. Use --no-start to disable that, and --start=force to ignore errors.',
+    default: true,
```
noomorph, line 34: Debatable. Should the `--start` option be `true` by default?

asafkorem:
Thinking of this API from a user perspective, it makes so much sense to define all necessary preparation actions for running Detox tests, from the same place where we define the `build` command. I LOVE IT 👑 

Are there any use-cases where defining what should happen on detox test completion would also make sense to define under the app config?

detox/local-cli/testCommand/builder.js
asafkorem: It should, IMO. If someone already defined the `start` property, so most-likely the `start` is necessary for the `test` command to work properly, and I can think of more cases where it will be more convenient to have it `true` by default than `false`.

asafkorem: Also, we have a CLI option to disable the automatic running of `start`. Maybe we should also introduce another property for that

```diff
 {
   type: '...',
   binaryPath: '...',
   build: '...',
   start: 'react-native start', // ← for example
+  skipStartOnTest: true
 }
```

asafkorem: (or just `skipStart: ...`)

asafkorem: not sure if the last suggestion is really necessary, up to you

asafkorem:
I think we should add 2 buildkite jobs for testing the start command and for general sanity with both iOS-debug and Android-debug builds (using demo-react-native example project).

PR #3776, opened 2022-12-23
lint(tests): fix trivial errors

noomorph:
Green CI is the best judge here. So far this PR does not need a special review.


detox/src/DetoxWorker.test.js
```diff
@@ -565,55 +565,6 @@ describe('DetoxWorker', () => {
     });
   });
 
-  // TODO: move it to the root realm
-  describe.skip('global context', () => {
```
noomorph, line 569: Already covered by Detox context tests.

PR #3778, opened 2022-12-23
fix(generation): invalid isArray assertion

noomorph:
## Description

I noticed thanks to the linter that our assertion here was never correct:

```diff
-if (typeof X !== 'object' || !X instanceof Array) {
+if (typeof X !== 'object' || !Array.isArray(X)) {
```

That's because `!X` gets evaluated as a boolean first, and then it is never an instance of an array, for an obvious reason.

### DIFF
--- a/detox/src/android/AndroidExpect.test.js
+++ b/detox/src/android/AndroidExpect.test.js
@@ -515,6 +515,15 @@ describe('AndroidExpect', () => {
         await e.web.element(e.by.web.tag('tag')).runScriptWithArgs(script, argsArr);
       });
 
+      it('runScriptWithArgs (unhappy paths)', async () => {
+        const script = 'function bar(a,b) {}';
+        const someElement = e.web.element(e.by.web.id('id'));
+
+        await expect(someElement.runScriptWithArgs(script, 42)).rejects.toThrowError(/args must be an array/);
+        await expect(someElement.runScriptWithArgs(script, null)).rejects.toThrowError(/args must be an array/);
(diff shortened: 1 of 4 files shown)


detox/src/android/espressoapi/EspressoDetox.js
```diff
@@ -52,7 +52,7 @@ class EspressoDetox {
   }
 
   static setURLBlacklist(urls) {
-    if (typeof urls !== 'object' || !urls instanceof Array) {
+    if (typeof urls !== 'object' || !Array.isArray(urls)) {
```
noomorph, line 55: I don't see that it is tested anywhere currently, so I didn't add an extra test here, but anyway the `runScriptWithArgs` covers it, so I hope it's okay.

detox/src/android/AndroidExpect.test.js
```diff
@@ -515,6 +515,15 @@ describe('AndroidExpect', () => {
         await e.web.element(e.by.web.tag('tag')).runScriptWithArgs(script, argsArr);
       });
 
+      it('runScriptWithArgs (unhappy paths)', async () => {
+        const script = 'function bar(a,b) {}';
+        const someElement = e.web.element(e.by.web.id('id'));
+
+        await expect(someElement.runScriptWithArgs(script, 42)).rejects.toThrowError(/args must be an array/);
+        await expect(someElement.runScriptWithArgs(script, null)).rejects.toThrowError(/args must be an array/);
```
noomorph, line 523: This one and the next assertion would fail without the fix.

PR #3783, opened 2022-12-27
fix: access to detox/internals from sandbox

noomorph:
## Description

Fixes an issue when accessing `require('detox/internals').whateverProperty` throws you `Cannot read ... of undefined`.

That happens only in sandboxed context (i.e. Jest test files themselves), but it is still rather a bug.

We don't recommend using internals from tests but this is not something forbidden technically.

For example, we ourselves would want to have that available, e.g. be able to get the unique `require('detox/internals').session.id` to integrate correctly with our new screenshot comparison service.


detox/package.json
```diff
@@ -123,6 +123,8 @@
       "__tests__",
       ".test.js$",
       ".mock.js$",
+      "index.js",
```
noomorph, line 126: The discussed logic, i.e. "use cached globals VS new require()", has zero meaning on unit tests level.

We cover it on integration and e2e tests level instead which test THE INTENT of this pull request, not the obvious implementation details.

detox/src/DetoxWorker.js
```diff
@@ -5,7 +5,7 @@ const Client = require('./client/Client');
 const environmentFactory = require('./environmentFactory');
 const { DetoxRuntimeErrorComposer } = require('./errors');
 const { InvocationManager } = require('./invoke');
-const symbols = require('./symbols');
+const symbols = require('./realms/symbols');
```
noomorph, line 8: Since "symbols" module is actually very sensitive to the realms (primary / secondary / sandbox ), I thought it is a good idea to keep it together.

detox/src/realms/DetoxContext.js
```diff
@@ -135,7 +135,10 @@ class DetoxContext {
    */
   async [symbols.installWorker](opts) {
     if (opts.global) {
-      opts.global['__detox__'] = this;
+      opts.global['__detox__'] = {
```
noomorph, line 138: One more interesting place.

Here I could use `this` and `new DetoxInternalsFacade(this)` instead of `require()`-ing both the facades, but, in my opinion, the latter approach is more idomatic. It just eliminates any chance that those instances could be non strict-equal.

I'd prefer to have a **total guarantee** that:

```js
require('detox') === global['__detox__'].clientApi
require('detox/internals') === global['__detox__'].internalsApi
```

detox/test/e2e/27.retries.test.js
```diff
@@ -9,6 +12,10 @@ describe('jest.retryTimes() support', () => {
   let counter = 3;
 
   beforeAll(async () => {
+    // This test won't work if you retry it via -R, --retries.
+    // Here we also assert that the session object is accessible from the sandbox.
+    jestExpect(session.testSessionIndex).toBe(0);
```
noomorph, line 17: Since this end-to-end test will break **anyway** if it gets retried via `-R, --retries`, I decided to add one more innocent safeguard to make sure that the subject of this pull request doesn't lose test coverage easily.

The core test for that belongs to integration test suite anyway.

detox/internals.js
```diff
@@ -1,4 +1,11 @@
...
+function create() {
+  const context = require('./index');
+  const DetoxInternalsFacade = require('./src/realms/DetoxInternalsFacade');
 
-module.exports = new DetoxInternalsFacade(realm);
+  return new DetoxInternalsFacade(context);
+}
+
+/** @type {DetoxInternals.Facade} */
+module.exports = global['__detox__']
```
noomorph, line 9: Basically it repeat the structure of the adjacent `index.js` file. I hope it makes it even easier to comprehend that way.

detox/test/integration/e2e/flaky.test.js
```diff
@@ -1,4 +1,6 @@
-describe('Stub1', () => {
+const { session } = require('detox/internals');
+
+describe('Flaky', () => {
```
noomorph, line 3: It is an important transition. Instead of having multiple `stub*.test.js` with an unclear intent, I divide them into two categories: **flaky** and **passing** tests.

That gives me a firmer ground: the pseudo-flaky test eventually turns 🟢 PASSING, which means that I can test both a happy and an unhappy scenario while expecting the test command (`detox test ...`) to exit with code = 0.

This way the integration tests become stricter and safer.

detox/test/integration/e2e/flaky.test.js
```diff
@@ -13,12 +15,10 @@ describe('Stub1', () => {
     } finally {
       detox.trace.endSection('Asserting various texts');
     }
-    throw new Error(`I'm only here to make things interesting!`);
-  });
 
-  it.skip('Skipped test', async () => {
-    // Checking that skipped tests are also traced
+    if (session.testSessionIndex === 0) {
```
noomorph, line 19: This serves as an extra check for `testSessionIndex` logic. If it does not increment in the consequent `-R, --retries` test runner calls via CLI, this test is going to fail and get the integration test suite red.

detox/test/integration/e2e/passing-sandbox.js
```diff
@@ -0,0 +1,9 @@
+const { config, session } = require('detox/internals');
+const { default: expect } = require('expect');
+
+describe('Sandbox', () => {
+  test('detox/internals should be accessible from the sandbox', () => {
+    expect(session.testSessionIndex).toBe(0);
```
noomorph, line 6: This is a rather basic expectation set for the Detox Internals API being accessible from the Jest sandbox.

On the other hand, I don't oblige myself to make it thorough as the goal of this pull request was only to enable that sandbox access it self. And this test exactly does the job.

detox/test/integration/initialization-test.test.js
```diff
@@ -1,15 +1,16 @@
-const { execCommand } = require('./utils/exec');
+const execa = require('execa');
 
 describe('Initialization (context) tests', () => {
   test.each([
     `cross-env DETOX_CONFIGURATION=stub node integration/utils/simplistic-runner.js`,
 
-    `detox test -c stub --config integration/e2e/config.js --runInBand --retries 1 stub1 stub2`,
-    `cross-env DETOX_CONFIGURATION=stub jest --config integration/e2e/config.js --runInBand stub2 stub3`,
+    `detox test -c stub --config integration/e2e/config.js --runInBand --retries 1 flaky passing`,
```
noomorph, line 7: The test runner invocations with `--retries` include running a flaky tests.

The invocations without retries are limited to nominally passing tests.

This way we can be as strict as possible and demand the `exitCode` to be 0.

That improves the value of the `initialization-test.test.js` suite in CI.

detox/test/integration/utils/simplistic-runner.js
```diff
@@ -3,8 +3,8 @@ const detox = require('detox/internals');
 async function main() {
   try {
     await detox.init();
-    await test1().catch(e => console.error(e));
-    await test2().catch(e => console.error(e));
+    await test1();
```
noomorph, line 6: Both `test1` and `test2` have nothing bad or specific, so they are expected to pass.

I decided to stop shooting own foot, and follow the Occam's razor principle.

detox/test/integration/utils/simplistic-runner.js
```diff
@@ -18,4 +18,7 @@ async function test2() {
   await device.terminateApp();
 }
 
-main().catch((e) => console.error(e));
+main().catch((e) => {
+  console.error(e.stack || e.message || e);
+  process.exit(1);
```
noomorph, line 23: As this is expected to be a passing test suite, catching an error must make it exit with a non-zero code.

That way the integration tests can be made stricter.

PR #3785, opened 2022-12-28
fix: logger termination and detox init

noomorph:
## Description

In this pull request, I fix two things at once – unfortunately, it is impossible to split them without making the build red:

1. The `simplistic-runner` integration test:
  * now it exits with a non-zero code when it catches an error
  * due to a freshly revealed error, I fixed the `!=` typo (and make `!== null` check strict as it was initially intended by me in Detox 20).
2. The logging after calling `detox.cleanup()`. When I started debugging the issues in the integration test, it turned out that `console.error(e)` call was not producing **anything**. So I had to go to the logger and see why, and I found two issues:
  * the `BunyanLogger#installDebugStream` fails to **replace** the stream - instead it just adds one more stream
  * the `BunyanLogger#closeFileStreams` closes `tty` streams (stdin=0, stdout=1, stderr=2) without my actual intent because they are technically file streams and I don't check what `fd` (file descriptor is)

### DIFF
--- a/detox/src/logger/utils/BunyanLogger.js
+++ b/detox/src/logger/utils/BunyanLogger.js
@@ -41,13 +41,14 @@ class BunyanLogger {
       streamOptions.out = new PassThrough().pipe(process.stderr);
     }
 
-    this._debugStream = {
+    this._bunyan.addStream({
       type: 'raw',
       level: config.level,
       stream: bds.default(streamOptions),
-    };
+    });
 
-    this._bunyan.addStream(this._debugStream);
+    this._debugStream = _.last(this._bunyan['streams']);
+
     return this;
   }
 
@@ -93,7 +94,7 @@ class BunyanLogger {
           : bunyanStream.stream;
 
         /* istanbul ignore next */
-        return stream.fd !== undefined && !stream.closed;
+        return stream.fd > 2 && !stream.closed;
     }
   };
 
--- a/detox/src/realms/DetoxPrimaryContext.js
+++ b/detox/src/realms/DetoxPrimaryContext.js
@@ -135,7 +135,7 @@ class DetoxPrimaryContext extends DetoxContext {
     process.env.DETOX_CONFIG_SNAPSHOT_PATH = this[_sessionFile];
     this[_lifecycleLogger].trace(`Serialized the session state at: ${this[_sessionFile]}`);
 
-    if (opts.workerId != null) {
+    if (opts.workerId !== null) {
       await this[symbols.installWorker](opts);
     }
   }
--- a/detox/src/realms/DetoxPrimaryContext.test.js
+++ b/detox/src/realms/DetoxPrimaryContext.test.js
@@ -66,6 +66,7 @@ describe('DetoxPrimaryContext', () => {
   const detoxWorker = () => latestInstanceOf(DetoxWorker);
   const logFinalizer = () => latestInstanceOf(logger.DetoxLogFinalizer);
   const getSignalHandler = () => lastCallTo(signalExit)[FIRST_ARGUMENT];
+  const facadeInit = () => facade.init({ workerId: null });
 
   backupProcessEnv();
 
@@ -102,7 +103,7 @@ describe('DetoxPrimaryContext', () => {
     it('should create an IPC server with a valid session state', async () => {
       const expectedIPCServerName = `primary-${process.pid}`;
 
-      await facade.init();
+      await facadeInit();
 
       expect(IPCServer).toHaveBeenCalledWith(expect.objectContaining({
         sessionState: expect.objectContaining({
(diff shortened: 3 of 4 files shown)


detox/src/logger/utils/BunyanLogger.js
```diff
@@ -41,13 +41,14 @@ class BunyanLogger {
       streamOptions.out = new PassThrough().pipe(process.stderr);
     }
 
-    this._debugStream = {
+    this._bunyan.addStream({
```
noomorph, line 44: Turned out, that `addStream` does not add the given object as-is. That's why you can't use it as a reference to remove it from the internal `_streams` array of the Bunyan logger.

detox/src/logger/utils/BunyanLogger.js
```diff
@@ -93,7 +94,7 @@ class BunyanLogger {
           : bunyanStream.stream;
 
         /* istanbul ignore next */
-        return stream.fd !== undefined && !stream.closed;
+        return stream.fd > 2 && !stream.closed;
```
noomorph, line 97: This way I check that the given `bunyan` stream is actually a FILE stream, not TTY stream (where fd can be 0, 1, 2).

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -135,7 +135,7 @@ class DetoxPrimaryContext extends DetoxContext {
     process.env.DETOX_CONFIG_SNAPSHOT_PATH = this[_sessionFile];
     this[_lifecycleLogger].trace(`Serialized the session state at: ${this[_sessionFile]}`);
 
-    if (opts.workerId != null) {
+    if (opts.workerId !== null) {
```
noomorph, line 138: It was intended to be so from the beginning. This is why you can see, for example, in `detox/local-cli/test.js` the safety measure:

```js
await detox.init({ workerId: null });
```

Thanks to the freshly fixed integration test I found this issue.

detox/test/integration/utils/simplistic-runner.js
```diff
@@ -11,11 +12,16 @@ async function main() {
 }
 
 async function test1() {
+  log.info('Test 1');
```
noomorph, line 15: These log checks help to verify that the logger works as intended when Detox content is active.

detox/test/integration/utils/simplistic-runner.js
```diff
@@ -11,11 +12,16 @@ async function main() {
...
 }
 
 async function test2() {
+  log.info('Test 2');
   await device.terminateApp();
 }
 
-main().catch((e) => console.error(e));
+main().catch((e) => {
+  console.error(e);
```
noomorph, line 25: Without the fix to the `BunyanLogger` class, the `console.error` just does not print anything. This is why I got interested and found the reason and fixed it.

detox/test/integration/utils/simplistic-runner.js
```diff
@@ -11,11 +12,16 @@ async function main() {
...
+  process.exit(1);
```
noomorph, line 26: Adding a non-zero exit code logic makes this test much stricter and useful. This helped me to find that the tests inside this file were actually crashing. That's how I fixed `!== null` check.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -194,12 +195,18 @@ describe('DetoxPrimaryContext', () => {
...
 
     it('should export context-shared file via DETOX_CONFIG_SNAPSHOT_PATH', async () => {
-      await facade.init();
+      await facadeInit();
 
       expect(process.env.DETOX_CONFIG_SNAPSHOT_PATH).toBeDefined();
       expect(process.env.DETOX_CONFIG_SNAPSHOT_PATH).toMatch(TEMP_FILE_REGEXP);
     });
 
+    it('should install a worker if called without options', async () => {
```
noomorph, line 204: Added an extra test for the most default call, i.e. `detox.init()`.

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -66,6 +66,7 @@ describe('DetoxPrimaryContext', () => {
   const detoxWorker = () => latestInstanceOf(DetoxWorker);
   const logFinalizer = () => latestInstanceOf(logger.DetoxLogFinalizer);
   const getSignalHandler = () => lastCallTo(signalExit)[FIRST_ARGUMENT];
+  const facadeInit = () => facade.init({ workerId: null });
```
noomorph, line 69: To make sure I don't need to go over each test and think about it, and since the presumption was the most basic init does not install the worker, I added an alias function, `facadeInit` which guarantees that the worker is not spawned.

Issue #3789, opened 2022-12-28
(WIP) Support setDatePickerDate for android as well


mauricedoepke:
### Describe your idea

The method is currently ios only, but it would be nice to make it work for android as well.

API wise I think it would be nice, to stay somewhat compatible to the ios variant. Which looks like this:
```
await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "ISO8601");
await element(by.id('datePicker')).setDatePickerDate('2019/02/06', "yyyy/MM/dd");
```
I suggest to support the first variant with the ISO8601 format as this is easy to parse and throw an error for the second as it does not really make sense to reimplement the exact behaviour of [NSDateFormatter](https://developer.apple.com/documentation/foundation/dateformatter). 

I did already do some research and it seems to me that this is quite doable. Apparently Detox uses espresso on Android.
And espresso supports updating date pickers: https://developer.android.com/reference/androidx/test/espresso/contrib/PickerActions

After some glance through the detox code, I think those files need changes:
- https://github.com/wix/Detox/blob/1dca8733b2446e66a386b0d522809cb156d36855/generation/index.js
This one apparently autogenerates some js bindings. I think one can use the DetoxAction and DetoxViewActions part as an example to add `androidx.test.espresso.contrib.PickerActions`

 - https://github.com/wix/Detox/blob/b1bd601e6d7407b7f1866e67542f49c31cfa59ea/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
 This file seems to be some kind of wrapper for all actions, so I guess the DatePicker action needs to be added here as well.
 
 - https://github.com/wix/Detox/blob/20034e5e6fef1c3bdb145311d7c9d1c36f71ec40/detox/src/android/actions/native.js
 - https://github.com/wix/Detox/blob/9824d4be23749e76d3e153ec17377c781618d6a5/detox/src/android/core/NativeElement.js
 Theese two seem to be yet another wrapper layer
 
 - https://github.com/wix/Detox/tree/b1bd601e6d7407b7f1866e67542f49c31cfa59ea/detox/android/detox/src/full/java/com/wix/detox/espresso
 We might need a PickerActions.java here similar to DetoxAction.java or DetoxViewActions.jave
 
 
 I'd be willing to invest some time into this as this feature would help me, but I would definitely need help from the detox core team.
 Are you interested in having this feature added? Did I roughly identify the right files to change? Are there any obivious ones missing in my list?

[...]

PR #3790, opened 2022-12-29
feat(android): support setDatePickerDate

mauricedoepke:
## Description
- This pull request addresses the issue described here: #3789 

In this pull request, I have extended the setDatePickerDate action to work on android as well.

setDatePickerDate is already working for android in this pr and I also updated one relevant test.
I have yet to investigate whether more tests need changes. 
I also have yet to update the docs.

Feel free to comment on how to improve this implementation.

> _For API changes:_

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
@@ -19,6 +19,7 @@ import com.wix.detox.espresso.scroll.ScrollHelper;
 import com.wix.detox.espresso.scroll.SwipeHelper;
 
 import org.hamcrest.Matcher;
+import java.time.ZonedDateTime;
 
 import androidx.test.espresso.UiController;
 import androidx.test.espresso.ViewAction;
@@ -31,6 +32,7 @@ import androidx.test.espresso.contrib.PickerActions;
 import static androidx.test.espresso.action.ViewActions.actionWithAssertions;
 import static androidx.test.espresso.matcher.ViewMatchers.isAssignableFrom;
 import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
+
 import static org.hamcrest.Matchers.allOf;
 
 
@@ -150,8 +152,9 @@ public class DetoxAction {
         return new ScrollToIndexAction(index);
     }
 
-    public static ViewAction setDatePickerDate(int year, int monthOfYear, int dayOfMonth) {
-        return PickerActions.setDate(year, monthOfYear, dayOfMonth);
+    public static ViewAction setDatePickerDate(String dateString) {
+        ZonedDateTime date = ZonedDateTime.parse(dateString);
+        return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
     }
 
     public static ViewAction adjustSliderToPosition(final double newPosition) {
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
@@ -26,6 +26,7 @@ import androidx.test.espresso.action.CoordinatesProvider;
 import androidx.test.espresso.action.GeneralClickAction;
 import androidx.test.espresso.action.GeneralLocation;
 import androidx.test.espresso.action.Press;
+import androidx.test.espresso.contrib.PickerActions;
 
 import static androidx.test.espresso.action.ViewActions.actionWithAssertions;
 import static androidx.test.espresso.matcher.ViewMatchers.isAssignableFrom;
@@ -147,6 +148,10 @@ public class DetoxAction {
 
     public static ViewAction scrollToIndex(int index) {
         return new ScrollToIndexAction(index);
+    }
+
+    public static ViewAction setDatePickerDate(int year, int monthOfYear, int dayOfMonth) {
+        return PickerActions.setDate(year, monthOfYear, dayOfMonth);
     }
 
     public static ViewAction adjustSliderToPosition(final double newPosition) {
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
@@ -20,9 +20,11 @@ import com.wix.detox.espresso.scroll.ScrollHelper;
 import com.wix.detox.espresso.scroll.SwipeHelper;
 
 import org.hamcrest.Matcher;
-import java.time.format.DateTimeFormatter;
-import java.time.LocalDate;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.time.ZonedDateTime;
+import java.util.Calendar;
+import java.util.Date;
 
 import androidx.test.espresso.UiController;
 import androidx.test.espresso.ViewAction;
@@ -155,19 +157,19 @@ public class DetoxAction {
         return new ScrollToIndexAction(index);
     }
 
-    public static ViewAction setDatePickerDate(String dateString, String formatString) {
-        if (Build.VERSION.SDK_INT < 26) {
-            throw new AssertionError("To use DatePicker on Android you must use API 26 or above");
-        }
-
+    public static ViewAction setDatePickerDate(String dateString, String formatString) throws ParseException {
         if (formatString.equals("ISO8601")) {
+            if (Build.VERSION.SDK_INT < 26) {
+                throw new AssertionError("To use DatePicker on Android you must use API 26 or above");
+            }
             ZonedDateTime date = ZonedDateTime.parse(dateString);
             return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
         } 
             
-        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(formatString);
-        LocalDate date = LocalDate.parse(dateString, formatter);
-        return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
+        Date date = new SimpleDateFormat(formatString).parse(dateString);
+        Calendar cal = Calendar.getInstance();
+        cal.setTime(date);
+        return PickerActions.setDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1, cal.get(Calendar.DAY_OF_MONTH));
     }
 
     public static ViewAction adjustSliderToPosition(final double newPosition) {
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
@@ -47,6 +47,8 @@ import static org.hamcrest.Matchers.allOf;
 
 public class DetoxAction {
     private static final String LOG_TAG = "detox";
+    private static final String ISO8601_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
+    private static final String ISO8601_FORMAT_NO_TZ = "yyyy-MM-dd'T'HH:mm:ss";
 
     private DetoxAction() {
         // static class
@@ -160,11 +162,11 @@ public class DetoxAction {
     public static ViewAction setDatePickerDate(String dateString, String formatString) throws ParseException {
         Date date;
         if (formatString.equals("ISO8601")) {
-            date = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ").parse(dateString);
+            date = parseDateISO8601(dateString);
         } else {
             date = new SimpleDateFormat(formatString).parse(dateString);
         }
-            
+
         Calendar cal = Calendar.getInstance();
         cal.setTime(date);
         return PickerActions.setDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1, cal.get(Calendar.DAY_OF_MONTH));
@@ -199,5 +201,13 @@ public class DetoxAction {
                 return (result == null ? null : result.asBase64String());
             }
         };
+    }
+
+    private static Date parseDateISO8601(String dateString) throws ParseException {
+        try {
+            return new SimpleDateFormat(ISO8601_FORMAT).parse(dateString);
+        } catch (ParseException e) {
+            return new SimpleDateFormat(ISO8601_FORMAT_NO_TZ).parse(dateString);
+        }
     }
 }
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
@@ -1,6 +1,7 @@
 package com.wix.detox.espresso;
 
 import android.view.View;
+import android.os.Build;
 
 import com.wix.detox.common.DetoxErrors.DetoxRuntimeException;
 import com.wix.detox.common.DetoxErrors.StaleActionException;
@@ -19,6 +20,8 @@ import com.wix.detox.espresso.scroll.ScrollHelper;
 import com.wix.detox.espresso.scroll.SwipeHelper;
 
 import org.hamcrest.Matcher;
+import java.time.format.DateTimeFormatter;
+import java.time.LocalDate;
 import java.time.ZonedDateTime;
 
 import androidx.test.espresso.UiController;
@@ -152,8 +155,18 @@ public class DetoxAction {
         return new ScrollToIndexAction(index);
     }
 
-    public static ViewAction setDatePickerDate(String dateString) {
-        ZonedDateTime date = ZonedDateTime.parse(dateString);
+    public static ViewAction setDatePickerDate(String dateString, String formatString) {
+        if (Build.VERSION.SDK_INT < 26) {
+            throw new AssertionError("To use DatePicker on Android you must use API 26 or above");
+        }
+
+        if (formatString.equals("ISO8601")) {
+            ZonedDateTime date = ZonedDateTime.parse(dateString);
+            return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
+        } 
+            
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(formatString);
+        LocalDate date = LocalDate.parse(dateString, formatter);
         return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
     }
 
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
@@ -158,15 +158,13 @@ public class DetoxAction {
     }
 
     public static ViewAction setDatePickerDate(String dateString, String formatString) throws ParseException {
+        Date date;
         if (formatString.equals("ISO8601")) {
-            if (Build.VERSION.SDK_INT < 26) {
-                throw new AssertionError("To use DatePicker on Android you must use API 26 or above");
-            }
-            ZonedDateTime date = ZonedDateTime.parse(dateString);
-            return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
-        } 
+            date = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ").parse(dateString);
+        } else {
+            date = new SimpleDateFormat(formatString).parse(dateString);
+        }
             
-        Date date = new SimpleDateFormat(formatString).parse(dateString);
         Calendar cal = Calendar.getInstance();
         cal.setTime(date);
         return PickerActions.setDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1, cal.get(Calendar.DAY_OF_MONTH));
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -1365,12 +1365,13 @@ declare global {
             setColumnToValue(column: number, value: string): Promise<void>;
 
             /**
-             * Sets the date of a date picker to a date generated from the provided string and date format. (iOS only)
+             * Sets the date of a date picker to a date generated from the provided string and date format.
              * @param dateString string representing a date in the supplied `dateFormat`
              * @param dateFormat format for the `dateString` supplied
              * @example
              * await expect(element(by.id('datePicker'))).toBeVisible();
-             * await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ");
+             * await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ"); //iOS example
+             * await element(by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "ISO8601"); //Android example
              */
             setDatePickerDate(dateString: string, dateFormat: string): Promise<void>;
 
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -1369,9 +1369,9 @@ declare global {
              * @param dateString string representing a date in the supplied `dateFormat`
              * @param dateFormat format for the `dateString` supplied
              * @example
-             * await expect(element(by.id('datePicker'))).toBeVisible();
-             * await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ"); //iOS example
-             * await element(by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "ISO8601"); //Android example
+             * await element(by.id('datePicker')).setDatePickerDate('2023-01-01T00:00:00Z', 'ISO8601');
+             * await element(by.id('datePicker')).setDatePickerDate(new Date().toISOString(), 'ISO8601');
+             * await element(by.id('datePicker')).setDatePickerDate('2023/01/01', "yyyy/MM/dd"); // iOS only
              */
             setDatePickerDate(dateString: string, dateFormat: string): Promise<void>;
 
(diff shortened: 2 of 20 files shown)


noomorph:
Looks good to me, at least at the first glance!

There is some missing unit test coverage, but I hope it is trivial enough to add it.

Could you post here a screenshot of the updated DatePicker screen?

If you need some help or have question, let us know.

mauricedoepke:
@noomorph 

I think my most recent PR should fix the unit test coverage and I also adjusted the docs to reflect setDatePickerDate working on android.

This is how the Android DatePicker screen looks like now. The ios one is unchanged.

detox/index.d.ts
```diff
@@ -1365,12 +1365,13 @@ declare global {
...
 
             /**
-             * Sets the date of a date picker to a date generated from the provided string and date format. (iOS only)
+             * Sets the date of a date picker to a date generated from the provided string and date format.
              * @param dateString string representing a date in the supplied `dateFormat`
              * @param dateFormat format for the `dateString` supplied
              * @example
              * await expect(element(by.id('datePicker'))).toBeVisible();
-             * await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ");
+             * await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ"); //iOS example
```
noomorph, line 1373: @mauricedoepke I wonder what example we could recommend that would work identically for both the platforms?

detox/src/android/AndroidExpect.test.js
```diff
@@ -237,6 +237,17 @@ describe('AndroidExpect', () => {
         await expectToThrow(() => e.element(e.by.id('ScrollView161')).scrollTo('noDirection'));
       });
 
+      it('should setDatePickerDate', async () => {
+        await e.element(e.by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', 'ISO8601');
+      });
+
+      it('should not setDatePickerDate given bad args', async () => {
+        await expectToThrow(() => e.element(e.by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', 'ISO8601'));
```
noomorph, line 245: Sorry I can't see what is invalid here comparing to the previous happy path... 🤔

docs/api/actions.md
```diff
@@ -226,20 +226,29 @@ await element(by.id('pickerView')).setColumnToValue(2, "Hello World");
...
 
 Sets the element’s date to the specified date string, parsed using the specified date format.
 
-The specified date string is converted by the system to an [`NSDate`](https://developer.apple.com/documentation/foundation/nsdate) object, using [`NSDateFormatter`](https://developer.apple.com/documentation/foundation/dateformatter) with the specified date format, or [`NSISO8601DateFormatter`](https://developer.apple.com/documentation/foundation/iso8601dateformatter) in case of ISO 8601 date strings. If you use JavaScript’s [Date.toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) or otherwise provide a valid ISO 8601 date string, set the date format to `"ISO8601"`, which is supported as a special case.
+In case of dateFormat set to `"ISO8601"`, the specified date string is converted using JavaScript’s `new Date(dateString)` on Android and [`NSISO8601DateFormatter`](https://developer.apple.com/documentation/foundation/iso8601dateformatter) on iOS. You can use [Date.toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) to format a JavaScript date accordingly.
```
noomorph, line 233: Thanks, I'll re-read your docs addendum tomorrow. Sorry, had no time today.

detox/src/android/AndroidExpect.test.js
mauricedoepke: Me neither. Looks like a copy and paste error. I remove this line.

I wonder if it makes sense to adjust the `expectToThrow ` in this case. Currently it optionally allows the callback to throw an error. It does not require it to do so. If it would, it would have complained in this case.

Does it make sense to adjust `expectToThrow ` to require an error to be thrown in a seperate pr?

detox/index.d.ts
mauricedoepke: How about this:
```
/**
  * Sets the date of a date picker to a date generated from the provided string and date format.
  * @param dateString string representing a date in the supplied `dateFormat`
  * @param dateFormat format for the `dateString` supplied
  * @example
  * const datePicker = device.getPlatform() === "ios" ? by.id('datePicker') : by.type('android.widget.DatePicker');
  * await element(datePicker).setDatePickerDate('2019-02-06T05:10:00-08:00', "ISO8601");
  */
```

This works both on android and ios. 

This still contains a case differentiation as @react-native-community/datetimepicker does not support testID's on Android yet. I will try to add get a pr into @react-native-community/datetimepicker that adds it thought. Then we can revisit this example here and unify it completely.

docs/api/actions.md
mauricedoepke: No hurry. Feel free to completely rewrite this if neccessary. I am not that good in writing docs.

detox/index.d.ts
mauricedoepke: I just started a PR on to enable testID's there. If I can get this merged, we can completely unify this example https://github.com/react-native-datetimepicker/datetimepicker/pull/705

detox/test/e2e/17.datePicker.test.js
```diff
@@ -19,9 +19,20 @@ describe(':ios: DatePicker', () => {
...
       await expect(element(by.id('utcDateLabel'))).toHaveText('Date (UTC): Feb 6th, 2019');
       await expect(element(by.id('utcTimeLabel'))).toHaveText('Time (UTC): 1:10 PM');
     });
+
+    it(':android: can select dates on a UIDatePicker', async () => {
+      await element(by.id('openAndroidDatePicker')).tap();
+
+      //rn-datepicker does not support testId's on android, so by.type is the only way to match the datepicker right now
+      //@see https://github.com/react-native-datetimepicker/datetimepicker#view-props-optional-ios-only
+      await element(by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "ISO8601");
```
noomorph, line 33: Actually I am wondering... 🤔 What result is supposed to be if I change a timezone to `+08:00` → 5th or 6th February?

detox/src/android/AndroidExpect.test.js
noomorph: Thanks, I created #3791 draft for that.

docs/api/actions.md
noomorph: Rewrote some parts.

detox/test/src/Screens/DatePickerScreen.js
```diff
@@ -1,6 +1,10 @@
 import moment from 'moment';
 import React, { Component } from 'react';
-import { Text, View, StyleSheet, DatePickerIOS } from 'react-native';
+import { Text, View, StyleSheet, Platform, Button } from 'react-native';
+import { default as DatePicker, DateTimePickerAndroid } from "@react-native-community/datetimepicker";
```
noomorph, line 4: @mauricedoepke I'm wondering 🧐 does their default `<DatePicker>` work on Android? What is the difference between `DatePicker` and `DateTimePickerAndroid`? 

Could you consider adding tests for both components if they are both implemented on Android?

detox/test/e2e/17.datePicker.test.js
noomorph, line 33: image

I see an error like that on CI:
https://buildkite.com/wix-mobile-oss/detox/builds/1871#018563a8-6f16-438e-ac69-4f1785fe8ec3/6-4104

detox/index.d.ts
noomorph: Hmm, I don't think that the way of getting a `datePicker` element matters here.
I mean that you should put ISO 8601 example first and maybe use a less verbose example for iOS like in the docs.

```js
await element(by.id('datePicker')).setDatePickerDate(new Date(2023, 0, 1).toISOString(), 'ISO8601');
await element(by.id('datePicker')).setDatePickerDate('2023/01/01', "yyyy/MM/dd"); // iOS only
```

mauricedoepke: Thanks, I changed it.

detox/test/src/Screens/DatePickerScreen.js
mauricedoepke: So how I understand it, it has to do with the IOS native DatePicker being a regular "component" that you can inline anywhere you want in your UI like a regular react component.

The native Android DatePicker however apparently is implemented as a dialog which is called by an imperative API by default.

DatePicker under the hood calls DateTimePickerAndroid on android, so I think it suffices if we just test one method, because we are testing detox here and not react-native-datetimepicker/datetimepicker

But I will try to adapt it so that both ios and android use DatePicker.

detox/test/e2e/17.datePicker.test.js
mauricedoepke: Weird. It worked on my machine. One possible explanation I can think of is, that detox's auto waiting might not work well together with how android updates the datepicker value.
This might cause the test to work on some systems, but not on others due to timing issues.

I will try to use a waitFor to verify this hypothesise.

mauricedoepke: If I understand it right, for iOS, the devices timezone determines this. So it would make sense to also let the devices timezone decide that on Android, I guess?

My current implementation for android uses the testers locale which is inconsistent.

We could try to use something like this:
```
import java.time.LocalDateTime;
LocalDateTime.parse(date)
```

To parse the date in java which should parse it using the devices timezone.

mauricedoepke: So after some research I think we need to use ZonedDateTime instead. But I am neither fluent in timezone handling nor in android programming.

I just pushed a commit that does what I think might be a solution. But please have a look at it and tell me what you think about it..

noomorph:
@mauricedoepke sure I'll resume tomorrow.

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
```diff
@@ -149,6 +152,11 @@ public static ViewAction scrollToIndex(int index) {
         return new ScrollToIndexAction(index);
     }
 
+    public static ViewAction setDatePickerDate(String dateString) {
+        ZonedDateTime date = ZonedDateTime.parse(dateString);
+        return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
+    }
```
asafkorem, line 158: As I said in the previous comment, I suggest to add a `dateFormat` / `pattern` parameter and to support the other case:
```java
ZonedDateTime date;
if (pattern.equals("ISO8601")) {
    date = ZonedDateTime.parse(dateString);
} else {
    DateTimeFormatter formatter = DateTimeFormatter.ofPattern(pattern);
    date = ZonedDateTime.parse(dateString, formatter);
}

return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
```

asafkorem (changes requested):
Thank you for your contribution to the project! Your code will be a valuable addition. We appreciate your support and for taking the time to open a pull request.

detox/test/e2e/17.datePicker.test.js
jonathanmos: My concern with this approach would be that we support api 21+, and `ZonedDateTime` was only introduced in api 26. I'm investigating whether we can use `coreLibraryDesugaring`, or whether it would be better to use another alternative

asafkorem: Great catch @jonathanmos, if this solution won't work, we can enable this feature only when running with API level 28 or higher. For lower API level, we can throw an exception to indicate that this feature is not available before API 28.

We should have the freedom to increase the minimum API level of our app as long as it is still compatible with the minimum supported version of React Native. However, since React Native currently supports API level 21 and above, we should not increase the minimum API level support in our app beyond that.

jonathanmos: While desugaring the aar does work, it forces the app to also apply desugaring. 

Let's put an assertion before ZonedDateTime that verifies that the api is 26+ and if not then throw an informative message. This will also require a small change to the documentation for the feature.

```
if (Build.VERSION.SDK_INT < 26) {
            throw new AssertionError("To use DatePicker on Android you must use API 26 or above");
}

File: detox/index.d.ts [hunk 7]
```diff
@@ -1365,12 +1365,13 @@ declare global {
...
-             * Sets the date of a date picker to a date generated from the provided string and date format. (iOS only)
+             * Sets the date of a date picker to a date generated from the provided string and date format.
              * @param dateString string representing a date in the supplied `dateFormat`
              * @param dateFormat format for the `dateString` supplied
              * @example
-             * await expect(element(by.id('datePicker'))).toBeVisible();
-             * await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', "yyyy-MM-dd'T'HH:mm:ssZZZZZ");
+             * await element(by.id('datePicker')).setDatePickerDate('2023-01-01T00:00:00Z', 'ISO8601');
+             * await element(by.id('datePicker')).setDatePickerDate(new Date().toISOString(), 'ISO8601');
+             * await element(by.id('datePicker')).setDatePickerDate('2023/01/01', "yyyy/MM/dd"); // iOS only
```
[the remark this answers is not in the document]
mauricedoepke: I close this, thread, as the next one discusses the same topic

detox/test/e2e/17.datePicker.test.js
mauricedoepke: Indeed got catch. I added your suggestion.

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
mauricedoepke: Thanks. I implemented this. I had to use LocalDate though for the pattern case as ZonedDateTime complains if it does not get a time or timezone.

detox/index.d.ts
mauricedoepke: @noomorph I adjusted the docs a bit, because a custom date format is now supported on android as well. You may want to check my changes.

mauricedoepke:
I adjusted the DateScrenn in the test app to use the devices local time instead of utc. I think it makes sense as setDatePickerDate parses the date in the local devices time as well. This way we don't have to do any conversions inside the tests.
I did this because my simulator and my mac were 1 hour apart and this made the tests break when using utc.

Screenshot_1672962756

detox/test/e2e/17.datePicker.test.js
noomorph: @jonathanmos @mauricedoepke 

1. In the current implementation, I see two different parsing strategies: `ZonedDateTime.parse(dateString)` and `LocalDate.parse(dateString, DateTimeFormatter.ofPattern(formatString))`. Are they **both unavailable** for API <26? If only one is unavailable, does it make sense to throw the assertion error in both cases? Maybe we could inform that only `ISO8601` is not supported on older APIs, WDYT? 🤔  
2. @jonathanmos, forgive me for unfamiliarity with the subject, but isn't the suggested runtime assertion `if (Build.VERSION.SDK_INT < 26) {` happening too late? I mean, this `DetoxAction.java` has an import of ZonedDateTime at the very top of the file. Don't we get an error right on an attempt to load that class? 🙏 Thanks in advance for the explanation.

mauricedoepke: I think we could try to use [simpledateformat](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html) for custom formats instead. This api is much older and should be supported by most devices

mauricedoepke: Looks like simpledateformat can support both custom and iso strings. I rewrote it. That should be supported from api level 1.

noomorph:
@mauricedoepke I reworked a couple of things:

1. Fixed the case on iOS: `new Date().toISOString()` – it turned out, iOS does not tolerate milliseconds, so I added ms cropping on the JS side.
2. I thought it would be better to add some variety of date pickers: sliders, calendars and changed the screen.
3. You used loops for generating tests – I changed that to `describe.each` DSL provided by Jest.

### iOS

| Simulator Screen Shot - iPhone 14 - 2023-01-06 at 19 35 28  | Simulator Screen Shot - iPhone 14 - 2023-01-06 at 19 35 24  | Simulator Screen Shot - iPhone 14 - 2023-01-06 at 19 35 18  |
|---|---|---|

### Android

https://user-images.githubusercontent.com/1962469/211065685-7535b3a1-be18-4b84-8405-2a8cd4819508.mp4

https://user-images.githubusercontent.com/1962469/211065704-4a74672a-e036-468a-9f84-07c448e1b04c.mp4

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
```diff
@@ -158,15 +158,13 @@ public static ViewAction scrollToIndex(int index) {
...
     public static ViewAction setDatePickerDate(String dateString, String formatString) throws ParseException {
+        Date date;
         if (formatString.equals("ISO8601")) {
-            if (Build.VERSION.SDK_INT < 26) {
-                throw new AssertionError("To use DatePicker on Android you must use API 26 or above");
-            }
-            ZonedDateTime date = ZonedDateTime.parse(dateString);
-            return PickerActions.setDate(date.getYear(), date.getMonthValue(), date.getDayOfMonth());
-        } 
+            date = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ").parse(dateString);
```
noomorph, line 163: Does it work with custom time zones? `+10:00` etc?

noomorph: Oh, I see you changed it again . Probably this is not relevant.

detox/test/e2e/17.datePicker.test.js
```diff
@@ -1,27 +1,49 @@
...
+      } else {
+        await element(by.id('openDatePicker')).tap();
+        //rn-datepicker does not support testId's on android, so by.type is the only way to match the datepicker right now
+        //@see https://github.com/react-native-datetimepicker/datetimepicker#view-props-optional-ios-only
+        await element(by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', 'ISO8601');
+        await element(by.text('OK')).tap();
 
-      if(failed === false) {
-        throw new Error('Test should have thrown an error, but did not');
+        await waitFor(element(by.id('utcDateLabel'))).toHaveText('Date (UTC): Feb 6th, 2019').withTimeout(3000);
```
mauricedoepke, line 37: @noomorph 
This does not test the custom pattern feature on android, does it?

Also what I learned during my local testing, when I ran the tests between 00:00 and 01:00 is that I think you have to check against:
 - utc for iso datestrings or  custom patterns that contain explicit timezones
 - local time for patterns that don't contain patterns.
 Otherwise tests can fail due to timezone errors.

The reason is when:
- using a iso string, you let the device parse the date with an explicit format. When you then display it in the local format and the local format differs from utc, you can have a day off error around midnight. When you hoever display it in utc, the results will be stable.
- using a pattern without timezone you parse it in local time. When you now display it in utc, the result can vary when the device has a non utc timezone. So in this case the result will only be stable when you also output it as local time.

As this is quite complex, I think it is easier to remove the loop, even though the resulting code duplication is not so nice.

This was my suggestion, you may have not seen it because we worked on similar times on this:
```
      it(':ios: can select dates on a UIDatePicker, format: ISO8601', async () => {
        await element(by.id('datePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', 'ISO8601');
        await expect(element(by.id('utcDateLabel'))).toHaveText('Date (UTC): Feb 6th, 2019');
        await expect(element(by.id('utcTimeLabel'))).toHaveText('Time (UTC): 1:10 PM');
      });

      it(':ios: can select dates on a UIDatePicker, format: yyyy/MM/dd HH:mm', async () => {
        await element(by.id('datePicker')).setDatePickerDate('2019/02/06 13:10', 'yyyy/MM/dd HH:mm');
        await expect(element(by.id('localDateLabel'))).toHaveText('Date (Local): Feb 6th, 2019');
        await expect(element(by.id('localTimeLabel'))).toHaveText('Time (Local): 1:10 PM');
      });
    
      it(':android: can select dates on a UIDatePicker, format: ISO8601', async () => {
        //rn-datepicker does not support testId's on android, so by.type is the only way to match the datepicker right now
        //@see https://github.com/react-native-datetimepicker/datetimepicker#view-props-optional-ios-only
        await element(by.type('android.widget.DatePicker')).setDatePickerDate('2019-02-06T05:10:00-08:00', 'ISO8601');
        await element(by.text('OK')).tap();

        await waitFor(element(by.id('utcDateLabel'))).toHaveText('Date (UTC): Feb 6th, 2019').withTimeout(3000);
      });

      it(':android: can select dates on a UIDatePicker, format: yyyy/MM/dd', async () => {
        await element(by.type('android.widget.DatePicker')).setDatePickerDate('2019/02/06', 'yyyy/MM/dd');
        await element(by.text('OK')).tap();

        await waitFor(element(by.id('localDateLabel'))).toHaveText('Date (Local): Feb 6th, 2019').withTimeout(3000);
      });
```

detox/test/e2e/17.datePicker.test.js
mauricedoepke: @jonathanmos I close this discussion now, as I think, using simpledateformat fixes all the here discussed issues. Feel free to open it again if you disagree.

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
mauricedoepke: As mentioned in another thread. This works now and is implemented with a simpledateformatter now.

mauricedoepke:
@jonathanmos @asafkorem @noomorph 

Can someone with deeper android knowledge have a look at this error?:
https://buildkite.com/wix-mobile-oss/detox/builds/1889#01858831-2d65-4c42-a6b1-8ecbace27c29

I have no idea what this means, how to debug this or if this is even related to this pr.

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java
noomorph: > I think the android implementation does not support ms either.

Well, it does not throw errors and it passes the tests. Do you think we need any extra action items in such case? 😕 ❓

jonathanmos:
> @jonathanmos @asafkorem @noomorph
> 
> Can someone with deeper android knowledge have a look at this error?: https://buildkite.com/wix-mobile-oss/detox/builds/1889#01858831-2d65-4c42-a6b1-8ecbace27c29
> 
> I have no idea what this means, how to debug this or if this is even related to this pr.

As discussed on Discord, could you remove the two androidTestImplementations of espresso-contrib and com.google.truth from examples/demo-pure-native-android/app/build.gradle? This will fix the error on buildkite.

detox/test/e2e/17.datePicker.test.js
```diff
@@ -1,8 +1,15 @@
 const jestExpect = require('expect').default;
 
 describe('DatePicker', () => {
+  let lastTestFailed = false;
+
   beforeEach(async () => {
-    await device.reloadReactNative();
+    if (lastTestFailed) {
```
d4vidi, line 7: I'm not happy about the implied nondeterministic behavior. What about changing this to `if (android) { ...` instead?
Apropos, `lastTestFailed` never gets reset back to `false` so its going to act more like a `anyTestFailed`, ATM.

noomorph: Okay, simplified it.

noomorph (changes requested):
So, from my side, I see only one remaining question:

**How much work is left to add time support (hh:mm:ss) for Android?** Maybe we could do it now and land the complete feature? 🙏 

As for the actual issues, it looks like the **simple date formatter cannot handle** `2023-01-11T10:41:26Z` for some reason, while it perfectly handles `2019-02-06T05:10:00-08:00`. The test fails with `Unparseable date: "2023-01-11T10:41:26Z"`.

noomorph:
@jonathanmos, I'm removing myself as an assignee for now – all JavaScript-related comments from my side are resolved at the moment.

Let's come up with the final resolution for that Zoned/Simple date formatter, and maybe let's cover that logic as a unit test in Java/Kotlin. Interesting that `2023-01-11T10:41:26Z` is parsed alright on my local machine, but it fails on CI. Maybe it is an API level issue, maybe not – let's locate it.

noomorph (approved):
Approving it all except for the `Test Failed: Unparseable date: "2023-01-11T10:41:26Z"` issue.

d4vidi:
Hi, I will try to also get into the loop of things here, sometime this week.

d4vidi:
I'm finally back into the loop of things - planning to high-prioritize this for the upcoming 2-week iteration.

@mauricedoepke It's starting to seem that by getting the `Test Failed: Unparseable date: "2023-01-11T10:41:26Z"` @noomorph has mentioned you could be crossing the last mile for getting this thing merged ✅ Do you think you could get around to it soon? I am likely to be here to help as needed, as I mentioned.

d4vidi:
Well hopefully I've fixed the case of ISO-8601 with no time-zone. I'm nonetheless not 100% satisfied with the documentation, so I'm inclined to hold back the merge a bit longer till this gets sorted out too.

detox/src/android/espressoapi/DetoxAction.js
```diff
@@ -233,6 +233,18 @@ class DetoxAction {
     };
   }
 
+  static parseDateISO8601(dateString) {
```
noomorph, line 236: @d4vidi looks like we need to remove `private` methods generation, right?

d4vidi: No, we have to switch to `invoke2` 😄

d4vidi:
Suggestion for a future rearchitecture: JS (node) should take on the single responsibility of converting the various formats to the unified time/date representation - which is in fact a unix time stamp. The JS will then deliver the various formats into the various native environments (android / iOS) in a unified way, and the native-side handling code will then be rendered super-simple.

PR #3804, opened 2023-01-03
fix(logger): complete events in sandbox

noomorph:
## Description

Consider this code in userbase:

```js
detox.log.trace.complete('Navigating to Sanity', async () => {
  detox.log.trace('Reloading app');
  detox.log.trace('Tap on Sanity');
  await element(by.text('Sanity')).tap();
  detox.log.trace('I am on Sanity screen');
});
```

### Expected result

Begin event, instant event x 3, End event
```
|B-- Navigating to Sanity --E|
|---i----------i---------i---|
```

### Actual result

```
|B-- E-----------------------|
|---i----------i---------i---|
```

### Analysis

The **sandboxed** `Promise` constructor fails the previously implemented `isPromise` check:

```
obj = [Promise*]
Promise.resolve(obj) !== obj
```

### Chosen solution

It is fair enough to expect any Promise-like object in such logger calls, so I switched the implementation to `isPromiseLike` with a duck typing check: if it has `.then` function, then probably it is a promise.

### Confirmation

| Before      | After |
| ----------- | ----------- |
|  |  |

#### Integration tests 

I added this case to the timeline snapshot test at the moment. For example, a regression would look like this:

image

#### Unit tests

1. Added tests for `isPromiseLike`.
2. Added an extra test for `DetoxLogger` which verifies it treats promise-like objects identically to the promises:

```js
const promiseLike = { then: jest.fn() };
const promiseLikeResult = l.fatal.complete('Fatal (pending)', promiseLike);
expect(promiseLikeResult).toBe(promiseLike);
// ...
expect.objectContaining({ msg: 'Fatal (pending)', ph: 'B', time, tid, level: 60 }),
```

### DIFF
--- a/detox/src/logger/__snapshots__/DetoxLogger.test.js.snap
+++ b/detox/src/logger/__snapshots__/DetoxLogger.test.js.snap
@@ -96,6 +96,7 @@ exports[`DetoxLogger - main functionality - should log complete duration events 
 00:00:00.000 detox[PID] E  Debug duration
 00:00:00.000 detox[PID] B  Trace duration
 00:00:00.000 detox[PID] E  Trace duration
+00:00:00.000 detox[PID] B  Fatal (pending)
 00:00:00.000 detox[PID] B  Error duration (sync)
 00:00:00.000 detox[PID] E  Error duration (sync)
   error: Error: Oops (sync)!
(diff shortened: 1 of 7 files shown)


detox/src/logger/__snapshots__/DetoxLogger.test.js.snap
```diff
@@ -96,6 +96,7 @@ exports[`DetoxLogger - main functionality - should log complete duration events
 00:00:00.000 detox[PID] E  Debug duration
 00:00:00.000 detox[PID] B  Trace duration
 00:00:00.000 detox[PID] E  Trace duration
+00:00:00.000 detox[PID] B  Fatal (pending)
```
noomorph, line 99: If it does not treat it as a promise, it would print erroneously the "E" (end) event for it.

PR #3834, opened 2023-01-11
chore: add reproduction required github action.

asafkorem:
Using https://github.com/peter-evans/create-or-update-comment action

### DIFF
--- a/.github/workflows/reproduction-required.yml
+++ b/.github/workflows/reproduction-required.yml
@@ -1,0 +1,27 @@
+name: Reproduction Required
+on:
+    issues:
+        types:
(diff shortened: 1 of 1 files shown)


noomorph (approved):
Looks good, but I'd suggest to ensure adding `user: more info needed` label if it does not exist.

d4vidi:
IIUC the [support requests action](https://github.com/wix/Detox/blob/master/.github/workflows/support.yml) could have reused here, couldn't it?

noomorph:
Yeah, maybe we could aggregate these actions into a single workflow. 🤔

PR #3847, opened 2023-01-17
[Website] Create a visual regression test suite

### LINKED ISSUE #3838: [Website] Create a visual regression test suite
## Description

- [x] Set up a test suite in the `website/` project using [Percy.io](https://percy.io).
- [ ] Create a corresponding step in [.github/workflows/website.html](https://github.com/wix/Detox/blob/master/.github/workflows/website.yml)

Promo-image

We could start from a simple set of test pages:

- home page
- documentation (demo page)
- blog page
- showcase page

Reportedly, the easiest way to start is **Percy Snapshots**:

https://docs.percy.io/docs/percy-snapshot

You can see a similar use case here:

* https://github.com/rohit-gohri/redocusaurus/tree/main/website
* https://github.com/rohit-gohri/redocusaurus/blob/main/website/package.json#L22
* https://github.com/rohit-gohri/redocusaurus/blob/main/website/snapshots.js

Although, I'd advise sticking with `snapshots.yml` (YAML format) just like the docs say.

moutro:
In this pull request, I have done the following:

- set up test suite in website project using Percy CLI
- create YAML test pages for homepage, showcase page, blog page and demo page
- light and dark modes

Resolves #3838.

image

image

NOTE:
Need to verify Percy.io dashboard functions correctly on image uploading. Some pages are not displaying correctly:
image

### DIFF
--- a/website/snapshots.yml
+++ b/website/snapshots.yml
@@ -1,0 +1,31 @@
+- name: Home page dark theme
+  url: https://wix.github.io/Detox/
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'dark')
+
+- name: Home page light theme
+  url: https://wix.github.io/Detox/
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'light')
+
+- name: Showcase page dark theme
+  url: https://wix.github.io/Detox/showcase
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'dark')
+
+- name: Showcase page light theme
+  url: https://wix.github.io/Detox/showcase
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'light')
+
+- name: Blog page dark theme
+  url: https://wix.github.io/Detox/blog
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'dark')
+
+- name: Blog page light theme
+  url: https://wix.github.io/Detox/blog
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'light')
+
+- name: Demo page dark theme
+  url: https://wix.github.io/Detox/docs/next/demo
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'dark')
+
+- name: Demo page light theme
+  url: https://wix.github.io/Detox/docs/next/demo
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'light')
--- a/website/snapshots.yml
+++ b/website/snapshots.yml
@@ -1,31 +1,23 @@
 - name: Home page dark theme
-  url: https://wix.github.io/Detox/
-  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'dark')
+  url: http://localhost:3000/Detox/
+  execute: document.documentElement.setAttribute("data-theme", "dark")
 
 - name: Home page light theme
-  url: https://wix.github.io/Detox/
-  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'light')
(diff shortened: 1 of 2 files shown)


website/snapshots.yml
```diff
@@ -0,0 +1,31 @@
+- name: Home page dark theme
+  url: https://wix.github.io/Detox/
+  execute: document.querySelector('div.toggle_vylO.colorModeToggle_DEke > button').click('data-theme', 'dark')
```
noomorph, line 3: There are three issues about this approach:

1. Button click function does not accept parameters. What you did was just a wild guess, and, fortunately for you, it was redundant (hence, ignored) – so that you did not notice that something was wrong.
2. `div.toggle_vylO.colorModeToggle_DEke` CSS selector is unreasonably specific. [Check out what is a specificity of a CSS selector.](https://specificity.keegan.st)
3. Even if you used only the really meaningful parts in that CSS selector, e.g.: `.toggle_vylO.colorModeToggle_DEke > button`, your problem is that `cssClasses_g1BBerIsh` are randomly suffixed, so the chances are that your click won't work after first change to website code or after updating a Docusaurus version.

noomorph: I again suggest to you to consider using an CSS attribute selector over the `aria-label` of that toggle button. Please re-read: https://www.w3schools.com/css/css_attribute_selectors.asp

website/snapshots.yml
```diff
@@ -0,0 +1,31 @@
+- name: Home page dark theme
+  url: https://wix.github.io/Detox/
```
noomorph, line 2: I would expect that you use `http://localhost:3000/Detox` instead. Please think again for yourself – what we are testing: local changes or the production version? If we assume we **always** want to test a production version, then how is this script going to **prevent** you from entering visual regressions before it's too late (so that you'd need to revert). Well, that's why it should be a local version. 🤷‍♂️

website/snapshots.yml
noomorph, line 3: AFAIR, the default theme is the dark one. According to the docs, `execute` script is run before capturing the snapshots. So, I assume you'll be storing on Percy service snapshots with the light theme called dark, and vice versa.

noomorph (changes requested):

website/snapshots.yml
moutro: @noomorph Thanks for the details regarding this approach. I have reviewed all and re-read the docs to understand better. 
Yes, instead of using a CSS selector, it would be better to use an attribute selector based on the aria-label of the toggle button. This would make the selector more robust and less likely to be affected by changes in the website's code. 

Decided to modify the approach to select the html attribute 'data-theme' to target dark and light modes due to a bug in Docusaurus on the toggle button.

moutro: Yes, I looked over the different themes to ensure dark > light were correctly coded.

website/snapshots.yml
moutro: After more research into it and reading your notes, I agree with your point and made the suggested change. Testing the local version will allow the user to catch visual regressions before they are pushed to production and avoid any need to revert changes later on (which is the point of doing this regression testing). Thanks for the explanation here!

moutro:
Leaving the Showcase page out of the regression testing plan in this PR. The Showcase page has dynamic data and the Percy dashboard editor does not allow sections to be omitted from future builds and diffs are misleading on the snapshots buiilds

asafkorem (approved):
LGTM

@noomorph FYI, we found a bug with docusaurus dark mode toggle.. if the page is opened on Light mode (because of caching), the aria-label still says it's on dark mode 🤦🏼‍♂️ 
Anyway, `document.documentElement.setAttribute("data-theme", ...)` is a better approach IMO. Because it is less-likely to change in the future.. also we don't need to use any `if button != null` check before calling the `click()` method.

noomorph:
As for the Showcase page, it is quite obvious – you should implement some `?shuffle=disabled` and don't shuffle if it is explicitly turned off in the query string. 🤷‍♂️  

The other solution is to use mocking in that way or another so that your data is always predictable but that's an overkill for now.

noomorph (approved):
Looks good enough. @moutro, @asafkorem, **please run builds sparingly (e.g. temporarily erase most of `snapshots.yml` and override the config with only 1 browser width) when your work is in progress.** We already have scorched 39% of free plan just by 3 days.

PR #3863, opened 2023-01-21
fix(jest): focused test with afterAll hook

noomorph:
## Description

Fixes issue:

```plain text
Invalid test summary status was passed to detox.beforeEach(testSummary). Valid values are: "running", "passed", "failed"
```

Minimal example to reproduce:

```js
describe('Focused', () => {
  afterAll(async () => {
    // Reproducing when hook_start is called after test_start
  });

  it.only('Only test', async () => {
    // Checking that skipped tests are also traced
  });

  it('Skipped test', async () => {
    // Causes issue in the previous implementation
  });
});
```

- This pull request addresses the issue reported internally inside Wix (thanks, @overengineered ).

In this pull request, I have added `test_skip` and `test_todo` listeners to `DetoxCoreListener` (dispatched decorator for `handleTestEvent` provided by `jest-circus`). This covers `it.only` case when `test.mode` is undefined in all the other tests (which is why `_isTestSkipped()` returns false). So, from now on, we'll maintain a `_skippedTests: Set<Test>` and add tests coming from `test_skip` and `test_todo` events. This change will make `_isTestSkipped()` more robust.


asafkorem:
Please approve that it is backward-compatible with previous Jest version (manual testing / add coverage)

asafkorem (approved):

noomorph:
Tested with Jest 27.x – also OK.

PR #3865, opened 2023-01-23
enable tests on RN 0.71

hawkrives:
> I think that there is an issue with Detox and RN 0.71, and my next step is to reproduce it in as close to the upstream testing environment as I can. I will open an issue for this if the test fails in the way that I expect that it will. I apologize for the PR noise!

## Description

- This pull request addresses the issue described here: (none)

In this pull request, I have enabled tests against React Native 0.71.

---

> _For API changes:_


hawkrives:
I will now close this. Evidently, my local testing was incomplete, as one of the 0.71 apps passed… 

Thank you all for Detox!

PR #3866, opened 2023-01-23
[Website]: Add GitHub Actions Step for Visual Regression Tests

### LINKED ISSUE #3838: [Website] Create a visual regression test suite

moutro:
In this pull request, I have:

- Created a GitHub actions step in the .github/workflows/website.yml file using an 'if' conditional for running Percy  regression test step and take snapshots if there are any changes in the 'website/src/css' files.
- This If condition uses the statndard way to check for changes in a specific file or folder in a GitHub Actions workflow as an array of file paths that were modified in the latest commit.

Resolves #3838 

cc: @asafkorem


moutro:
@noomorph @asafkorem - Split up the snapshots.yml into separate files for each page and left all inside the custom css file. Added the Showcase page this time with Yarik's recommendation to use `?shuffle=disabled` in the query string so a valid screenshot can be taken of all the cards when testing.

PR #3871, opened 2023-01-25
[Website]: Migrating CSS styles to SCSS format

### LINKED ISSUE #3845: [Website] Migrate the styles to SCSS
Currently, we use `website/styles/custom.css`.

We could split it (since it is considerably large already) into modularized `_name.scss` files and import back all of them in a `custom.scss` (mind the new extension :)) file.

Better to complete (and merge) after setting up the visual regression flow on CI (Github Actions + Percy).
https://github.com/wix/Detox/issues/3838

Reference: https://docusaurus.io/docs/styling-layout#sassscss

moutro:
## Description

Resolves #3845
 
In this pull request I have completed the following for review:

1. Installed `docusaurus-plugin-sass`
2. Include in the docusaurus.config.js file
3. Set customCss in docusaurus.config.js file, the property to point to the SCSS file
4. Renamed all css files including the custom.css file to .scss and update all component js files import statements to reference the new .scss files

Docusaurus doc for reference on steps: https://docusaurus.io/docs/styling-layout#sassscss

5. Inspected the network payloads for any changes and see the sizes (`styles.css` files) are still roughly the same 
       - live site is 26kb
       - testing site localhost is 32.7kb
6. Inspected pages (homepage, showcase page, docs, etc) and page styles look the same and loading correctly


package.json
```diff
@@ -54,5 +54,8 @@
     "shell-utils": "1.x.x",
     "unified": "^10.1.0"
   },
-  "version": "20.1.2"
+  "version": "20.1.2",
+  "dependencies": {
+    "docusaurus-plugin-sass": "^0.2.3"
```
noomorph, line 59: Why do we need it here at all in the root project?

noomorph (changes requested):
Overall, looks good. Please fix my comment and let's wait until the Percy is merged.

package.json
moutro: Thanks for the catch there. Removed and it's now in the website root package.json.

PR #3872, opened 2023-01-25
ci(github): visual test for Showcase

noomorph:
## Description

- [x] add new snapshots to YAML
- [x] implement `?shuffle=disabled` in the Showcase component


website/src/components/Showcase/Showcase.js
```diff
@@ -6,9 +6,9 @@ import styles from './Showcase.module.css';
...
 import { TailSpin } from 'react-loader-spinner';
 
-function Showcase() {
+const Showcase = () => {
   const [cards, setCards] = useState([]);
-  const [toRender, setToRender] = useState([]);
+  const [toRender, setToRender] = useState([<TailSpin />]);
 
   useEffect(() => {
     setCards(_makeShuffledCardList(cardList).map(_makeCard));
```
noomorph, line 14: I don't think that all those numerous changes were necessary. Consider this direction:

```suggestion
    const resultingCardList = shouldShuffle() ? _makeShuffledCardList(cardList) : cardList;
    setCards(cardList.map(_makeCard));
```

It is a bad idea to make `_makeShuffledCardList` so smart – despite a straightforward name, this function starts to decide whether it wants to shuffle or not. Besides, nothing implies that such a function would be DOM-dependent as its goal is just to shuffle the list.

Please go with a simpler version as suggested above. If there some things I didn't notice, let me know.

asafkorem: I agree with Yarik, the logic of whether we should call `_makeShuffledCardList` or not should be extracted from `_makeShuffledCardList` to another method, otherwise the method's name is wrong.

website/snapshots.yml
```diff
@@ -14,10 +22,11 @@
...
-- name: Demo page dark theme
-  url: http://localhost:3000/Detox/docs/next/demo
+- name: Showcase page dark theme
+  url: http://localhost:3000/Detox/showcase?shuffle=disabled
   execute: document.documentElement.setAttribute("data-theme", "dark")
 
-- name: Demo page light theme
-  url: http://localhost:3000/Detox/docs/next/demo
+- name: Showcase page light theme
+  url: http://localhost:3000/Detox/showcase?shuffle=disabled
```
asafkorem, line 30: what happened here? did we had duplications?

asafkorem (changes requested):

website/snapshots.yml
noomorph: no it was just reordering

website/src/components/Showcase/Showcase.js
moutro: Very good advice here and thank you both. I see now. The logic to shuffle or not, should be handled outside of the _makeShuffledCardList function. I removed it and placed it inside useEffect and included your suggestion @noomorph 
Will push changes.

website/src/components/Showcase/Showcase.js
```diff
@@ -6,20 +6,22 @@ import styles from './Showcase.module.scss';
...
+  const [toRender, setToRender] = useState([<TailSpin />]);
 
   useEffect(() => {
-    setCards(_makeShuffledCardList(cardList).map(_makeCard));
+    const shouldShuffle = new URLSearchParams(window.location.search).get('shuffle') !== 'disabled';
+    const resultingCardList = shouldShuffle ? _makeShuffledCardList(cardList) : cardList;
+    setCards(resultingCardList.map(_makeCard));
   }, []);
 
   useEffect(() => {
```
noomorph, line 19:
```suggestion
return cards.length > 0 ? _makeCardsContainer(cards) : _makeLoadingSpinner();
```

Could you explain why do we need such advanced manipulations with that `toRender` and `setToRender` state? Why can't we just return those values in a simple return statement?

moutro: There was a spinner loader implemented to the Showcase page that added a `toRender `and `setToRender` state.  I will work on simplifying the code to just return the desired component based on the length of the cards array without the intermediate state there.

PR #3883, opened 2023-01-29
[Website]:  Split custom.scss file into multiple files

### LINKED ISSUE #3820: [Website] Create global theme for styling components
### Describe your idea

Detox website's Figma: [https://www.figma.com/file/QGhhB7djq0OxvLzU…](https://www.figma.com/file/QGhhB7djq0OxvLzULhs7u5/Detox)

- [ ] Make a global styling PR that will fix the fonts, colors, etc. in 'custom.css'
- [ ] Current website we are mixing the old and new design and some components are using old fonts and some are using the new fonts (like in the admonitions)
- [ ] make the most global docusaurus theme related cusomizations in a PR
- [ ] Next step to then go back and style the specific components using old styles to match new Figma styles

moutro:
## Description

Subtask of #3820 

- **_Split_** custom.scss file into multiple .scss files into separate .scss files.
- Import all back into custom.scss file.

Note: Live Sass Watcher Extension added a comment at the bottom of all CSS files. Left them there for now but please let me know if you want me to remove. This is ths reason there looks to be many files modified in this PR.

cc: @asafkorem


noomorph (changes requested):
There is a problem. I expected that all the `*.css` would have gone away. Instead, I see good old CSS files along with the new SCSS files.

@moutro, could you explain how you ended up with these quasi-duplicates?

website/src/css/_benefits-section.scss
```diff
@@ -0,0 +1,24 @@
...
+[data-theme='dark'] .benefits-header {
+  color: #ffffff;
+}
+
+.social-buttons {
+  justify-content: flex-end;
+  display: flex;
+}
+
+.flex-jc-fe {
```
noomorph, line 21: Does it really belong to the benefits section? Looks generic on a first glance.

website/src/css/custom.scss
```diff
@@ -0,0 +1,26 @@
...
+@import 'benefits-section';
+@import 'footer-socials';
+@import 'footer';
+@import 'header-github';
+@import 'navbar';
+@import 'showcase-link';
+@import 'showcase-section';
+@import 'utilities';
+
+.docusaurus-highlight-code-line {
```
noomorph, line 14: This can be extracted to `_markdown.scss`.

moutro:
> There is a problem. I expected that all the `*.css` would have gone away. Instead, I see good old CSS files along with the new SCSS files.
> 
> @moutro, could you explain how you ended up with these quasi-duplicates?

@noomorph - I believe that what I did was to keep the `*.css` files includig the `custom.css` file in the projects but should have deleted all of them since I am using the SASS plugin and also including it in the config file to point to the `custom.scss` file.  Based on this, I should have deleted the css files and will do so now. Thanks!

website/src/css/_benefits-section.scss
moutro: @noomorph  - This is being using just in the Homepage SubHeader for the social icons. It's why I put it here. Is it not correct?
image

website/src/css/custom.scss
moutro: @noomorph - do you mean lines 14 - 18 and 24 - 26?

noomorph: Looks like it.

As for`[data-theme='light'] .banner {`, I doubt it is used at all.

moutro: I think it belongs to the UA banner

noomorph: Maybe someone it belongs by intent, but not by any actual effect. See:

image

The class `banner_SLir` != `banner`, so this rule looks useless.

moutro: Thanks. I will remove it since it's not doing anything.

noomorph (approved):

PR #3917, opened 2023-02-08
Apply DETOX_DBG_SELECTED_CFG env-var to allow for IDE jest-debugging

d4vidi:
## Description

- This pull request addresses the issue described here: N/A

In this pull request, I have introduced `DETOX_DBG_SELECTED_CFG` which allows for an env-var based selection of configuration. This opens the door to debug-based execution of Detox tests in IDE's (e.g. webstorm), which jest must be executed explicitly:
image

@noomorph if there's a better way to do this which I've overlooked, please let me know

---

> _For API changes:_

### DIFF
--- a/detox/src/realms/DetoxPrimaryContext.js
+++ b/detox/src/realms/DetoxPrimaryContext.js
@@ -51,7 +51,7 @@ class DetoxPrimaryContext extends DetoxContext {
     }
   }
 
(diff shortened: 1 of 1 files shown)


noomorph:
Use `DETOX_CONFIGURATION` var. Please don't create duplicates.

d4vidi:
Noted. We should probably add that to the contribution/debug guides

PR #3919, opened 2023-02-08
[Website]: Global styling for website

moutro:
## Description

Subtask of #3820 

This pull request addresses the issues:

- Global styling PR that fixes the fonts, social buttons, showcase buttons, colors.
- Add Sass styling to the Admonitions and change the admonitions to Sass format
- Create mixins to be able to reuse global styles in next task to style the specific components to match new Figma styles (e.g. admonitions, etc.) once approved in this PR

cc: @asafkorem


.gitignore
```diff
@@ -231,3 +231,6 @@ detox/test/artifacts
 Detox-android/
 Detox-ios-src.tbz
 Detox-ios.tbz
+
+# Git VCS 
+website/**/*.css
```
asafkorem, line 236: VCS?

Also, why do we need that under .gitignore?

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+:root {
+  @include detox-colors('detox-color');
+  @include ifm-colors('ifm');
+
+  &[data-theme='dark'] {
+    @include detox-colors-dark('detox-color');
+  }
+}
+
+// Fonts
```
asafkorem, line 90: I think you can extract this to another scss file (`_fonts.scss`).

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
+// All of the following configs are used in the website and documentation provided by the Design team Glyph
+
+// Colors
```
asafkorem, line 3: Same (`_colors.scss`)

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  font-weight: 400;
+}
+
+@mixin t3-20 {
+  font-family: 'Nunito Sans';
+  font-size: 20px;
+  font-weight: 400;
+}
+
+// Showcase buttons
```
asafkorem, line 188: Isn't that part of the Showcase? Is that the card's buttons? This needs to be under showcase-specific file

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  &:active {
+    background: #5c25af;
+    color: #ffffff;
+    text-decoration: none;
+  }
+}
+
+// Social buttons
```
asafkorem, line 244: Same, needs to be under homepage-specific file (<=> context)

asafkorem (changes requested):
Thanks @moutro! 🥇 
Looks good overall, but needs some extractions.
Could you please share here some screenshots?

noomorph:
@asafkorem there is Percy for that. Jason does not need to share screenshots manually

.gitignore
noomorph: @moutro , I'd move it before `## Final outputs` closer to `/examples/...`.
Also, you can proceed `/website` with a slash in the beginning.

@asafkorem, this is correct. We migrated to SCSS. All the resulting CSS files are just temporary artifacts needed for serving and deploying the website.

website/package.json
```diff
@@ -15,10 +15,10 @@
     "write-heading-ids": "docusaurus write-heading-ids"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.2.0",
-    "@docusaurus/preset-classic": "^2.2.0",
-    "@docusaurus/remark-plugin-npm2yarn": "^2.2.0",
-    "@docusaurus/theme-search-algolia": "^2.2.0",
+    "@docusaurus/core": "^2.3.1",
```
noomorph, line 18: Nice idea to upgrade!

website/src/css/_config.scss
noomorph: Agree.

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  --#{$prefix}-color-primary-darkest: #150a51;
+  --#{$prefix}-color-primary-light: #6c43a3;
+  --#{$prefix}-color-primary-lighter: #8859c9;
+  --#{$prefix}-color-primary-lightest: #9f6de5;
+  --#{$prefix}-code-font-size: 95%;
+  --#{$prefix}-highlighted-code-line-bg: #d9e5f6;
+}
+
+@mixin detox-colors-dark($prefix) {
+  --#{$prefix}-highlighted-code-line-bg: #414458;
```
noomorph, line 70: Looks wrong. It should be `docusaurus-highlighted-code...`.

Neither `--#{$prefix}-highlighted-code-line-bg: #414458;` nor `-ifm-background-color: #12161b;` don't belong here btw.

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  --#{$prefix}-u4a: #7633e526;
+  --#{$prefix}-e4a: #3ebf7226;
+
+  --#{$prefix}-r5a: #f14c5212;
+  --#{$prefix}-y5a: #f7ba2912;
+  --#{$prefix}-b5a: #3e77e412;
+  --#{$prefix}-u5a: #7633e512;
+  --#{$prefix}-e5a: #3ebf7212;
+
+  --#{$prefix}-a0: #12161bff;
```
noomorph, line 42: I would separate them to detox-colors-light.

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
+// All of the following configs are used in the website and documentation provided by the Design team Glyph
+
+// Colors
+
+@mixin detox-colors($prefix) {
```
noomorph, line 5: Using mixins is not correct AT ALL here. You want to use mixins when there is some repetitive aspect. 

Why would you want to use those color mixins more than twice? You define them once in `:root` as CSS variables and there you go. Every descendant can access them via `var(...)`.

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  --#{$prefix}-r0: #f14c52ff;
```
noomorph, line 6: On the other hand, I see that you wanted to avoid repetitiveness. Okay, fine, I understand, but there are other options if you really want that.

See a nice example here: https://stackoverflow.com/a/61191193

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  --#{$prefix}-y0: #f7ba29ff;
+  --#{$prefix}-b0: #3e77e4ff;
+  --#{$prefix}-u0: #7633e5ff;
+  --#{$prefix}-e0: #3ebf72ff;
+
+  --#{$prefix}-r1a: #f14c5289;
```
noomorph, line 12: Actually if we are talking about some serious auto-generation business, you could try to define transparency levels like (1, 0.8. 0.6. 0.5) and then generate all those `r1a, r2a, r3a` in a loop.

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+
+  &[data-theme='dark'] {
+    @include detox-colors-dark('detox-color');
+  }
+}
+
+// Fonts
+
+@mixin h1-80 {
+  font-family: 'Nunito Sans';
```
noomorph, line 93: Looks not bad. Here I agree that mixins make sense! We are likely to use them multiple times.

What I like less is that you define only a few properties. That seems very risky - what if there are some previously defined things like: `font-style: italic` etc.?

I'd vote for defining `font` shorthands in such mixins. That should be a safer option:

```
font: 400 normal 80px/96px 'Nunito Sans', sans-serif.
```

noomorph: Au fait, you haven't defined line-heights in the mixins. Looks very risky 😬

website/src/css/_config.scss
```diff
@@ -0,0 +1,348 @@
...
+  &:active {
+    background: #5c25af;
+    color: #ffffff;
+    text-decoration: none;
+  }
+}
+
+// Social buttons
+
+@mixin discord-btn {
```
noomorph, line 246: Very doubtful that it should be a mixin. The same goes for twitter-btn, etc.

website/src/css/custom.scss
```diff
@@ -9,5 +8,7 @@
 @import 'navbar';
 @import 'showcase-link';
 @import 'showcase-section';
-@import 'utilities';
 @import 'markdown';
+
+@import url('https://fonts.googleapis.com/css2?family=Nunito+Sans:wght@300;700&display=swap');
```
noomorph, line 13: Please follow the suggestion of Asaf and move also these imports to `_fonts.scss`.

moutro:
> > @asafkorem there is Percy for that. Jason does not need to share screenshots manually
> 
> Right..right, I honestly forgot about that while writing this line lol, I was a bit distracted probably.. 🙂
> 
> @moutro Percy's website and screenshots look fine to me, but I don't understand why it was passed on the GitHub action (there are 3 unreviewed changes on Percy). That's out of this task scope of course, but the check should fail (or "in progress") while there's no approval for the changes. See: [conclusion](https://github.com/marketplace/actions/github-checks#conclusion) and [status](https://github.com/marketplace/actions/github-checks#status).

@asafkorem - Looks like from the docs links you provided, that the conclusion and status checks can be added as fields inside the `performance` step at the end of the job (conclusion: ${{ always() } /   status: ${{ always() }} ). I can do that in a separate PR. Please let me know.

.gitignore
moutro: Done. Thanks!

website/package.json
moutro: Thanks! :) 
Want to make sure we are always up-to-date

website/src/css/custom.scss
moutro: Thanks Yarik. Moved both google fonts imports to the `_fonts.scss` file

website/src/css/_config.scss
moutro: Done!

website/src/css/_config.scss
moutro: Yeah, that's true. They aren't being used anywhere else. Thanks!

website/src/css/_config.scss
moutro: Hmm. Yes, looks incorrect.

Not sure they need to be as their own mixins so I moved them into the :root. Does that sound like a better approach? @noomorph

website/src/css/_config.scss
moutro: Good idea. Done - thanks!

website/src/css/_admonitions.scss
```diff
@@ -1,5 +1,5 @@
 .alert {
-  @include detox-colors(--detox-color-a0);
+  @include detox-colors-light(--detox-color-a0);
```
noomorph, line 2: I don't see a reason to include colors as a mixin here.

website/src/css/_showcase-section.scss
```diff
@@ -0,0 +1,55 @@
+// Showcase buttons
+
+@mixin showcase-default-btn {
+  background: radial-gradient(103.29deg, #7633e6 -14.83%, #5d71e6 102.26%);
```
noomorph, line 4: I think Asaf meant that you could have moved it entirely somewhere under Showcase component.

website/src/css/custom.scss
```diff
@@ -1,4 +1,6 @@
 @import 'config';
```
noomorph, line 1: The _config file looks empty now. Maybe you should remove it?

website/src/css/_fonts.scss
```diff
@@ -0,0 +1,100 @@
+// Fonts
+
+@import url('https://fonts.googleapis.com/css2?family=Nunito+Sans:wght@300;700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@300;600&display=swap');
+
+@mixin h1-80 {
+  font-family: 'Nunito Sans';
+  font-size: 80px;
+  font-weight: 400;
```
noomorph, line 9: Remember what I told you earlier about using font shorthand properties.

website/src/css/_colors.scss
```diff
@@ -0,0 +1,93 @@
+// Colors
+
+@mixin detox-colors-light($prefix) {
+  --#{$prefix}-r0: #f14c52ff;
```
noomorph, line 4: Remember what I told you earlier about generating colors from a SASS map variable.

website/src/css/_colors.scss
```diff
@@ -0,0 +1,93 @@
...
+  --#{$prefix}-y0: #f7ba29ff;
```
noomorph, line 5: Besides, those r,y,b,u,e colors are not inherently dark or light. They are a common part 🤷‍♂️

website/src/css/_colors.scss
```diff
@@ -0,0 +1,93 @@
...
+@mixin ifm-colors($prefix) {
+  --#{$prefix}-color-primary: #048fc0;
+  --#{$prefix}-color-primary-dark: #253282;
+  --#{$prefix}-color-primary-darker: #161360;
+  --#{$prefix}-color-primary-darkest: #150a51;
+  --#{$prefix}-color-primary-light: #6c43a3;
+  --#{$prefix}-color-primary-lighter: #8859c9;
+  --#{$prefix}-color-primary-lightest: #9f6de5;
+  --#{$prefix}-code-font-size: 95%;
+  --#{$prefix}-highlighted-code-line-bg: #d9e5f6;
```
noomorph, line 64: Prefix is wrong for `--docusaurus-highlighted-code-line-bg: #d9e5f6;` because it is not infima. I wrote about it previously. Please don't forget.

website/src/css/_colors.scss
```diff
@@ -0,0 +1,93 @@
...
+
+@mixin docusaurus-dark {
+  --docusaurus-highlighted-code-line-bg: #414458;
+}
+
+:root {
+  @include detox-colors-light('detox-color');
+  @include ifm-colors('ifm');
+
+  &[data-theme='dark'] {
```
noomorph, line 88: If docusaurus sets `data-theme=light` explicitly, maybe you could split colors to shared+[dark|light] parts and make it clear in your SCSS code as well. 

```
:root {
@include detox-colors

[data-theme=light] {
  @include detox-colors-light();
  ...
[data-theme=dark] {
  @include detox-colors-dark();
  ...
```

.gitignore
```diff
@@ -226,11 +226,11 @@ detox/test/artifacts
 /examples/demo-react-native/artifacts
 /examples/demo-react-native-jest/artifacts
 
+# Git VCS
```
noomorph, line 229: I think `Git VCS` comment does not serve any explanatory purpose here

noomorph (changes requested):
Looks better, but we need one more round at the very least. Most likely two.

website/src/css/_config.scss
moutro: @noomorph - using the font shorthand property is a 😎 cool idea! You can put a few values in there. When designers specify the font size and line-height in the format "24px / 30px", it means that the font size is 24 pixels and the line-height is 30 pixels. So the line-heights are there. and no need to specify "normal" since 400 is a regular/normal font weight.

Is this correct? If so, I have adjusted the fonts mixins to follow your suggestion and added only the color property of white

website/src/css/_config.scss
moutro: Hi, thanks again for the detailed explanation. I worked on it and my solution seems to work except for admonitions colors, navbar links colors. Could be more depending on Percy run. Sent in latest commit.

website/src/css/_admonitions.scss
moutro: Thanks. Removed and modified the admonitions based on the changes in the _colors.scss file map() function.

website/src/css/_showcase-section.scss
moutro: This entire file, decided to remove it as it was not looking to be adding any useful changes.

website/src/components/Showcase/Hero.module.scss
```diff
@@ -46,36 +46,36 @@
...
-  /* T/T2 */
-  font-family: 'Nunito Sans', sans-serif;
-  font-style: normal;
-  font-weight: 100px;
-  font-size: 16px;
-  line-height: 25px;
-  /* identical to box height, or 156% */
-  color: #ffffff;
+    /* T/T2 */
+    font-family: 'Nunito Sans', sans-serif;
```
noomorph, line 71: Aren't you supposed to use T2 font here?

noomorph:
@moutro I fixed fonts and colors.

website/src/css/_config.scss
noomorph: Cool

Issue #3818, opened 2023-01-04
[Website] Fix navbar mobile


moutro:
### Describe your idea

Detox website's Figma Design Final tab for Mobile Styles: https://www.figma.com/file/QGhhB7djq0OxvLzULhs7u5/Detox?node-id=98%3A2284&t=hPcPyD8iBOlyDjmc-0

- [x] Place searchbox and hamburger on RHS on mobile screens

Sample Images of changes needed:

Mobile screens:

![Screenshot 2023-01-04 at 17 08 09](https://user-images.githubusercontent.com/34761997/210585455-b981b752-20d8-417e-b01d-536ed97f673d.png)

larger screens:

![Screenshot 2023-01-04 at 17 08 37](https://user-images.githubusercontent.com/34761997/210585557-86f3369d-2df1-4c53-89a0-287a7191e222.png)

noomorph:
Thanks, @IrbisKronos. The issue is resolved.

PR #3922, opened 2023-02-09
docs(website): apply mobile navbar design

IrbisKronos:
## Description

This pull request addresses the issue described here: [#3818](https://github.com/wix/Detox/issues/3818) 

In this pull request, I have the hamburger and the search box buttons stylized according to the provided figma layout.

image

### DIFF
--- a/website/src/css/_navbar.scss
+++ b/website/src/css/_navbar.scss
@@ -24,12 +24,27 @@
 }
 /*===========*/
 
+/* Restyle gamburger button */
+
 .navbar__toggle {
   width: 40px;
   height: 40px;
   background-color: var(--detox-color-a3);
   border-radius: 10px;
-  display: flex;
   justify-content: center;
   align-items: center;
+}
+
+/* Restyle search button */
+
+@media (max-width: 769px) {
+  button.DocSearch-Button {
+    background-color: var(--detox-color-a3);
+    height: 40px;
+    border-radius: 10px;
+    padding: 0 10px;
+  }
+  button.DocSearch-Button svg {
+    opacity: 60%;
+  }
 }
--- a/website/src/css/_navbar.scss
+++ b/website/src/css/_navbar.scss
@@ -24,34 +24,36 @@
 }
 /*===========*/
 
-/* Fix navbar logo */
-.navbar__logo {
-  height: 26px;
-  margin: 0 10px 0 8px;
-}
-
-/* Restyle hamburger button */
+/* Restyle logo & hamburger button */
 
 .navbar {
   height: 70px;
+
+  &__items {
+    &--right {
+      position: relative;
+    }
+  }
+
+  &__logo {
+    height: 26px;
+    margin: 0 10px 0 8px;
+  }
+
+  &__toggle {
+    width: 40px;
+    height: 40px;
+    background-color: var(--detox-color-a3);
+    border-radius: 10px;
+    justify-content: center;
+    align-items: center;
+  }
 }
 
-.navbar__items--right {
-  position: relative;
-}
-
-.navbar__toggle {
-  width: 40px;
-  height: 40px;
-  background-color: var(--detox-color-a3);
-  border-radius: 10px;
-  justify-content: center;
-  align-items: center;
-}
 @media (max-width: 768px) {
   .navbar__toggle {
     position: absolute;
-    margin-right: 0;
+    margin: 0;
     top: 15px;
     right: 25px;
   }
@@ -65,9 +67,6 @@
     height: 40px;
     border-radius: 10px;
     padding: 0 10px;
-    position: absolute;
-    top: -16px;
-    right: 48px;
   }
   .DocSearch-Search-Icon {
     opacity: 0.6;
--- a/website/src/css/_navbar.scss
+++ b/website/src/css/_navbar.scss
@@ -32,10 +32,4 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  svg {
-    width: 22px;
-    path {
-      stroke-width: 2.5;
-    }
-  }
 }
--- a/website/src/css/_navbar.scss
+++ b/website/src/css/_navbar.scss
@@ -45,6 +45,6 @@
     padding: 0 10px;
   }
   button.DocSearch-Button svg {
-    opacity: 60%;
+    opacity: 0.6;
   }
 }
--- a/website/src/css/_navbar.scss
+++ b/website/src/css/_navbar.scss
@@ -22,3 +22,20 @@
   padding: 5px;
   font-weight: 300;
 }
+/*===========*/
+
+.navbar__toggle {
+  width: 40px;
+  height: 40px;
+  background-color: var(--detox-color-a3);
+  border-radius: 10px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  svg {
+    width: 22px;
+    path {
+      stroke-width: 2.5;
+    }
+  }
+}
--- a/website/src/css/_navbar.scss
+++ b/website/src/css/_navbar.scss
@@ -30,11 +30,14 @@
   margin: 0 10px 0 8px;
 }
 
-/* Restyle gamburger button */
+/* Restyle hamburger button */
 
 .navbar {
-  position: relative;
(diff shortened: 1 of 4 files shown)


website/src/css/_navbar.scss
```diff
@@ -22,3 +22,29 @@
...
+  height: 40px;
+  background-color: var(--detox-color-a3);
+  border-radius: 10px;
+  justify-content: center;
+  align-items: center;
+}
+
+/* Restyle search button */
+
+@media (max-width: 769px) {
```
noomorph, line 40: Looks like you hypercorrected yourself and made a mistake involuntarily.
```suggestion
@media (max-width: 768px) {
```

All similar media queries in Docusaurus source code use `768px`, not `769px`.

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,29 @@
...
+  button.DocSearch-Button {
+    background-color: var(--detox-color-a3);
+    height: 40px;
+    border-radius: 10px;
+    padding: 0 10px;
+  }
```
noomorph, line 46: SCSS syntax allows you to express the same intent via [nesting](https://www.w3schools.com/sass/sass_nesting.php) to avoid repetition:

```suggestion
    svg {
      opacity: 0.6;
    }
  }
```

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,29 @@
...
+  button.DocSearch-Button svg {
+    opacity: 0.6;
+  }
+}
```
noomorph, line 50: I reviewed your changes visually: https://percy.io/66fcbcad/detox-website/builds/25010682/changed/1396573909

Unfortunately, you forgot to override the positioning of `.navbar__toggle` for mobile viewports. Please add `position: absolute` override and pick a correct value for `right: ??px`.

| Actual | Expected |
|--------|----------|
|  |  |

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,29 @@
...
+  background-color: var(--detox-color-a3);
+  border-radius: 10px;
+  justify-content: center;
+  align-items: center;
+}
+
+/* Restyle search button */
+
+@media (max-width: 769px) {
+  button.DocSearch-Button {
```
noomorph, line 41: Side note for other reviewers passing by – adding a specific CSS class was **necessary** due to a clash between styling rules from the vendor and our custom ones.

noomorph (changes requested):
Overall, looks very good as for the first try – you almost nailed it, @IrbisKronos!

See my comments above, and good luck with finishing the PR!

IrbisKronos:
Thanks for the feedback. I took into account the comments, corrected and styled the navigation panel completely according to the layout.

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,47 @@
...
+/* Fix navbar logo */
+.navbar__logo {
+  height: 26px;
+  margin: 0 10px 0 8px;
+}
+
+/* Restyle gamburger button */
+
+.navbar {
+  position: relative;
```
noomorph, line 36: Hmm, was it necessary to define `position: relative` here? I think it only complicated your task, because you had to apply an ugly override below:

```
div.searchBox_node_modules-\@docusaurus-theme-classic-lib-theme-Navbar-Search-styles-module {
   right: 80px;
 }
 ```
 
I think it's better if you remove that `position: relative`, so that your absolute positions are calculated against the outmost container (like it was before).

File: website/src/css/_navbar.scss [hunk 6]
```diff
@@ -22,3 +22,47 @@
...
+    background-color: var(--detox-color-a3);
+    height: 40px;
+    border-radius: 10px;
+    padding: 0 10px;
+  }
+  .DocSearch-Search-Icon {
+    opacity: 0.6;
+  }
+}
+div.searchBox_node_modules-\@docusaurus-theme-classic-lib-theme-Navbar-Search-styles-module {
```
noomorph, line 66: Ugly and unreliable selector. It won't work in the production mode. 

image

noomorph: Module (scoped) CSS classes are autogenerated. If you are unsure when working on such changes, you can use two commands: `npm run build` and `npm run serve`. The first will build the production-ready static website, and the second one will run a simple HTTP server serving the compiled files.

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,47 @@
...
+
+.navbar__toggle {
+  width: 40px;
+  height: 40px;
+  background-color: var(--detox-color-a3);
+  border-radius: 10px;
+  justify-content: center;
+  align-items: center;
+  position: absolute;
+  margin-right: 0;
```
noomorph, line 48: I don't think that `margin-right` will anyhow work if the position is absolute.

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,47 @@
...
+  top: 15px;
+  right: 25px;
```
noomorph, line 50: Actually I'm not sure that the hamburger should move to the right in compact mode. 

I'd rather put it under media query 768px.

On the other hand, we don't have specific designs for that intermediate resolution so it's worth checking with our designers.

image

website/src/css/_navbar.scss
IrbisKronos: Noticing that the height of the "navbar" block does not match the template, I determined it. And then I decided to move away from this block in determining indents, because then you can make corrections without taking into account the "padding" of the intermediate element. Come to think of it, it doesn't seem like such a good idea)

website/src/css/_navbar.scss
IrbisKronos: With
image
Without
image

The "margin-right: 0.5rem" property comes into play and this, I suspect, is due to "display: inherit" which inherits from "flex".

website/src/css/_navbar.scss
IrbisKronos: I will fix it according to your suggestion.

website/src/css/_navbar.scss
noomorph: Hmm, that can be true. 🤔  Maybe you should just reset `margin: 0` as simple as it is? I mean you won't be happy if some other CSS rule adds margin to any of the side as it can affect your positioning.

website/src/css/_navbar.scss
noomorph: Cool.

website/src/css/_navbar.scss
IrbisKronos: Sorry, didn't know it didn't work in production mode. In general, it is not a problem to bind to the nearest parent element. Although top/right calculations will not be so obvious.

IrbisKronos:
I look forward to the next feedback )

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,54 @@
...
+/* Restyle search button */
+
+@media (max-width: 768px) {
+  button.DocSearch-Button {
+    background-color: var(--detox-color-a3);
+    height: 40px;
+    border-radius: 10px;
+    padding: 0 10px;
+    position: absolute;
+    top: -16px;
```
noomorph, line 69: Looks too dirty (-16px 😬 ). Try swizzling `Navbar/Search` – use wrapper strategy.
Near the index.js file create `styles.module.css` file where you can define a style overriding position to `right: 80px` (or whatever it is).

```
src/theme/Navbar/Search/styles.module.scss
src/theme/Navbar/Search/index.js
```

website/src/css/_navbar.scss
```diff
@@ -22,3 +22,54 @@
...
+.navbar__logo {
+  height: 26px;
+  margin: 0 10px 0 8px;
+}
+
+/* Restyle hamburger button */
+
+.navbar {
+  height: 70px;
+}
```
noomorph, line 37: Good opportunity to use nesting here

```
.navbar {
  // ...
  &__items {
    &--right {
    }
  }
  
  &__toggle {
  }
}
```

noomorph (approved):
Brilliant. All's good! Thanks!!

PR #3926, opened 2023-02-10
[Genymotion SaaS] Update gmsaas start flag to fix wrong device timeout

lilinor:
## Description

- When starting a device, `--stop-when-inactive` flag was added to the `gmsaas start` command in the past when the global timeout didn't existed but only the front end inactivity timeout. When the global timeout was released, this flag still pointed to the front end inactivity timeout which is the wrong time out for devices started from gmsaas. Removing it solves the issue.


d4vidi:
Closing in favor of #3929

PR #3954, opened 2023-02-26
Fix detoxURLBlacklistRegex arg for launchApp() on iOS

### LINKED ISSUE #3953: detoxURLBlacklistRegex is broken in device.launchApp()
### What happened?

As reported by @awinograd on [Discord](https://discord.com/channels/957617863550697482/957619870919442432/1078058934839017623), `detoxURLBlacklistRegex` appears to be broken when passed in through `launchArgs` in `device.launchApp()`.

I will submit a potential fix shortly.

Should adhere to [the docs](https://wix.github.io/Detox/docs/api/device/#11-detoxurlblacklistregexinitialize-the-url-blacklist-at-app-launch) and work in an analogue way to Android.

- [ ] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

_No response_

Detox version: 20.2.1
React Native version: 
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 
Device model: 
iOS version: 
macOS version: 
Xcode version: 
Test-runner (select one): jest / other

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
paste your device.log here!
```

</details>

### More data, please!

_No response_

d4vidi:
## Description

- This pull request fixes #3953

In this pull request, I have aligned the usage of the `detoxURLBlacklistRegex` to the docs (and Android).
Apparently, `detoxURLBlacklistRegex` cannot be extracted from `NSUserDefaults` as an `arrayForKey` (`nil` is returned). Therefore, I've switched to `stringForKey` and implemented a simplified means of string-splitting.

I honestly don't know whether this has ever worked before - perhaps in older iOS versions. In any case, it is what it is... Thanks to the added coverage we shouldn't be having this regression in the future.

---

> _For API changes:_

### DIFF
--- a/detox/test/e2e/14.network.test.js
+++ b/detox/test/e2e/14.network.test.js
@@ -46,23 +46,25 @@ describe('Network Synchronization', () => {
     await waitFor(element(by.text('Long Network Request Working!!!'))).toBeVisible().withTimeout(4000);
     await expect(element(by.text('Long Network Request Working!!!'))).toBeVisible();
 
-    await device.setURLBlacklist([]);
+    await device.launchApp({ newInstance: true });
   });
 
-  it(':android: launchArgs with detoxURLBlacklistRegex should set the blacklist', async () => {
+  it('launchArgs with detoxURLBlacklistRegex should set the "black" (synchronization-ignore) list', async () => {
+    const blackListRegexp = '^http://localhost:\\d{4}?\/[a-z]+\/\\d{4}?$';
+
     await device.launchApp({
       newInstance: true,
-      launchArgs: { detoxURLBlacklistRegex: ' \\("^http:\/\/localhost:\\d{4}?\/[a-z]+\/\\d{4}?$"\\)' },
+      launchArgs: { detoxURLBlacklistRegex: `\\("http://meaningless\.first\.url","${blackListRegexp}"\\)` },
     });
 
-    await device.reloadReactNative();
-
-    await element(by.text('Network')).tap();
-    await element(by.id('LongNetworkRequest')).tap();
-    await expect(element(by.text('Long Network Request Working!!!'))).not.toBeVisible();
-    await waitFor(element(by.text('Long Network Request Working!!!'))).toBeVisible().withTimeout(4000);
-    await expect(element(by.text('Long Network Request Working!!!'))).toBeVisible();
-
-    await device.setURLBlacklist([]);
+    try {
+      await element(by.text('Network')).tap();
+      await element(by.id('LongNetworkRequest')).tap();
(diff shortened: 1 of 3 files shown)


detox/test/e2e/14.network.test.js
```diff
@@ -46,23 +46,25 @@ describe('Network Synchronization', () => {
...
-    await waitFor(element(by.text('Long Network Request Working!!!'))).toBeVisible().withTimeout(4000);
-    await expect(element(by.text('Long Network Request Working!!!'))).toBeVisible();
-
-    await device.setURLBlacklist([]);
+    try {
+      await element(by.text('Network')).tap();
+      await element(by.id('LongNetworkRequest')).tap();
+      await expect(element(by.text('Long Network Request Working!!!'))).not.toBeVisible();
+      await waitFor(element(by.text('Long Network Request Working!!!'))).toBeVisible().withTimeout(4000);
+      await expect(element(by.text('Long Network Request Working!!!'))).toBeVisible();
```
asafkorem, line 65: The first expectation will throw on timeout, so I believe we don't need the second check.
```suggestion
      await waitFor(element(by.text('Long Network Request Working!!!'))).toBeVisible().withTimeout(4000);
```

detox/test/e2e/14.network.test.js
```diff
@@ -46,23 +46,25 @@ describe('Network Synchronization', () => {
...
+    } finally {
+      await device.launchApp({ newInstance: true });
+    }
```
asafkorem, line 68: Same thing could (should) have done for the test `setURLBlacklist() should disable synchronization for given endpoint`, right?
Let's put them together under a shared describe block called `"URL Blacklist"` with `afterEach` block that relaunches the app with new instance, instead of using try-finally

asafkorem (approved):
Oh, wow. It appears that this API never worked on iOS, and I suspect that it was dysfunctional even in previous versions of iOS. 
Minor comments regarding tests, the implementation LGTM(!)
GG, Resolve & merge 🙂

d4vidi:
Yikes, missed those 2 comment. Will review / fix now 🙏🏻

detox/test/e2e/14.network.test.js
d4vidi: Much agreed, great idea 🤘🏻

tiagoacastro:
This still seems to have issues on 20.11.2

Issue #3977, opened 2023-03-12
Text property/matching: Derive from child-views


d4vidi:
### Describe your idea

As reported via an [internal support case](https://wix.slack.com/archives/C1Y4D9BK8/p1678275248117969) by @bunkib, we could do better in terms of Android equivalence (to iOS) by aligning text-based matching logic:

On iOS, the _text_ property of a view is defined not only by the text associated with view itself - provided that it is in fact a text view, but rather by recursively concatenating the text values of all children. This is not because this is how we decided to introduce that behavior on iOS, but because this is how React Native sees it (IIRC).

This is relevant on 2 levels:
1. `.toHaveText()` matching logic: Look up the text based on the view's textual children, recursively.
2. `text` attribute: Assign based on the children's texts, similarly.

d4vidi:
TODO Reiterate [this thread](https://wix.slack.com/archives/C1Y4D9BK8/p1678275050951799)

PR #3979, opened 2023-03-14
Improve quality of testing over Detox assertions (expects etc.)

d4vidi:
## Description

- This pull request is related the issue described here: #3977

This pull request introduces some preliminaries to #3977: It cleans up some Detox self-tests related to expectations (assertions). That allows, in turn, to introduce further tests related to that issue.

- Eliminated assertion E2E tests that have already been covered on unit-test level (i.e. invalid args detection).
- Improved expect unit-testing:
  - Replaced custom `expectToThrow()` with jest's `expect().toThrow()` which does the same thing
  - Removed the `by.traits()` test for unsupported traits at unit level, which only make sense on E2E-level because the error should come from native-detox code (todo: add such E2E's)
  - Added missing invalid-args assertions in some `expect()` calls which were unit-tested to throw but didn't fail because `expectToThrow()` was "penetrable".

---

> _For API changes:_

### DIFF
--- a/detox/src/android/AndroidExpect.test.js
+++ b/detox/src/android/AndroidExpect.test.js
@@ -1,4 +1,7 @@
 // @ts-nocheck
+
+const jestExpect = require('expect').default; // eslint-disable-line
+
 describe('AndroidExpect', () => {
   let e;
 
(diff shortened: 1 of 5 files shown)


d4vidi:
@noomorph Please review post-factum if you wish

detox/src/android/AndroidExpect.test.js
```diff
@@ -1,4 +1,7 @@
 // @ts-nocheck
+
+const jestExpect = require('expect').default; // eslint-disable-line
```
noomorph, line 3: In this context, it looks weird – you can use regular `expect`, can't you??

detox/src/android/core/NativeMatcher.js
```diff
@@ -1,27 +1,42 @@
+const { DetoxRuntimeError } = require('../../errors');
 const invoke = require('../../invoke');
 const DetoxMatcherApi = require('../espressoapi/DetoxMatcher');
 
 class NativeMatcher {
+  static _assertMatcher(matcher) {
+    if (!(matcher instanceof NativeMatcher)) {
+      throw new DetoxRuntimeError(`Expected a matcher, got ${typeof matcher}`);
```
noomorph, line 8: Nah, use something like `require('util').inspect()`. `typeof object` is not interesting

detox/test/e2e/04.assertions.test.js
```diff
@@ -1,91 +1,56 @@
...
   });
 
   it('assert toggle-switch widget', async () => {
-    await expect(element(by.id('UniqueId146'))).toHaveToggleValue(false);
-    await element(by.id('UniqueId146')).tap();
-    await expect(element(by.id('UniqueId146'))).toHaveToggleValue(true);
-    await expect(element(by.id('UniqueId146'))).not.toHaveToggleValue(false);
-  });
-
-  it('should throw exception for visibility threshold out of range (lower than 1)', async () => {
```
noomorph, line 52: You removed this because those tests didn't actually work, right?

d4vidi: Because the use cases are fully covered under `expect` tests and there's no added value in running them as E2E's using futile matchers

detox/src/android/core/NativeMatcher.js
d4vidi: https://github.com/wix/Detox/pull/3980/files#diff-af99071757b1d50b8db1299cb540dd373a3d6346b826447a408b34007fb0115aR10

PR #3980, opened 2023-03-15
Android accessibility-label alignment with iOS

### LINKED ISSUE #3977: Text property/matching: Derive from child-views

d4vidi:
## Description

- This pull request fixes #3977

In this pull request, I have applied a more extensive algorithm for resolving a `View`'s `contentDescription` so as to bridge over the RN gap in https://github.com/facebook/react-native/issues/32826 (as explained in #3977).

---

> _For API changes:_


detox/android/detox/src/full/java/com/wix/detox/reactnative/ui/UIExtensions.kt
```diff
@@ -0,0 +1,39 @@
...
+        if (rawLabel != null) {
+            subLabels.add(rawLabel)
+        } else {
+            view.forEachChild { child ->
+                collectChildAccessibilityLabels(child, subLabels)
+            }
+        }
+    }
+    return subLabels
+}
```
asafkorem, line 32: You could avoid this recursion using a queue:
```kotlin
private fun collectChildAccessibilityLabels(
    view: View,
): List<CharSequence> {
    val labels = mutableListOf<CharSequence>()
    val queue = mutableListOf<View>()

    queue.add(view)

    while (queue.isNotEmpty()) {
        val currentView = queue.removeAt(0)
        getRawAccessibilityLabel(currentView)?.let { rawLabel ->
            labels.add(rawLabel)
        } ?: run {
            view.forEachChild { child ->
                queue.add(child)
            }
        }
    }

    return labels
}

```

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxMatcher.java
```diff
@@ -47,6 +48,14 @@ public static Matcher<View> matcherForContentDescription(String contentDescripti
         return allOf(withContentDescription(contentDescription), withEffectiveVisibility(Visibility.VISIBLE));
     }
 
+    public static Matcher<View> matcherForContentDescriptionOrText(String contentDescription) {
+        return allOf(anyOf(withContentDescription(contentDescription), withText(contentDescription)), withEffectiveVisibility(Visibility.VISIBLE));
+    }
```
asafkorem, line 53: I don't really get what you're doing here to be honest. Let's discuss this part..

asafkorem (changes requested):
image

detox/src/android/matchers/index.js
```diff
@@ -2,9 +2,9 @@ const native = require('./native');
 const web = require('./web');
 
 module.exports = {
-  accessibilityLabel: (value) => new native.LabelMatcher(value),
   id: (value) => new native.IdMatcher(value),
-  label: (value) => new native.LabelMatcher(value),
+  label: (value) => new native.ExtendedValueMatcher(value),
+  accessibilityLabel: (value) => new native.ExtendedValueMatcher(value),
```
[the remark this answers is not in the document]
d4vidi: I don't know either - `accessibilityLabel` isn't even documented. I kept it for the benefit of the doubt that it is used by someone, somewhere. We can deprecated for Detox 21/22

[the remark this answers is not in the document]
d4vidi: ... https://github.com/wix/Detox/discussions/3406#discussion-4081876

detox/android/detox/src/full/java/com/wix/detox/reactnative/ui/UIExtensions.kt
d4vidi: I liked the idea of traversing the view hierarchy as a general method.

detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxMatcher.java
d4vidi: I'll try to improve conventions

asafkorem (approved):

Issue #3768, opened 2022-12-19
Random `session.sessionId` is not enough random


noomorph:
### What happened?

This report comes from the engine team at Wix. They have a very specific use case – they use explicit bundle IDs for installing, launching, terminating, and uninstalling apps like this:

```
await device.launchApp('com.wix.appId');
await device.terminateApp('com.wix.appId');
```

This bug arose on the boundary of two adjacent test suites:
* the previous suite was testing `com.wix.app1`
* the next suite was testing `com.wix.app2`

The previous suite did not terminate the app – and by the way, technically, you are not obliged to. 
The next suite launched another app with the same `-sessionId <UUID>` parameter as the previous launch.

That created a situation when:
* the `app1` was in the background, still connected to the Detox server in `sessionId=<UUID>`
* the `app2` tried to connect to it using the same ID.

### What was the expected behaviour?

The apps should have been using different `sessionId`s.

### Was it tested on latest Detox?

- [ ] I have tested this issue on the latest Detox release and it still reproduces.

### In what environment did this happen?

Detox version: 20.1.0
Node version: 14.x
Test-runner (select one): jest

d4vidi:
@noomorph Are you sure that this is the right approach? Technically, we don't support multiple apps yet, and all the more so - `device.selectApp()` starts by terminating the currently selected one. I understand that this would not imply supporting multiple apps per se, but I'm a bit concerned that without hard-terminating all left-over apps in the beginning of each test (i.e. the alternative approach) we are in fact introducing numerous edge-cases that we don't know how to handle, regardless of the session ID.

noomorph:
@d4vidi, well, it is part of a code that will be refactored heavily sooner or later. We could live with it as fine as without it.

PR #3982, opened 2023-03-15
fix: randomize session.id between runs

noomorph:
## Description

- This pull request addresses the issue described here: #3768

In this pull request, I have moved random UUID generation to Client.js so as not to mutate the sessionConfig (as it will be preserved for the next test suites within the same Jest worker).

## Before

Two consequent test file runs: see before-app-launch events – detoxSessionId is the same, `c87c6cee-8409-1998-b434-cbe71a6ce286`:

<table><tr><td>

before-1

</td><td>

before-2

</td></tr></table>

## After

Two consequent test file runs: see before-app-launch events – detoxSessionId is random:

<table><tr><td>

after-1

</td><td>

after-2

</td></tr></table>

### DIFF
--- a/detox/src/DetoxWorker.js
+++ b/detox/src/DetoxWorker.js
@@ -96,8 +96,6 @@ class DetoxWorker {
     this._behaviorConfig = behaviorConfig;
     this._deviceConfig = deviceConfig;
     this._sessionConfig = sessionConfig;
-    // @ts-ignore
-    this._sessionConfig.sessionId = sessionConfig.sessionId || uuid.UUID();
     this._runtimeErrorComposer.appsConfig = this._appsConfig;
 
     this._client = new Client(sessionConfig);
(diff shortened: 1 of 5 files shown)


detox/src/DetoxWorker.js
```diff
@@ -96,8 +96,6 @@ class DetoxWorker {
     this._behaviorConfig = behaviorConfig;
     this._deviceConfig = deviceConfig;
     this._sessionConfig = sessionConfig;
-    // @ts-ignore
-    this._sessionConfig.sessionId = sessionConfig.sessionId || uuid.UUID();
```
d4vidi, line 100: The lint rule is right, `uuid` is not longer required 😄

detox/src/client/Client.js
```diff
@@ -29,7 +30,7 @@ class Client {
     this._onUnhandledServerError = this._onUnhandledServerError.bind(this);
     this._logError = this._logError.bind(this);
 
-    this._sessionId = sessionId;
+    this._sessionId = sessionId || uuid.UUID();
```
d4vidi, line 33: I'm not entirely sure about this but wouldn't it make more sense for the worker (i.e. consumer) to regenerate the UUID per each client instance? Arguably, choosing a random UUID can be considered a client "feature" but I'd feel more comfortable with the `preconfigured-sessionID-or-fallback-to-random-uuid` logic to be contained in one place (i.e. the DetoxWorker).

d4vidi (approved):
I generally approve of this but please review the comments first and also let's wrap up [this discussion](https://github.com/wix/Detox/issues/3768#issuecomment-1477600993)

Issue #3139, opened 2021-12-21
Detox cant find element by RTL text


kholood-ea:
### Description
detox cant seem to find element by text when text is written in other then english for ex:

this test passes ✅
```
<View flex><Text>Home page</Text> </View>

it('should show home page text ', async () => {
    await expect(element(by.text('Home page'))).toBeVisible();
});
```
but this one doesn't ❌
```
<View flex><Text>الرئيسية</Text> </View>

it('should show home page text', async () => {
    await expect(element(by.text('الرئيسية'))).toBeVisible();
  });
```
### Environment

 - Detox:19.3.1
 - React Native:0.64.1
 - Node:12.19.0
 - Device: iphone 13 pro max
 - OS: ios
 - Test-runner (select one): `jest-circus`

sheydarezaee:
experiencing the same ☝🏽

noomorph:
Assigning myself before I create a test case in Detox E2E repo.

noomorph:
Hmm, I'll put this into iteration scope.

noomorph:
I'm afraid I have to close this issue because no demo project was provided.  I tried the same in https://github.com/wix/Detox/pull/3985, but it works. 🤷

PR #3985, opened 2023-03-15
test(e2e): add RTL text samples

noomorph:
Related to: https://github.com/wix/Detox/issues/3139

Changes the sanity screen like this:

<table><tr><td>

screen

</td><td>

screen-2

</td></tr></table>

P.S. At the first glance, seems that the tests pass. Let's check for the CI result.


asafkorem (approved):
Great idea to add those tests!
Although, I'm not sure if that goes under "Sanity" or needs a designated suite for "RTL" or something..
However.. since we aren't dealing with RTL / LTR texts in different ways, it doesn't really matter and can be part of the Sanity suite.

noomorph:
Yeah, I've been wondering exactly the same. Ideally, it could move to the matchers suite probably.

d4vidi:
Just as long as you also add `Ласкаво просимо` 😄

PR #3987, opened 2023-03-17
Update chalk dependency to version 4

stianjensen:
https://github.com/chalk/chalk/releases

This version of Chalk is faster, and drags in less transitive dependencies.

Version 4 still supports node back to v10, and I see that Detox declares Node 14 as a minimum.

I did not upgrade all the way to Chalk 5 since it is ESM only.

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -61,7 +61,7 @@
     "bunyan": "^1.8.12",
     "bunyan-debug-stream": "^3.1.0",
     "caf": "^15.0.1",
-    "chalk": "^2.4.2",
+    "chalk": "^4.0.0",
     "child-process-promise": "^2.2.0",
     "execa": "^5.1.1",
     "find-up": "^4.1.0",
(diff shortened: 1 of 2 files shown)


asafkorem:
Hi @stianjensen. Thanks for the PR! Upgrading this dependency makes sense. 
Builds from external repositories that change our scripts or dependencies are automatically failed by our CI pipeline. I'll manually run this check.

@noomorph FYI

asafkorem:
I manually triggered our CI checks for this upgrade on: https://buildkite.com/wix-mobile-oss/detox/builds/2267
Seems like we get some type errors like:
```
TypeError: Cannot read property 'grey' of undefined
```
so tests are failing.. :/

stianjensen:
> so tests are failing.. :/

Sorry about that! I think I managed to set up lerna correctly locally now and run the full test suite (had some issues with running the native tests, but I believe this should only impact the js-tests anyway).

asafkorem (approved):
Thanks for resolving those issues @stianjensen. It passes our CI checks now 👍🏼

asafkorem:
@stianjensen Thanks and congratulations on your first Detox contribution! This will be part of Detox next release.

PR #3991, opened 2023-03-20
Docs: fix android/build.gradle indentation

CodingItWrong:
This docs-only PR fixes indentation of the changes to `android/build.gradle` in Project Setup.

Currently in 20.x and Next what is shown is:

image

Note that `google` and `maven` are indented the same amount as `allrepositories`, even though they are nested under it.

I actually made a mistake in setting up Detox in my Android project as a result, because I'm not familiar with native Android development. I thought `google` and `maven` _did_ belong at the same level as `allrepositories` instead of underneath it. When I tried this, my project would not build.

This update makes the indentation more clear so users are less likely to be confused or make the same error as I did. Now it shows:

image

### DIFF
--- a/docs/introduction/project-setup.mdx
+++ b/docs/introduction/project-setup.mdx
@@ -171,14 +171,15 @@ Open your main build script and apply the suggested changes:
 
  allprojects {
    repositories {
-   …
-   google()
+     …
+     google()
 // highlight-start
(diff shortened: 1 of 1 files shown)


d4vidi:
Thanks @CodingItWrong 😄

Issue #3973, opened 2023-03-07
cannot find symbol "Detox" on Android RN 0.70


arcln:
### What happened?

```
$ ./gradlew assembleAndroidTest -DtestBuildType=debug

[...]

> Task :app:compileDebugAndroidTestJavaWithJavac FAILED
/Users/arthur/Code/Mobile/temp/Sample/android/app/src/androidTest/java/com/sample/DetoxTest.java:3: error: cannot find symbol
import com.wix.detox.Detox;
                    ^
  symbol:   class Detox
  location: package com.wix.detox
/Users/arthur/Code/Mobile/temp/Sample/android/app/src/androidTest/java/com/sample/DetoxTest.java:4: error: package com.wix.detox.config does not exist
import com.wix.detox.config.DetoxConfig;
                           ^
/Users/arthur/Code/Mobile/temp/Sample/android/app/src/androidTest/java/com/sample/DetoxTest.java:22: error: cannot find symbol
        DetoxConfig detoxConfig = new DetoxConfig();
        ^
  symbol:   class DetoxConfig
  location: class DetoxTest
/Users/arthur/Code/Mobile/temp/Sample/android/app/src/androidTest/java/com/sample/DetoxTest.java:22: error: cannot find symbol
        DetoxConfig detoxConfig = new DetoxConfig();
                                      ^
  symbol:   class DetoxConfig
  location: class DetoxTest
/Users/arthur/Code/Mobile/temp/Sample/android/app/src/androidTest/java/com/sample/DetoxTest.java:27: error: cannot find symbol
        Detox.runTests(mActivityRule, detoxConfig);
        ^
  symbol:   variable Detox
  location: class DetoxTest
5 errors
```

### What was the expected behaviour?

The test APK builds successfully.

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [ ] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### Help us reproduce this issue!

```bash
# create 0.70.7 project / this also occurs on 0.71.3
npx react-native@0.70.7 init Sample --version 0.70.7
cd Sample

# add detox
npm i detox

# follow step 4 of https://wix.github.io/Detox/docs/introduction/project-setup#step-4-additional-android-configuration

# build test apk
[...]

PR #3995, opened 2023-03-23
Update project-setup.mdx

bruteforks:
To clarify setup, as currently it's very easy for your eyes to just see the google() portion and put it in the wrong place. Spent 2 days struggling with this, and others have too. https://github.com/wix/Detox/issues/3973

## Description

- This pull request addresses the issue described here: #3973 

---

> _For API changes:_

### DIFF
--- a/docs/introduction/project-setup.mdx
+++ b/docs/introduction/project-setup.mdx
@@ -148,6 +148,7 @@ patch or create if they are missing:
 ### 4.1. Patching build scripts
 
 Open your main build script and apply the suggested changes:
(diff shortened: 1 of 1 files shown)


d4vidi:
Thanks @bruteforks! I think we should revisit whether `allprojects` is in fact needed at all. It is very likely that simply adding the maven url to the top-level `repositories` section (which everyone has) could suffice.

d4vidi:
> sure thing! glad it can help someone else!

This is the power of the community 💪🏻

Issue #3655, opened 2022-10-28
[Genymotion SaaS] Device name conflict after environment setup timeout


noomorph:
### What happened?

There were two tests: `test1` and `test2`.

Detox started a Genymotion SaaS device for `test1`. The environment setup failed with a timeout.

Then we switched to `test2`. Detox tried to allocate a device with the same name and failed, because Genymotion SaaS does not allow to create devices with the same name.

### What was the expected behaviour?

Perhaps, Detox should have destroyed the first device if there were some problems with it.

Then the second test would have created a device with the same name again without any problem.

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [X] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### In what environment did this happen?

Detox version: 20.x.x
Test-runner (select one): jest-circus

### More data, please!

![image](https://user-images.githubusercontent.com/1962469/198636868-0e9da8bd-469b-4501-9e30-bce7328151c3.png)

noomorph:
We should not forget to fix this issue as a part of https://github.com/wix/Detox/issues/3573

PR #4005, opened 2023-04-10
feat(drivers): add postAllocate() method

noomorph:
## Description

- This pull request addresses the issue described here: #3655

In this pull request, I have added a new method, `postAllocate` to the device allocation drivers, which is an optional method to handle operations after the device is allocated, such as waiting until the boot animation is finished, configuring the device, etc.

```diff
class MyNewAllocationDriver {
  constructor(deps) {
    this.emitter = deps.eventEmitter;
  }
  async allocate(deviceConfig) {
    // ...
    return new Cookie(id); // This is where a cookie is formed once for the entire process
  }
+  async postAllocate(cookie) {
+   // Optional method to handle operations after the device is allocated
+    // like waiting until the boot animation is finished, configuring the device, etc.
+  }
  async free(cookie, options) {
    // ...
  }
}
```
This helps to separate "acquiring a _cookie_" from "finishing the preparation" so that even in case the preparation goes wrong, the layer responsible for the teardown is guaranteed to receive the cookie and be able to clean up.

Previously, a broken post-cookie preparation resulted in us not having a cookie to use when freeing the device – i.e., not freeing the device. So, I rewrote our allocation drivers to leverage this method. This simplified a few `try-catch` places where we were aware of this problem yet applied a workaround (and missed this problem with Genymotion drivers). 

### DIFF
--- a/detox/src/devices/allocation/drivers/android/genycloud/GenyAllocDriver.js
+++ b/detox/src/devices/allocation/drivers/android/genycloud/GenyAllocDriver.js
@@ -17,6 +17,7 @@ class GenyAllocDriver extends AllocationDriverBase {
     this._recipeQuerying = recipeQuerying;
     this._instanceLauncher = instanceLauncher;
     this._instanceAllocationHelper = allocationHelper;
+    this._launchInfo = {};
   }
 
   /**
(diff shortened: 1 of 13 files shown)


detox/src/devices/allocation/drivers/android/genycloud/GenyAllocDriver.js
```diff
@@ -28,21 +29,24 @@ class GenyAllocDriver extends AllocationDriverBase {
...
+   * @param {GenycloudEmulatorCookie} cookie
+   * @returns {Promise<void>}
+   */
+  async postAllocate(cookie) {
+    const { instance } = cookie;
+    const { isNew } = this._launchInfo[instance.uuid];
+    const readyInstance = cookie.instance = await this._instanceLauncher.launch(instance, isNew);
 
+    const { adbName } = readyInstance;
     await this._adb.disableAndroidAnimations(adbName);
```
noomorph, line 47: @d4vidi, you might review this pull request post-factum and wonder why I "kinda applied an overkill", but there was a real reason.

In case with that Genymotion bug (referenced in this PR), the problem was that it DID NOT throw any errors, it was just stuck for minutes-minutes (eternity). So the deallocation just never happened. This `try-catch` was simply not enough to handle this kind of errors.

`init()` - `cleanup()` cycles in DetoxWorker are controlled much tighter (we have `yield`, CAF cancellation, and Jest environment timers which correctly manage even the messiest situations), so I had no better choice than to go up one level and join the YIELD* party in init.

noomorph: And if you ever wonder why I didn't try to fix this solely via #4006 – well, in my opinion that's a too local and specific fix which conceptually doesn't guarantee that this situation won't happen again. Splitting code into pre-cookie and post-cookie is much safer when it comes to deallocating via cookie, and that provides a generic solution to this issue.

PR #4008, opened 2023-04-12
Update docs related to detoxURLBlacklistRegex

arthuralee:
## Description

According to the v18 migration guide here https://wix.github.io/Detox/docs/guide/migration/#180, the escaping of the regex for detoxURLBlacklistRegex is no longer required. This PR updates the docs to reflect this

### DIFF
--- a/docs/api/device.md
+++ b/docs/api/device.md
@@ -229,7 +229,7 @@ Useful if the app makes frequent network calls to blacklisted endpoints upon sta
 ```js
 await device.launchApp({
   newInstance: true,
-  launchArgs: { detoxURLBlacklistRegex: '\\("^http://192\.168\.1\.253:d{4}/.*","https://e\.crashlytics\.com/spi/v2/events"\\)' },
+  launchArgs: { detoxURLBlacklistRegex: '("http://192.168.1.253:d{4}/.*","https://e.crashlytics.com/spi/v2/events")' },
 });
 ```
 


docs/api/device.md
```diff
@@ -229,7 +229,7 @@ Useful if the app makes frequent network calls to blacklisted endpoints upon sta
 ```js
 await device.launchApp({
   newInstance: true,
-  launchArgs: { detoxURLBlacklistRegex: '\\("^http://192\.168\.1\.253:d{4}/.*","https://e\.crashlytics\.com/spi/v2/events"\\)' },
+  launchArgs: { detoxURLBlacklistRegex: '("http://192.168.1.253:d{4}/.*","https://e.crashlytics.com/spi/v2/events")' },
```
noomorph, line 232:
```suggestion
  launchArgs: { detoxURLBlacklistRegex: '("^http://192.168.1.253:d{4}/.*","https://e.crashlytics.com/spi/v2/events")' },
```

noomorph: Nevertheless, I think we should revisit this code in Swift and check out its implementation. I already have forgotten how it works 😅 

Sorry, it will take some time. Maybe @asafkorem could verify this correction to docs?

asafkorem, line 232: @arthuralee Unfortunately, that's no longer valid. We updated this documentation recently.

The backward slashes are required before the brackets, since we're passing it as launch argument with a string type. Removing the slashes means that this type will be passed as an array, and [this line](https://github.com/wix/Detox/blob/2bd2f4a79ce8dbfdae06ffe9d6921122bb03aeb7/detox/ios/Detox/DetoxInit.m#L70) will end with `blacklistRegex == nil` (since there's no string with that key, but an array of strings).

Also, we need to escape those dots with backward slashes to avoid handling them as regex patterns.

Regardless, based on the [current implementation](https://github.com/wix/Detox/blob/2bd2f4a79ce8dbfdae06ffe9d6921122bb03aeb7/detox/ios/Detox/DetoxInit.m#L70), the real issue with this `detoxURLBlacklistRegex` is that the implementation doesn't interpret or handle the regex pattern `:d{4}`. The code will not recognize that pattern as regex and will include it as plain text in the blacklisted urls.

asafkorem, line 232:
```suggestion
  launchArgs: { detoxURLBlacklistRegex: '\\("^http://192\.168\.1\.253:\\d{4}/.*","https://e\.crashlytics\.com/spi/v2/events"\\)' },
```

asafkorem (changes requested):

asafkorem:
The error I mentioned [on this comment](https://github.com/wix/Detox/pull/4008#discussion_r1170038251) was fixed here: https://github.com/wix/Detox/pull/4017 (the port component).

I am closing this PR because the changes mentioned do not fix the documentation.

PR #4011, opened 2023-04-13
Update fs-extra from 4.x to 9.x

stianjensen:
## Description

This preserves compat with node versions back to 10.x

https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md

There are no API-level breaking changes that seem to affect behavior (tests pass), only deprecated node versions and some refactored the behavior of a few APIs – for instance more use of native node APIs instead of polyfills.


stianjensen:
Superseded by #4013

PR #4020, opened 2023-04-18
chore(dependencies): lock node-ipc version to 9.2.1.

asafkorem:
According to https://github.com/RIAEvangelist/node-ipc/issues/3#issuecomment-1457937499:

> The original (before wipe) latest versions without the "protestware" were:
> * For version v9: v9.2.1
> * For version v10/v11: v10.1.0


asafkorem:
close due to duplication https://github.com/wix/Detox/pull/4019

PR #4060, opened 2023-05-16
Restore usage of launch-args lib

d4vidi:
## Description

- This pull request addresses the issue described here: https://github.com/iamolegga/react-native-launch-arguments/issues/44

In this pull request, I have restored the usage of `react-native-launch-arguments` in Detox's self-test project.

Context: While revisiting https://github.com/iamolegga/react-native-launch-arguments/issues/44, it has occurred to me through the details of the problem that its root could be in a react native bug. This is due to the following circumstances:
- The bug stems from `LaunchArgumentsModule.getConstants()` returning an empty map
- Source code shows this can happen if there is no intent available (unlikely) or otherwise in case there's no activity available (might be?).
- Deduced: because `getConstants()` is a react-native framework-level hook, it doesn't make sense for it to be called before there's an activity involved. Ergo, an RN bug is likely.

All of this has led me to test out again from scratch whether this happens with our current default, RN `0.70.x`. So far, seems stable enough - in particular, when compared to the flakiness seen with older RN versions.

---

> _For API changes:_

### DIFF
--- a/detox/test/android/app/build.gradle
+++ b/detox/test/android/app/build.gradle
@@ -98,6 +98,7 @@ dependencies {
     implementation project(':react-native-community-checkbox')
     implementation project(':react-native-community-geolocation')
     implementation project(':react-native-datetimepicker')
+    implementation project(':react-native-launcharguments')
 
     androidTestImplementation(project(path: ':detox'))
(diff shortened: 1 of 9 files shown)


d4vidi:
Long story short, waiting for this to be released 🤞🏻: https://github.com/iamolegga/react-native-launch-arguments/pull/72

PR #4086, opened 2023-05-24
fix(iOS): add compatibility for building with Xcode 14.3

asafkorem:


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "detox/ios/COSTouchVisualizer"]
 	path = detox/ios/COSTouchVisualizer
-	url = https://github.com/joeblau/COSTouchVisualizer
+	url = https://github.com/wix-incubator/COSTouchVisualizer
```
asafkorem, line 3: FYI @d4vidi, this repo isn't active for years.. There were a few fixes that needed to be made. Of course I made a PR to the original repo.

Issue #3366, opened 2022-05-02
(WIP) Support regex in matchers


asafkorem:
#### Support regular expression in Detox matchers
Specifically in `by.text()`, `by.label()`.
This can be useful since text and labels may include dynamic content (e.g. date).

Also, I suggest to change the current API, from:
https://github.com/wix/Detox/blob/246317d3ee5b7587bb966c78204b5def5b72dfaf/detox/index.d.ts#L804

To:
```ts
text(text: string | RegExp): NativeMatcher;
```

---

#### Usage example

This code will find an element with the text `"The year is 2023"`, if exists:

```js
const someElement = element(by.text(/The year is: \d{4}/));
```

While this code won't:

```js
const someElement = element(by.text('The year is: \d{4}'));
```

noomorph:
Why not dynamically recognize `instanceof Regexp` versus string?

asafkorem:
> Why not dynamically recognize `instanceof RegExp` versus string?

Sounds great, let's do it this way.

noomorph:
@asafkorem the only pitfall I see is that we might need to standardize RegExp mechanisms between iOS/Android/JS...

<table>
<td>
<img width=300 src=https://user-images.githubusercontent.com/1962469/166209143-90048c25-cf1b-4c5e-b613-b55fc574c84d.png>
</td>
<td>
<img src=https://user-images.githubusercontent.com/1962469/166209174-597d49b0-9746-4d5a-8341-b0510405d892.png>
</td>
</table>

d4vidi:
@noomorph that is spot on. We've somehow manged to overcome that in the context of the URL's blacklist - we could revisit and hopefully replicate.

d4vidi:
Nevertheless I'd like this kind of any API to come with a warning of some sort, that would explains that ideally, whichever text we trying to match should be deterministic to the point it can be predetermined down to the character.
Of course, this is not always the case in unmocked tests, but generally this should be the preferred situation.

boxabirds:
This is super useful when working with iOS elements because you can't seem to expose testIDs (or equivalent) so I'm using `accessibilityLabel` to perform the function of both an identifier and additional state.

mariushorvat:
Is someone working on this issue?

noomorph:
Not really, @asafkorem 's been busy last months trying to release Detox 21 with XCUITest.

@asafkorem would you mind to take this issue to the final mile? Looks like there is demand for it, 10*👍 already

[...]

PR #4100, opened 2023-06-07
feat: support regex in matchers

andrius-girdzius:
## Description

**WIP**

- This pull request addresses the issue described here: #3366 

In this pull request, I have added support for Regex text/label/testID matching in Detox.

---

> _For API changes:_


detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/ViewMatchers.kt
```diff
@@ -54,6 +64,42 @@ fun toHaveSliderPosition(expectedValuePct: Double, tolerance: Double): Matcher<V
...
+    object: BoundedMatcher<View, View>(View::class.java) {
+        override fun describeTo(description: Description) {
+            description.appendText("withRegexContentDescription($jsRegex)")
+        }
+
+        override fun matchesSafely(view: View): Boolean {
+            val contentDescription = view.contentDescription?.toString()
+            return contentDescription?.isMatch(jsRegex) ?: false
+        }
+    }
```
asafkorem, line 101: You can avoid this duplicated structure by adding a factory method for those regex-matchers:

```
private fun createRegexMatcher(matcherDescription: String, regex: String, extractStringFromView: (View) -> String?): Matcher<View> =
    object: BoundedMatcher<View, View>(View::class.java) {
        override fun describeTo(description: Description) {
            description.appendText("$matcherDescription($regex)")
        }

        override fun matchesSafely(view: View): Boolean {
            val extractedString = extractStringFromView(view)
            return extractedString?.isMatch(regex) ?: false
        }
    }
```

It might even be worth extracting it to another file 🤔

asafkorem: Usage example:

```kotlin
private fun withRegexText(jsRegex: String): Matcher<View> =
    createRegexMatcher("withRegexText", jsRegex) { view -> (view as? TextView)?.text?.toString() }
```

asafkorem: Also, those methods should be private since you're using them in this file context only

asafkorem: Another small note, sorry if it seems nitpicky, but personally, I prefer ordering methods according to their "caller-callee" relationship, where a method follows its caller when possible (if the method has not been declared before). This approach improves readability and mirrors the logical flow of execution in the code

For example, in this order, `withRegexText` will follow `withText`

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/ViewMatchers.kt
```diff
@@ -54,6 +64,42 @@ fun toHaveSliderPosition(expectedValuePct: Double, tolerance: Double): Matcher<V
...
 
+fun withRegexText(jsRegex: String): Matcher<View> =
+    object: BoundedMatcher<View, TextView>(TextView::class.java) {
+        override fun describeTo(description: Description) {
+            description.appendText("withRegexText($jsRegex)")
+        }
+
+        override fun matchesSafely(view: TextView): Boolean {
+            val text = view.text?.toString()
+            return text?.isMatch(jsRegex) ?: false
```
asafkorem, line 75: Have you considered replacing the custom `withText` method with this one? Instead of deciding between different matchers (`withRegexText` and Espresso's `withText`) prior to their invocation, you could simply use an if-else in the return value, like this: return `isRegex ? (value?.isMatchJSRegex(expected) ?: false) : value == expected`. Same Q for the other regex-matchers

asafkorem: Honestly I'm not sure if it will be better, WDYT?

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/RegexMatcher.kt
```diff
@@ -0,0 +1,42 @@
...
+// Returns whether the whole string matches the given `jsRegex`.
+// JS flags has the format of `/<pattern>/<flags>`.
+// Flags can be either:
+// - i: With this flag the search is case-insensitive: no difference between A and a (see the example below).
+// - s: Enables “dotall” mode, that allows a dot . to match newline character \n (covered in the chapter Character classes).
+// - m: Multiline mode (covered in the chapter Multiline mode of anchors ^ $, flag "m").
+// Other flags (e.g. g,u,s) are not supported as they do not have equivalents in Kotlin.
+//
+// - See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
+fun String.isMatch(jsRegex: String): Boolean {
```
asafkorem, line 12: I think calling it `matchesJSRegex` would be more self-explanatory. `isMatch` is too broad

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/RegexMatcher.kt
```diff
@@ -0,0 +1,42 @@
...
+    val flagsChars = getRegexFlags(jsRegex)
+    val options = getRegexOptions(flagsChars)
+    val pattern = getRegexPattern(jsRegex)
+    return Regex(pattern, options).matches(this)
+}
+
+fun getRegexFlags(jsRegex: String): CharSequence {
```
asafkorem, line 19: add the `private` modifier where possible

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/RegexMatcher.kt
asafkorem: I know I know, that's what I called it in iOS; I take the responsibility. The same comment applies 😅

detox/src/android/AndroidExpect.test.js
```diff
@@ -43,6 +43,9 @@ describe('AndroidExpect', () => {
       await e.expect(e.element(e.by.accessibilityLabel('test'))).toHaveText('text');
       await e.expect(e.element(e.by.accessibilityLabel('test'))).toNotHaveText('text');
       await e.expect(e.element(e.by.accessibilityLabel('test'))).not.toHaveText('text');
+      await e.expect(e.element(e.by.accessibilityLabel('test'))).toHaveText(/text/);
+      await e.expect(e.element(e.by.accessibilityLabel('test'))).toNotHaveText(/text/);
+      await e.expect(e.element(e.by.accessibilityLabel('test'))).not.toHaveText(/text/);
```
asafkorem, line 48: add the same for label and test-id matchers

detox/src/android/espressoapi/DetoxMatcher.js
```diff
@@ -14,39 +14,51 @@ function sanitize_matcher(matcher) {
   return originalMatcher.type ? originalMatcher.value : originalMatcher;
 } 
 class DetoxMatcher {
-  static matcherForText(text) {
+  static matcherForText(text, isRegex) {
     if (typeof text !== "string") throw new Error("text should be a string, but got " + (text + (" (" + (typeof text + ")"))));
+    if (typeof isRegex !== "boolean") throw new Error("isRegex should be a boolean, but got " + (isRegex + (" (" + (typeof isRegex + ")"))));
```
asafkorem, line 19: Consider extracting these assertions into a separate method. Doing so would:
- Minimize code duplication
- Adhere to the Single Responsibility Principle (SRP)

asafkorem (changes requested):
You've done some good work here, I'm really excited to see this! 
Kindly let me know when it's ready for another review, post exiting draft mode

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/ViewMatchers.kt
andrius-girdzius: Yeah I believe it would be nicer code-wise to eliminate use of Espresso's `withText`.

My main concern was changing too much - I only briefly looked at the Espresso's implementation so I decided to simply add on regex matching and use their implementation elsewhere, just in case my own implementation would break existing Detox consumers. 

But I think I can open the source code again and have a deeper look to eliminate use of Espresso here, if you think that's okay 👍

detox/src/android/espressoapi/DetoxMatcher.js
andrius-girdzius: this file is autogenerated based on params on `DetoxMatcher.java`. Did you mean I should edit the generation code?

asafkorem: ahh, right. never mind then 😆

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/ViewMatchers.kt
asafkorem: Generally, that's a great thinking. It makes sense not to disrupt existing usage. 
A thorough implementation, however, poses no risk if done correctly

asafkorem: The simpler their implementation, the safer it is to "copy-paste" it and change it according to our needs..
If it adds too much code to our implementation it's suspicious, because it looks like a simple `value == expected` check, then we might miss something in our regex implementation 😅 

I prefer the simpler solution for readability and cleanliness, as you see fit

asafkorem:
Upon quick review, I must say that this is a masterpiece. Great work! 🙌🏼 👑 
I might have some minor comments but it's quite there, I'll do another round later today / tomorrow

andrius-girdzius:
> Upon quick review, I must say that this is a masterpiece. Great work! 🙌🏼 👑 I might have some minor comments but it's quite there, I'll do another round later today / tomorrow

@asafkorem Sounds great! Looking forward to the comments, fixing and merging! 🥳

docs/api/matchers.md
```diff
@@ -42,6 +42,20 @@ Match elements with the specified text.
...
+| Flag | Name          | Modification                                                                                                                                       |
+|------|---------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| `i`  | Ignore Casing | Makes the expression search case-insensitively.                                                                                                    |
+| `s`  | Dot All       | Makes the wild character . match newlines as well.                                                                                                 |
+| `m`  | Multiline     | Makes the boundary characters ^ and $ match the beginning and ending of every single line instead of the beginning and ending of the whole string. |
+
+**Note:** JS flags like `g` (global) and `y` (sticky) that are not supported, as well as `u` (unicode) which is always implied, are ignored when parsing input.
+
+```js
+element(by.text(/Tap [A-Za-z]+/i));
```
asafkorem, line 56: I love it. Minor suggestion for improvement:

```suggestion
You can also utilize regex (Regular Expressions) alongside certain flags. Here's a table with the supported flags:

| Flag | Name          | Modification                                                                                                                                       |
|------|---------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
| `i`  | Ignore Casing | Makes the expression search case-insensitively.                                                                                                    |
| `s`  | Dot All       | Makes the wild character `.` match newlines as well.                                                                                                 |
| `m`  | Multiline     | Makes the boundary characters (`^` and `$`) match the beginning and ending of every single line instead of the beginning and ending of the whole string. |

**Note:** JS flags like `g` (global) and `y` (sticky) that are not supported, as well as `u` (unicode) which is always implied, are ignored when parsing input.

The following sample code snippet matches text starting with "Tap" followed by any number of alphabetic characters, case-insensitively:

```js
element(by.text(/Tap [A-Za-z]+/i));
```

asafkorem (changes requested):
Again, great work!! I'm super excited with this! Just a few small tweaks and we can ship it :shipit:

detox/android/detox/src/full/java/com/wix/detox/espresso/matcher/RegexMatcher.kt
```diff
@@ -0,0 +1,62 @@
...
+        val stringItem = item.toString()
+        return stringItem.matchesJSRegex(jsRegex)
+    }
+
+    override fun describeTo(description: Description) {
+        description.appendText("should match the pattern: $jsRegex")
+    }
+
+    companion object {
+        fun <T> matchesRegex(jsRegex: String): RegexMatcher<T> {
```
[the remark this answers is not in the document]
asafkorem: However not sure if we need that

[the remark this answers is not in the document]
andrius-girdzius: ah, that's true, it's not really needed here 💡

andrius-girdzius:
> Again, great work!! I'm super excited with this! Just a few small tweaks and we can ship it :shipit:

I've added the fixes, also modified `matchers.md` slightly to have a separate section on regex matching (since it is relevant for multiple matchers) and added links to it from the supported methods. 🥼

asafkorem (approved):

asafkorem:
> > Again, great work!! I'm super excited with this! Just a few small tweaks and we can ship it :shipit:
> 
> I've added the fixes, also modified `matchers.md` slightly to have a separate section on regex matching (since it is relevant for multiple matchers) and added links to it from the supported methods. 🥼

GG!

detox/src/ios/expectTwo.js
```diff
@@ -10,6 +10,7 @@ const { assertEnum, assertNormalized } = require('../utils/assertArgument');
 const { removeMilliseconds } = require('../utils/dateUtils');
 const { actionDescription, expectDescription } = require('../utils/invocationTraceDescriptions');
 const log = require('../utils/logger').child({ cat: 'ws-client, ws' });
+const isRegExp = require('../utils/regexp');
```
noomorph, line 13: For consistency, I would suggest:

1) to rename the file to `utils/isRegExp.js`
2) use named exports, e.g.: `const { isRegExp } = require(...)`

detox/src/utils/regexp.js
```diff
@@ -0,0 +1,5 @@
+function isRegExp(obj) {
+  return Object.prototype.toString.call(obj) === '[object RegExp]';
```
noomorph, line 2: I assume this is because of the different sandboxes (Jest), right?

andrius-girdzius: Yes, simply checking `instanceof` fails in this case, so I had to use a workaround

docs/api/matchers.md
```diff
@@ -117,3 +135,21 @@ On iOS, matched elements are sorted by their x and y axes.
...
+
+| Flag | Name          | Modification                                                                                                                                             |
+|------|---------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `i`  | Ignore Casing | Makes the expression search case-insensitively.                                                                                                          |
+| `s`  | Dot All       | Makes the wild character `.` match newlines as well.                                                                                                     |
+| `m`  | Multiline     | Makes the boundary characters (`^` and `$`) match the beginning and ending of every single line instead of the beginning and ending of the whole string. |
+
+**Note:** JS flags like `g` (global) and `y` (sticky) that are not supported, as well as `u` (unicode) which is always implied, are ignored when parsing input.
```
noomorph, line 149: I'd suggest to create an admonition and warn about possible issues with look-behind expressions (iOS/Swift doesn't support them where as Kotlin supports), e.g.:

image

asafkorem: @noomorph sounds good. Great catch with the "look behind" assertions! IMO we should provide this information inline.

```suggestion
:::caution Note

Regular expression flags such as `g` (global) and `y` (sticky) that are not supported, as well as `u` (unicode) which is always implied, are ignored when parsing input. 

Pay attention that as of writing this note, Android supports lookbehind assertions in its regular expression implementation, while iOS does not. It's advisable to check the platform-specific documentation for limitations.

:::
```

asafkorem: Due to the fact that this is only a button tap (+ to avoid the bikeshedding), I will allow myself to commit this change 🙂

asafkorem:
Released with v20.11.0 🎉 
Thanks again @andrius-girdzius! 

7rioa1

PR #4124, opened 2023-07-10
Ensure correct working directory

martintreurnicht:
# Problem
 this Gradle script assumes that the working directory is always the android working directory. When you run something like fastlane, that runs gradle scripts from your react-native project root, the working directory is NOT the android directory

# Solution
Use gradle's `project.rootDir` variable instead to ensure that we're always in the right working directory instead of assuming this to be true

### DIFF
--- a/detox/android/rninfo.gradle
+++ b/detox/android/rninfo.gradle
@@ -1,6 +1,6 @@
 import groovy.json.JsonSlurper
 
-def rnVersion = getRNVersion()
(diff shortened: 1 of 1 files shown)


noomorph:
@martintreurnicht, hi and thanks for your contribution! Our team has low availability today due to various reasons. I hope Amit or Asaf can look soon.

gosha212 (approved):

Issue #1331, opened 2019-04-24
Support for single reinstall in Jest multi worker mode


META-DREAMER:
**Is your feature request related to a problem? Please describe.**
Carrying on from the discussion here: https://github.com/wix/Detox/issues/1279

**Describe the solution you'd like**
Detox should only install the app once per worker instead of reinstalling for every test file.

**Describe alternatives you've considered**
[There's some workarounds](https://github.com/wix/Detox/pull/1328) using `--reuse`, but they only work with `--runInBand` or `--workers 1`.

hammadj:
@rotemmiz The workaround with `--reuse` could work, but there's some caveats. With Jest, the global `beforeAll` and `afterAll` hooks will also run for each worker, so that needs to be taken into account since users will have `detox.init()` in `beforeAll` as well as `adapter.afterAll()` and `detox.cleanup()` in `afterAll`.

noomorph:
I have some ideas on this but I need to conduct my own investigation.

sarthakmis:
detox test --reuse worked for me

sbycrosz:
Hey, we managed to solve this based on [jest-environment-puppeteer](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-environment-puppeteer) implementation

What we did:

1. Spawn n amount of detox web-socket servers, where n is the amount is the number of jest workers specified
1. Pass generated session (web-socket-server-url and uuid) to each test suite with the process.env hack (which should be easy to change when jest add a new API to pass primitives https://github.com/facebook/jest/issues/7184#issuecomment-492122367)
1. On each test suite, initialize detox with session generated from step 1

Proof of concept: [jest-environment-detox](https://github.com/stashaway-engineering/jest-environment-detox)

What do you think? @rotemmiz

hammadj:
@noomorph The solution by @sbycrosz works really well! Any way we can integrate it into detox itself?

noomorph:
@hammadj, must have been fallen between the cracks, sorry. Looks interesting, I'll look closer.

noomorph:
Can't look in the nearest future, sorry, maybe on Friday but not likely. Got many other tasks at the moment.

hammadj:
@noomorph No worries, we are using it in our project successfully, no rush to integrate right away. Thanks for your attention and support on this!

sbycrosz:
Hey, can we integrate this into detox? We found there are some changes in the internal APIs that prevents the workaround from working 😢

[...]

Issue #2106, opened 2020-06-07
Launch emulators sequentially


d4vidi:
**Is your feature request related to a problem? Please describe.**
The Android emulator binary doesn't seem to handle many launches taking place at the same time very well. In particular, seems that emulator launches touch some of the configuration, and - if race conditions are met, that evidently causes some emulators either boot from scratch (overlook snapshots) due to an alleged config change or possibly boot with the wrong specs.

Note: This somewhat associates with the `adb jam` problem (#1857), as it is possible that as a side effect, improving the launch behavior would help reduce the frequency of "jams": as the tests-run timeline changes, it should become less likely that multiple `adb install` commands are executed simultaneously, which seems to be the soft spot of adb with respect to that specific bug.

**Describe the solution you'd like**
Make emulators launch as a mutually-excluded sequence.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

d4vidi:
Following recent futile attempts at getting this to work properly, I'm "back-logging" it again with this possible change as the next potential PoC: Stop using the emulators' `-read-only` and clone emulator instances on-the-fly, instead (as implemented for iOS).

d4vidi:
Closing; We will revisit if necessary

PR #4173, opened 2023-08-25
feat: new device allocation lifecycle

### LINKED ISSUE #2894: Fix cleanup in multi-worker envs (DAS project)
## Description

**DAS** = **D**evice **A**llocation **S**erver

A rearchitecture of Detox' device allocation concern. In a nutshell - we want to have that concern either implemented as a global-scope managed server, or integrated into the existing Detox-server itself*.

The issue address the **1st phase** in what could be an inception of further future improvements. It's scope is (roughly):
- [x] Make existing Detox server managed at global scope (rather than having distinct servers bootstrapped and killed over and over again based on workers' lifecycles).
- [x] Setting up this kind of a server at global-scope (either as an independent entity, or bundled into the existing Detox server). It will strictly be running at local host, for the time being, so as to avoid adding this up into a bigger, hard-to-manage change.
- [x] De-facto activating the server...

(@noomorph this is what I had in mind, obviously you're the one to decide the best way to move forward!)

### LINKED ISSUE #3573: [Epic] DAS Phase 1: Global device allocation
### Description

When using the `android.genycloud` device type, it is not possible to run multiple `detox test …` commands in parallel like that:

```bash
detox test -c android.geny.release -n Recipe1 & \ # session.id = e54490fd-7cb7-e05a-586e-c3ec23540496
detox test -c android.geny.release -n Recipe2 & \ # session.id = 1acee489-e9d3-4a82-8e50-24f723f60ff8
wait
```

The problem is that the fastest command would go over the `~/Library/Detox/genycloud-cleanup.lock`, e.g.:

```json
[
  {"id":"d9d09cbf-df0b-4044-ae07-32d794a65d49","data":{"name":"Detox.e54490fd-7cb7-e05a-586e-c3ec23540496.w1"}},
  {"id":"39178451-9b6f-43b9-9053-7321294255ab","data":{"name":"Detox.1acee489-e9d3-4a82-8e50-24f723f60ff8.w1"}},
]
```

and stop all the instances, including the ones that do not belong to its test session.

In this example, if the session `e54490fd-7cb7-e05a-586e-c3ec23540496` completes faster and cleans up, then the slower one,`1acee489-e9d3-4a82-8e50-24f723f60ff8`, will start failing after this, because its device will have been stopped by that moment.

### Suggested solution

Move the device allocation to the primary (global) context and stop using lock files in the Genymotion SaaS drivers. The global allocation mechanism will track all the devices created throughout the test session, but since it will be happening in memory, there will no longer be a shared file to fight over. Voila! No file – no problem.

### Gameplan

```[tasklist]
- [ ] #3655
- [ ] #2106
- [x] #2894 (`--u, --cleanup` works incorrectly in Jest (devices are being shut down after each test suite)
- [x] Race conditions when running multiple "detox test ..." commands in parallel
```

### LINKED ISSUE #3655: [Genymotion SaaS] Device name conflict after environment setup timeout

noomorph:
Resolves #2894.
Resolves #3573.
Resolves #3655.

- [ ] #3573
- [ ] #3655
- [ ] #2894 (`--u, --cleanup` works incorrectly in Jest (devices are being shut down after each test suite)

Enabler for:

- [ ] #1331 (likely, the Android-specific PR won't be very needed after that: #3497 )

## Description

1. Device allocation now happens in the Detox primary context (the top parent process). That allows storing all the allocation data and history in memory and opens many technical possibilities. The Detox worker (via the secondary context) sends IPC commands to allocate and deallocate a device respectively.
2. Device allocator drivers now have `init`, `cleanup` and `emergencyCleanup` methods for the global lifecycle handling. 
3. Device registry (lock file wrapper) can register a device, release a device (without unregistering it), and unregister it completely. Every device is associated with Detox test session id to prevent stealing from a concurrent Detox run.
4. Genymotion SaaS allocation driver keeps track of all the allocated devices in memory, making it impossible to cause conflicting lock file resets. It is not limited to one device per worker - if device allocation fails, the worker can retry and allocate one more device.
5.  Limited `deviceShutdown` event only to `device.resetContentAndSettings()` on iOS for the artifacts manager. Also, `bootDevice` event is now synthetic - used only to make sure that the artifacts manager knows a correct device id. 
6. Minor refactors in the utility classes.


d4vidi:
@noomorph I've switched to tracking these tasks under #3573, which is an issue rather than a PR.

d4vidi:
Is #2106 in fact necessary here? Is it necessary at all? It used to address issues with local emulators running on Linuxes and I don't think it's of high interest anymore.

noomorph:
@d4vidi #2106 was cheap enough to implement, and I believe it has positive impact on community.

noomorph:
Merged via 9bb74739f1fbdcdf6678926fccfb806c50b2c8a7.

PR #4180, opened 2023-09-06
feat(android): add support for multiple-views actions.

asafkorem:
This PR enhances Detox's ability to work with actions on multiple matching views on Android. The aim is to provide greater flexibility for actions, allowing them to handle both single and multiple view scenarios declaratively.

### Motivation

As of now, Detox struggles with scenarios where `getAttributes()` is required to operate on several matched elements on Android, even though this is supported on iOS. This PR lays the groundwork to rectify this discrepancy, paving the way for full implementation in a subsequent PR.

See: https://github.com/wix/Detox/issues/3196

### Problem

- The existing `perform` method in the `EspressoDetox` class relies on `ViewInteraction`, which doesn't naturally cater to multiple matching views.

### Solution

- To sidestep the challenge posed by `viewInteraction = onView(matcher)`, we transitioned the `perform` method in `EspressoDetox` to directly use `Matcher<View>` instead of `ViewInteraction`.
  
- This adjustment enable us to iterate on the the matching views by using the by-index functionality of `Matcher`.
  
- Furthermore, a new marker interface, `MultipleViewsAction`, has been introduced. Actions having this mark can handle and act upon all views that match.

---

### Usage Examples

```java
public class SomeMultiViewAction implements ViewAction, MultipleViewsAction {
    @Override
    public void perform(UiController uiController, View view) {
        // Your implementation goes here...
    }
}

perform(someMatcher, new SomeMultiViewAction());
```

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/EspressoDetox.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/EspressoDetox.java
@@ -1,5 +1,7 @@
 package com.wix.detox.espresso;
 
+import com.wix.detox.espresso.performer.ViewActionPerformer;
+
 import android.app.Activity;
 import android.content.Context;
 import android.content.ContextWrapper;
@@ -20,6 +22,7 @@ import java.util.ArrayList;
 import androidx.test.espresso.UiController;
 import androidx.test.espresso.ViewAction;
 import androidx.test.espresso.ViewInteraction;
+import androidx.test.espresso.NoMatchingViewException;
 import androidx.test.platform.app.InstrumentationRegistry;
 
 import static androidx.test.espresso.Espresso.onView;
@@ -32,13 +35,8 @@ public class EspressoDetox {
     private static final String LOG_TAG = "detox";
 
     public static Object perform(Matcher<View> matcher, ViewAction action) {
-        ViewInteraction interaction = onView(matcher);
-        interaction.perform(action);
-
-        if (action instanceof ViewActionWithResult) {
-            return ((ViewActionWithResult) action).getResult();
-        }
-        return null;
+        ViewActionPerformer performer = ViewActionPerformer.forAction(action);
+        return performer.performOn(matcher);
     }
 
     public static Activity getActivity(Context context) {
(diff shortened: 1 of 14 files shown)


asafkorem:
This change still needs some written tests. But we can still start the review @d4vidi

detox/src/android/core/NativeElement.js
```diff
@@ -20,7 +20,7 @@ class NativeElement {
   }
 
   _selectElementWithMatcher(matcher) {
-    this._call = invoke.call(invoke.Espresso, 'onView', matcher._call);
+    this._call = matcher._call;
```
asafkorem, line 23: Refactor

asafkorem, line 23: Refactor

detox/src/android/core/NativeElement.js
```diff
@@ -15,39 +15,37 @@ class NativeElement {
...
     this._emitter = emitter;
-    this._originalMatcher = matcher;
-    this._selectElementWithMatcher(this._originalMatcher);
+    this._matcher = matcher;
   }
 
-  _selectElementWithMatcher(matcher) {
-    this._call = invoke.call(invoke.Espresso, 'onView', matcher._call);
+  get _call() {
+    return invoke.call(invoke.Espresso, 'onView', this._matcher._call);
```
asafkorem, line 22: consider refactor detox-assertions as well

detox/android/detox/src/full/java/com/wix/detox/espresso/EspressoDetox.java
```diff
@@ -31,15 +32,35 @@
...
-        interaction.perform(action);
+    public static Object perform(Matcher<View> matcher, ViewAction action) {
+        if (action instanceof MultipleViewsAction) {
+            performOnViews(matcher, action);
+        } else {
+            onView(matcher).perform(action);
+        }
 
         if (action instanceof ViewActionWithResult) {
             return ((ViewActionWithResult) action).getResult();
```
asafkorem, line 43: aggregate the the results for multiple elements

detox/src/android/core/NativeElement.js
asafkorem: in a following PR.

PR #4115, opened 2023-06-26
fix(jest): basic --bail support

noomorph:
## Description

Related to https://github.com/jestjs/jest/issues/14143

This PR attempts to bridge the challenging integration between Jest's `-bail`, Detox's `--retries`, and the intertwined bugs and constraints. Here's a brief overview of the obstacles addressed:

1. **`DetoxReporter#onRunComplete` invocation**: Ensure it's called only once using cached promises and deferreds. Concurrent failures in other workers might trigger redundant invocations of the same callback.
2. **Tracking Reported Test Files**: Handle scenarios where `numTotalTestSuites` is greater than the count of known test files. In such cases, critical information for scheduling another re-run via Detox's `--retries` becomes unavailable.
3. **`globalTeardown` Management**: Ensure our global teardown is executed. Given Jest's apathy towards this, we manually invoke Detox's internal `cleanup()` function during early exits.
4. **Ensure Jest workers exit completely**: Facilitate smooth exits for Jest workers equipped with our custom Detox environment. By broadcasting an early-exit signal, we enable them to fail promptly upon encountering Jest being interrupted. Then they're able to clean up (or uninstall Detox workers) properly before halting.

While these interventions significantly improve the use of Jest's --bail, a notable limitation remains detailed in point (2). The best we can do is print a warning and give up on retrying.

PR #4199, opened 2023-09-25
fix: missing Jest summary

noomorph:
## Description

This fix addresses the bug highlighted in the discussion #4121.

The core issue stems from Jest's behavior: [it automatically adds a summary reporter when the default string is found in reporters](https://github.com/jestjs/jest/blob/a640c6513f08f6d97827050010a8a28fff530ac9/packages/jest-core/src/TestScheduler.ts#L344-L356). We recommend users exclusively use the Detox report as the primary one, given the challenges and potential problems arising with the default reporter, especially in contexts like `--bail`, and that's why we have to solve this problem on our side once again.

So the proposed solution is:
- Check if users have not yet specified a `summary` string in the reporters.
- If not, we morph DetoxReporter into a composition of Verbose and Summary reporters.

Furthermore, this fix tackles the problem of the test summary being redundantly printed. This duplication arises when onRunComplete is triggered several times, notably when Jest operates with --bail and multiple workers, and when onRunComplete unfolds as an asynchronous extended operation—a typical situation with custom reporters. Addressing this is pivotal for the integration of PR #4115.

### DIFF
--- a/detox/runners/jest/reporters/DetoxReporter.js
+++ b/detox/runners/jest/reporters/DetoxReporter.js
@@ -1,24 +1,77 @@
 const resolveFrom = require('resolve-from');
-/** @type {typeof import('@jest/reporters').VerboseReporter} */
+/** @type {new (globalConfig: any) => import('@jest/reporters').VerboseReporter} */
 const JestVerboseReporter = require(resolveFrom(process.cwd(), '@jest/reporters')).VerboseReporter;
+/** @type {new (globalConfig: any) => import('@jest/reporters').SummaryReporter} */
+const SummaryReporter = require(resolveFrom(process.cwd(), '@jest/reporters')).SummaryReporter;
 
 const { config, reportTestResults } = require('../../../internals');
 
 class DetoxReporter extends JestVerboseReporter {
+  constructor(globalConfig) {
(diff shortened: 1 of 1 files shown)


detox/runners/jest/reporters/DetoxReporter.js
```diff
@@ -1,24 +1,77 @@
...
+
+  /**
+   * @returns {import('@jest/reporters').SummaryReporter | null}
+   * @private
+   */
+  _initSummaryReporter() {
+    /** @type {(config: import('@jest/types').Config.ReporterConfig) => boolean} */
+    const isSummaryReporter = (config) => config[0] === 'summary';
+    if (this._globalConfig.reporters.some(isSummaryReporter)) {
+      return null;
```
d4vidi, line 71: `return StubReporter`? 🤔

d4vidi (commented):
OK, just one subtle suggestion (see inline).

Also, adding this on top of the review of #4115 (in retrospect, I should have reviewed in the opposite order...) -
Now that DetoxReporter aggregates multiple reporting needs, I strongly suggest we figure out a way to more clearly separate the various concerns, thus allowing for an easier comprehension of the what-belongs-to-who type of questions...

Issue #2846, opened 2021-06-18
getAttributes() does not work in detox 18.14.0 in Android


sandeep-bastola:
### Describe the bug
The `getAttributes()` function does not work in Android with Detox `18.14.0`. 

It throws an error :
`java.lang.ClassNotFoundException: com.google.android.material.slider.Slider`

### Steps To Reproduce

- [x] I have tested this issue on the latest Detox release (18.17.0) and it still reproduces

1. Issue appeared in detox version 18.14.0 . Release note fo this version mentions that the getAttributes() function works in Android.
2. Issue does not appear in iOS.

### Detox Trace-Logs

N/A

### Device logs (adb logcat)

```
06-18 09:40:11.001  9652  9773 D DetoxWSClient: Received action 'invoke' (ID #33, params={"target":{"type":"Class","value":"com.wix.detox.espresso.EspressoDetox"},"method":"perform","args":[{"type":"Invocation","value":{"target":{"type":"Class","value":"androidx.test.espresso.Espresso"},"method":"onView","args":[{"type":"Invocation","value":{"target":{"type":"Class","value":"com.wix.detox.espresso.DetoxMatcher"},"method":"matcherForTestId","args":["homeScreen_stars_counter_value"]}}]}},{"type":"Invocation","value":{"target":{"type":"Class","value":"com.wix.detox.espresso.DetoxAction"},"method":"getAttributes","args":[]}}]})
06-18 09:40:11.001  9652  9771 I DetoxDispatcher: Handling action 'invoke' (ID #33)...
06-18 09:40:11.002  9652  9771 D Detox   : class com.wix.detox.espresso.DetoxMatcher, matcherForTestId, [homeScreen_stars_counter_value]
06-18 09:40:11.004  9652  9771 D Detox   : class androidx.test.espresso.Espresso, onView, [(with tag value: is "homeScreen_stars_counter_value" and view has effective visibility=VISIBLE)]
06-18 09:40:11.004  9652  9771 D Detox   : class com.wix.detox.espresso.DetoxAction, getAttributes, []
06-18 09:40:11.007  9652  9771 D Detox   : class com.wix.detox.espresso.EspressoDetox, perform, [androidx.test.espresso.ViewInteraction@4ba662b, com.wix.detox.espresso.action.GetAttributesAction@ec0ca88]
06-18 09:40:11.013  9652  9652 D IdlingResourceRegistry: All idling resources are idle.
[...]

PR #4203, opened 2023-09-27
fix(android): extract material slider value attribute with reflection.

asafkorem:
This fixes a `ClassNotFoundException` error that is thrown when the Slider class from the `com.google.android.material.slider` package is not found at runtime.

The solution is to use reflection instead of importing the Slider class, which may not necessarily exist in the app.

Also, I made a light refactoring for get-attributes related code.


asafkorem:
Resolves https://github.com/wix-private/detox/issues/198 and https://github.com/wix/Detox/issues/2846

PR #4205, opened 2023-10-01
(Android) Make activity launch more supervised, debuggable

d4vidi:
## Description

- This pull request addresses the issue described here: N/A

In this pull request, I have made adjustments to the way the app's activity is launched by Detox, making it in a more supervised way and thus more debuggable. That includes the following:

1. React markers, noting important things such as Javascript loading, bundle downloading, native package initialization, etc. - are now logged by Detox. To avoid log spamming, only major "root" events are logged.
1. Before this change, the app's activity used to be launched in the very beginning of Detox (i.e. the beginning of the Detox native test, namely, in `Detox.java`). This means that things such as app crashes / ANR's were not reported to the test runner, if happening at a very early stage. This PR introduces a fundamental change where the connection to the Detox test-runner is made as the first step, with activity launching happening only following that. This also helps make sure that the react markers are not subscribed-to too late, resulting in the missing-out of some of the earlier markers.

This is all happening as the result of some internal incidents.

_Note: This could potentially pave way to making `MonitoredInstrumentation` scarce. @noomorph_

#### TODO

Introduce this for iOS.

---

> _For API changes:_

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/ActivityLaunchHelper.kt
+++ b/detox/android/detox/src/full/java/com/wix/detox/ActivityLaunchHelper.kt
@@ -1,0 +1,76 @@
+package com.wix.detox
+
+import android.app.Instrumentation.ActivityMonitor
+import android.content.Context
+import android.content.Intent
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.rule.ActivityTestRule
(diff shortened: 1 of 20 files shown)


d4vidi:
Successfully ran this on the internal wix-engine project ==> merge

Issue #4207, opened 2023-10-04
Interacting with Redux State in tests


micahdasMA:
### Description

I am using Detox to test a react-native app I have. Part of the app flow involves logging in in order to access certain parts of the app. I would like to test specific aspects of the app past the login screen, but I don't want to have to login in each time that the test runs as that introduces additional complexity and is not actually the traditional user flow (users would typically stay logged in after they initially logged in to the app).

I am trying to find a way to set the redux state before launching the app, or some way to specifically mock the state of the app after opening it.

I looked through the docs on Mocking, but it did not seem like there was anything about directly interacting with redux

I tried launching the app as a new instance, then calling the redux store to set the state that a logged in user would have, and then reloading the app, but this did not result in the user logged in

### Your environment

Detox version: 20.11.0
React Native version: 0.72.1
Node version: 16.2.0
Device model: iPhone 14
OS: iOS
Test-runner (select one): **jest** / other

noomorph:
@micahdasMA, if such an API existed, do you think you would have been able to resolve your problem?

![image](https://github.com/wix/Detox/assets/1962469/f2d8d74a-49d4-4cac-a23b-164114e741ea)

noomorph:
Speaking about officially available solutions,  I'd advise going with static mock files (per the Mocking Guide) where you would register test-only [app protocol handlers](https://reactnavigation.org/docs/deep-linking/), and then you could have used the `device.openURL(...)` API with a URL like `myapp://quickLogin?email=someone@example.com`.

micahdasMA:
@noomorph I think that would do exactly what I want! Thanks for sharing.

In the meantime, I will see what I can do with the app protocol handlers.

noomorph:
Minor correction: device.launchApp({ url: '...' }), not device.openURL() - although the latter might work too as well, but I am not 100% sure.

Issue #4088, opened 2023-05-26
How to simulate change of Date on Simulator/Emulator


abajajct:
### Description

Hello, I am trying to put up a test case which requires me to change the date and then open the app again to test my case. I did some research and see that iOS simulator date depends upon the HOST on which it runs. This makes it impossible to test such scenarios without impacting the machine running the simulator. I wanted to check if there is something I am missing here. Any suggestions will be appreciated.

### Your environment

Detox version: 19.2
React Native version: 0.71
Node version: 16
Device model: iOS
OS: 16.4
Test-runner (select one): jest / other Jest

noomorph:
If this is a visual test, I can suggest: https://wix.github.io/Detox/docs/api/device/#devicesetstatusbar-ios-only

But if you really need to have another time (e.g. you use it in network requests or business logic), well, I can suggest just only mocking: https://wix.github.io/Detox/docs/guide/mocking . 

You need to be creative here: create some TimeService class (or some timeUtils module) and use it instead of just calling Date.now or new Date(). Create an alternative implementation based on a hardcoded value, mock server or command-line argument (e.g. parse it via react-launch-arguments) using e.g. `.e2e.ts` extension. Build your app (or run Metro) with that extension override and that should be it.

I hope my team allows someday to merge `backdoor` feature here: https://github.com/wix/Detox/pull/3245
If it were functioning, you could implement a small handler inside your app and set time right from Detox tests as you go. Unfortunately, you'll have to make a bit longer way (I told you your current options above), but if you think that would be nice to have, please put thumbs up in #3245. I hope one day Amit and Asaf get convinced that the feature is very nice.

d4vidi:
@abajajct have you found the elaborate answer by @noomorph sufficient? Could you nevertheless explain what changing the date is required for? Is it due to the visibility of the time (which is programmable in other ways) or is it related to the app itself / back-end service of some sort?

abajajct:
@d4vidi Hello, I want to test a date based application where from the day you start up, each day unlocks a new part of the application. @noomorph's answer is helpful and I really appreciate that but is there a more straightforward way to acheive this ?

abajajct:
@noomorph The approach you mentioned, would it allow me to change the date in between a running tests. For example: I start with today's date -> Perform fews steps -> Change Date -> Perform steps > Change date again

noomorph:
@abajajct, yes, why not?

asafkorem:
[...]

Issue #3986, opened 2023-03-16
Mocking post/get requests with different responses


ElliotWood13:
### Description

Hello, I can't see any documentation on how to mock post/get requests and pass in different responses. Please can you point me to where I'll find how to do this so I can test different happy/sad paths in my application? Thanks.

### Your environment

Detox version: 
React Native version: 
Node version: 
Device model: 
OS: 
Test-runner (select one): jest / other

asafkorem:
@ElliotWood13 there's no designated Detox API for mocking network requests directly, however - you can mock the module that runs your network requests, see our [mocking guide](https://wix.github.io/Detox/docs/next/guide/mocking) for guidance.

Sid-Turner-Ellis:
@asafkorem would this be static though? sounds like they are asking for dynamic mocks

noomorph:
@asafkorem @d4vidi how hard is it for us to set up a global proxy server forwarding on a device? If we could find a good proxy implementation with some sweet mocking API and combine it with the said `device.enableProxy('http://localhost:3333')` API - that could enable this scenario very easily.

ElliotWood13:
yeah exactly @Sid-Turner-Ellis. Ideally being able to mock like in Jest with a resolved/rejected response would be great.

Sid-Turner-Ellis:
My requirements are different to yours (https://github.com/wix/Detox/issues/3996) but it sounds like you just need to set up a mock server that you can post the desired responses to before the request:

1. In your detox test, make a post request to your server describing what its next return value should be
2. Mock the API url with the mock server url with a .mock.js file 
3. The response should be the value mocked

The same logic would follow for checking the requests, you could look into Nock or https://www.mocks-server.org/ if you want something more fleshed out.

I would recommend really thinking whether you should do this though - I'm not saying you shouldn't but whilst asking my question I realised I was trying to use the e2e tests to test things that my unit/integration tests should cover

d4vidi:
> @asafkorem @d4vidi how hard is it for us to set up a global proxy server forwarding on a device? If we could find a good proxy implementation with some sweet mocking API and combine it with the said `device.enableProxy('http://localhost:3333')` API - that could enable this scenario very easily.

That's definitely an idea that would deserve research. I estimate we might be able to employ `nock` to work inside the app but we would still likely have to allow for a means to configure or preconfigure it through the test code. Another caveat with nock would be that we would not be mocking pure native calls. Anyways, definitely doable, though probably not entirely trivial.

[...]

PR #4208, opened 2023-10-05
[WiP] feat: add device.backdoor() API

noomorph:
## 🚨  Blockers🚨  (why it is WiP)

The underlying mechanisms seem to have changed, and **this PR does not work already with RN73. 😢** 

## Description

This is a revived pull request by @codebutler (#3245). **All props go to him.**  🏆 

Inspired by [functionality in the Xamarin test framework](https://learn.microsoft.com/en-us/appcenter/test-cloud/frameworks/uitest/features/backdoors), this PR proposes a new API within Detox to allow tests to send messages or commands directly to the app, modifying its state during runtime without UI interactions.

The **"Detox as a platform"** vision discussed in issue #207 underlines the potential benefits and use cases of enabling dynamic behaviours and state manipulations during test executions. The Backdoor API is a step in this direction, providing a mechanism for tests to instruct internal changes within the app from E2E tests dynamically.

### Implementation:

A simple API call within a test sends a specific action to the app.
The message should be an object with `action` property and any other custom data:

```js
await device.backdoor({
  action: '...',
  /* ... optional serializable payload ... */
});
```

On the application side, you have to implement a handler for the backdoor message, e.g.:

```js
import { detoxBackdoor } from 'detox/react-native'; // to be used only from React Native

detoxBackdoor.registerActionHandler('my-testing-action', ({ arg1, arg2 }) => {
  // ...
});

// There can be only one handler per action, so you're advised to remove it when it's no longer needed
detoxBackdoor.clearActionHandler('my-testing-action');

// You can supress errors about overwriting existing handlers
detoxBackdoor.strict = false;

// If you want to have multiple listeners for the same action, you can use `registerActionListener` instead
const listener = ({ arg1, arg2 }) => { /* Note that you can't return a value from a listener */ };
detoxBackdoor.addActionListener('my-testing-action', listener);
// You can remove a listener in a similar way
detoxBackdoor.removeActionListener('my-testing-action', listener);

// You can set a global handler for all actions without a handler and listeners
detoxBackdoor.onUnhandledAction = ({ action, ...params }) => {
  // By default, it throws an error or logs a warning (in non-strict mode)
};
```

### Use Case Example

Using the Backdoor API, we can mock time from a Detox test, assuming the app takes the current time indirectly from a time service which has a [mocked counterpart (see the guide)](https://wix.github.io/Detox/docs/next/guide/mocking), e.g.:

- `src/services/TimeService.ts`
- `src/services/TimeService.e2e.ts`

From a Detox test, this may look like:

```js
await device.backdoor({
  action: "set-mock-time",
  time: 1672531199000,
});
```

Assuming the `TimeService` provides the current time via `now()` method, the mocked `TimeService.e2e.ts` will be  listening to `set-mock-time` actions and return the received time value in its `now()` implementation:

```js
// TimeService.e2e.ts

import { detoxBackdoor } from 'detox/react-native';

export class FakeTimeService {
  #now = Date.now();

  constructor() {
    detoxBackdoor.registerActionListener('set-mock-time', ({ time }) => this.setNow(time));
  }

  // If you have a single instance through the app, you might not need any cleanup.
  dispose() {
    detoxBackdoor.clearActionListener('set-mock-time');
  }

  setNow(time) {
    this.#now = time;
  }

  now() {
    return this.#now;
  }
}
```

This allows any test to dynamically set the time to a desired value, thereby controlling the app behaviour for precise testing scenarios without multiple compiled app variants or complex mock server setups.

If you still would like to have multiple listeners for the same action, you can use the `detoxBackdoor.addActionListener` method instead:

```js
const listener = ({ time }) => console.log(`Received time: ${time}`);
detoxBackdoor.addActionListener('set-mock-time', listener);
detoxBackdoor.removeActionListener('set-mock-time', listener);
```

The weaker side of action listeners is that they will never be able to return a value to the caller by design.

### Further development

While this PR introduces the ability to send instructions from tests to the app, it lacks bi-directional communication, enabling the app to report state or responses back to the executing test. This can be a future development vector for this feature.

Here's a draft version of the future BackdoorEmitter (omitting listeners and non-relevant code):

```js
export class BackdoorEmitter {
// ...

  /** @private */
  _onBackdoorEvent = ({ id, event }) => {
    const handler = this._handlers[event.action] || this.onUnhandledAction;
    const promise = invokeMaybeAsync(handler, event);

    promise.then((result) => {
      this._emitter.emit('detoxBackdoorResult', { id, result });
    }, (error) => {
      this._emitter.emit('detoxBackdoorResult', { id, error: serializeError(error) });
    });
  };
}
```

As you can see, the missing part to finish it on the native side is to tap on the `detoxBackdoorResult` events and wait for them before returning the result to the Detox tester (client).

### Summary

This PR facilitates dynamic mocking and state configuration during Detox test runs, enhancing testing capability and flexibility. Feedback, especially regarding practical use cases and potential improvements, is welcomed.

It can give a simple enough solution for many support requests like in #4088, #4207, #3986 and others.

> _For API changes:_


noomorph:
Okay, so far, I have studied the existing concerns from the team about the previous attempt to introduce Backdoor API.

The following points have been raised:

1. Backdoor API naming is undesirable. (@d4vidi)

On the contrary, this is a very accurate name because it reminds the user about the danger of leaving any detoxBackdoor listeners in the production code. See a screenshot from the updated Mocking guide:

image

2. Backdoor API (at this implementation stage) is not bidirectional – i.e. you can't send the results back from the app (including "I am done" confirmation).

This makes sense to me – such functionality would be valuable. On the other hand, I don't view this as a critical flaw but as a minor shortcoming acceptable for Phase 1. Since any consequent activity like re-rendering, network requests, timers, etc., falls under the general synchronization handling in Detox, I doubt that we'll see any new idling timeouts due to the backdoor mechanism.

markhomoki:
This is a really good idea, and will simplify certain flows. 

We are using a similar flow, but we also expect data from the application and use it throughout the test. Application I'm working on is white label, and tests are running on multiple prod environments that have different config. While some parts can use mock files, I'm wondering if this backdoor action could also help us get data back into Detox.

I see this PR sends data to the RN app, but could it be modified to receive data as well? 
Thanks, and great work. 🚀

noomorph:
Well, if there is a way to subscribe to RN emitter the same way JS does, then certainly that is possible.

After a couple of thoughts, the weakest side of this pull request is that this API is not encapsulated:

```js
import {DeviceEventEmitter, NativeAppEventEmitter, Platform} from 'react-native';

const RNEmitter = Platform.OS === "ios" ? NativeAppEventEmitter : DeviceEventEmitter;

RNEmitter.addListener("detoxBackdoor", ({ action, time }) => {
  // Handle actions as needed
});
```

If it turns out that we can listen to these events on the native side, then it may transform into a pair of  detoxBackdoor/detoxBackdoorResult events with some event ids for tracking, and that's why releasing this feature as-is (the way the original author intended), is not the best idea ever.

Maybe a good compromise for now would be to create a facade for this API that end users can import as a part of Detox, and make sure it can be upgraded to bidirectional flow without breaking compatibility. 🤔

detox/test/src/Screens/ActionsScreen.js
```diff
@@ -1,18 +1,19 @@
...
+  ScrollView,
   StyleSheet,
+  Text,
+  TouchableOpacity,
+  View,
   Slider as LegacySlider,
   SafeAreaView,
   requireNativeComponent,
 } from 'react-native';
+// import { detoxBackdoor } from 'detox/react-native';
```
noomorph, line 16: Unfortunately, here Metro bundler blocks us :(

noomorph:
I'll take it for the next sprint – hopefully I'll be able to set up https://github.com/wix-incubator/wml with our e2e project.

noomorph:
It seems there is a simpler solution nowadays to this:

https://metrobundler.dev/docs/configuration/#unstable_enablesymlinks-experimental

detox/test/src/Screens/ActionsScreen.js
noomorph: No longer :)

renatop7:
Hello @noomorph, this solution is very cool, do you know if this works in RN 0.71?

I'm desperate for something like this, to be able to control my testing without a bunch of flags inside my app.

noomorph:
@renatop7 , just as we are speaking, my colleague is working on this pull request right now. I hope you will see it soon.

P. S. As for RN71, yes, it worked, but I don't recommend to recompile/patch unless you are very proficient in these things.

PR #4212, opened 2023-10-10
internal(jest): use jest-environment-emit

noomorph:
## Description

Internal change to ensure future compatibility with `jest-metadata` and `jest-allure2-reporter`.


d4vidi (approved):
@noomorph I essentially approve, but I would like to understand better why `WithEmitter` is needed at this point. I see you've added more event listeners (de-facto, this PR for the first time integrates `jest-metadata`), but don't yet do something about it. I understand what you wrote - about this being a prep for the future, but what should we expect to come next?

d4vidi:
BTW I've allowed myself to retry one of the failing jobs, which failed on account of agent issues

noomorph:
@d4vidi,

`WithEmitter` is a hollow thing at the moment (zero meaningful event listeners by default), so it is hopefully **safe to deploy** in a public release without breaking anyone.

Its goal is to provide a standard interface for connecting environment listeners in the _entire Jest ecosystem_, and it will replace our `DetoxCircusEnvironment.addListeners` trick in the near future. We stop writing those object-oriented class listeners, and all the new environment hooks will be written as event listeners.

**The first beneficiary is DWP** because we'll connect `jest-metadata`, `jest-allure2-reporter`, and `detox-allure2-adapter` with their listeners to ensure a first-class reporting experience. I'm going to open-source some listeners from DWP to `wix-incubator` as well, so it's a bi-directional plan. As more tutorials appear, the enthusiasts from the community will be able to use the new ecosystem as much as we do.

---

As for Detox OSS itself, I decided not to provide `jest-metadata` out of the box right now, because:

1) Its huge potential is going to be wasted at this simplistic level of integration – this library gives a foundation to reduce our artifacts manager code by 70% – https://github.com/wix/Detox/issues/4233 . I'm looking forward to this task, you can't even imagine how much.
2) I still want to look out for potential issues in the long run – perhaps that's the most honest answer here, so I'd feel safer to make it default only in Detox 21. It has been working just perfectly all the time we use it here on the Buildkite OSS, but I want to multiply it by dozens of mobile developers and hundreds of CI builds daily to make sure it does not pose any problem.

PR #4213, opened 2023-10-10
docs(android-dev-env): document how to list available device specs

monholm:
## Description

Added documentation on how to list all available device specs - for convenience when setting up a new emulator.


noomorph:
Thanks for contribution, @monholm !

PR #4215, opened 2023-10-11
fix(ios): preserve logs after device.resetContentAndSettings()

noomorph:
## Description

Fixes `--record-logs ...` artifact issue after calling `device.resetContentAndSettings()`.

Regression caused by #4173.


noomorph:
✅  **The fix is available in [`detox@20.13.1`](https://github.com/wix/Detox/releases/tag/20.13.1).**

Issue #4127, opened 2023-07-13
Something is odd about `runScriptWithArgs` signature


noomorph:
### What happened?

@elgrogz reports:

> We're trying to use the runScriptWithArgs webview function to access a webview's window object but we're having a bit of difficulty. Ideally we'd just run the script against the entire webview but the runScriptWithArgs seems to be run against a particular element in the webview. Is that correct? Maybe then this isn't a valid use case for the function. If anyone knows how we can get the window object then please let me know 🙂 This is our current attempt:

```
send(params: Record<string, unknown>) {
    const webview = web.element(by.web.id("root"));
    const id = randomUUID();
    const json = JSON.stringify({
      id,
      jsonrpc: "2.0",
      ...params,
    });

    webview.runScriptWithArgs(
      `function(json, element) {
        alert("yo");
        window.ledger.e2e.walletApi.send('${json}');
        alert("yoo");
      }`,
      [json],
    );
  }
```

I also noticed that the expected arguments don't match up with what the docs say: <https://wix.github.io/Detox/docs/api/webviews/#runscriptwithargsscript-args>

It says it expects ...args which implies any amount of args, but it actually expects an array of any[]. When I pass an array I get an Unhandled arg typeArrayList from Java. And if I use the exact function in the docs it says there are too many arguments. Is this a bug, and if not how should I be handling it?

![image](https://github.com/wix/Detox/assets/1962469/7cdbe452-1845-4f8d-af18-dd97c3f7e5af)
![image](https://github.com/wix/Detox/assets/1962469/cf1807a4-258b-4225-b074-2b7f8b5930f0)

### Was it tested on latest Detox?

- [ ] I have tested this issue on the latest Detox release and it still reproduces.

### In what environment did this happen?

Detox version: 
React Native version: 
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 
Test-runner (select one): jest / other

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
paste logs here!
```

</details>

17Amir17:
Seeing this too

pagman77:
Also running into the same issue here.

noomorph:
The fix is on its way. I left you a question for you here regarding the updated signature: https://github.com/wix/Detox/pull/4221

noomorph:
[...]

PR #4221, opened 2023-10-16
fix(android): webElement.runScript(fn, args)

noomorph:
## Description

- This pull request addresses the issue described here: #4127.

In this pull request, I removed `runScriptWithArgs` from Web Elements API since it has never worked.

Instead, I fixed `runScript` and extended its capabilities (in a backward-compatible way) as follows:

### `runScript(script[, args])`

Run the specified script on the element.
The script should be a string that contains a valid JavaScript function.
It will be called with that element as the first argument:

```js
const webElement = web.element(by.web.id('identifier'));
await webElement.runScript('(el) => el.click()');
```

For convenience, you can pass a function instead of a string, but please note that this will not work if the function uses any variables from the outer scope:

The script can accept additional arguments and return a value. Make sure the values are primitive types or serializable objects, as they will be converted to JSON and back:

```js
const text = await webElement.runScript(function get(element, property) {
  return element[property];
}, ['textContent']);
```

> _For API changes:_

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/web/WebElement.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/web/WebElement.java
@@ -62,12 +62,12 @@ public class WebElement {
         return getWebViewInteraction().withElement(get()).perform(DriverAtoms.getText()).get();
     }
 
-    public Evaluation runScript(String script) {
-        return getWebViewInteraction().withElement(get()).perform(new SimpleAtom(script)).get();
+    public Object runScript(String script) {
+        return getWebViewInteraction().withElement(get()).perform(new SimpleAtom(script)).get().getValue();
     }
 
-    public Evaluation runScriptWithArgs(String script, final ArrayList<Object> args) {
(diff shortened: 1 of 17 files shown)


noomorph:
@elGrogz, @17Amir17, @pagman77, @ggilchrist-ledger, since you've expressed interest on this topic, could you please review the updated signature? In addition to:

```js
await webElement.runScript('function (element) { /* ... */ }');
```

You get this also:

```js
const result = await webElement.runScript((element, arg1, argN) => value, [arg1, argN]);
```

I'll summarize once again:

* new convenience: function to string conversion (anonymous, named and arrow functions are all supported, provided they don't use outer scope – the convenience boils down to `fn.toString()`, so no wonders to expect)
* scripts can return synchronous results (provided they are primitive or serializable JSON values)
* discarded weird convention about the element being received as the last argument – it is always the first one now.

noomorph:
✅  **The fix is available in [`detox@20.13.1`](https://github.com/wix/Detox/releases/tag/20.13.1).**

Issue #4223, opened 2023-10-17
Avoid globals in the typescript type definitions


ggilles:
### Describe your idea

Currently, the Detox typescript definitions start with something like this:

```typescript
declare global {
    const detox: Detox.DetoxExportWrapper;
    const device: Detox.DetoxExportWrapper['device'];
    const element: Detox.DetoxExportWrapper['element'];
    const waitFor: Detox.DetoxExportWrapper['waitFor'];
    // ...
```

This makes those declarations available in the global namespace.

However, it still required to:

```typescript
import { by, element, expect, waitFor, device } from 'detox';
```

Because just declaring those functions as global doesn't make them available at runtime in the context of typescript compilation.

For example: 
```typescript
describe('Example', () => {
    it('runtime fail !', async () => {
        await device.launchApp();
    });
});
```
will fail at runtime like this:
```
  Example
    ✕ runtime fail ! (7929 ms)

  ● Example › runtime fail !

    ReferenceError: device is not defined

      1 | describe('Example', () => {
      2 |     it('runtime fail !', async () => {
    > 3 |         await device.launchApp();
        |         ^
      4 |     });
      5 | });
      6 |
```

The reason it fails is because we are missing the proper `import { device } from 'detox';` of course.

The problem here is that the **presence of the global declaration of `device`** prevents proper imports resolution from happening correctly: VS code will not suggest any missing import, and `tsc` will not complain about it either. This is an annoying usability issue.

noomorph:
By the way, it looks like you have some local configuration issue there. For the most part the typings should work, the only known issue is `expect`. Do you have skipLibCheck: true? Try to run your TypeScript with --listFiles when without explicit detox import, and see if it picks our index.d.ts at all. It should, actually.

ggilles:
@noomorph Well, we actually tried doing the setup several times 🤔 

Again, the typing **does work**, but the **automatic import** does not.

For example, picking https://github.com/ixartz/React-Native-Boilerplate, here is a video of what happens : in the video I am first demonstrating what should happen when import resolution works fine by editing the `Welcome.test.tsx` file, and then I edit a detox e2e test to show the problem: in this case, the `element` import is not properly added back.

https://github.com/wix/Detox/assets/1478233/297625f3-0748-4a70-943b-ff531bca064a

I had a look and tried several repositories, including https://github.com/software-mansion/react-native-screens/tree/main/Example which is fairly recent and they all show the same problematic behavior.

[...]

PR #4224, opened 2023-10-17
Remove globals from typescript interface definition

ggilles:
## Description

This pull request addresses the issue described here: #4223

This is as straightforward as can be: the globals definitions prevent proper automatic imports by IDE to happen correctly, and imports must be explicit anyway to avoid runtime bugs. Removing those definitions fixes the problem.

### DIFF
--- a/detox/index.d.ts
+++ b/detox/index.d.ts
@@ -11,14 +11,6 @@
 import { BunyanDebugStreamOptions } from 'bunyan-debug-stream';
 
 declare global {
-    const detox: Detox.DetoxExportWrapper;
-    const device: Detox.DetoxExportWrapper['device'];
-    const element: Detox.DetoxExportWrapper['element'];
-    const waitFor: Detox.DetoxExportWrapper['waitFor'];
-    const expect: Detox.DetoxExportWrapper['expect'];
-    const by: Detox.DetoxExportWrapper['by'];
-    const web: Detox.DetoxExportWrapper['web'];
-
     namespace NodeJS {
         interface Global {
             detox: Detox.DetoxExportWrapper;


noomorph:
@ggilles, I believe this issue isn't _that_ straightforward to resolve. To start, Detox _does inject_ global variables (albeit by default), and the type definitions primarily mirror this characteristic.

The solution I prefer involves moving the global definitions to the `detox/runners/jest` package and redefining Detox's `expect` interface on top of Jest's Matchers API. This approach would eliminate any ongoing conflicts.

By "redefining Detox's `expect` interface on top of Jest's Matchers API" I mean this:

```ts
import { Matchers } from '@jest/types';

declare global {
  namespace jest {
    interface Matchers<R, T> {
      // Define your custom matcher functions here
      toBeVisible(): Promise<R>;
      toHaveText(text: string | RegExp): Promise<R>;
    }
  }
}
```

whereas the Jest implementation would be like a wrapper over Detox's own `expect`:

```ts
import { expect as detoxExpect } from 'detox';

expect.extend({
  async toBeVisible(received) {
    try {
      await detoxExpect(received).toBeVisible();
      return {
        message: () => `expected ${received} not to be visible`,
        pass: true
      };
    } catch (error) {
      return {
        message: () => `expected ${received} to be visible`,
        pass: false
      };
    }
  },

  async toHaveText(received, argument) {
    try {
      await detoxExpect(received).toHaveText(argument);
      return {
        message: () => `expected ${received} not to have text ${argument}`,
        pass: true
      };
    } catch (error) {
      return {
        message: () => `expected ${received} to have text ${argument}`,
        pass: false
      };
    }
  }
});
```

noomorph (changes requested):
See my comment above.

noomorph:
As I mentioned in your original issue, it more looks like a TypeScript issue in your project.

I tested manually two times this guide: https://wix.github.io/Detox/docs/guide/typescript

I know what I am talking about. Please try to examine your TS config - there must be something else interfering.

Since this global typing **IS** working in many projects, we can't change it without bumping a major version, first and foremost, so this PR the way it is composed now, it cannot be merged, sorry.

PR #4229, opened 2023-10-20
fix(android): add namespace to build.gradle for android gradle plugin 8 compatibility

mikehardy:
## Description

I may be wrong, but I believe this is necessary for android gradle plugin 8 compatibility, which is what react-native 0.73 will ship with. I'm working through compat issues right now as I bump into them, testing release candidates

> fix(android): add namespace to build.gradle for android gradle plugin 8 compatibility

Related: https://github.com/invertase/react-native-firebase/commit/19ae9b8cb66d16fa20c7ff00d6da1a81438c7030

### DIFF
--- a/detox/android/detox/build.gradle
+++ b/detox/android/detox/build.gradle
@@ -21,6 +21,7 @@ def _rnNativeArtifact = rnInfo.isRN71OrHigher
             : 'com.facebook.react:react-native:+'
 
 android {
+    namespace "com.wix.detox"
     compileSdkVersion _compileSdkVersion
     buildToolsVersion _buildToolsVersion
 
--- a/detox/android/detox/build.gradle
+++ b/detox/android/detox/build.gradle
@@ -21,7 +21,10 @@ def _rnNativeArtifact = rnInfo.isRN71OrHigher
             : 'com.facebook.react:react-native:+'
 
 android {
-    namespace "com.wix.detox"
+    def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
+    if (agpVersion >= 7) {
+        namespace "com.wix.detox"
+    }
     compileSdkVersion _compileSdkVersion
     buildToolsVersion _buildToolsVersion
 


detox/android/detox/build.gradle
```diff
@@ -21,6 +21,7 @@ def _rnNativeArtifact = rnInfo.isRN71OrHigher
             : 'com.facebook.react:react-native:+'
 
 android {
+    namespace "com.wix.detox"
```
mikehardy, line 24:
```suggestion
    def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
    if (agpVersion >= 7) {
        namespace "com.wix.detox"
    }
```

mikehardy:
squash merge - I didn't do the complete / backwards-compatible fix first, sorry :-). Ignore if you all are already handling this another way, of course

noomorph:
@mikehardy , thanks a lot for the effort! Should we verify this with 0.73.0-rc.2?

mikehardy:
Any app that switches to android gradle plugin V8+ will verify, for example a test app, for example the react-native-firebase e2e app 😂

This is true even while still on older gradle or rn 72, it's the android Gradle plugin version in main app module that drives the change

detox/android/detox/build.gradle
```diff
@@ -21,6 +21,10 @@ def _rnNativeArtifact = rnInfo.isRN71OrHigher
             : 'com.facebook.react:react-native:+'
 
 android {
+    def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
+    if (agpVersion >= 7) {
```
noomorph, line 25: One more question, @mikehardy. You state in the title about Gradle Plugin 8+, but here you check >= 7, could you explain, please?

mikehardy: Sorry I didn't see this until just now - this wasn't just a breaking change, it was a non-backwards-compatible breaking change (that is, it will *only* work on AGP8+), but people may still use it on AGP7! So this was the only way the community came up with for it to work on old and new at the same time. Pain in the butt! 😆

PR #4260, opened 2023-10-30
feat(cli): detached runner

noomorph:
## Description

In this pull request, I have added a new feature, **detached mode** for test runners.

### `testRunner.detached` \[boolean]

Default: `false`.

When true, tells `detox test` to spawn the test runner in a detached mode.

This is useful in CI environments, where you want to intercept SIGINT and SIGTERM signals to gracefully shut down the test runner and the device.

Instead of passing the kill signal to the child process (the test runner), Detox will send an emergency shutdown request to all the workers, and then it will wait for them to finish.

### Example

```diff
     args: {
       $0: 'jest',
       config: 'e2e/jest.config.js',
+      forceExit: process.env.CI ? true : undefined,
       _: ['e2e/'],
     },
+    detached: !!process.env.CI,
     retries: process.env.CI ? 1 : undefined,
```

Detox will exit indirectly through Jest using the same mechanism as in our `--bail` workaround (#4115). If some tests don't clean up well, `--forceExit` will help to shut down in time.

---

> _For API changes:_

### DIFF
--- a/detox/local-cli/test.test.js
+++ b/detox/local-cli/test.test.js
@@ -5,6 +5,7 @@ if (process.platform === 'win32') {
 
 jest.mock('../src/logger/DetoxLogger');
 jest.mock('./utils/jestInternals');
+jest.mock('./utils/interruptListeners');
 
 const cp = require('child_process');
 const cpSpawn = cp.spawn;
@@ -17,6 +18,8 @@ const _ = require('lodash');
 const { buildMockCommand, callCli } = require('../__tests__/helpers');
 
 const { DEVICE_LAUNCH_ARGS_DEPRECATION } = require('./testCommand/warnings');
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
 
 describe('CLI', () => {
   let _env;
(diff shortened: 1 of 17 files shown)


detox/local-cli/test.test.js
```diff
@@ -143,6 +146,38 @@ describe('CLI', () => {
...
+    beforeEach(() => {
+      detoxConfig.testRunner.detached = true;
+    });
+
+    test('should be able to run as you would normally expect', async () => {
+      await run();
+      expect(_.last(cliCall().argv)).toEqual('e2e/config.json');
+    });
+
+    test('should intercept SIGINT and SIGTERM', async () => {
```
asafkorem, line 159: WDYT about having this as 2 tests (one for `SIGINT` and one for `SIGTERM`), and another one for `SIGKILL` (no interception)?

asafkorem: anyway the implementation is very straightforward, so it's not very important

asafkorem (approved):
LGTM 👍🏼

detox/local-cli/test.test.js
noomorph: @asafkorem the tests here are rather for decoration and hitting code coverage :D if we be honest.

noomorph: I tested both manually – that's the best way in such affairs. 😊

asafkorem: LOL, OK :) SGTM

PR #4283, opened 2023-11-17
docs: update Launch Args guide

MattFoley:
## Description
The existing docs describe how to grab the launchArgs natively on Android, but not on iOS, so after doing some research, I wanted to add this here so iOS is also described. 

Very simple change to explain what is meant by "normal means", which required a bit of research on my end. Since this is a very small docs change, I hope you don't mind that I didn't flesh out the entire PR checklist.


noomorph:
Thank you, @MattFoley! 👍

PR #4290, opened 2023-11-23
refactor: dynamic device allocation

noomorph:
## Description

This is an internal pull request to enable driver-less initialization in the primary context unless it is really needed (when a worker allocates a device).

As an implementation choice, I allow the creation of multiple device allocators simultaneously in the primary context in a lazy fashion. To map device allocator implementation to a specific cookie, we host a small map of {cookie.id}-{deviceAllocator instance}. On the cleanup, they are all destroyed as expected, in a resilient `for-of` loop...

In theory, this pull request should allow picking a device type dynamically right from the test (e.g., via test environment pragmas), but this is not a public API anyway, and we don't recommend such adventures unless you have no other choice. Full-fledged multi-device support will come significantly later. Most importantly, this pull request doesn't stand in the way of the future architecture evolution – Detox device allocation server will be agnostic to "current test file" or "current selected configuration", it will just listen to precise device configurations requested over network and will provide devices accordingly.

### DIFF
--- a/detox/src/realms/DetoxPrimaryContext.js
+++ b/detox/src/realms/DetoxPrimaryContext.js
@@ -23,7 +23,9 @@ const _emergencyTeardown = Symbol('emergencyTeardown');
 const _lifecycleLogger = Symbol('lifecycleLogger');
 const _sessionFile = Symbol('sessionFile');
 const _logFinalError = Symbol('logFinalError');
-const _deviceAllocator = Symbol('deviceAllocator');
+const _cookieAllocators = Symbol('cookieAllocators');
+const _deviceAllocators = Symbol('deviceAllocators');
+const _createDeviceAllocator = Symbol('createDeviceAllocator');
 //#endregion
 
 class DetoxPrimaryContext extends DetoxContext {
@@ -32,7 +34,8 @@ class DetoxPrimaryContext extends DetoxContext {
 
     this[_dirty] = false;
     this[_wss] = null;
-    this[_deviceAllocator] = null;
+    this[_cookieAllocators] = {};
+    this[_deviceAllocators] = {};
 
     /** Path to file where the initial session object is serialized */
     this[_sessionFile] = '';
@@ -85,7 +88,6 @@ class DetoxPrimaryContext extends DetoxContext {
     const detoxConfig = await this[symbols.resolveConfig](opts);
 
     const {
-      device: deviceConfig,
       logger: loggerConfig,
       session: sessionConfig
     } = detoxConfig;
@@ -108,16 +110,6 @@ class DetoxPrimaryContext extends DetoxContext {
     });
 
     await this[_ipcServer].init();
-
-    const environmentFactory = require('../environmentFactory');
-
-    const { deviceAllocatorFactory } = environmentFactory.createFactories(deviceConfig);
-    this[_deviceAllocator] = deviceAllocatorFactory.createDeviceAllocator({
-      detoxConfig,
-      detoxSession: this[$sessionState],
-    });
-
-    await this[_deviceAllocator].init();
 
     // TODO: Detox-server creation ought to be delegated to a generator/factory.
     const DetoxServer = require('../server/DetoxServer');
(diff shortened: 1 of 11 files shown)


d4vidi:
@noomorph Could you elaborate on the use case where such a primary-context initialization is needed _right now_? This is actually future proof and so as a bottom line I like this change (seeing that it is not going to cost extra in maintenance). Nevertheless, I'd be happy to better understand the current context.

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -191,11 +194,18 @@ class DetoxPrimaryContext extends DetoxContext {
...
+        const deviceAllocator = this[_deviceAllocators][key];
+        delete this[_deviceAllocators][key];
+        try {
+          await deviceAllocator.cleanup();
+        } catch (err) {
+          this[symbols.logger].error({ cat: 'device', err }, `Failed to cleanup device allocator for ${key}`);
+        }
       }
 
+      this[_cookieAllocators] = {};
```
d4vidi, line 207: This begs for serious test coverage

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -191,11 +194,18 @@ class DetoxPrimaryContext extends DetoxContext {
         await this[symbols.uninstallWorker]();
       }
     } finally {
-      if (this[_deviceAllocator]) {
-        await this[_deviceAllocator].cleanup();
-        this[_deviceAllocator] = null;
+      for (const key of Object.keys(this[_deviceAllocators])) {
+        const deviceAllocator = this[_deviceAllocators][key];
+        delete this[_deviceAllocators][key];
```
d4vidi, line 199: Protect with `!== undefined`?

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -227,11 +237,18 @@ class DetoxPrimaryContext extends DetoxContext {
...
-      this[_deviceAllocator].emergencyCleanup();
-      this[_deviceAllocator] = null;
+    for (const key of Object.keys(this[_deviceAllocators])) {
+      const deviceAllocator = this[_deviceAllocators][key];
+      delete this[_deviceAllocators][key];
+      try {
+        deviceAllocator.emergencyCleanup();
+      } catch (err) {
+        this[symbols.logger].error({ cat: 'device', err }, `Failed to emergency cleanup device allocator for ${key}`);
+      }
```
d4vidi, line 247: Could you try to avoid the (subtle) iteration-code duplication?

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -227,11 +237,18 @@ class DetoxPrimaryContext extends DetoxContext {
...
     }
 
+    this[_cookieAllocators] = {};
+
```
d4vidi, line 251: Test (unit) coverage is a must here.

d4vidi (changes requested):
Please review both my comments here and in the issue discussion 🙏🏻

detox/src/realms/DetoxPrimaryContext.test.js
```diff
@@ -205,6 +205,11 @@ describe('DetoxPrimaryContext', () => {
   });
 
   describe('when initialized', () => {
+    beforeEach(async () => {
+      /** Allocate a device to enable more assertions in the tests */
+      await context[symbols.allocateDevice]({ type: 'some.device' });
```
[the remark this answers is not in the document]
noomorph: Thank you! I've addressed both the comments.

detox/src/realms/DetoxPrimaryContext.js
```diff
@@ -253,6 +270,24 @@ class DetoxPrimaryContext extends DetoxContext {
     }
   };
 
+  [_initDeviceAllocator] = async (deviceConfig) => {
```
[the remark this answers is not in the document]
noomorph: Renamed to _createDeviceAllocator

detox/src/realms/DetoxPrimaryContext.js
noomorph: All the newly added rare `try-catch` branches have been covered.

The `cookieAllocators = {}` cleanup is covered by a new test, `DetoxPrimaryContext when initialized when a device is being allocated and then the context has been cleaned up should not be able to find that cookie anymore`.

detox/src/realms/DetoxPrimaryContext.js
noomorph: I reviewed this a couple of times, but I don't see a good reason for the suggested extra caution.

I made sure that a device allocator (when created) cannot be null or anything falsy; hence, if it exists in the map, it just should be available as expected.

detox/src/realms/DetoxPrimaryContext.js
noomorph: Unfortunately not. At least, without obfuscating the code further. The goal is to make sure that the function has 0% `async` statements inside, and the duplication here is the lesser of two evils.

detox/src/realms/DetoxPrimaryContext.js
noomorph: Both cleanup and emergency cleanup were thoroughly covered. Thanks for bringing attention to that.

noomorph:
> Could you elaborate on the use case where such a primary-context initialization is needed right now? This is actually future proof and so as a bottom line I like this change (seeing that it is not going to cost extra in maintenance). Nevertheless, I'd be happy to better understand the current context.

Yes, sure.

1. The added laziness helps to avoid initialization errors in non-friendly CI environments when we don't even need to actually use this or that allocation driver. You know that I rely on this in my Detox test bundler.

2. The parametrization is not needed right now. However, this should give some freedom for DWP – e.g. if some test suite requires a specific OS version, we'll be able to request an exact device via a custom test environment:

```js
/**
 * Web view tests
 * Due to crashes on iOS 17.x, this test suite should be run on older runtimes
 * @iosVersion 16.4
 */
 
describe('Web view tests', () => {
  // ..
});
```

PR #4301, opened 2023-12-19
fix(iOS): migrate (some) device commands to Simctl.

asafkorem:
Use Simctl instead of ASU (AppleSimUtils) where possible, utilizing Simctl capabilities like setting the Simulator geo-location (which wasn't possible before).
Also, add E2E tests for setting device permissions.


asafkorem:
CI fails due to build issues, Xcode 15 is required. Already asked @igorgn to upgrade OSS agents to latest Xcode.

Issue #3512, opened 2022-07-28
Add start position parameters to `scrollTo(edge)` action API


asafkorem:
### Describe your idea

Add start position parameters (`startPositionX`, `startPositionY`) to [scrollTo(edge)](https://github.com/wix/Detox/blob/master/docs/APIRef.ActionsOnElement.md#scrolltoedge) action API, enabling to start scrolling from a desired position, same as we have in [scroll(offset, direction, startPositionX, startPositionY)
](https://github.com/wix/Detox/blob/master/docs/APIRef.ActionsOnElement.md#scrolloffset-direction-startpositionx-startpositiony) action API.

This feature request came from this GitHub discussion: https://github.com/wix/Detox/discussions/3423#discussion-4103607.

The motivation for this feature, quoting @iMaupin words:
> Basically, I am trying to avoid elements on the screen getting tapped unintentionally when I use the scrollTo('edge') methods. Being able to set the x or y coordinate as you can on scroll() should mitigate this issue.

asafkorem:
Release with Detox v20.15 🚀 
Thanks @andrey-wix for this contribution!

PR #4309, opened 2023-12-22
Add start position as an optional parameter to scrollTo action

andrey-wix:
## Description

- This pull request addresses the issue described here: #3512 

---

 - [X] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files.

> _For API changes:_

### DIFF
--- a/detox/ios/Detox/Invocation/Action.swift
+++ b/detox/ios/Detox/Invocation/Action.swift
@@ -129,6 +129,15 @@ class Action : CustomStringConvertible {
 		}
 	}
 	
+	func startPosition(forIndex index: Int, in params: [Any]?) -> Double {
+		guard params?.count ?? 0 > index,
+			  let param = params?[index] as? Double,
+			  param.isNaN == false else {
+			return Double.nan
+		}
+		return param
+	}
+	
 	var description: String {
 		let paramsDescription: String
 		if let params = params {
@@ -434,19 +443,9 @@ class ScrollToEdgeAction : Action {
 			fatalError("Unknown scroll direction")
 			break;
 		}
-		
-		let startPositionX : Double
-		if params?.count ?? 0 > 1, let param1 = params?[1] as? Double, param1.isNaN == false {
-			startPositionX = param1
-		} else {
-			startPositionX = Double.nan
-		}
-		let startPositionY : Double
-		if params?.count ?? 0 > 2, let param2 = params?[2] as? Double, param2.isNaN == false {
-			startPositionY = param2
-		} else {
-			startPositionY = Double.nan
-		}
+
+		let startPositionX = startPosition(forIndex: 1, in: params)
+		let startPositionY = startPosition(forIndex: 2, in: params)
 		let normalizedStartingPoint = CGPoint(x: startPositionX, y: startPositionY)
 
 		element.scroll(to: targetEdge, normalizedStartingPoint: normalizedStartingPoint)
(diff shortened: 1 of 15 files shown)


asafkorem:
Great work @andrey-wix! 👑 Huge thanks!!
I will start reviewing this today

gosha212 (approved):
LGTM

detox/ios/Detox/Invocation/Action.swift
```diff
@@ -435,8 +435,22 @@ class ScrollToEdgeAction : Action {
...
+			startPositionX = param1
+		} else {
+			startPositionX = Double.nan
+		}
+		let startPositionY : Double
+		if params?.count ?? 0 > 2, let param2 = params?[2] as? Double, param2.isNaN == false {
+			startPositionY = param2
+		} else {
+			startPositionY = Double.nan
+		}
```
asafkorem, line 449: We should extract this repetitive code to a function (outside of this class so it will be reused in `ScrollAction` as well
```
func startPosition(forIndex index: Int, in params: [Any]?) -> Double {
    guard params?.count ?? 0 > index, 
          let param = params?[index] as? Double, 
          param.isNaN == false else {
        return Double.nan
    }
    return param
}
```

Then use it this way:
```swift
let startPositionX = startPosition(forIndex: 1, in: params)
let startPositionY = startPosition(forIndex: 2, in: params)
```

asafkorem (changes requested):
Great work @andrey-wix! 🥇 🙌🏼 
I have some small suggestions before we merge this :shipit:

detox/src/android/actions/native.js
```diff
@@ -81,10 +81,10 @@ class ScrollAmountStopAtEdgeAction extends Action {
 }
 
 class ScrollEdgeAction extends Action {
-  constructor(edge) {
+  constructor(edge, startPositionX = -1, startPositionY = -1) {
```
noomorph, line 84: This is somewhat confusing for me. Documentation says that `NaN` acts like `auto`, but if someone indeed decides to pass that `NaN` explicitly, this implementation will not hold water.

I'm not sure if this should be addressed in this PR, but overall this is a bug.

asafkorem: Confusing indeed, but how is that a bug @noomorph?
If someone will pass NaN, `startPositionX` will be converted to `null` in JSON, and Java will handle it the same as it will handle the default `-1` value passed from here, see: https://github.com/wix/Detox/blob/4332464d6b3e8fdb0188cfe04a35100ede94ac6c/detox/android/detox/src/full/java/com/wix/detox/espresso/DetoxAction.java#L88
please correct me if I'm wrong @d4vidi


I generally agree with @noomorph , it's a bit confusing that we use both `-1` or `NaN` for the same behaviour. However it's a copied code from different methods we have that does exactly the same, we can probably replace the default `-1` with `NaN` and avoid the `< 0` check from the line I mentioned.

noomorph: I was left under the impression that passing NaN directly won't work – otherwise, why would Android implement such a weird check for `< 0`?

In any case, I stated that this issue can be addressed in a separate PR.

asafkorem (approved):
The build now passes after I rebased your branch (seems like it failed because it was outdated with recent changes on master).
Code LGTM 💯  Great work and thank you for this contribution @andrey-wix!!

PR #4315, opened 2023-12-28
fix(ios): increase maxBuffer for applesimutils --list

noomorph:
## Description

Resolves issues on Wix's private CI, where there are too many runtimes installed simultaneously.


detox/src/utils/childProcess/exec.js
```diff
@@ -14,9 +14,10 @@ async function execWithRetriesAndLogs(bin, options = {}) {
     interval = 1000,
     prefix = null,
     args = null,
-    timeout = 0,
+    timeout,
```
noomorph, line 17: Zero is a default value, so there is no need to pass that explicitly.
This helps to make tests a bit cleaner.

PR #4318, opened 2024-01-02
fix(iOS): locate all windows before hitability test.

asafkorem:
The resolved issue mainly affected iOS 17 (regression due to UI changes in iOS).


noomorph:
Did you remove `.skip`?

PR #4335, opened 2024-01-15
fix(ios): streaming parsing for applesimutils --list

noomorph:
## Description

This solves an issue with very dirty iOS environments, where there are just way too many simulators.

As this solution is implemented via a streaming parser, we should no longer have issues with the magic `maxBuffer` number.

### DIFF
--- a/detox/src/devices/common/drivers/ios/tools/AppleSimUtils.js
+++ b/detox/src/devices/common/drivers/ios/tools/AppleSimUtils.js
@@ -2,6 +2,8 @@
 const path = require('path');
 
 const _ = require('lodash');
+const makeParser = require('stream-json');
+const { streamArray } = require('stream-json/streamers/StreamArray');
 
 const DetoxRuntimeError = require('../../../../../errors/DetoxRuntimeError');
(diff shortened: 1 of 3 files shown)


asafkorem (approved):
LGTM 🙂

d4vidi:
@noomorph 

>I need some product manager-like decision on the fix in Detox:

IIUC this is going to be agnostic to there user, because we [await the cp](https://github.com/wix/Detox/pull/4335/files#diff-2379781c8ffbd3dcb1607c2e3083a2f8b4203a0d6a86e51244da2f0c2f662505R145) anyways. So, not sure what to consider here, have I missed something?

In addition, some extra thoughts here: Would it make sense to enable this on the util-level - namely, alongside (or bundled-in) `cp.spawnAndLog()`? I wonder.

In any case, a great improvement.

d4vidi:
@noomorph let's take a brave decision here to make to break.

d4vidi:
Closing in favor of the discussion in https://github.com/wix/AppleSimulatorUtils/issues/122

PR #4365, opened 2024-02-10
fix(ios): track hierarchy artifacts

noomorph:
## Description

This pull request addresses absent UI hierarchy artifacts in our Allure reports.

It should be helpful for anyone who's using: https://github.com/wix-incubator/detox-allure2-adapter

### DIFF
--- a/detox/src/artifacts/uiHierarchy/IosUIHierarchyPlugin.js
+++ b/detox/src/artifacts/uiHierarchy/IosUIHierarchyPlugin.js
@@ -51,6 +51,7 @@ class IosUIHierarchyPlugin extends ArtifactPlugin {
     if (this.enabled) {
       const scope = this.context.testSummary ? 'perTest' : 'perSession';
       setUniqueProperty(this._artifacts[scope], name, artifact);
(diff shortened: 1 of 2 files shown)


d4vidi (approved):
Approving even though this makes me wonder: Can artifact creation be made more seamlessly? Namely, improve the infra such that this would happen in a very fundamental place, where every artifact creation is reported as an event. 

I know this might be a bit naive on my part, but I need to ask this.

noomorph:
This is a mere fix, @d4vidi, not an attempt to reimagine the subsystem. I have a vision regarding the new iteration of artifacts, and it is based on making all artifact creating APIs public, plus a thin integration attaching to test runner lifecycle to call those API.

Anyways, here I just align the implementation with my initial API design in 2019. All artifacts undergo track/untrack, but it seems that I was a bit slack last time (the UI hierarchy is one of the last plugins we added), and simply forgot to add it.

We use tracking and untracking to make sure that artifacts get cleaned up even if the artifact plugin forgets to do that.

Allure integration just hops on trackArtifact to spy on artifact.doSave whenever an artifact is created. I hope that explains.

PR #4376, opened 2024-02-20
fix: communicate crash errors more clearly

noomorph:
## Description

- This pull request addresses recurring internal support tickets where our users seem to use the following pattern:

```js
try {
 await tapOnSomething();
} finally {
 await tapOnSomethingElse();
}
```

If `await tapOnSomething()` makes the application crash, the consequent action, `await tapOnSomethingElse()` executes when no app is running. This flow results into misleading errors like this:

```
Detox can't seem to connect to the test app(s)!

HINT: 

Have you forgotten to call 'device.launchApp()' in the beginning of your test?
Refer to our troubleshooting guide, for full details: https://wix.github.io/Detox/docs/troubleshooting/running-tests#tests-execution-hangs

---
The following package could not be delivered:

{
  type: 'invoke',
  params: [Object],
  messageId: 5
}
```

So, this PR improves crash error reporting – the Detox client memorizes the last crash error and doesn't clear it until it sees that the app is again connected to the server.

This change will ensure rethrowing the same crash error consistently, regardless of how many nested try-catch or try-finally statements exist.

### DIFF
--- a/detox/src/client/Client.test.js
+++ b/detox/src/client/Client.test.js
@@ -676,7 +676,7 @@ describe('Client', () => {
   });
 
   describe('on AppWillTerminateWithError', () => {
-    it('should schedule the app termination in 5 seconds, and reject pending and future requests until the app reconnects', async () => {
+    it('should schedule the app termination in 5 seconds, and reject pending', async () => {
       jest.spyOn(client, 'terminateApp');
 
       await client.connect();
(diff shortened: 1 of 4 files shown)


detox/test/e2e/19.crash-handling.test.js
```diff
@@ -18,14 +18,27 @@ describe('Crash Handling', () => {
   it('Should throw error upon internal app crash', async () => {
     await device.reloadReactNative();
     await expectToThrow(() => element(by.text('Crash')).tap(), 'The app has crashed');
-    await expectToThrow(() => element(by.text('Crash')).tap(), 'Detox can\'t seem to connect to the test app(s)!');
```
d4vidi, line 21: Oh wow. Feels weird now that it was ever tested this way.

detox/src/client/Client.test.js
```diff
@@ -676,7 +676,7 @@ describe('Client', () => {
   });
 
   describe('on AppWillTerminateWithError', () => {
-    it('should schedule the app termination in 5 seconds, and reject pending', async () => {
+    it('should schedule the app termination in 5 seconds, and reject pending and future requests until the app reconnects', async () => {
```
d4vidi, line 679: Doesn't this justify splitting to 2 separate tests, as with the prod-tests equivalent?

d4vidi (approved):
Please review discussions before merging

detox/test/e2e/19.crash-handling.test.js
noomorph: Actually I'm satisfied to see this :) This means that here we had to re-evaluate our product decision based on the real feedback.

PR #4394, opened 2024-03-11
feat(ios): add web-view testing support.

asafkorem:
In this pull request, I've:
- Added support for web-view testing on iOS, achieving feature parity with Android.
- Introduced a new fancy test suite for **full coverage** of Detox web-view APIs.
- Updated documentation to reflect these changes.

TODOS (in this PR):
- [x] Add missing unit tests (JS).
- [x] Ensure Android pass the new test suite.

Follow-ups:
- [ ] Write a post about this new support in our Blog page.


detox/test/e2e/utils/snapshot.js
```diff
@@ -0,0 +1,58 @@
+const fs = require('fs-extra');
+const { ssim } = require('ssim.js');
```
asafkorem, line 2: I've added this util for image comparison

asafkorem:
It will be released today in the upcoming version.

asafkorem:
Follow-up on Android issues: https://github.com/wix/Detox/issues/4398

detox/src/android/core/WebElement.js
```diff
@@ -122,6 +129,11 @@ class WebViewElement {
 
     throw new DetoxRuntimeError(`element() argument is invalid, expected a web matcher, but got ${typeof webMatcher}`);
   }
+
+  atIndex(_index) {
+    // Not implemented yet
+    throw new Error('atIndex() is not supported for Android WebViewElement');
```
noomorph, line 135: Please use `DetoxRuntimeError`.

detox/test/e2e/utils/snapshot.js
noomorph: I would advise eventually to use https://github.com/americanexpress/jest-image-snapshot instead of low-level wrappers.

detox/test/e2e/utils/waitForCondition.js
```diff
@@ -0,0 +1,22 @@
+const jestExpect = require('expect').default;
+
+
+async function waitForCondition (func, condition, timeout = 5000) {
```
noomorph, line 4: This can be rewritten to `Promise.race`.

docs/api/webviews.md
```diff
@@ -1,373 +1,130 @@
...
-Match elements with the specified CSS selector.
-
-```js
-web.element(by.web.cssSelector('#cssSelector'));
-```
+1. The `web()` function and `by.id()` matcher locate the web view by its accessibility identifier.
+2. The `myWebView.element()` method and `by.web.id()` matcher locate an HTML element within the web view.
+3. The expectation to verify the element's text is the same as in the single web view scenario.
 
-### `by.web.name(name)`
```
noomorph, line 79: Why did you so brutally cut down the article? I don't understand.

detox/test/e2e/29.webview.test.js
```diff
@@ -1,194 +1,304 @@
...
-  });
+        await expect(headline).toHaveText('Changed');
+      });
 
-  it('should replace text', async () => {
-    const textInput = await webview_1.element(by.web.id('textInput'));
-    await textInput.scrollToView();
-    await textInput.tap();
-    await textInput.typeText('first text');
+      it('should run script with arguments', async () => {
```
noomorph, line 203: Why did you remove tests from scripts passed as regular functions? This will reduce test coverage.

detox/test/e2e/utils/waitForCondition.js
noomorph: Yeah, `Promise.race` may be not the best way. Maybe something like this, then?

```js
function waitForCondition(func, condition, timeout = 5000) {
  const abortController = new AbortController();
  const timeoutId = setTimeout(() => abortController.abort(), timeout);

  const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms, null, abortController.signal));

  const checkCondition = async () => {
    try {
      while (!abortController.signal.aborted) {
        const result = await func();
        if (condition(result)) {
          clearTimeout(timeoutId);
          return result;
        }
        await sleep(100);
      }
      throw new Error('Timeout reached before condition was met');
    } catch (error) {
      clearTimeout(timeoutId);
      throw error;
    }
  };

  return checkCondition();
}
```

docs/api/webviews.md
asafkorem: Agreed to revert to the old document (with some changes) and create a guide based on the new document (which is more of a guide than an API document)

asafkorem:
Release with v20.19

PR #4411, opened 2024-03-16
fix(iOS, WebViews): support testing inner-frames.

asafkorem:
In this PR, I made some iOS fixes for WebView testing APIs:
- Introduced support for matching and interacting with elements in inner frames on
- We now bypass CORS restrictions on tests, specifically to avoid issues with inner-frames ([origin](https://gist.github.com/zhuowei/0b7074b3803d72609c028ab5723d9c28))
- Done some refactoring and fixed a minor issue to keep the cursor at the end of the text field while typing
- Added tests for the new enhancements / fixes


detox/test/e2e/29.webview.test.js
```diff
@@ -269,14 +287,43 @@ describe('Web View', () => {
...
+    /** @type {Detox.WebViewElement} */
+    let webview;
+    const mockServer = new MockServer();
+
+    beforeAll(async () => {
+      mockServer.init();
+
+      if (device.getPlatform() === 'android') {
+        // Android needs to reverse the port in order to access the mock server
+        await device.reverseTcpPort(mockServer.port);
```
asafkorem, line 300: not necessary at the moment, since this is iOS only suite, but I prefer to leave it here for forward compatibility.

PR #4412, opened 2024-03-19
chore: upgrade @react-native-community/datetimepicker

noomorph:
## Description

This should fix a compilation error in `detox/test` on Xcode 15.3.
I confirmed that it works on Xcode 15.2 locally as well.


asafkorem:
See: https://github.com/react-native-datetimepicker/datetimepicker/pull/868

PR #4422, opened 2024-03-26
fix: revert accidental Node 18.x bump

noomorph:
## Description

image

In this pull request, I have reverted the unintended breaking change.


noomorph:
The build is passing (except for one error related to a weird retry attempt — which needs an investigation in a sprint related to Allure integration).

Issue #4408, opened 2024-03-14
Detox postinstall fails in non-React Native projects


noomorph:
### What happened?

If I try to install `detox` in a fresh or standalone project (without RN dependency), I see this error:

```
npm ERR! code 1
npm ERR! path /Users/yaroslavs/Projects/1/node_modules/detox
npm ERR! command failed
npm ERR! command sh -c node scripts/postinstall.js
npm ERR! node:internal/modules/cjs/loader:1147
npm ERR!   throw err;
npm ERR!   ^
npm ERR!
npm ERR! Error: Cannot find module 'react-native/package.json'
....
```

### What was the expected behaviour?

Detox should be able to install without errors.

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Help us reproduce this issue!

1. Create an empty folder, e.g. `my-project`.
2. Go inside `my-project`.
3. Run: `npm init -y`
4. Run: `npm install detox`

### In what environment did this happen?

Detox version: 20.19.1
React Native version: N/A
Has Fabric (React Native's new rendering system) enabled: N/A
Node version: 20.11.1
Test-runner (select one): N/A

### Detox logs

<details><summary>Detox logs</summary>

```
➜  1 npm install detox
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm ERR! code 1
npm ERR! path /Users/yaroslavs/Projects/1/node_modules/detox
npm ERR! command failed
npm ERR! command sh -c node scripts/postinstall.js
npm ERR! node:internal/modules/cjs/loader:1147
npm ERR!   throw err;
npm ERR!   ^
npm ERR!
npm ERR! Error: Cannot find module 'react-native/package.json'
npm ERR! Require stack:
npm ERR! - /Users/yaroslavs/Projects/1/node_modules/detox/src/utils/rn-consts/rn-consts.js
npm ERR! - /Users/yaroslavs/Projects/1/node_modules/detox/scripts/updateGradle.js
npm ERR! - /Users/yaroslavs/Projects/1/node_modules/detox/scripts/postinstall.js
npm ERR!     at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
npm ERR!     at Module._load (node:internal/modules/cjs/loader:985:27)
[...]

PR #4430, opened 2024-03-31
fixed postintall for non react native projects

gosha212:
## Description

- This pull request addresses the issue described here: #4408<?>

In this pull request, I have fixed the postinstall for non RN projects

---

> _For API changes:_


noomorph (approved):
How much work will it be to handle the version as "maybe null" object?

Issue #4437, opened 2024-04-02
Textfield not working 


alexandravychytill:
### What happened?

With new release (20.19.5) type text for web textfield not working on iOS. It has no error, but there gets no text inserted. 

### What was the expected behaviour?

 await web.element(by.web.id(elementId)).typeText(text) should actually insert the text

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [X] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### In what environment did this happen?

Detox version: 20.19.5
React Native version: 0.72.0
Has Fabric (React Native's new rendering system) enabled: yes
Node version: >=16
Device model: iPhone 15
iOS version: 17.2
macOS version: 14.4
Xcode version: 15.2
Test-runner (select one): jest

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
paste your device.log here!
```

</details>

gosha212:
@asafkorem

asafkorem:
@alexandravychytill 
Please try to upgrade to [20.19.5](https://github.com/wix/Detox/releases/tag/20.19.5) since we made a related fix on this version.

If it still happens after the version upgrade, can you please provide a snippet of the text field that you're trying to tap into so we'll be able to reproduce that?

asafkorem:
Oh, you're the original reporter of this issue https://github.com/wix/Detox/issues/4426. 
We made a test that (allegedly?) reproduces your issue and it passes with the fix from the patch version 🙁
I'll need a code snippet or something I can work with

alexandravychytill:
So I just updated to new version and the logs say "passed " but actually there is no insert of any text. 

Textfield: 
`<input aria-invalid="false" id="P0-0" type="email" class="MuiInputBase-input MuiInput-input css-mnn31">`

asafkorem:
Thanks, I see you're using MUI. What happens if you change the type attribute to `text` instead of `email`? Could you please try that just for the investigation (in case you have access for the page's code)?

alexandravychytill:
Hello
So unfortunately I do not have access to the page code.

asafkorem:
[...]

PR #4444, opened 2024-04-10
fix(iOS): dispatch input event when changing web inner-element's value.

asafkorem:
Hopefully fixes: https://github.com/wix/Detox/issues/4437

TODO:
- [ ] add a test that uses a React web page since our current test suite uses a simple HTML page..


asafkorem:
according to https://github.com/wix/Detox/issues/4437#issuecomment-2051056991, this fix resolves the issue.

Issue #4445, opened 2024-04-10
No static field REACT_BRIDGE_LOADING_START of type Lcom/facebook/react/bridge/ReactMarkerConstants


Rolando-Barbella:
### What happened?

After successfully doing all the setup, I want to run my first, when I do `detox test --configuration android.emu.debug`, I get the following error:
```
 @Thread main(2):
    java.lang.NoSuchFieldError: No static field REACT_BRIDGE_LOADING_START of type Lcom/facebook/react/bridge/ReactMarkerConstants; in class Lcom/facebook/react/bridge/ReactMarkerConstants; or its superclasses (declaration of 'com.facebook.react.bridge.ReactMarkerConstants'
```
I have found little information about this type of error, I have doubled checked the whole setup instruction but still 

My react native vs is a bit  old unfortunately: 0.63.2
I also have: _androidTestImplementation('com.wix:detox-legacy:+')_  on my app/build.gradle

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [ ] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### In what environment did this happen?

Detox version: ^20.19.4
React Native version: : 0.63.2
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 16.14.0
Device model: Pixel_3a_API_31_arm64-v8a
Android version: 
Test-runner (select one): jest 

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
The app has crashed, see the details below:

    @Thread main(2):
    java.lang.NoSuchFieldError: No static field REACT_BRIDGE_LOADING_START of type Lcom/facebook/react/bridge/ReactMarkerConstants; in class Lcom/facebook/react/bridge/ReactMarkerConstants; or its superclasses (declaration of 'com.facebook.react.bridge.ReactMarkerConstants' appears in /data/app/~~7xSG0zETQAWbbbbcV7dYzw==/com.osapiens.operations.cona.tnr-b7e5AMXH7v6iCZzqF53AcQ==/base.apk)
        at com.wix.detox.reactnative.ReactMarkersLogger.logMarker(ReactMarkersLogger.kt:20)
[...]

PR #4447, opened 2024-04-11
Added condition wrapper for only R.N vs supported

Rolando-Barbella:
## Description

Following issue [#4445](https://github.com/wix/Detox/issues/4445) I added a condition that limits react logger for not supported versions as @d4vidi recommend it

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/reactnative/ReactMarkersLogger.kt
+++ b/detox/android/detox/src/full/java/com/wix/detox/reactnative/ReactMarkersLogger.kt
@@ -4,6 +4,7 @@ import android.util.Log
 import com.facebook.react.bridge.ReactMarker
 import com.facebook.react.bridge.ReactMarkerConstants
 import com.facebook.react.bridge.ReactMarkerConstants.*
+import com.facebook.react.modules.core.ReactChoreographer
 
 object ReactMarkersLogger : ReactMarker.MarkerListener {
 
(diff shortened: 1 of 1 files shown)


d4vidi (changes requested):
Hey @Rolando-Barbella,
thanks a lot for this request!
It's the right direction, but I was aiming for an `if` to be implemented under the `attach()` method (i.e. making the logger a no-op). I will definitely accept and merge that. Also, why is the condition `minor ≥ 71000` and not `≥71`?

Rolando-Barbella:
Hi @d4vidi thanks for the feedback,  I just updated the branch with the changes, hopefully it makes more sense now

d4vidi:
@Rolando-Barbella you don't need the isLoggerEnabled toggle. The entire class is a listener which unless registered, it does nothing.

Rolando-Barbella:
@d4vidi  sorry, I have little experience with Kotlin, if the logic inside the attach method is not needed, what do you mean by _an if to be implemented under the attach() method (i.e. making the logger a no-op)_ ?


Do you consider something like this?




------
```
object ReactMarkersLogger : ReactMarker.MarkerListener {

    fun attach() {
        ReactMarker.addListener(this)
    }

    override fun logMarker(marker: ReactMarkerConstants, p1: String?, p2: Int) {
        val isMarkerValid = when (marker) {
            DOWNLOAD_START,
            DOWNLOAD_END,
            BUILD_REACT_INSTANCE_MANAGER_START,
            BUILD_REACT_INSTANCE_MANAGER_END,
            REACT_BRIDGE_LOADING_START,
            REACT_BRIDGE_LOADING_END,
            REACT_BRIDGELESS_LOADING_START,
            REACT_BRIDGELESS_LOADING_END,
            CREATE_MODULE_START,
            CREATE_MODULE_END,
            NATIVE_MODULE_SETUP_START,
            NATIVE_MODULE_SETUP_END,
            PRE_RUN_JS_BUNDLE_START,
            RUN_JS_BUNDLE_START,
            RUN_JS_BUNDLE_END,
            CONTENT_APPEARED,
            CREATE_CATALYST_INSTANCE_START,
            CREATE_CATALYST_INSTANCE_END,
            DESTROY_CATALYST_INSTANCE_START,
            DESTROY_CATALYST_INSTANCE_END,
            CREATE_REACT_CONTEXT_START,
            CREATE_REACT_CONTEXT_END,
            PROCESS_PACKAGES_START,
            PROCESS_PACKAGES_END -> true
            else -> false
        }

        if (isMarkerValid && ReactNativeInfo.rnVersion().minor >= 71) {
            Log.d("Detox.RNMarker", "$marker ($p1)")
        }
    }
}

```

detox/android/detox/src/full/java/com/wix/detox/reactnative/ReactMarkersLogger.kt
```diff
@@ -7,38 +7,44 @@ import com.facebook.react.bridge.ReactMarkerConstants.*
...
 
+    private var isLoggerEnabled = false
+    
     fun attach() {
-        ReactMarker.addListener(this)
+        if (ReactNativeInfo.rnVersion().minor >= 71) {
+            isLoggerEnabled = true
+        } else {
+            ReactMarker.addListener(this)
+        }
```
d4vidi, line 17:
```suggestion
        if (ReactNativeInfo.rnVersion().minor >= 71) {
            ReactMarker.addListener(this)
        }
```

d4vidi (changes requested):
@Rolando-Barbella check out my suggestion. This is everything you need.

detox/android/detox/src/full/java/com/wix/detox/reactnative/ReactMarkersLogger.kt
Rolando-Barbella: @d4vidi much simpler approach! I updated the branch, thank you

d4vidi (approved):

Issue #4434, opened 2024-04-01
[iOS Dialogs] Support matching system dialog buttons.


asafkorem:

asafkorem:
Released with [Detox v20.22](https://github.com/wix/Detox/releases/tag/20.22.0)

Issue #4275, opened 2023-11-08
Detox cannot interact with system dialogs


urbanclap-admin:
### What happened?

We have a google phone number hint dialog that automatically invokes when tapping on the login field, and we cannot dismiss the dialog via detox as it cannot find the view.

This causes the text to fail unless we manually dismiss the dialog.

Is there any workaround for this?

### What was the expected behaviour?

The google phone number dialog should dismiss when tapped on the dismiss button

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [X] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### In what environment did this happen?

Detox version: 20.13.5
React Native version: 0.67
Has Fabric (React Native's new rendering system) enabled: No
Node version: v21
Device model: Any android device
Android version: any android version
Test-runner (select one): jest 

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
paste your device.log here!
```

</details>

urbanclap-admin:
@noomorph any workaround for this?

aaronpliu:
not supported! it's mentioned on v21 RC3.  Would see #3208 Next Phrase

aaronpliu:
you would find simulator I/O - keyboard to chose "Connect Hardware Keyboard" or not. If you want to use soft keyboard or just wanna input by text, you are able to setup I/O on simulator.
For really system dialog interaction, it was not supported in current Detox release version. I just saw it in Detox v21 RC3 (with XCUITest). 
For iOS real device, hopefully Detox can support it a.s.a.p

asafkorem:
Support released with [Detox v20.22](https://github.com/wix/Detox/releases/tag/20.22.0) (System API)

Issue #3227, opened 2022-02-18
How to interact with system alert dialogs


davidjonesdialexa:
### Description

Our app calls a `Cancel/Continue` system alert dialog. The Matcher can't seem to locate the button to tap. Any ideas how to handle an element like this?

![Screen Shot 2022-02-18 at 4 08 59 PM](https://user-images.githubusercontent.com/84142414/154771553-7a86d2b2-7fad-4c64-9e69-b4ea918a9da3.png)
![Screen Shot 2022-02-18 at 4 51 21 PM](https://user-images.githubusercontent.com/84142414/154771699-cfe9ed2d-473c-4f7b-a141-8e4722d6ffe3.png)

### Detox version

19.4.5

### React Native version

0.66.3

### Device model

iphone 11

### Xcode version

13.2.1

### iOS version

15.2

### macOS version

12.2.1

### Test runner

jest-circus

asafkorem:
Hi @davidjonesdialexa, unfortunately, Detox is not supporting interactions with system dialog, we are still working on this as part of [XCUITest](https://github.com/wix/Detox/issues/3208) integration.

In the meanwhile, I can suggest to use a launch argument that will disable the login while testing with Detox.

cortisiko:
@davidjonesdialexa As a work around maybe you can you try something like this:

`element(by.label("Continue")).atIndex(0).tap();`

activebiz:
> @davidjonesdialexa As a work around maybe you can you try something like this:
> 
> `element(by.label("Continue")).atIndex(0).tap();`

Tried this but doesnt work.

nero2009:
Please what is the fix for this?

sebbyrum:
Is this still something that is being developed? Would be very helpful for a lot of projects I'm sure

sjors98:
This works sometimes for me: 
```
await element(by.label('Continue').and(by.type('_UIAlertControllerActionView'))).tap();
```
Where Continue is the text in the button you want to tap :)

mariushorvat:
Is there a fix for this issue? Thanks!

alfonsosn:
Bump.

IvayloMetodiev:
Is there any development or workaround here? Finding elements `by.text` works in some cases but if there is a system alert on the screen nothing helps.

hoangtrunnam:
I have the same problem, I can't fix this, Does anyone have an idea?

KaterinaUK:
[...]

Issue #3017, opened 2021-10-13
Support for XCTest addUIInterruptionMonitor for iOS system alerts and dialogs


antgly:
I'm aware that the XCTest framework on iOS supports intercepting and interacting with system alerts, which is currently not a part of Detox.

It would be great if we could overload `expect` to await on an alert by text, instead of an element, passing in the text of the button to tap and/or the alert text to verify.

An example method call would look like this:

`await expect(alert(by.text('This is an alert')).toBeVisible()`

`await alert(by.button('OK').tap()`

If this has to be done as a setup step, then maybe an API like this could be useful:

```
beforeEach(async () => {
  // existing code
  await interceptAlert('This is an alert', buttons -> {
    buttons['OK'].tap()
  }
})
```

It could reject the promise if no button with label appears or if the alert doesn't appear within `expect` timeframe.

Apple's resource on the matter: https://developer.apple.com/documentation/xctest/xctestcase/handling_ui_interruptions

d4vidi:
Thanks for the suggestion, @antgly! We will keep this in mind in the context of our work on XCTest integ. in #2741.

@asafkorem FYI 😄

antgly:
Still interested in this (as are many following this project) 👍

antgly:
This is a blocker for #2741, so I think this should stay open for contextual reasons.

antgly:
Can we keep this issue open as it’s a blocker for many apps’ system interactions?

asafkorem:
Released with [Detox v20.22](https://github.com/wix/Detox/releases/tag/20.22.0)

PR #4457, opened 2024-04-21
feat(iOS): add system-dialogs interaction support.

asafkorem:
Resolves: #4433, #4434, #4435, https://github.com/wix/Detox/issues/4275, https://github.com/wix/Detox/issues/3227, https://github.com/wix/Detox/issues/3017

This PR add support for system-dialogs interactions (system alerts and permission requests).

There are several changes in this PR:
- **Initial integration with XCUITest**: add new scripts for detox-cli (`xcuitest-cache-build`/`rebuild`/`clean`), and refactored existing scripts.
- Add initial system-matchers: `by.system.label`, `by.system.type`. 
- Add initial system-action (`tap()`).
- Add initial system-expectations (`toExist()`, `atIndex(..)`, `not` modifier).

Check the [`system` test suite for example](https://github.com/wix/Detox/blob/95b489fa0f8fd1e34de9c997e6d7fe4538c62ed4/detox/test/e2e/36.system.test.js).

Check the rationale behind the new API design decision [here](https://github.com/wix/Detox/pull/4457#issuecomment-2085308233).


docs/api/system.md
```diff
@@ -0,0 +1,78 @@
+# System
+
+System APIs allows you to interact with elements in the system-level, such as dialogs, alerts, OS browser, push notifications, etc.
```
asafkorem, line 3: need to fix: only system dialogs ATM, other features in the future

noomorph:
The elephant in the room (i.e. the longest answer to write) is:

This pull request current lacks elaboration on:
1) why `by.system` and `system.element` are the best possible solution in today’s context,
2) what were the other solutions considered,
3) why the latter are less suitable or subpar.

It’s unlikely you’ll be happy to write an essay on this, but this kind of engineering note really helps the colleagues and fellow contributors, so please 🙏

noomorph (changes requested):
Unless there was a really good reason to do so, I find it pointless to inflate the CLI toolkit with typical parameterless commands like:

1. build-framework-cache
2. build-xcuitest-cache
3. rebuild-framework-cache
4. rebuild-xcuitest-cache

I see that it created a cascade of changes in documentation, and quite many copy&pastes between files.

First of all, I don't see significant value in exposing these implementation details to user on top level. Our users may be more than happy in ignorance what exactly the magical `build-framework-cache` command does in terms of troubleshooting.

Secondly, I recognize the importance of isolating the build/cleanup for these different projects, and this is why I'd suggest introducing a couple of useful CLI arguments:

```
detox build-framework-cache --xcuitest # build only XCUITest
detox build-framework-cache --detox # build only Detox.framework
detox build-framework-cache --clean # clean before rebuilding this or that
```

In that case:

* build-xcuitest-cache, clean-xcuitest-cache, rebuild-xcuitest-cache can be omitted
* rebuild-framework-cache can serve as a mere alias to`build-framework-cache` with an extra `--clean` option
* maybe even clean-framework-cache can reuse build-framework-cache with (clean=true, xcuitest=false, detox=false), if this doesn't seem "too clever" to maintain.
* we won't be adding any new CLI commands
* just small additions to the docs will be required

**I will be adding a few more notes but overall I must admit that this a very clean and nice PR.** I decided not to wait with other notes, because this one implies some bit of rework, and the earlier you see it the better for us all it is.

detox/src/utils/environment.js
```diff
@@ -187,7 +187,31 @@ async function getFrameworkPath() {
...
+  if (!_iosXCUITestRunnerPath) {
+    _iosXCUITestRunnerPath = _doGetXCUITestRunnerPath();
+  }
+
+  return _iosXCUITestRunnerPath;
+}
+
+async function _doGetXCUITestRunnerPath() {
+  const detoxVersion = getDetoxVersion();
+  const sha1 = (await exec(`(echo "${detoxVersion}" && xcodebuild -version) | shasum | awk '{print $1}'`)).stdout.trim();
```
noomorph, line 205: 1. I don't like the fact we're calling a pretty complex command twice. This could have been memoized.
2. Obviously, this is something that requires a separate function.
3. It's quite trivial to calculate a sha1 sum in Node.js itself (in case of doubt, ask any AI assistant), so maybe we could get away with just calling `xcodebuild -version` in Shell? This should be perfectly enough, no need to `echo`, `awk` and `shasum` it, right?

detox/src/utils/environment.js
```diff
@@ -187,7 +187,31 @@ async function getFrameworkPath() {
...
+
+  const derivedDataPath = `${DETOX_LIBRARY_ROOT_PATH}/ios/xcuitest-runner/${sha1}`;
+
+  const xctestrunPath = (await exec(`find ${derivedDataPath} -name "*.xctestrun" -print -quit`)).stdout.trim();
```
noomorph, line 209: Is it because the path is dynamic (random, guidish something)?

detox/src/ios/expectTwo.test.js
```diff
@@ -754,6 +757,42 @@ describe('expectTwo', () => {
...
   });
 
+  describe('system', () => {
+    it(`should parse system.element(by.system.label('tapMe')).atIndex(1).tap()`, async () => {
+      const testCall = await e.system.element(e.by.system.label('tapMe')).atIndex(1).tap();
+      const jsonOutput = {
+        invocation: {
+          type: 'systemAction',
+          systemAction: 'tap',
+          systemPredicate: {
```
noomorph, line 767: I wonder what caused you to rename literally every property inside the object?

On the first glance, an object like 

```js
{
  type: 'systemAction',
  action: 'tap',
  predicate: {
   // ...
  },
  atIndex: 1
}
```

...could have perfectly co-exist with other invocations. Wouldn't `type: 'systemAction'` be enough already for distinguishing those?

noomorph (commented):
I think I've finished the review for now. Thanks a lot for keeping it compact. Great job overall. See my comments. 🙏

detox/src/utils/environment.js
asafkorem: 1+2. It's already memoized
https://github.com/wix/Detox/blob/7bfba227375ab4241a66c08c5f6895d45c267eb9/detox/src/utils/environment.js#L193-L201

detox/src/ios/expectTwo.test.js
asafkorem: You make a good point, and I had considered that too. However, I chose to follow the naming "conventions" established in the WebView APIs (`webAction`, `webPredicate`, `webExpectation`..) to maintain a consistent design approach

detox/src/utils/XCUITestRunner.js
```diff
@@ -0,0 +1,51 @@
+const DetoxRuntimeError = require('../errors/DetoxRuntimeError');
+
+const { execWithRetriesAndLogs } = require('./childProcess/exec');
+const environment = require('./environment');
+
+class XCUITestRunner {
```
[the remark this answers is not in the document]
asafkorem: Will do

detox/src/utils/environment.js
asafkorem: 3. you're right.

detox/src/utils/environment.js
asafkorem: yeah, it can be dynamic based on the env.. e.g.
`DetoxXCUITestRunner_DetoxXCUITestRunner_iphonesimulator17.4-arm64-x86_64.xctestrun`

detox/src/utils/environment.js
asafkorem: done

detox/src/utils/XCUITestRunner.js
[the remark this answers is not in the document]
asafkorem: done

detox/src/utils/XCUITestRunner.js
```diff
@@ -0,0 +1,51 @@
...
+
+        const flags = [
+            '-xctestrun', runnerPath,
+            '-sdk', 'iphonesimulator',
+            '-destination', `"platform=iOS Simulator,id=${this.simulatorId}"`,
+            'test-without-building',
+        ];
+
+        try {
+            return await execWithRetriesAndLogs(
```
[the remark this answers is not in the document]
asafkorem: `TEST_RUNNER_PARAMS` isn't recognized when passing from `env` in options, it might be because of a limitation or specific handling of environment variables by `xcodebuild`.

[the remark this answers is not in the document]
asafkorem: btw, they also have specific requirement that all relevant environment variables must start with `TEST_RUNNER_`, and any variables not adhering to this naming convention may not be passed through 🤷🏼‍♂️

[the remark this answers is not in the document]
asafkorem: > If you originally meant to keep some payload details in logs for later troubleshooting, I'd advise just to log your payload separately as a plain text via our logger – this way or that way, reading plain text is nicer than decoding Base64 with eyes in the logs. :)

Done

noomorph:
@asafkorem, I'm particularly concerned about the next things:

1) How well does it translate to Android? Will there be 1:1 `system.element` and `by.system` mapping, and will it be needed at all? For example, the way you're adding the XCUI test driver interface doesn't align with the Android counterpart, UIAutomator:

    https://wix.github.io/Detox/docs/api/device/#devicegetuidevice-android-only

    We should understand how we align `device.getUiDevice()` and `system.element` – either UIAutomator gets a rewrite to `*.system.*`, or XCUITest gets implemented for iOS in `device.getUiDevice()`, or something like that.

2) How sure are we that the XCUI test driver won't be needed elsewhere except for interacting with system dialogs? As we found out recently, cross-iframe WebView interactions can be done only via XCUI. Do you intend to mix `web` and `system` APIs in that case, e.g.:?

```js
await web.element(by.system.type('button')).tap()
```

3) Are blind (x,y) taps and swipes on the device screen also an example of system elements and actions, in your opinion?

asafkorem:
> Unless there was a really good reason to do so, I find it pointless to inflate the CLI toolkit with typical parameterless commands like:
> 
> 1. build-framework-cache
> 2. build-xcuitest-cache
> 3. rebuild-framework-cache
> 4. rebuild-xcuitest-cache
> 
> I see that it created a cascade of changes in documentation, and quite many copy&pastes between files.
> 
> First of all, I don't see significant value in exposing these implementation details to user on top level. Our users may be more than happy to live in blissful ignorance about what exactly the magical `build-framework-cache` command does in terms of troubleshooting.
> 
> Secondly, I recognize the importance of isolating the build/cleanup for these different projects, and this is why I'd suggest introducing a couple of useful CLI arguments:
> 
> ```
> detox build-framework-cache --xcuitest # build only XCUITest
> detox build-framework-cache --detox # build only Detox.framework
> detox build-framework-cache --clean # clean before rebuilding this or that
> ```
> 
> In that case:
> 
> * build-xcuitest-cache, clean-xcuitest-cache, rebuild-xcuitest-cache can be omitted
> * rebuild-framework-cache can serve as a mere alias to`build-framework-cache` with an extra `--clean` option
> * maybe even clean-framework-cache can reuse build-framework-cache with (clean=true, xcuitest=false, detox=false), if this doesn't seem "too clever" to maintain.
> * we won't be adding any new CLI commands
> * just small additions to the docs will be required
> 
> **I will be adding a few more notes but overall I must admit that this a very clean and nice PR.** I decided not to wait with other notes, because this one implies some bit of rework, and the earlier you see it the better for us all it is.

xcuitest-cache commands were removed, and replaced with args for `build-framework-cache` and extended behaviour.

New options (args for detox-build-framework):

- **--build**: Specifies binaries to build (_all / detox / xcuitest / none_). Defaults to _all_.
- **--clean**: Specifies binaries to clean (_all / detox / xcuitest / none_). Defaults to _none_.

`detox clean-framework-cache` and `detox rebuild-framework-cache` are now aliases for `detox build-framework-cache` with specific flags.

asafkorem:
> @asafkorem, I'm particularly concerned about the next things:
> 
> 1. How well does it translate to Android? Will there be 1:1 `system.element` and `by.system` mapping, and will it be needed at all? For example, the way you're adding the XCUI test driver interface doesn't align with the Android counterpart, UIAutomator:
>    https://wix.github.io/Detox/docs/api/device/#devicegetuidevice-android-only
>    We should understand how we align `device.getUiDevice()` and `system.element` – either UIAutomator gets a rewrite to `*.system.*`, or XCUITest gets implemented for iOS in `device.getUiDevice()`, or something like that.

The current `getUIDevice` API is specific to Android and not aligned with our goal for a cohesive API across platforms. It essentially wraps [Android's UIAutomator](https://developer.android.com/reference/androidx/test/uiautomator/UiDevice), which doesn't fit well (at all) with Detox's design. Instead of seeking a direct alignment with this, I plan to treat UIDevice as an underlying detail for the System APIs on Android. Both feature parity and clean API design are goals here.

> 2. How sure are we that the XCUI test driver won't be needed elsewhere except for interacting with system dialogs? As we found out recently, cross-iframe WebView interactions can be done only via XCUI. Do you intend to mix `web` and `system` APIs in that case, e.g.:?
>
> ```js
> await web.element(by.system.type('button')).tap()
> ```

System APIs should definitely provide more than interactions with system dialogs.

My approach is to expand system APIs to include functionalities like push notifications, photo library interactions, system browsers (Safari, Chrome), and interactions with complex WebViews within apps. 
Regarding special webview cases, you raised a nice idea (to mix between system matcher and web-view apis), another solution will be to extend the current (and future) system API abilities to match and interact with elements within the app. IMO that's a discussion for itself. Anyway we will definitely leverage the new system abilities to interact with web-views in special scenarios as we are already familiar with one within the company.

> 3. Are blind (x,y) taps and swipes on the device screen also an example of system elements and actions, in your opinion?

First, we need to evaluate if there are valid use cases that require blind interactions on the device screen.

Although these interactions are performed on system elements, integrating them into the System API could complicate its design. The Device API, already provides a good abstraction for these types of gestures (special kind of system element with unique actions). Therefore, implementing those blind actions within the device object will help to maintain a clear distinction in functionalities although the overlapping aspects of device and system interactions.

detox/local-cli/build-framework-cache.js
```diff
@@ -1,16 +1,95 @@
...
+module.exports.handler = async function buildFrameworkCache(argv) {
+  if (!isMacOS()) {
     detox.log.info(`The command is supported only on macOS, skipping the execution.`);
+    return;
   }
+
+  const frameworkPath = path.join(os.homedir(), '/Library/Detox/ios/framework');
+  const xcuitestPath = path.join(os.homedir(), '/Library/Detox/ios/xcuitest-runner');
+
+  await handleCleaning(
```
noomorph, line 39: Uh-oh... 😢  

Actually, what I meant was:

* to call `handleBuilding` in `build-framework-cache`
* to call `handleCleaning` in `clean-framework-cache`
* to call both `handleCleaning` and `handleBuilding` in `rebuild-framework-cache`

I envisioned that those utility functions will reside in separate files, and be imported by each CLI tool independently, instead of dirty tricks with using `yargs` while you are using `yargs`:

```js
  const argv = hideBin(process.argv);
  yargs(argv)
  .command(require('./build-framework-cache'))
  .parse(['build-framework-cache', '--clean="all"', '--build="all"']);
  ```

In such case, the implementation should have been dead simple:

```
handleCleaning({ xcui: boolean; framework: boolean; }) => Promise<void>
handleBuilding({ xcui: boolean; framework: boolean; }) => Promise<void>
```

Sorry for nitpicking, but wouldn't `detox (re|build|clean)-framework-cache --this --that --those` look and feel more natural and simple?

File: detox/scripts/build_universal_framework_legacy.sh [hunk 8]
```diff
@@ -1,76 +0,0 @@
-#!/bin/bash -e
```
noomorph, line 1: @asafkorem I wonder, won't removing `legacy` scripts break someone?

detox/src/ios/XCUITestRunner.js
```diff
@@ -0,0 +1,49 @@
...
+
+        const flags = [
+            '-xctestrun', runnerPath,
+            '-sdk', 'iphonesimulator',
+            '-destination', `"platform=iOS Simulator,id=${this.simulatorId}"`,
+            'test-without-building',
+        ];
+
+        try {
+            return await exec(`TEST_RUNNER_PARAMS="${base64InvocationParams}" xcodebuild ${flags.join(' ')}`);
```
noomorph, line 33: Can we pass `TEST_RUNNER_PARAMS` to `{ env }` parameter to exec?

detox/src/ios/XCUITestRunner.js
```diff
@@ -0,0 +1,49 @@
...
+const environment = require('../utils/environment');
+const log = require('../utils/logger').child({ cat: 'xcuitest-runner' });
+
+class XCUITestRunner {
+    constructor({ simulatorId }) {
+        this.simulatorId = simulatorId;
+    }
+
+    async execute(invocationParams) {
+        log.trace({ event: 'XCUITEST_RUNNER' }, `invocation params: ${JSON.stringify(invocationParams)}`);
```
noomorph, line 13: Could you please try using:

```
log.trace({ event: 'XCUITEST_RUNNER' }, 'invocation params: %j', invocationParams}`);
```

detox/src/ios/system.js
```diff
@@ -0,0 +1,125 @@
...
+      type: 'systemExpectation',
+      systemPredicate: this.element.matcher.predicate,
+      ...(this.element.index !== undefined && { systemAtIndex: this.element.index }),
+      ...(this.modifiers.length !== 0 && { systemModifiers: this.modifiers }),
+      systemExpectation
+    };
+  }
+
+  expect(expectation, traceDescription) {
+    assert(traceDescription, `must provide trace description for expectation: \n ${JSON.stringify(expectation)}`);
```
noomorph, line 37: I feel uneasy about eager object stringifying each time we try to send something. To my embarrassment, I would even be fine with stupid `if (!traceDescription) { assert(...) }`, just not to spend CPU time for things that won't be needed in 99.99% cases.

I see two good options:
1) check if stack traces are usable even without this stringification – this really can be so
2) if not, use homebrewed assertion utils from `./src/utils` (if needed, add some)

Sorry for nitpicking. 🙏

detox/src/ios/system.js
```diff
@@ -0,0 +1,125 @@
...
+    const traceDescription = systemActionDescription.tap();
+    return this.withAction('tap', traceDescription);
+  }
+
+  withAction(action, traceDescription) {
+    assert(traceDescription, `must provide trace description for action: \n ${JSON.stringify(action)}`);
+
+    const invocation = {
+      type: 'systemAction',
+      systemPredicate: this.matcher.predicate,
```
noomorph, line 68: Not something to fix, but rather out of curiosity? Was there a really good reason to prefix literally every property in the invocation object with `system` (systemThis, systemThat)? Didn't that complicate the implementation in some places?

detox/src/utils/environment.js
```diff
@@ -171,24 +172,36 @@ function throwMissingGmsaasError() {
   throw new DetoxRuntimeError(`Failed to locate Genymotion's gmsaas executable. Please add it to your $PATH variable!\nPATH is currently set to: ${process.env.PATH}`);
 }
 
-function getDetoxVersion() {
+const getDetoxVersion = _.memoize(() => {
```
noomorph, line 175: Rather a nitpicking, but you don't need to use `memoize` here. You'll be absolutely fine with `_.once` since your functions don't require any arguments. It is a more lightweight implementation, and its mental model is a bit easier/lighter.

noomorph (approved):
The biggest concern is that CLI redirection logic between those scripts looks too heavy-weight and clumsy. Reusing and calling helper functions would have been much cleaner and simpler to comprehend. I seriously recommend reconsider that implementation. 🙏

noomorph (changes requested):
A false approval that was. 😅  Sorry. Alas, requesting changes.

detox/src/ios/system.js
noomorph: Okay, sorry, found your answer.

detox/local-cli/build-framework-cache.js
noomorph: I mean that this API doesn't make sense to me:

> --build: Specifies binaries to build (all / detox / xcuitest / none). Defaults to all.
> --clean: Specifies binaries to clean (all / detox / xcuitest / none). Defaults to none.

What sense does it bear to call `detox ... --build xcuitest --clean detox` ?

d4vidi:
Following up on the [comment](https://github.com/wix/Detox/pull/4457#issuecomment-2085308233) explaining the reasoning behind the API, I'd like to propose another approach that would offer some alleviation to the "does the API align with Android / future platform?" problem:

While the `element.system` API offers a very powerful mechanism, it does, to some degree, leak the fact that some elements are system-oriented while other are app-based. This is indeed not necessarily coherent across all platforms.

In the price of more boilerplate work, we could go a different way and extend the API with mini-drivers that are tailored to specific usage cases. For example, a permissions dialog can be a Detox built-in mini-driver, that offers actions for approving or disproving the permission. Similarly, we could do this for other system-ish elements, such as system alerts, toasts (android), and even the photo gallery / camera. To some extent, this approach has already been introduced (implicitly) by specifying a dedicated API for interacting with [date-pickers](https://wix.github.io/Detox/docs/api/actions/#setdatepickerdatedatestring-dateformat) (Also true for [Espresso](https://developer.android.com/reference/androidx/test/espresso/contrib/PickerActions), not just Detox).

While slightly cumbersome for us as project maintainers, this approach does force the mechanics of system interaction vs. app interaction onto mere implementation details. It therefore has the potential of being more sustainable for the user, all-the-while more straightforward (arguably).

detox/ios/DetoxXCUITestRunner/DetoxXCUITestRunner/Extensions/XCUIApplication+Extensions.swift
```diff
@@ -0,0 +1,17 @@
...
+import XCTest
+
+extension XCUIApplication {
+  static var springboard: XCUIApplication {
+    return XCUIApplication(bundleIdentifier: "com.apple.springboard")
+  }
+
+  static var safari: XCUIApplication {
+    return XCUIApplication(bundleIdentifier: "com.apple.mobilesafari")
+  }
```
d4vidi, line 16: Dead code?

detox/ios/DetoxXCUITestRunner/DetoxXCUITestRunner/Handlers/PredicateHandler.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  PredicateHandler.swift (DetoxXCUITestRunner)
+//  Created by Asaf Korem (Wix.com) on 2024.
+//
+
+import Foundation
+import XCTest
+
+class PredicateHandler {
+  let springboardApp: XCUIApplication
```
d4vidi, line 10: Is the `springboard` prefix on purpose? Can this be renamed to `app`? It's as though a general class (handler or whatever) is implemented over something very specific, but I don't fully understand what I'm seeing here.

d4vidi (commented):
Some additional questions/suggestions from my end

detox/test/e2e/36.system.test.js
```diff
@@ -0,0 +1,79 @@
...
+    it('should throw on expectation call', async () => {
+      await expectToThrow(async () => {
+        await expect(system.element(by.system.label('Allow'))).toExist();
+      }, 'System interactions are not supported on Android, use UiDevice APIs directly instead');
+    });
+
+    it('should throw on action call', async () => {
+      await expectToThrow(async () => {
+        await system.element(by.system.type('button')).atIndex(0).tap();
+      }, 'System interactions are not supported on Android, use UiDevice APIs directly instead');
```
d4vidi, line 76: Oh nice

detox/test/e2e/36.system.test.js
d4vidi, line 1: Nicely done

detox/ios/DetoxXCUITestRunner/DetoxXCUITestRunner/Extensions/XCUIApplication+Extensions.swift
asafkorem: lol, yeah.. will revert and return this when implementing safari support

detox/local-cli/build-framework-cache.js
asafkorem: Here's a more polite and grammatically corrected version of your message:

For `detox build-framework-cache --clean --xcuitest --detox`, the default setting should enable both `xcuitest` and `detox` enabled. However, if only one of these options is specified, the other should be disabled. This setup may not be very intuitive, but I'm okay with it. I'd prefer not to spend too much time on this internal utility to avoid unnecessary bike-shedding. I'm fine with that, will change

asafkorem: For `detox build-framework-cache --clean --xcuitest --detox`, the default setting should enable both `xcuitest` and `detox` enabled. However, if only one of these options is specified, the other should be disabled. This setup may not be very intuitive, but I'm okay with it. I'd prefer not to spend too much time on discussing this internal utility to avoid bike-shedding. I'm totally fine with that, will change

detox/scripts/build_universal_framework_legacy.sh
asafkorem: > Who's our vulnerable audience?

Allegedly, Xcode < 12 if such thing even exist (current Xcode version is 15), though they can update their xcodebuild (CLI) version and they'll be fine.

Xcode 11 is quite outdated. Xcode 12 was released in 2020, and I have no confidence that Detox will compile on such an old version (it also comes with the respective old simctl and xcodebuild command).
IMO there is no good reason to continue maintaining or supporting this script

detox/src/utils/environment.js
asafkorem: oh, great

detox/test/e2e/36.system.test.js
asafkorem: ❤️

asafkorem:
I'll mark this API as experimental, so it will be subject to changes at least until we collect the initial feedback. Thanks @d4vidi for this suggestion.

detox/local-cli/build-framework-cache.js
asafkorem: Actually, something bothers me. What's the point of adding the `--clean` flag here if we have the `rebuild-...` command? It's like using `--build` from `clean-...`
I'll just add the `--xcuitest` and `--detox` flags for all commands, and the `clean` / `rebuild` command will call the clean method.

detox/src/utils/environment.js
asafkorem: Same :)

detox/src/ios/XCUITestRunner.js
asafkorem: done both ✅

detox/src/ios/system.js
asafkorem: I wonder if some workaround like

```js
const lazyMessage = {
    toString: function() {
        return `... ${JSON.stringify(...)}`;
    }
};
```

will do the trick when passing it to the assert function. however that's a bit dirty.

asafkorem: you're right about 1, the JSON is redundant here.. there's enough information without it

detox/ios/DetoxXCUITestRunner/DetoxXCUITestRunner/Extensions/ElementType+Extensions.swift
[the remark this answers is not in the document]
asafkorem: To be honest, I see no reason to trim the list.. I prefer to align with the Accessibility Inspector's element types

detox/src/ios/system.js
```diff
@@ -0,0 +1,125 @@
...
+
+  toExist() {
+    const traceDescription = expectDescription.toExist();
+    return this.expect('toExist', traceDescription);
+  }
+
+  get not() {
+    this.modifiers.push('not');
+    return this;
+  }
```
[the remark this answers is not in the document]
asafkorem: Extract what? It's a very lean code. 3± lines of code in this file.. wrap an array with another API?
I do agree it's weird that we have this modifiers object just for the `not`. I think we should consider refactoring it with all other APIs, and just use a boolean flag for `shouldNegate`, or `evaluatesTo` (true/false). But that's a different task (small, but I prefer not to include that with this change).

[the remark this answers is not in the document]
asafkorem: Ah, you're referring to the entire JSON generation, not just the `not`. I agree with that. We should implement a shared infrastructure for iOS and Android and there's a lot of code we can save not just between platform but within the infrastructure

detox/ios/DetoxXCUITestRunner/DetoxXCUITestRunner/Handlers/PredicateHandler.swift
asafkorem: Yes, on purpose. Once it will support interactions with more apps, I will add them as params as well (e.g. `safariApp`, `appleLoginApp`)

asafkorem: The name is to clarify that this is the springboard application.

detox/local-cli/build-framework-cache.js
```diff
@@ -1,16 +1,24 @@
...
+    default: false
+  }),
 
-module.exports.handler = async function buildFrameworkCache() {
-  if (os.platform() === 'darwin') {
-    cp.execSync(path.join(__dirname, '../scripts/build_framework.ios.sh'), { stdio: 'inherit' });
-  } else {
-    detox.log.info(`The command is supported only on macOS, skipping the execution.`);
+  handler: async function(argv) {
+    await build(argv.detox, argv.xcuitest);
```
noomorph, line 22: Oh, that's very lean! Thanks!

detox/local-cli/utils/frameworkUtils.js
```diff
@@ -0,0 +1,78 @@
...
+const fs = require('fs-extra');
+const _ = require('lodash');
+
+const detox = require('../../internals');
+
+const frameworkBuildScript = '../../scripts/build_local_framework.ios.sh';
+const xcuitestBuildScript = '../../scripts/build_local_xcuitest.ios.sh';
+
+const getFrameworkPath = _.once(() => path.join(os.homedir(), '/Library/Detox/ios/framework'));
+const getXcuitestPath = _.once(() => path.join(os.homedir(), '/Library/Detox/ios/xcuitest-runner'));
```
noomorph, line 14: @asafkorem  I wonder, can't that be taken from `src/utils/environment` ?

detox/local-cli/utils/frameworkUtils.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+async function execBuildScript(targetPath, scriptPath, descriptor) {
+  detox.log.info(`Building ${descriptor} cache at ${targetPath}..`);
+
+  try {
+    const result = await exec(path.join(__dirname, scriptPath), { capture: ['stdout', 'stderr'] });
+    detox.log.info(result.stdout);
+  } catch (error) {
+    detox.log.error(`Error while building ${descriptor}: ${error.stderr}`);
```
noomorph, line 32: I'm unsure if the error logs will always be in `stderr`. Could you make sure that both stdout/stderr get included?

Also, is there a good reason why you aim to **capture** stdout/stderr? Why not just let it flow through? I mean `stdio: 'inherit`? 

In this mode, detox doesn't create any log files, so we're not losing any precious info, be it good or not.

detox/src/ios/web.js
```diff
@@ -170,7 +169,7 @@ class WebElement {
   }
 
   withAction(action, traceDescription, ...params) {
-    assert(traceDescription, `must provide trace description for action: \n ${JSON.stringify(action)}`);
+    assertDefined(traceDescription);
```
noomorph, line 172: Not sure how it will work if you need to print a specific error about `traceDescription === undefined` – how do you understand that, let's say, a variable1 is `trace description`? 🤔

noomorph (changes requested):
Still needs some ❤️ love and work, but now it looks way better! 🙏

Den1sKa-aNd1:
Good day,

Any ETA on this?

detox/local-cli/utils/frameworkUtils.js
asafkorem: I'll expose something from there.

asafkorem: done

detox/src/ios/web.js
asafkorem: done

detox/local-cli/utils/frameworkUtils.js
asafkorem: migrated to `stdio: 'inherit'`

docs/api/system.md
```diff
@@ -0,0 +1,90 @@
+# System
+
+System APIs allows you to interact with dialogs in the system-level (e.g. permissions, alerts, etc.).
+
+:::caution Experimental
+
+System APIs are currently in an experimental phase.
+This means that the API is not yet stable and may change in the near future.
+**Do not rely on the current API as it may change in future minor releases.**
```
noomorph, line 9: This line `Do not rely on this API` sounds almost like shooting in your own leg. Needs rephrasing. Maybe just mention that "This API may change over minor releases", because "Do not rely" sounds like "Please don't use it :) "

noomorph (approved):
Looks really good except one small thing in the `system.md` article.

zerobertelprivat:
Hello people,

thanks for your amazing work in advance. Unfortunately i dont get this system api work...

Maybe i am stupid, but:

I updated detox to v20.23.0
I am able to import:
import { by, ..., system } from 'detox';

But, if i run a test:

console.log(system); -> delivers undefined
console.log(by.system); -> delivers an object

i tried to fix the issue via several ways:

deleting node_modules, re add them
call "detox clean-framework-cache && detox build-framework-cache" (deleted "Detox" folder in Library before that in
another attempt)
call "detox build"

The test runs, but "system" is still undefined, no matter what i am trying... what did i wrong?

I appreciate your help,
Greets and thanks,
Robert

PR #4475, opened 2024-04-29
feat: add `duration` and `point` (optional) parameters for `longPress()` action.

asafkorem:
This pull request introduces optional parameters for the `longPress()` action **on both iOS and Android**:

- `duration: number`: Specifies the duration of the long press (in ms). This parameter was previously supported on iOS but has now been extended to Android.
- `point: {x: number, y: number}`: Defines the position of the long press within the element's coordinate space.

Documentation has been updated accordingly, and relevant end-to-end tests have been added.


detox/detox.d.ts
```diff
@@ -1287,18 +1287,23 @@ declare global {
...
+             * @param point coordinates in the element's coordinate space. Optional (default is the center of the element).
              * @example await element(by.id('tappable')).tap();
              * @example await element(by.id('tappable')).tap({ x:5, y:10 });
              */
             tap(point?: Point2D): Promise<void>;
 
             /**
              * Simulate long press on an element
-             * @param duration (iOS only) custom press duration time, in milliseconds. Optional (default is 1000ms).
+             * @param duration custom press duration time, in milliseconds. Optional (default is 1000ms).
```
asafkorem, line 1298: This is now different on iOS and Android, according to the platform. TODO: update.

asafkorem, line 1298: This is now different on iOS and Android, according to the platform. TODO: update.

asafkorem, line 1298:
```suggestion
             * @param duration custom press duration time, in milliseconds. Optional (defaults to the standard long-press duration for the platform).
```

docs/api/actions.md
```diff
@@ -51,15 +51,18 @@ Simulates multiple taps on the element at its activation point. All taps are app
...
 ```
 
-### `longPress(duration)`
+### `longPress(duration, point)`
 
-Simulates a long press on the element at its activation point.
+Simulates a long press on the element at its activation point or at the specified point.
 
-`duration` (iOS only) — press during time, in milliseconds. Optional (default is 1000 ms).
+`duration` — press during time, in milliseconds. Optional (default is 1000 ms).
```
asafkorem, line 58:
```suggestion
`duration` — press during time, in milliseconds. Optional (defaults to the standard long-press duration for the platform).
```

detox/android/detox/src/main/java/com/wix/detox/espresso/action/DetoxMultiTap.kt
```diff
@@ -37,8 +37,9 @@ open class DetoxMultiTap
             private val longTapMinTimeMs: Long = getLongTapMinTime(),
             private val tapEvents: TapEvents = TapEvents(),
             private val uiControllerCallSpy: UiControllerSpy = UiControllerSpy.instance,
-            private val log: DetoxLog = DetoxLog.instance)
-    : Tapper {
+            private val log: DetoxLog = DetoxLog.instance,
+            private val duration: Long? = null
```
asafkorem, line 41: I'm thinking to rename `DetoxMultiTap` to `DetoxCustomTapper` / `DetoxTapper` since it will better reflect its role.

detox/detox.d.ts
```diff
@@ -1287,18 +1287,23 @@ declare global {
...
-             * @param duration (iOS only) custom press duration time, in milliseconds. Optional (default is 1000ms).
+             * @param duration custom press duration time, in milliseconds. Optional (defaults to the standard long-press duration for the platform).
+             * @param point coordinates in the element's coordinate space. Optional (default is the center of the element).
              * @example await element(by.id('tappable')).longPress();
+             * @example await element(by.id('tappable')).longPress(2000);
+             * @example await element(by.id('tappable')).longPress(1500, { x:5, y:10 });
+             * @example await element(by.id('tappable')).longPress({ x:5, y:10 });
              */
-            longPress(duration?: number): Promise<void>;
+            longPress(duration?: number, point?: Point2D): Promise<void>;
```
asafkorem, line 1305: reverse params order

noomorph: I'm torn between `point, duration` and by adding `longPress(params)`, where params is `{ x, y, duration }`. 😅

detox/src/android/core/NativeElement.js
```diff
@@ -42,12 +42,31 @@ class NativeElement {
     return await new ActionInteraction(this._invocationManager, this._matcher, action, traceDescription).execute();
   }
 
-  async longPress() {
-    const action = new actions.LongPressAction();
-    const traceDescription = actionDescription.longPress();
+  async longPress(arg1, arg2) {
```
noomorph, line 45: In case we decide to leave things as they are (see my comment about a single complex parameter with named options), I'd suggest naming them as:

```js
async longPress(maybePoint, maybeDuration) {
  const duration = maybeDuration != null ? maybeDuration : maybePoint;
  const point = maybeDuration != null ? maybePoint : null;
  
  // ... and then you validate the types of `point` and `duration`.
```

detox/src/android/core/NativeElement.js
```diff
@@ -42,12 +42,31 @@ class NativeElement {
...
+    let duration = null;
+    let point = null;
+
+    if (arg1 !== undefined && typeof arg1 === 'number' && arg2 === undefined) {
```
noomorph, line 49: There's so much validation logic that I'd suggest extracting it to one more validation function.

detox/src/android/core/NativeElement.js
```diff
@@ -42,12 +42,31 @@ class NativeElement {
...
+      throw new Error('longPress accepts either a duration (number) or a point ({x: number, y: number}) as ' +
+          'its first argument, and optionally a point as its second argument (if the first argument is a duration).');
+    }
+
+    const action = new actions.LongPressAction(duration, point && point.x, point && point.y);
+    const traceDescription = actionDescription.longPress(duration, point);
     return await new ActionInteraction(this._invocationManager, this._matcher, action, traceDescription).execute();
   }
 
+  _checkIfPointIsValid(point) {
```
noomorph, line 66: It doesn't sit well with me that iOS and Android param arg validation logic is written twice.
See `src/utils/assertArgument.js` we use in iOS API. I think you can come up with a common logic for both platforms in this case.

detox/src/android/espressoapi/DetoxAction.js
```diff
@@ -271,6 +271,92 @@ class DetoxAction {
     };
   }
 
+  static longPress() {
```
noomorph, line 274: Oh... wow.. is this generated by `generation` project? If not, it won't work as we overwrite this file automatically on each build.

detox/src/ios/expectTwo.js
```diff
@@ -800,6 +798,37 @@ class IosExpect {
...
+    point = arg1;
+  } else {
+    throw new Error('longPress accepts either a duration (number) or a point ({x: number, y: number}) ' +
+        'as its first argument, and optionally a point as its second argument (if the first argument is a duration).');
+  }
+
+  return { duration, point };
+}
+
+function _assertValidPoint(point) {
```
noomorph, line 821: You definitely should visit `utils/assertArgument.js` and reuse the logic for both platforms.

detox/src/android/actions/native.js
```diff
@@ -25,9 +25,11 @@ class TapAtPointAction extends Action {
 }
 
 class LongPressAction extends Action {
-  constructor() {
+  constructor(duration, x, y) {
```
noomorph, line 28: Wouldn't it be nicer to keep arguments as an object here `{ duration, x, y}`? This is regardless of what is chosen for the public user-facing API – I guess that transfering it to the native side should be much more pleasant, then.

docs/api/actions.md
```diff
@@ -51,15 +51,18 @@ Simulates multiple taps on the element at its activation point. All taps are app
 await element(by.id('tappable')).multiTap(3);
 ```
 
-### `longPress(duration)`
+### `longPress(duration, point)`
```
noomorph, line 54: Don't forget to revisit if/when you change the API. 🙏

generation/core/generator.js
```diff
@@ -107,13 +107,12 @@ module.exports = function getGenerator({
...
     const overloadFunctionExpressions = json.instances.map(({ args }) => {
       const params = [...(json.prefixedArgs || []), ...args];
-        return t.functionDeclaration(
-          t.identifier(sanitizedName + params.length),
-          params.filter(filterBlacklistedArguments).map(({ name }) => t.identifier(name)),
-          createMethodBody(classJson, Object.assign({}, json, { args }))
-        )
-      }
-    );
+      return t.functionDeclaration(
```
noomorph, line 110: While the overload generation is a brilliant mechanism and this is worth its own pull request, can we stop for a a second and discuss if a DTO argument could be a better idea, still?

noomorph (changes requested):
🙏 Looks very promising, but I see a couple of things that should be addressed early. Please see my comments.

detox/src/android/espressoapi/DetoxAction.js
asafkorem: it is

detox/src/android/core/NativeElement.js
asafkorem: Regarding https://github.com/wix/Detox/pull/4475#discussion_r1584330256
`longPress(duration)` was already supported in iOS so we won't break this API, also I prefer to extend it in a flat way as we're doing with the rest of our Action APIs

detox/detox.d.ts
asafkorem: `longPress(duration)` was already supported in iOS so we won't break this API, also I prefer to extend it in a flat way as we're doing with the rest of our Action APIs

detox/android/detox/src/main/java/com/wix/detox/espresso/action/DetoxMultiTap.kt
asafkorem: Done.

detox/detox.d.ts
asafkorem: done - order replaced.

asafkorem: now: `longPress(point, duration)`

detox/src/android/core/NativeElement.js
asafkorem: It doesn't make sense that `maybePoint` can also be `duration`..
I'll call them `optionalPointOrDuration` and `optionalDuration`.

detox/src/android/core/NativeElement.js
asafkorem: yep

detox/src/android/core/NativeElement.js
asafkorem: done

detox/src/ios/expectTwo.js
asafkorem: done

docs/api/actions.md
asafkorem: done

detox/src/android/actions/native.js
asafkorem: (point, duration) fits better here. changed

generation/core/generator.js
asafkorem: I want to align with the current `longPress(duration)` API we have (iOS-only)

asafkorem: we can, possibly, enable both DTO and `duration` only but it feels a bit dirty.
also, it makes sense to align with `tap({x: number, y: number})` for the point param.

detox/detox.d.ts
```diff
@@ -1287,18 +1287,25 @@ declare global {
...
              * @example await element(by.id('tappable')).tap();
              * @example await element(by.id('tappable')).tap({ x:5, y:10 });
              */
             tap(point?: Point2D): Promise<void>;
 
             /**
              * Simulate long press on an element
-             * @param duration (iOS only) custom press duration time, in milliseconds. Optional (default is 1000ms).
+             * @param point coordinates in the element's coordinate space. Optional (default is the center of the element).
+             * @param duration custom press duration time, in milliseconds. Optional (defaults to the standard long-press duration for the platform).
```
d4vidi, line 1299: Please elaborate on the fact that the usage of `duration` should be restricted to cases that are off the standard, namely, when a duration is explicitly specified over the clickability properties of the view (e.g. as with RN gesture handler)

detox/android/detox/src/main/java/com/wix/detox/espresso/action/common/TapEvents.kt
```diff
@@ -19,12 +19,20 @@ private const val EVENTS_TIME_GAP_MS = 30
...
+        coordinates: FloatArray,
+        precision: FloatArray,
+        downTimestamp: Long?,
+        duration: Long?
+    ): List<MotionEvent> {
         val (x, y) = coordinates
         val downEvent = motionEvents.obtainDownEvent(x, y, precision, downTimestamp)
-        val upEvent = motionEvents.obtainUpEvent(downEvent, downEvent.eventTime + EVENTS_TIME_GAP_MS, x, y)
+
+        val upEventDuration = (duration ?: EVENTS_TIME_GAP_MS).toLong()
```
d4vidi, line 33: `EVENTS_TIME_GAP_MS = 30L`

detox/android/detox/src/main/java/com/wix/detox/espresso/action/DetoxCustomTapper.kt
```diff
@@ -104,8 +106,8 @@ open class DetoxMultiTap
 
     private fun verifyTapEventTimes(upEvent: CallInfo, downEvent: CallInfo) {
         val delta: Long = (upEvent - downEvent)!!
-        if (delta >= longTapMinTimeMs) {
+        if (delta >= longTapMinTimeMs && duration == null) {
```
d4vidi, line 109: Maybe split into 2 classes, one dedicated to long press related things (w/ or w/o duration).

d4vidi (commented):
commenting my 2 cent notes

detox/detox.d.ts
asafkorem: ✅

detox/android/detox/src/main/java/com/wix/detox/espresso/action/common/TapEvents.kt
asafkorem: done

PR #4498, opened 2024-05-25
docs(troubleshooting): suggest double-checking NetworkSecurityConfig if app won't load

ace-n:
## Description

[This page](https://wix.github.io/Detox/docs/troubleshooting/running-tests#if-you-do-not-see-your-app-running-on-the-device) provides a _very_ helpful list of reasons why a Detox app might not load.

**However, I managed to run into a new one!** 😨

-------------------

Thanks to an errant `git` command, I mistakenly removed [my `network_security_config.xml` file](https://wix.github.io/Detox/docs/introduction/project-setup#43-enabling-unencrypted-traffic-for-detox) and the reference to it in `AndroidManifest.xml`. In the process, the app I was testing stopped loading on my physical Android device.

> I confirmed this was the root cause by:
> 1) trying lots of other things that didn't work 🙃
> 2) restoring my Network Security Config
> 3) confirming that my app launched as expected 🎉 

I thought I'd add this scenario to the list mentioned above, lest somebody else fall down the same rabbit hole. 🙂


asafkorem (approved):
Hey @ace-n, thanks. LGTM.

PR #4505, opened 2024-06-10
feat(iOS): support interactions with secured web-view elements.

asafkorem:


detox/detox.d.ts
```diff
@@ -1112,6 +1112,31 @@ declare global {
...
              */
             tag(tagName: string): WebMatcher;
+
+            /**
+             * (iOS Only) Find an element on the DOM tree by its value
+             * @param value
+             * @example
+             * web.element(by.web.value('hello'))
+             */
+            value(value: string): WebMatcher;
```
noomorph, line 1122: Sounds like `<input value="...">` right? Why is it iOS-only, then? 🤔

detox/detox.d.ts
```diff
@@ -1112,6 +1112,31 @@ declare global {
...
+             */
+            label(text: string): MaybeSecuredWebMatcher;
+
+            /**
+             * (iOS Only) Find a secured element on the web-view by its accessibility type.
+             * @note Secured-Web APIs are still in experimental phase and are subject to changes in the near future.
+             * @example
+             * web(by.web.type('textField')).asSecured()
+             */
+            type(type: string): SecuredWebMatcher;
```
noomorph, line 1139: How do you plan to solve issue that iOS and Android types are likely to be very different?

`device.getPlatform() === 'ios' ? ... ` or ... how?

detox/detox.d.ts
```diff
@@ -1608,14 +1672,67 @@ declare global {
...
+             * @example await web.element(by.web.type('textField')).asSecured().clearText();
+             */
+            clearText(): Promise<void>;
+        }
+
         interface IndexableWebElement extends WebElement {
             /**
              * Choose from multiple elements matching the same matcher using index.
-             * @example await web.element(by.web.tag('p')).atIndex(2).tap();
+             * @example await web.element(by.web.tag('p')).asSecured().atIndex(2).tap();
```
noomorph, line 1715: What's the sense to use `asSecured()` for a matcher (`by.web.tag`) which likely won't work for a secured context?

detox/test/src/Screens/WebViewScreen.js
```diff
@@ -119,7 +119,7 @@ const webViewFormWithScrolling = `
         </style>
     </head>
     <body>
-        <h1 id="pageHeadline">First Webview</h1>
+        <h1 id="pageHeadline" aria-label="first-webview">First Webview</h1>
```
noomorph, line 122: So, do we have secured webview in the test here? If not, how do we gain confidence in the API??

detox/src/matchers/factories/index.js
```diff
@@ -15,7 +15,7 @@ class Ios extends MatchersFactory {
   createMatchers({ invocationManager, runtimeDevice, eventEmitter }) {
     const IosExpect = require('../../ios/expectTwo');
     const XCUITestRunner = require('../../ios/XCUITestRunner');
-    const xcuitestRunner = new XCUITestRunner({ simulatorId: runtimeDevice.id });
+    const xcuitestRunner = new XCUITestRunner({ runtimeDevice: runtimeDevice });
```
noomorph, line 18: You can use just `runtimeDevice`, if the name matches

detox/detox.d.ts
```diff
@@ -1112,6 +1112,31 @@ declare global {
              * web.element(by.web.tag('mark'))
              */
             tag(tagName: string): WebMatcher;
+
```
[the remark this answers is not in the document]
asafkorem: Totally a reasonable idea. We already use the same names in most places, and semantically it's already under the web.element context, so it shouldn't add any confusion. Let's follow up on that. We might want to break it in Detox 21 or deprecate it sooner..

detox/detox.d.ts
asafkorem: We should have it for Android as well. As soon as we rewrite the infra to use same JS code builder for Android + iOS, we'll have this API there as well. Right now, we just wrap Espresso on Android, and it doesn't have it

asafkorem: actually, we don't have to wait and we can simply wrap runscript from the javascript side for Android and support it there as well, however there are many other things that can be done like this, and I prefer to keep it and do it properly for all

detox/detox.d.ts
asafkorem: This uses the accessibility types from the iOS system level (see XCUITest, Accessibility Inspector). I believe the best approach will be to understand what heuristics we can use for Android accessibility types to classify different elements and come up with new shared types. At the moment, I prefer to simply expose the iOS types to avoid premature optimization. This is an experimental API, so once we have a good candidate to replace it, we'll do it.

detox/detox.d.ts
asafkorem: no sense, this is an error obviously, I'll replace it with by.type / label. good catch

detox/test/src/Screens/WebViewScreen.js
asafkorem: I have tests written in a different branch (`temp-secured-webview-2`), our test app needs some changes to make them work because currently on iOS, things aren't accessible* with XCUITest. 
*we have that overlay that blocks any accessibility tool. Just check it with the Inspector, and you'll see that you can't tap on elements. Similarly, XCUITest is unable to see those elements. This will require a change to the test app, I already solved this problem before and plan to push it here or in a separated change)
Regardless our tests, I also tested it with the Payments module, which also gives us high confidence (probably the most important)

detox/src/matchers/factories/index.js
asafkorem: sure

detox/test/src/Screens/WebViewScreen.js
asafkorem: added.

PR #4508, opened 2024-06-17
feat(config): multi-app build and start scripts

noomorph:
## Description

In this pull request, I have extended Detox config resolution so that it can take `build` and `start` scripts not only from every app, but it can take configuration-scoped overrides, where a single command does all the building or app server running.

### DIFF
--- a/detox/detox.d.ts
+++ b/detox/detox.d.ts
@@ -33,6 +33,10 @@ declare global {
             logger?: DetoxLoggerConfig;
             session?: DetoxSessionConfig;
             testRunner?: DetoxTestRunnerConfig;
+            /** Build command for the entire configuration, overriding individual app build commands. */
+            build?: string;
+            /** Start command for the entire configuration, overriding individual app start commands. */
+            start?: string;
         };
 
         interface DetoxArtifactsConfig {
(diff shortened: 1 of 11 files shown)


asafkorem:
@noomorph, can't we do that by default?

noomorph:
@asafkorem the example in our E2E is merely a self-test.

The real use case is complex build scripts like the one we use for building multiple Wix apps in parallel, where Gradle cache is simply not applicable.

However, I might think on some other solution... maybe around `resolveConfig`.

detox/src/configuration/composeScriptsConfig.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * @param {{
+ *  appsConfig: Record<string, Detox.DetoxAppConfig>
+ *  localConfig: Detox.DetoxConfiguration;
+ * }} options
+ * @returns {DetoxInternals.RuntimeScriptConfig[]}
+ */
+function composeScriptsConfig(options) {
```
asafkorem, line 8: I think the term "commands" fits here better than "scripts" and is more aligned with other uses of it

asafkorem (approved):
Good stuff. I think it would be better to align with one term, such as "build command," "start command," "commandsConfig," etc.
Anyhow, it's a green from me 🟢

PR #4520, opened 2024-07-03
fix(iOS): add `@retroactive` for protocol conformance (Xcode 16 compliance).

asafkorem:
Our `String` extension declares a conformance to the protocols `LocalizedError` and `Error`. 
This will not behave correctly if Swift introduces this conformance in the future. Xcode 16 started to warn when not using this new syntax.

See original Swift proposal: 
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0364-retroactive-conformance-warning.md


asafkorem:
Released with [20.23.1](https://github.com/wix/Detox/releases/tag/20.23.1)

PR #4523, opened 2024-07-06
fix(iOS): resolve keyboard error for iOS 18 simulators.

asafkorem:
Due to private API changes in UIKit, tested with iOS 18 (beta).


asafkorem:
Released with [20.23.1](https://github.com/wix/Detox/releases/tag/20.23.1)

PR #4534, opened 2024-07-28
feat(android): add `device.tap()` and `device.longPress()`.

gayablau:
## Description

---

> _For API changes:_


asafkorem:
@gayablau please fix the lint errors and unit tests.

detox/src/devices/runtime/drivers/DeviceDriverBase.js
```diff
@@ -51,6 +51,10 @@ class DeviceDriverBase {
     return '';
   }
 
+  async tap(point) {
+    return '';
+  }
+
```
asafkorem, line 57: add a similar method for `longPress(arg1, arg2)` (in `DeviceDriverBase.js`).

detox/test/e2e/06.device.test.js
```diff
@@ -84,4 +84,35 @@ describe('Device', () => {
...
+    await expect(element(by.text('Screen Long Pressed'))).toBeVisible();
+  });
+
+  it(':android: long press on screen by coordinates with duration', async () => {
+    await device.launchApp({ newInstance: true });
+    await element(by.text('Device Tap')).tap();
+    const point = {x: 200, y: 190}
+    await device.longPress(point, 2000);
+    await expect(element(by.text('Screen Long Custom Duration Pressed'))).toBeVisible();
+  });
```
asafkorem, line 117: let's extract these to a new suite (e.g. `06.device-tap.test.js`)

asafkorem: also add e2e tests for long-press with (1) duration only and (2) no params

asafkorem: I would also add more tests:
- `should long press on specified point` (expect for **specific area** to be pressed when passing the `point` param)
- `should not succeed in long pressing with point outside the target area`

Check how we added a similar tests on `03.actions.test.js`.

asafkorem: Same for `device.tap(point)`

asafkorem (commented):
Great work! 👏🏼 🤩 
We also need to [document](https://wix.github.io/Detox/docs/api/device) this (`device.md`).

detox/src/devices/runtime/drivers/android/AndroidDriver.js
```diff
@@ -245,6 +245,20 @@ class AndroidDriver extends DeviceDriverBase {
...
   }
 
+  async tap(point) {
+    let x = point?.x || 100;
+    let y = point?.y || 100;
+    const call = EspressoDetoxApi.tap(x, y);
+    await this.invocationManager.execute(call);
+  }
+
+  async longPress(point, duration) {
```
asafkorem, line 255: please add the relevant unit tests (`androiddriver.test.js`)

asafkorem (changes requested):

asafkorem:
Also let's add assertion when calling this API on iOS (raise unsupported message) - we should have that until adding the support on iOS.

detox/test/e2e/06.device.test.js
gayablau: about the spetific point test these are `tap on button by coordinates ignoring status bar` and `long press on button by coordinates with duration ignoring status bar`. about failing on press outside the are these are `long press on screen with duration without specifying coordinates` and `tap on screen without specifying coordinates`
tell me if It is what you meant

gayablau:
united android and ios branches. fixes of review pushed to https://github.com/wix/Detox/pull/4542

PR #4558, opened 2024-08-18
build: run xcodebuild in isolated env.

asafkorem:
`env -i bash -c <command>` runs the specified command in an isolated environment with no inherited environment variables.

- **`env -i`**: Starts a command with an empty environment, ignoring all inherited environment variables.
- **`bash -c`**: Executes a given command string in a new `bash` shell instance.

I've tested this locally on our xcodebuild with @igorgn 👑


d4vidi (approved):
I approve because this is definitely an improvement.
Nevertheless, I see `xcodebuild` triggered all over the place throughout the numerous projects; I think we oughtta narrow their scope just as much

PR #4559, opened 2024-08-21
test(e2e): refactor skip mechanism

noomorph:
## Description

Internal change to get rid of weird `custom.ifRN...` stuff.

This will help @gosha212  with his snapshot tests as well.


gosha212 (approved):
Amazing

detox/test/e2e/utils/rnSkipper.js
```diff
@@ -0,0 +1,21 @@
+const rn = require('../../../src/utils/rn-consts/rn-consts').rnVersion.minor;
+
+/** @type {import('jest-environment-emit').EnvironmentListenerFn} */
+const listener = ({ testEvents }) => {
+  testEvents
+    .on('start_describe_definition', ({ event: { blockName }, state: { currentDescribeBlock }}) => {
+      const match = blockName.match(/@rn(\d+)/i);
+      if (match && match[1] != rn) {
+        currentDescribeBlock.mode = 'skip';
+      }
```
asafkorem, line 10: So that means we can use multiple `@rn` versions on a test, right?

detox/test/e2e/utils/rnSkipper.js
```diff
@@ -0,0 +1,21 @@
...
+    .on('add_test', ({ event: { testName }, state: { currentDescribeBlock }}) => {
+      const match = testName.match(/@rn(\d+)/i);
+      if (match && match[1] != rn) {
+        const n = currentDescribeBlock.children.length;
+        currentDescribeBlock.children[n - 1].mode = 'skip';
+      }
+    });
+};
+
+module.exports = listener;
```
asafkorem, line 21: how about some more robust syntax:

```suggestion
const { rnVersion } = require('../../../src/utils/rn-consts/rn-consts');

/** @type {import('jest-environment-emit').EnvironmentListenerFn} */
const listener = ({ testEvents }) => {
  const rn = Number(rnVersion.minor);
  if (isNaN(rn)) {
    throw new Error(`Invalid RN version: ${rnVersion.minor}`);
  }

  const patterns = {
    exact: /^(\d+)$/,
    range: /^(\d+)\.\.\.(\d+)$/,
    greaterEqual: /^(\d+)\.\.\.$/, 
    lessEqual: /^\.\.\.(\d+)$/
  };

  const parseVersionPattern = (pattern) => {
    if (patterns.exact.test(pattern)) {
      return rn === parseInt(pattern);
    } else if (patterns.range.test(pattern)) {
      const [, start, end] = pattern.match(patterns.range);
      return rn >= parseInt(start) && rn <= parseInt(end);
    } else if (patterns.greaterEqual.test(pattern)) {
      const [, version] = pattern.match(patterns.greaterEqual);
      return rn >= parseInt(version);
    } else if (patterns.lessEqual.test(pattern)) {
      const [, version] = pattern.match(patterns.lessEqual);
      return rn <= parseInt(version);
    } else {
      throw new Error(`Invalid @rn pattern: ${pattern}`);
    }
  };

  const checkVersion = (str) => {
    const match = str.match(/@rn([0-9.]+(?:\.\.\.[0-9.]*)?)/i);
    if (!match) return true;
    try {
      return parseVersionPattern(match[1]);
    } catch (error) {
      console.error(`RN version check error: ${error.message}`);
      return false;
    }
  };

  const skipIfVersionMismatch = (name, block) => {
    if (!checkVersion(name)) {
      block.mode = 'skip';
    }
  };

  testEvents
    .on('start_describe_definition', ({ event: { blockName }, state: { currentDescribeBlock } }) => {
      skipIfVersionMismatch(blockName, currentDescribeBlock);
    })
    .on('add_test', ({ event: { testName }, state: { currentDescribeBlock } }) => {
      const currentTest = currentDescribeBlock.children[currentDescribeBlock.children.length - 1];
      skipIfVersionMismatch(testName, currentTest);
    });
};

module.exports = listener;
```

asafkorem (approved):
This is definitely the right approach. I think we should make that a Detox feature (and not just internally for our test app)m I'm sure more RN **frameworks/libs** have RN-specific features.
I think we should also check for range, check my suggestion.

detox/test/e2e/utils/rnSkipper.js
noomorph: Actually, no, it is 1 version only at the moment.

detox/test/e2e/utils/rnSkipper.js
noomorph: Looks interesting, but I'd suggest using `semver` and covering this with unit tests, because parsing logic becomes far from trivial.

PR #4562, opened 2024-08-25
build(ios): new XCUITest runner build.

asafkorem:
Implemented a more secure and efficient method for generating temporary build artifacts, ensuring only essential files are copied to the XCUITest build directory. Adhering to the principle of least privilege (PoLP).
- Reduced the XCUITest build artifact size from `73.6 MB` to `6.6 MB`.
- Use `mktemp` and `trap` to remove all unnecessary artifacts even if the script fails.


raldred:
That's a serious reduction! Thanks @asafkorem

PR #4628, opened 2024-11-03
Use alternative selector script

gosha212:
## Description

---

> _For API changes:_

### DIFF
--- a/detox/android/detox/src/full/java/com/wix/detox/espresso/web/DetoxDriverAtoms.java
+++ b/detox/android/detox/src/full/java/com/wix/detox/espresso/web/DetoxDriverAtoms.java
@@ -1,0 +1,625 @@
+package com.wix.detox.espresso.web;
+
+import static androidx.test.espresso.web.model.Atoms.castOrDie;
+import static androidx.test.espresso.web.webdriver.Locator.*;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
+
+import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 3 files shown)


detox/android/detox/src/full/java/com/wix/detox/espresso/web/DetoxDriverAtoms.java
```diff
@@ -0,0 +1,642 @@
...
+     * Returns {@code true} if the desired element is in view after scrolling.
+     */
+    public static Atom<Boolean> webScrollIntoView() {
+        return new WebScrollIntoViewAtom(new WebScrollIntoViewSimpleAtom(), castOrDie(Boolean.class));
+    }
+
+    /**
+     * Finds multiple elements given a locator strategy.
+     */
+    public static Atom<List<ElementReference>> findMultipleElementsDetox(
```
asafkorem, line 151: except for this method I see lots of methods here that we're not using, copied from the original that we're still using. Can we keep only the specific custom method we're using?

asafkorem (changes requested):
@gosha212 is that the solution we're moving forward or we're still testing this approach?
Anyhow if it works on Android, and the same script from iOS, let's plan together how we can do that from the same source (JS level)

detox/android/detox/src/full/java/com/wix/detox/espresso/web/DetoxDriverAtoms.java
gosha212: This is still WIP @asafkorem

gosha212:
This POC is not relevant. Couldn't make it work better that the espresso solution

PR #4638, opened 2024-11-24
Improved android build scripts

gosha212:
## Description

In this pull request, I have improved the android build script in order to be sure it could work in any project structure

---

> _For API changes:_

### DIFF
--- a/detox/android/rninfo.gradle
+++ b/detox/android/rninfo.gradle
@@ -1,11 +1,24 @@
 import groovy.json.JsonSlurper
 
-def getRNVersion = { workingDir ->
-    println("RNInfo: workingDir=$workingDir")
+def getRNPackageJsonDir() {
+    def currentDir = rootDir
+    while (currentDir != null) {
+        def nodeModulesDir = new File(currentDir, "node_modules/react-native")
+        def file = new File(nodeModulesDir, 'package.json')
+        if (file.exists()) {
+            return file.path
+        }
+        currentDir = currentDir.parentFile
+    }
+    throw new GradleException("Unable to find module: $moduleName")
+}
+
+def getRNVersion = { rnPackageJsonPath ->
+    println("RNInfo: package.json=$rnPackageJsonPath")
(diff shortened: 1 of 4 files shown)


examples/demo-react-native/android/settings.gradle
```diff
@@ -3,14 +3,12 @@ rootProject.name = 'DetoxRNExample'
 
 include ':app'
 
-println("RNInfo: rootDir=$rootDir")
```
d4vidi, line 6: on purpose?

detox/android/rninfo.gradle
```diff
@@ -1,11 +1,24 @@
...
-    println("RNInfo: workingDir=$workingDir")
+def getRNPackageJsonDir() {
+    def currentDir = rootDir
+    while (currentDir != null) {
+        def nodeModulesDir = new File(currentDir, "node_modules/react-native")
+        def file = new File(nodeModulesDir, 'package.json')
+        if (file.exists()) {
+            return file.path
+        }
+        currentDir = currentDir.parentFile
```
d4vidi, line 11: Will this continue all the way up to the filesystem's root? I think we should stop at the root project's dir 🤔

gosha212: We don't know who is the root. It could be in any structure of projects in mono repo. It is logn. not so bad

examples/demo-react-native/android/settings.gradle
gosha212: Yes. It is a wrong log. WE don't user a root dir any more

detox/android/rninfo.gradle
d4vidi: Yeah I didn't have performance in mind when I wrote this, rather possible resolutions of wrong node module folders or attempting access to non-priviledged directories (think CI agents).
Anyways, smt to think abt but might be an overkill.

d4vidi (approved):

PR #4659, opened 2024-12-15
chore(deps): upgrade detox-copilot.

LironMShemen:
Due to recent changes in Copilot, the "before all" functionality using Copilot Perform is no longer supported.


detox/test/detox_copilot_cache.json
```diff
@@ -1,915 +1 @@
...
-  "{\"step\":\"Verify that a compact date picker is now display\",\"previous\":[{\"step\":\"Restart the React Native state\",\"code\":\"await device.reloadReactNative();\"},{\"step\":\"Navigate to the DatePicker screen\",\"code\":\"await element(by.text('DatePicker')).tap();\"},{\"step\":\"Verify there is an element with the text \\\"Compact Date Picker\\\"\",\"code\":\"await expect(element(by.text('Compact Date Picker'))).toBeVisible();\"},{\"step\":\"Verify there is an element with today`s date at the buttom of the screen\",\"code\":\"const dateElement = element(by.text('20 Oct 2024'));\\nawait expect(dateElement
```
asafkorem, line 1: we need to update

asafkorem (changes requested):

detox/src/DetoxWorker.js
```diff
@@ -1,6 +1,8 @@
 const CAF = require('caf');
 const _ = require('lodash');
 
+const PromptHandler = require('../test/e2e/utils/PromptHandler');
```
asafkorem, line 4: That's why the test is failing. Why is it here?

detox/src/DetoxWorker.js
```diff
@@ -1,6 +1,8 @@
...
+
```
asafkorem, line 5:
```suggestion
```

detox/src/DetoxWorker.js
```diff
@@ -226,9 +226,18 @@ class DetoxWorker {
     yield this._artifactsManager.onRunDescribeStart(...args);
   };
 
-  onTestStart = function* (_signal, testSummary) {
+  onTestStart = function* (_signal, testSummary){
+    try{
+        this.copilot.init(new PromptHandler());
```
asafkorem, line 231: Shouldn't be here. It should be on the custom describe.

asafkorem: why do we have this duplicated calls?

asafkorem (changes requested):

asafkorem (approved):

PR #4689, opened 2025-01-19
fix(copilot): scroll is not aware of start position param type

yohai109:


detox/src/copilot/detoxCopilotFrameworkDriver.js
```diff
@@ -99,6 +99,7 @@ const detoxCopilotFrameworkDriver = {
             guidelines: [
               'Direction can be "up", "down", "left", or "right".',
               'Use `startPositionX` and `startPositionY` to specify the starting point of the scroll gesture.',
+              '`startPositionX` and `startPositionY` are relative to the element\'s width and height, respectively.',
```
asafkorem, line 102: let's clarify the fact that it uses a `[0,1]` range value (normalized).

asafkorem (changes requested):

asafkorem (approved):

Issue #4678, opened 2025-01-05
When updating from v20.28 to v20.31 android build fails with "This declaration needs opt-in. Its usage must be marked with '@kotlin.ExperimentalStdlibApi"


Overtorment:
### What happened?

Trying to update Detox, and cant compile android build , get error:

```
e: file:///xxxxx/node_modules/detox/android/detox/src/full/java/com/wix/detox/reactnative/idlingresources/ReactNativeIdlingResources.kt:115:29 This declaration needs opt-in. Its usage must be marked with '@kotlin.ExperimentalStdlibApi' or '@OptIn(kotlin.ExperimentalStdlibApi::class)'
```

looks like minor issue, _but_ it is inside detox internals, in node_modules.

tried to bump kotlin version to latest, but that didnt solve it.

managed to solve it by adding compiler args to my build.gradle:

```
diff --git a/android/build.gradle b/android/build.gradle
index 1615d0e3cb..bfe95e8f6c 100644
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -66,6 +66,7 @@ subprojects {
             }
         }
         tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile) {
+            kotlinOptions.freeCompilerArgs += ["-Xopt-in=kotlin.ExperimentalStdlibApi"]
             if (project.plugins.hasPlugin("com.android.application") || project.plugins.hasPlugin("com.android.library")) {
                 kotlinOptions.jvmTarget = android.compileOptions.sourceCompatibility
             } else {
```

but that doesnt feel right

### Was it tested on latest Detox?

- [X] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test throw out a timeout?

- [x] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

### Help us reproduce this issue!

the issue reproduces pretty reliably on our CI:  https://github.com/BlueWallet/BlueWallet/pull/7431

### In what environment did this happen?

Detox version: 20.31
React Native version: 0.75.4
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: na
Device model: na
Android version: na
Test-runner (select one): na

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
[...]

PR #4698, opened 2025-02-01
Fix(android): gradle ExperimentalStdlibApi

Gregoirevda:
fixes issue #4678


mikehardy (approved):
For what it's worth, this looks like the right fix in the right spot and fixed compile for me.
Using this via patch-package
Thanks @Gregoirevda !

gosha212 (changes requested):
A minor fix and we good to go

PR #4699, opened 2025-02-01
Update android compiling dependency

Gregoirevda:


docs/introduction/partials/_debugging-native-code-compliing-android.mdx
```diff
@@ -27,6 +27,9 @@
      }
     ```
 
+    Make sure to remove `androidTestImplementation('com.wix:detox:+')`
```
d4vidi, line 30: @Gregoirevda thanks, but with this approach we ask to both add and remove the Detox native dep. We aim for the docs to be as friendly as possible and so I think we should rephrase and explain that the dep should be _replaced_; Something like:

1. In your app's build script, replace ... <etc>
```diff title="android/app/build.gradle"
     dependencies {
         // ...
    -     androidTestImplementation('com.wix:detox:+')`
    +    androidTestImplementation(project(path: ":detox"))
     }
```

Gregoirevda: @d4vidi Agree, updated

PR #4721, opened 2025-02-26
Alleviate geny-cloud failures due to network hiccups

d4vidi:
## Description

In this pull request, I have introduced a work-around for recurring issues with Genymotion tunnel connections. Empirically, the connection shows hiccups from time to time, which makes ADB disconnect the device altogether. This PR tackles that by setting longer timeouts and more retries, so as to allow for the gmsaas tunnel to reconnect automatically.

---

> _For API changes:_

### DIFF
--- a/detox/package.json
+++ b/detox/package.json
@@ -1,7 +1,7 @@
 {
   "name": "detox",
   "description": "E2E tests and automation for mobile",
-  "version": "20.32.0",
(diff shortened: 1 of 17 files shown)


d4vidi:
@lilinor FYI we're currently analyzing the effect of this kind of a fix

PR #4728, opened 2025-03-06
feat(ios): add booted query support

noomorph:
## Description

This pull request includes changes to add support for the `booted` property in the iOS simulator configuration within the Detox project. This translates to the call of `applesimutils` to find **any booted device** (when specified without any extra query parameters like type, OS, etc.):

```
applesimutils --list --booted
```

> _For API changes:_

### DIFF
--- a/detox/detox.d.ts
+++ b/detox/detox.d.ts
@@ -408,6 +408,7 @@ declare global {
             type: string;
             name: string;
             os: string;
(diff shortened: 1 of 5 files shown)


d4vidi:
@noomorph I'm not convinced this is the right approach, product-wise. Let's schedule a discussion if you think otherwise.

PR #4743, opened 2025-03-26
fix(genycloud): reallocate unresponsive devices

noomorph:
## Description

When Detox is in the middle of something (i.e., it does not need to shutdown), we're adding an additional check — if Genymotion device is answering at all (`echo ok`). And if it is not responsive, we're forcing shutdown so that the next Detox worker, in case it requires a device, will trigger a fresh allocation (of hopefully a less problematic device).


d4vidi (commented):
ok cool

d4vidi:
@lilinor

PR #4758, opened 2025-04-09
WIP] Update react native 0.77.2

mauricedoepke:
While working on my screenshot PR I wanted to also somewhat test the ios build
If found out, that apple just broke react native < 0.77, see:
https://expo.dev/changelog/xcode-16-3-patches

For this reason I tried upgrading the demo/example apps to react native 0.77.2. I did not check everything yet. But so far it seems that most this is mostly working.

I think there might be very few tests broken. Bot not many. Some of your tests seem to fetch data from private endpoints which seems to be working from your ci only. Maybe you can let the tests run on this if you are interested in updating the rn version.


gosha212:
@mauricedoepke Thanks for the collaboration. This change is in our roadmap for sure. Can you elaborate which tests you failed to run locally? It should work on every machine without any issues

mauricedoepke:
@gosha212 

I do not remember all:
but on android it was:
- "should handle tap during storage stress"  this one sounds like it might be flaky and very dependent on enrionment conditions

and on ios:
I think more or less all the "pilot" tests. But if I remember correctly this failed because they tried to fetch stuff frome some wix server which failed for me probably due to authorization.

But I did not observe naything major.

gosha212 (changes requested):
You did amazing job! 
Please add the configurations for RN 76 back in order to insure that it still works. I will remove it after merging in order to save us some build time.

gosha212 (commented):
looks like the build is not working for ios 73

mauricedoepke:
> looks like the build is not working for ios 73

Sorry, totally forgot, that for such a project background compatibility is super important. I totally forgot that the newly added code might break that. I can check this out on the weekend.

.buildkite/pipeline_common.sh
gosha212, line 1: There is a lot of inaccuracy in the naming of the jobs and this file. Please align the filenames this script. Also please add rn 76 configurations in order to validate that everything is still fine

gosha212:
@mauricedoepke I've taken this PR in my hands in order to speed up the process. I will share the result with you.
You were right about the pilot tests. They are using a local server. Thanks for your help

mauricedoepke:
@gosha212 thanks a lot for doing so

PR #4777, opened 2025-06-09
Replace node-module 'tempfile' seamlessly with 'tmp'

### LINKED ISSUE #4418: Replace tempfile with something like tempy
### Describe your idea

Currently, Detox relies on `tempfile`, which relies on `uuid@3.4.0`. On running `npm i` this generates a warning. Checking the `tempfile` repo it hasn't been updated since last year, and the README recommends switching to the `tempy` package.

d4vidi:
## Description

- This pull request fixes #4418

(WIP) I've replaced `tempfile` with an internal util which uses `tmp` internally

---

> _For API changes:_

### DIFF
--- a/detox/src/utils/tempfile.test.js
+++ b/detox/src/utils/tempfile.test.js
@@ -2,51 +2,65 @@ describe('tempfile', () => {
   let tmp;
   let tempfile;
 
-  beforeEach(() => {
-    jest.mock('tmp');
-    tmp = require('tmp');
-    tempfile = require('./tempfile');
+  describe('in full integration with tmp module', () => {
+    beforeEach(() => {
+      jest.restoreAllMocks();
+      tempfile = require('./tempfile');
+    });
+
+    it('should work with the real tmp module', () => {
+      const result = tempfile('.log');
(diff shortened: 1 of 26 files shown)


d4vidi:
@noomorph, I could use a piece of advice about making `tempfile.js` easily accessible for usage outside `detox/src/` (e.g. in `local-cli/`)

detox/src/utils/tempfile.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+  it('should handle empty extension', () => {
+    tempfile('');
+    expectTmpCalled();
+  });
+
+  it('should handle undefined extension', () => {
+    tempfile();
+    expectTmpCalled();
+  });
```
noomorph, line 71: Can we add one real test please? We need to make sure it runs okay on Linux/Windows on Github actions, too.

noomorph (changes requested):
Looks awesome, but I strongly advise to add one real test for `tempfile.js` which will be run in Github Actions environment, and make sure that our Windows coverage is ok.

detox/src/utils/tempfile.test.js
d4vidi: On it

d4vidi:
@noomorph whenever you're ready

noomorph (approved):
Awesome!

PR #4779, opened 2025-06-12
Upgrade to RN 0.78

gosha212:
## Description

Upgraded detox to support rn 78


.buildkite/pipeline_common.sh
```diff
@@ -3,13 +3,15 @@
 echo "steps:"
 
 cat .buildkite/jobs/pipeline.android_demo_app_rn_77.yml
```
d4vidi: since there's not much difference between 77 and 78 in terms of packaging and shipping, I think we can drop 77 for the _demo_ project (android & iOS)

.buildkite/pipeline_common.sh
```diff
@@ -3,13 +3,15 @@
...
+cat .buildkite/jobs/pipeline.android_demo_app_rn_78.yml
 cat .buildkite/jobs/pipeline.android_rn_73.yml
-cat .buildkite/jobs/pipeline.android_rn_76.yml
 cat .buildkite/jobs/pipeline.android_rn_77.yml
 cat .buildkite/jobs/pipeline.android_rn_77_old_arch.yml
```
d4vidi: maybe switch to 78_old_arch (and then later - 79_old_arch, 80, ...)? We decided not to have 77 with old-arch internally, anyways

scripts/change_react_native_version.js
```diff
@@ -18,7 +18,7 @@ async function run() {
     packageJson.devDependencies['react-native'] = reactNativeVersion;
   } else {
     const data = await fetch(`https://registry.npmjs.org/react-native/${reactNativeVersion}`);
-    const reactVersion = data.peerDependencies.react;
```
d4vidi, line 21: please let's not start managing these versions with carets. We revisit these every once in a while manually because of the RN upgrades, anyways, so we should be good IMO

d4vidi (commented):
Thanks @gosha212!
- Please see notes / req for changes
- Please add a proper PR description

.buildkite/pipeline_common.sh
d4vidi: @gosha212 still a to-do

d4vidi (approved):

Issue #4788, opened 2025-07-01
Semantic matching basics


d4vidi:
### Describe your idea

This is an issue meant to specifically hold the idea mentioned here: https://github.com/wix/Detox/issues/4145#issuecomment-1689920814

We want to allow for a more intuitive matching API that would allow for a middle-ground solution between the "traditional" matching techniques (e.g. test ID / label based) and the fully blown Detox pilot. Meaning:

- `by.type('list')` // or horizontal-list / vertical-list, or all of those
- `by.type('image')` // or icon
- `by.type('input-field')`

etc.

---

- TODO: This kind of an API really raises the need for an easier disambiguation. This ideas could be introduced further down the road based on more discussions in the original issue (#4145)
- Also related: #4686

noomorph:
## Proposal about possible semantic UI types

Here is my attempt to outline semantic UI types relevant for mobile E2E testing (e.g. with Detox). Below you can find primary names, common aliases, source references (ARIA, HIG, Material, Open UI), and typical mappings to React Native, Android, iOS components, and accessibility roles.

@markdevocht @d4vidi 

---

### 📄 Static Content

* **text**
  Aliases: label, static-text
  Sources: ARIA `text`, iOS `UILabel`, Android `TextView`
  RN: `Text`, `RCTText`, `ReactTextView`
  a11y: `accessibilityRole="text"`, `StaticText` in XCUI

* **image**
  Aliases: image-view, icon
  Sources: ARIA `img`, iOS `UIImageView`, Android `ImageView`
  RN: `Image`, `RCTImageView`, `ReactImageView`
  a11y: `accessibilityRole="image"`, `Image` in XCUI

---

### ✍️ Input Controls

* **text-field**
  Aliases: input, textbox
  Sources: ARIA `textbox`, HIG `UITextField`, Android `EditText`
  RN: `TextInput`, `RCTTextInput`, `ReactEditText`
  a11y: `accessibilityRole="textbox"`, `TextField` in XCUI

* **slider**
  Aliases: seek-bar, range-slider
  Sources: ARIA `slider`, `UISlider`, Android `SeekBar`
  RN: `Slider`, `RCTSlider`, `ReactSlider`
  a11y: `accessibilityRole="adjustable"`, `Adjustable` in XCUI

* **stepper**
  Aliases: incrementor
  Sources: HIG `UIStepper`, Android `+/- Button` pattern
  RN: custom composed buttons
[...]

PR #4793, opened 2025-07-09
Semantic matching basics #4788

markdevocht:
## Description

---

> _For API changes:_

### DIFF
--- a/detox/detox.d.ts
+++ b/detox/detox.d.ts
@@ -1123,10 +1123,17 @@ declare global {
             label(label: string | RegExp): NativeMatcher;
 
             /**
-             * Find an element by native view type.
-             * @example await element(by.type('RCTImageView'));
+             * Find an element by native view type OR semantic type.
+             * Supports both platform-specific class names and cross-platform semantic types.
+             * @example 
+             * // Platform-specific class names:
+             * await element(by.type('RCTImageView')); // iOS
+             * 
+             * // Cross-platform semantic types:
+             * await element(by.type('image'));
              */
-            type(nativeViewType: string): NativeMatcher;
+            type(typeOrSemanticType: string): NativeMatcher;
+
 
             /**
              * Find an element with an accessibility trait. (iOS only)
--- a/detox/src/android/matchers/native.js
+++ b/detox/src/android/matchers/native.js
@@ -1,6 +1,7 @@
 const DetoxRuntimeError = require('../../errors/DetoxRuntimeError');
 const invoke = require('../../invoke');
 const { isRegExp } = require('../../utils/isRegExp');
+const { getClassNamesForSemanticType, getAvailableSemanticTypes } = require('../../utils/semanticTypes');
 const { NativeMatcher } = require('../core/NativeMatcher');
 const DetoxMatcherApi = require('../espressoapi/DetoxMatcher');
 
@@ -31,7 +32,26 @@ class IdMatcher extends NativeMatcher {
 class TypeMatcher extends NativeMatcher {
   constructor(value) {
     super();
-    this._call = invoke.callDirectly(DetoxMatcherApi.matcherForClass(value));
+    // Check if it's a known semantic type first
+    if (getAvailableSemanticTypes().includes(value)) {
+      // It's a semantic type, create matcher for all class names
+      const classNames = getClassNamesForSemanticType(value, 'android');
+      
+      let combinedMatcher = null;
+      for (let i = 0; i < classNames.length; i++) {
+        const matcher = new NativeMatcher(invoke.callDirectly(DetoxMatcherApi.matcherForClass(classNames[i])));
+        combinedMatcher = combinedMatcher ? combinedMatcher.or(matcher) : matcher;
+      }
+      
+      if (!combinedMatcher) {
+        throw new DetoxRuntimeError(`No class names found for semantic type: ${value}`);
+      }
+      
+      this._call = combinedMatcher._call;
+    } else {
+      // Not a semantic type, treat as regular class name
+      this._call = invoke.callDirectly(DetoxMatcherApi.matcherForClass(value));
+    }
   }
 }
 
--- a/detox/src/ios/expectTwo.js
+++ b/detox/src/ios/expectTwo.js
@@ -11,6 +11,7 @@ const { actionDescription, expectDescription } = require('../utils/invocationTra
 const { isRegExp } = require('../utils/isRegExp');
 const log = require('../utils/logger').child({ cat: 'ws-client, ws' });
 const mapLongPressArguments = require('../utils/mapLongPressArguments');
+const { getClassNamesForSemanticType, getAvailableSemanticTypes } = require('../utils/semanticTypes');
 const tempfile = require('../utils/tempfile');
 const traceInvocationCall = require('../utils/traceInvocationCall').bind(null, log);
 
@@ -442,7 +443,17 @@ class Matcher {
 
   type(type) {
     if (typeof type !== 'string') throw new Error('type should be a string, but got ' + (type + (' (' + (typeof type + ')'))));
-    this.predicate = { type: 'type', value: type };
+    // Check if it's a known semantic type first
+    if (getAvailableSemanticTypes().includes(type)) {
+      // It's a semantic type
+      const classNames = getClassNamesForSemanticType(type, 'ios');
+      const predicates = classNames.map(className => ({ type: 'type', value: className }));
+      this.predicate = { type: 'or', predicates };
+    } else {
+      // Not a semantic type, treat as regular class name
+      this.predicate = { type: 'type', value: type };
+    }
+    
     return this;
   }
 
--- a/detox/src/utils/semanticTypes.js
+++ b/detox/src/utils/semanticTypes.js
@@ -1,0 +1,99 @@
+/**
+ * Semantic type mappings for cross-platform component matching
+ */
+
+const SEMANTIC_TYPE_MAPPINGS = {
+  // Images
+  'image': {
+    ios: ['RCTImageView', 'UIImageView'],
+    android: ['android.widget.ImageView', 'com.facebook.react.views.image.ReactImageView']
+  },
+
+  // Input fields
+  'input-field': {
+    ios: ['RCTTextInput', 'RCTMultilineTextInput', 'UITextField', 'UITextView'],
+    android: ['android.widget.EditText', 'com.facebook.react.views.textinput.ReactEditText']
+  },
+
+  // Text elements
+  'text': {
+    ios: ['RCTText', 'UILabel'],
+    android: ['android.widget.TextView', 'com.facebook.react.views.text.ReactTextView']
+  },
+
+  // Button elements
+  'button': {
+    ios: ['RCTButton', 'UIButton'],
+    android: ['android.widget.Button', 'com.facebook.react.views.view.ReactViewGroup']
+  },
+
+  // Scroll containers
+  'scrollview': {
+    ios: ['RCTScrollView', 'UIScrollView'],
+    android: ['android.widget.ScrollView', 'androidx.core.widget.NestedScrollView', 'com.facebook.react.views.scroll.ReactScrollView']
+  },
+
+  // Lists
+  'list': {
+    ios: ['RCTFlatList', 'UITableView', 'UICollectionView'],
+    android: ['android.widget.ListView', 'androidx.recyclerview.widget.RecyclerView', 'com.facebook.react.views.scroll.ReactScrollView']
+  },
+
+  // Switches/Toggles
+  'switch': {
+    ios: ['RCTSwitch', 'UISwitch'],
+    android: ['android.widget.Switch', 'androidx.appcompat.widget.SwitchCompat', 'com.facebook.react.views.switchview.ReactSwitch']
+  },
+
+  // Sliders
+  'slider': {
+    ios: ['RCTSlider', 'UISlider'],
+    android: ['android.widget.SeekBar', 'com.facebook.react.views.slider.ReactSlider']
+  },
+
+  // Picker/Selector
+  'picker': {
+    ios: ['RCTPicker', 'UIPickerView'],
+    android: ['android.widget.Spinner', 'com.facebook.react.views.picker.ReactPickerManager']
+  },
+
+  // Activity indicators/Progress
+  'activity-indicator': {
+    ios: ['RCTActivityIndicatorView', 'UIActivityIndicatorView'],
+    android: ['android.widget.ProgressBar', 'com.facebook.react.views.progressbar.ReactProgressBarViewManager']
+  }
+};
+
+/**
+ * Get platform-specific class names for a semantic type
+ * @param {string} semanticType - The semantic type (e.g., 'image', 'input-field')
+ * @param {string} platform - The platform ('ios' or 'android')
+ * @returns {string[]} Array of class names for the platform
+ */
+function getClassNamesForSemanticType(semanticType, platform) {
+  const mapping = SEMANTIC_TYPE_MAPPINGS[semanticType];
+  if (!mapping) {
+    throw new Error(`Unknown semantic type: ${semanticType}. Available types: ${Object.keys(SEMANTIC_TYPE_MAPPINGS).join(', ')}`);
+  }
+
+  const classNames = mapping[platform];
+  if (!classNames) {
+    throw new Error(`Platform ${platform} not supported for semantic type ${semanticType}`);
+  }
+
+  return classNames;
+}
+
+/**
+ * Get all available semantic types
+ * @returns {string[]} Array of available semantic type names
+ */
+function getAvailableSemanticTypes() {
+  return Object.keys(SEMANTIC_TYPE_MAPPINGS);
+}
+
+module.exports = {
+  SEMANTIC_TYPE_MAPPINGS,
+  getClassNamesForSemanticType,
+  getAvailableSemanticTypes
+};   
--- a/detox/src/utils/semanticTypes.test.js
+++ b/detox/src/utils/semanticTypes.test.js
@@ -1,0 +1,246 @@
+const { getClassNamesForSemanticType, getAvailableSemanticTypes } = require('./semanticTypes');
+
+describe('Enhanced by.type() with semantic types', () => {
+  describe('getClassNamesForSemanticType', () => {
+    it('should return iOS class names for image type', () => {
+      const classNames = getClassNamesForSemanticType('image', 'ios');
+      expect(classNames).toEqual(['RCTImageView', 'UIImageView']);
+    });
+
+    it('should return Android class names for image type', () => {
+      const classNames = getClassNamesForSemanticType('image', 'android');
+      expect(classNames).toEqual(['android.widget.ImageView', 'com.facebook.react.views.image.ReactImageView']);
+    });
+
+    it('should return iOS class names for input-field type', () => {
+      const classNames = getClassNamesForSemanticType('input-field', 'ios');
+      expect(classNames).toEqual(['RCTTextInput', 'RCTMultilineTextInput', 'UITextField', 'UITextView']);
+    });
+
+    it('should return Android class names for input-field type', () => {
+      const classNames = getClassNamesForSemanticType('input-field', 'android');
+      expect(classNames).toEqual(['android.widget.EditText', 'com.facebook.react.views.textinput.ReactEditText']);
+    });
+
+    it('should throw error for unknown semantic type', () => {
+      expect(() => {
+        getClassNamesForSemanticType('unknown-type', 'ios');
+      }).toThrow('Unknown semantic type: unknown-type');
+    });
+
+    it('should throw error for unsupported platform', () => {
+      expect(() => {
+        getClassNamesForSemanticType('image', 'windows');
+      }).toThrow('Platform windows not supported for semantic type image');
+    });
+
+    it('should return iOS class names for scrollview type', () => {
+      const classNames = getClassNamesForSemanticType('scrollview', 'ios');
+      expect(classNames).toEqual(['RCTScrollView', 'UIScrollView']);
+    });
+
+    it('should return Android class names for scrollview type', () => {
+      const classNames = getClassNamesForSemanticType('scrollview', 'android');
+      expect(classNames).toEqual([
+        'android.widget.ScrollView',
+        'androidx.core.widget.NestedScrollView',
+        'com.facebook.react.views.scroll.ReactScrollView'
+      ]);
+    });
+  });
+
+  describe('getAvailableSemanticTypes', () => {
+    it('should return all available semantic types', () => {
+      const types = getAvailableSemanticTypes();
+      expect(types).toEqual([
+        'image',
+        'input-field',
+        'text',
+        'button',
+        'scrollview',
+        'list',
+        'switch',
+        'slider',
+        'picker',
+        'activity-indicator'
+      ]);
+    });
+
+    it('should return an array', () => {
+      const types = getAvailableSemanticTypes();
+      expect(Array.isArray(types)).toBe(true);
+    });
+
+    it('should not be empty', () => {
+      const types = getAvailableSemanticTypes();
+      expect(types.length).toBeGreaterThan(0);
+    });
+  });
+
+  describe('Integration with TypeMatcher (Android)', () => {
+    // Test the logic flow by simulating the matcher behavior
+    it('should differentiate between semantic types and regular class names', () => {
+      // Test semantic type detection
+      expect(getAvailableSemanticTypes().includes('image')).toBe(true);
+      expect(getAvailableSemanticTypes().includes('com.example.CustomView')).toBe(false);
+
+      // Test that semantic types have multiple class names
+      const imageClassNames = getClassNamesForSemanticType('image', 'android');
+      expect(imageClassNames.length).toBeGreaterThan(1);
+      expect(imageClassNames).toContain('android.widget.ImageView');
+      expect(imageClassNames).toContain('com.facebook.react.views.image.ReactImageView');
+    });
+
+    it('should handle semantic type with minimum class names correctly', () => {
+      // Test a semantic type that has the minimum number of class names (2)
+      const imageClassNames = getClassNamesForSemanticType('image', 'android');
+      expect(imageClassNames.length).toBe(2);
+      expect(imageClassNames).toContain('android.widget.ImageView');
+      expect(imageClassNames).toContain('com.facebook.react.views.image.ReactImageView');
+    });
+
+    it('should handle multiple class names for scrollview', () => {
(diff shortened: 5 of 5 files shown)


detox/src/android/matchers/native.js
```diff
@@ -31,7 +32,26 @@ class IdMatcher extends NativeMatcher {
...
+      
+      if (!combinedMatcher) {
+        throw new DetoxRuntimeError(`No class names found for semantic type: ${value}`);
+      }
+      
+      this._call = combinedMatcher._call;
+    } else {
+      // Not a semantic type, treat as regular class name
+      this._call = invoke.callDirectly(DetoxMatcherApi.matcherForClass(value));
+    }
```
d4vidi: Comments feel redundant here, code will be better without

detox/src/ios/expectTwo.js
```diff
@@ -442,7 +443,17 @@ class Matcher {
...
+    if (getAvailableSemanticTypes().includes(type)) {
+      // It's a semantic type
+      const classNames = getClassNamesForSemanticType(type, 'ios');
+      const predicates = classNames.map(className => ({ type: 'type', value: className }));
+      this.predicate = { type: 'or', predicates };
+    } else {
+      // Not a semantic type, treat as regular class name
+      this.predicate = { type: 'type', value: type };
+    }
+
```
d4vidi: Comments feel redundant here, code will be better without

detox/src/android/matchers/native.js
```diff
@@ -31,7 +32,26 @@ class IdMatcher extends NativeMatcher {
...
+    // Check if it's a known semantic type first
+    if (getAvailableSemanticTypes().includes(value)) {
+      // It's a semantic type, create matcher for all class names
+      const classNames = getClassNamesForSemanticType(value, 'android');
+      
+      let combinedMatcher = null;
+      for (let i = 0; i < classNames.length; i++) {
+        const matcher = new NativeMatcher(invoke.callDirectly(DetoxMatcherApi.matcherForClass(classNames[i])));
+        combinedMatcher = combinedMatcher ? combinedMatcher.or(matcher) : matcher;
+      }
```
d4vidi: This would more likely be better with one of the func-programming options (map / reduce / ...)

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Get all available semantic types
+ * @returns {string[]} Array of available semantic type names
+ */
+function getAvailableSemanticTypes() {
+  return Object.keys(SEMANTIC_TYPE_MAPPINGS);
+}
+
+module.exports = {
+  SEMANTIC_TYPE_MAPPINGS,
```
d4vidi: no need to export this
```suggestion
```

detox/src/android/matchers/native.js
```diff
@@ -31,7 +32,26 @@ class IdMatcher extends NativeMatcher {
 class TypeMatcher extends NativeMatcher {
   constructor(value) {
     super();
-    this._call = invoke.callDirectly(DetoxMatcherApi.matcherForClass(value));
+    // Check if it's a known semantic type first
+    if (getAvailableSemanticTypes().includes(value)) {
```
d4vidi:
```suggestion
    if (isSemanticType(value)) {
```
Since the API's scope is very specific, I think it can be better- (and not over-) fitted for its specific usage, which would yielding a more idiomatic code.

detox/detox.d.ts
```diff
@@ -1123,10 +1123,17 @@ declare global {
...
+             * Supports both platform-specific class names and cross-platform semantic types.
+             * @example 
+             * // Platform-specific class names:
+             * await element(by.type('RCTImageView')); // iOS
+             * 
+             * // Cross-platform semantic types:
+             * await element(by.type('image'));
              */
-            type(nativeViewType: string): NativeMatcher;
+            type(typeOrSemanticType: string): NativeMatcher;
```
d4vidi: I know it's going to be redundant but there's going to be great value in having the various semantic types defined as an explicit typescript type.
Something like:

```
            type(typeOrSemanticType: SemanticMatchingTypes | string): NativeMatcher;
```

detox/src/utils/semanticTypes.js
d4vidi, line 1: feels weird to think of this as a "general-purpose" util. it's super-coupled with the Detox matchers, and should therefore reside somewhere that would be more matcher-specific.

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+  getClassNamesForSemanticType,
```
d4vidi: The API names are a bit long. We can probably make them more concise. For example:
```suggestion
  getClasses,
```

And then on the call site:
```diff
- const { getClassNamesForSemanticType, getAvailableSemanticTypes } = require('../../utils/semanticTypes');
+ const semanticTypes = require('../../utils/semanticTypes');

semanticTypes.getClasses(type, platform)
```

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+function getClassNamesForSemanticType(semanticType, platform) {
+  const mapping = SEMANTIC_TYPE_MAPPINGS[semanticType];
+  if (!mapping) {
+    throw new Error(`Unknown semantic type: ${semanticType}. Available types: ${Object.keys(SEMANTIC_TYPE_MAPPINGS).join(', ')}`);
+  }
+
+  const classNames = mapping[platform];
+  if (!classNames) {
+    throw new Error(`Platform ${platform} not supported for semantic type ${semanticType}`);
+  }
```
d4vidi: I'm having doubts about whether these errors are the right approach for this. Perhaps we need a change in granularity, or even consider dropping them. Let's discuss.

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+    android: ['android.widget.SeekBar', 'com.facebook.react.views.slider.ReactSlider']
+  },
+
+  // Picker/Selector
+  'picker': {
+    ios: ['RCTPicker', 'UIPickerView'],
+    android: ['android.widget.Spinner', 'com.facebook.react.views.picker.ReactPickerManager']
+  },
+
+  // Activity indicators/Progress
```
d4vidi: How about adding `progress` as an alias of `activity-indicator`?

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+  'activity-indicator': {
+    ios: ['RCTActivityIndicatorView', 'UIActivityIndicatorView'],
+    android: ['android.widget.ProgressBar', 'com.facebook.react.views.progressbar.ReactProgressBarViewManager']
```
d4vidi: I don't think `ReactProgressBarViewManager` is a view

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+  // Sliders
+  'slider': {
+    ios: ['RCTSlider', 'UISlider'],
+    android: ['android.widget.SeekBar', 'com.facebook.react.views.slider.ReactSlider']
+  },
+
+  // Picker/Selector
+  'picker': {
+    ios: ['RCTPicker', 'UIPickerView'],
+    android: ['android.widget.Spinner', 'com.facebook.react.views.picker.ReactPickerManager']
```
d4vidi: I don't think `ReactPickerManager` is a view

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+  // Switches/Toggles
+  'switch': {
+    ios: ['RCTSwitch', 'UISwitch'],
+    android: ['android.widget.Switch', 'androidx.appcompat.widget.SwitchCompat', 'com.facebook.react.views.switchview.ReactSwitch']
+  },
+
+  // Sliders
+  'slider': {
+    ios: ['RCTSlider', 'UISlider'],
+    android: ['android.widget.SeekBar', 'com.facebook.react.views.slider.ReactSlider']
```
d4vidi: I think ReactSlider inherits SeekBar
Let's revisit the androids in whole list item by item

detox/src/utils/semanticTypes.js
```diff
@@ -0,0 +1,99 @@
...
+  // Text elements
+  'text': {
+    ios: ['RCTText', 'UILabel'],
+    android: ['android.widget.TextView', 'com.facebook.react.views.text.ReactTextView']
+  },
+
+  // Button elements
+  'button': {
+    ios: ['RCTButton', 'UIButton'],
+    android: ['android.widget.Button', 'com.facebook.react.views.view.ReactViewGroup']
```
d4vidi: `ReactViewGroup` is likely a generic base class, and not a button.
In any case, this brings up a bigger question: Is a React button on Android even implemented using a Button, to begin with? IIRC it is a composition of some other views that are not a native button.

If the answer is yes (and maybe even if it's not), we should probably consider introducing more sophisticate techniques in order to identify things based on their "type" (accessibility, maybe).

detox/src/utils/semanticTypes.js
d4vidi: > I think ReactSlider inherits SeekBar

Actually, if the answer is no, we'd have to better understand how tolerant the fundamental class-based matching we rely on is. If we find ourselves having to exact-match `com.facebook.react.view.slider.ReactSlider` (or other **react-package** classes) specifically, it means we'd be overly coupled to the class name and therefore super sensitive to class name changes amongst react versions.

For example, what happens if going from RN80 to RN81 the following change is made:
```diff
- com.facebook.react.view.slider.ReactSlider
+ com.facebook.react.views.slider.ReactSlider
```
or even as a part of a bigger bulk change:
```diff
- com.facebook.react.view.slider.ReactSlider
+ com.meta.react.view.slider.ReactSlider
```

A _strict_ class-based matcher which would enforce the existence of `com.facebook.react.view.slider.ReactSlider` would not allow us to hold both options (old and new class).
This would make it difficult for us to be backwards compatible.

detox/src/utils/semanticTypes.test.js
```diff
@@ -0,0 +1,246 @@
...
+        getClassNamesForSemanticType('image', 'windows');
+      }).toThrow('Platform windows not supported for semantic type image');
+    });
+
+    it('should return iOS class names for scrollview type', () => {
+      const classNames = getClassNamesForSemanticType('scrollview', 'ios');
+      expect(classNames).toEqual(['RCTScrollView', 'UIScrollView']);
+    });
+
+    it('should return Android class names for scrollview type', () => {
```
d4vidi: I think we could implement a for-each over a params matrix (type, platform) and perform a Jest snapshot-match assertion, and that would do it.

detox/src/utils/semanticTypes.test.js
```diff
@@ -0,0 +1,246 @@
...
+        // Both platforms should always return at least one class name
+        expect(androidClassNames.length).toBeGreaterThan(0);
+        expect(iosClassNames.length).toBeGreaterThan(0);
+      });
+    });
+  });
+
+  describe('Integration with Matcher.type() (iOS)', () => {
+    // Simple mock for iOS Matcher class that replicates the actual logic
+    class MockMatcher {
```
d4vidi: This is definitely code-smell

detox/detox.d.ts
noomorph: it is `... | string | {}`, IIRC, if someone wants the autocompletion 🤔

detox/src/android/matchers/native.js
markdevocht: Agreed!

detox/src/utils/semanticTypes.js
markdevocht: moved the semanticTypes.js and its test to src/matchers

detox/src/utils/semanticTypes.js
markdevocht: you are correct, the com.facebook.react.views.slider.ReactSlider inherits from android.widget.SeekBar and thus I have removed com.facebook.react.views.slider.ReactSlider from the list.
Also I build a more complex system to negate some wrong matching because of the inheritance matching. This means that is a user wants to match "activity-indicator" in android we should exclude the "android.widget.AbsSeekBar" because that is definitly not what the user is looking for but the ProgressBar inherits from the AbsSeekBar so it would create a false match.

detox/src/utils/semanticTypes.js
markdevocht: I'm going to leave this, as you suggested, up for discussion as to what else we could do instead of throwing an exception.

detox/test/e2e/27.semantic-types.test.js
```diff
@@ -0,0 +1,78 @@
...
+      .toBeVisible()
+      .withTimeout(2000);
+  });
+
+  it('should support progress alias for activity-indicator', async () => {
+    // Both should work identically
+    await element(by.text('Actions')).tap();
+    // Note: This test assumes the app has progress indicators
+    // In a real test app, you'd verify both 'progress' and 'activity-indicator' work the same
+    expect(() => element(by.type('progress'))).not.toThrow();
```
noomorph: Would not it be much more reasonable to run `expect(...).not.toExist()`? 

Because if you use `element(...)` - we don't even do a roundtrip from Node.js to the device and back.

detox/src/matchers/semanticTypes.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  const classNames = mapping[platform];
+  if (!classNames) {
+    throw new Error(`Platform ${platform} not supported for semantic type ${semanticType}`);
+  }
+
+  return classNames.map(item => {
+    if (typeof item === 'string') {
+      return item;
```
noomorph: > _Disclaimer_: This one is partially AI-generated based on my fiddling with the code locally and prompt to create a comment which explains what I'd like to be changed...

---

I’d like you to consider two related changes that will materially simplify the consumers on both iOS and Android (`expectTwo.js` and `native.js`):

- Normalize `getClasses` to always return `Array<{ className: string, excludes: string[] }>`
- Replace the “unknown semantic type” throw with a normalized pass-through

#### Why this helps in practice

- iOS (`expectTwo.js`) and Android (`native.js`) both currently duplicate logic to branch on mixed shapes (strings vs objects) and to special-case exclusions. With a normalized return, they can be made linear and branch-free.
- If we treat unknown keys as pass-through class names, both platforms can remove the “is it a semantic type?” pre-check and just use one path.

### iOS: concrete simplification in `expectTwo.js`

Today, iOS needs shape checks and fallback handling:

```js
  type(typeOrSemanticType) {
    if (typeof typeOrSemanticType !== 'string') throw new Error('type should be a string, but got ' + (typeOrSemanticType + (' (' + (typeof typeOrSemanticType + ')'))));

    if (semanticTypes.includes(typeOrSemanticType)) {
      const classNames = semanticTypes.getClasses(typeOrSemanticType, 'ios');
      const predicates = classNames.map(item => {
        if (typeof item === 'string') return createTypePredicate(item);
        if (!item.className || !item.excludes) return createTypePredicate(item);
        return createExclusionPredicate(item.className, item.excludes);
      });
      this.predicate = createOrPredicate(predicates);
    } else {
      this.predicate = { type: 'type', value: typeOrSemanticType };
    }

    return this;
  }
```

With a normalized `getClasses` and pass-through on unknowns, this collapses to a single straight path:

```js
type(typeOrSemanticType) {
  if (typeof typeOrSemanticType !== 'string') throw new Error('type should be a string, but got ' + (typeOrSemanticType + (' (' + (typeof typeOrSemanticType + ')'))));

  const matchDescriptors = semanticTypes.getClasses(typeOrSemanticType, 'ios'); // always [{ className, excludes }]
  const predicates = matchDescriptors.map(({ className, excludes }) =>
    excludes.length ? createExclusionPredicate(className, excludes) : createTypePredicate(className)
  );

  this.predicate = predicates.length > 1 ? createOrPredicate(predicates) : predicates[0];
  return this;
}
```

- No `includes()` pre-check.
- No `typeof item` or “missing excludes” branches.
- One mental model: iterate normalized descriptors and map to predicates.

### Android: concrete simplification in `native.js`

Current Android mapping also branches on shape and excludes:

```js
    if (semanticTypes.includes(typeOrSemanticType)) {
      const classNames = semanticTypes.getClasses(typeOrSemanticType, 'android');

      const matchers = classNames.map(item => {
        if (typeof item === 'string') return createClassMatcher(item);
        if (!item.className || !item.excludes) return createClassMatcher(item);

        const includeMatcher = createClassMatcher(item.className);
        const excludeCombined = combineWithOr(item.excludes.map(createClassMatcher));

        return includeMatcher.and(excludeCombined.not);
      });

      const combinedMatcher = combineWithOr(matchers);
      if (!combinedMatcher) {
        throw new DetoxRuntimeError(`No class names found for semantic type: ${typeOrSemanticType}`);
      }
      this._call = combinedMatcher._call;
    } else {
      this._call = invoke.callDirectly(DetoxMatcherApi.matcherForClass(typeOrSemanticType));
    }
```

With normalization and pass-through, this also becomes linear:

```js
const descriptors = semanticTypes.getClasses(typeOrSemanticType, 'android'); // always [{ className, excludes }]

const matchers = descriptors.map(({ className, excludes }) => {
  const includeMatcher = createClassMatcher(className);
  return excludes.length
    ? includeMatcher.and(combineWithOr(excludes.map(createClassMatcher)).not)
    : includeMatcher;
});

const combinedMatcher = combineWithOr(matchers);
if (!combinedMatcher) {
  throw new DetoxRuntimeError(`No class names found for: ${typeOrSemanticType}`);
}
this._call = combinedMatcher._call;
```

- No `includes()` check or string/object branching.
- Exclusion logic is uniformly applied.
- Error handling focuses on “nothing resolved” rather than “was this key in a registry.”

### API contract proposal

- Output: `getClasses(key, platform)` always returns `Array<{ className: string, excludes: string[] }>`; `excludes` may be empty.
- Unknown keys: treat as pass-through literal class names, returning `[ { className: key, excludes: [] } ]`.
- Invalid platform for a known semantic type: still throw (that’s a configuration error).
- Input shapes we expand:
  - `string[]`
  - `{ include: string|string[], exclude?: string[] }`
  - `Array<{ include/exclude }>` (flattened)
  - Anything else: fail early in dev (clear error). This avoids silently leaking malformed descriptors into consumers.

### Why this is better for us

- Less branching and duplication in both `expectTwo.js` and `native.js`.
- A single, predictable data flow → easier to read, test, and extend.
- Callers don’t need to know whether a key is a semantic alias or an actual class; they just resolve-and-iterate.
- Stronger ergonomics for TypeScript/JSDoc and snapshot tests.

detox/src/android/matchers/native.js
```diff
@@ -1,9 +1,16 @@
...
+const semanticTypes = require('../../matchers/semanticTypes');
 const { isRegExp } = require('../../utils/isRegExp');
 const { NativeMatcher } = require('../core/NativeMatcher');
 const DetoxMatcherApi = require('../espressoapi/DetoxMatcher');
 
+const createClassMatcher = (className) =>
+  new NativeMatcher(invoke.callDirectly(DetoxMatcherApi.matcherForClass(className)));
+
+const combineWithOr = (matchers) =>
+  matchers.reduce((acc, matcher) => acc?.or(matcher) ?? matcher, null);
```
noomorph:
```suggestion
  matchers.reduce((acc, matcher) => acc.or(matcher));
```

Since you will have at least 1 element in matchers, better use this shorthand for `reduce`

detox/test/e2e/27.semantic-types.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+  it('should work with waitFor operations', async () => {
+    await waitFor(element(by.type('button')).atIndex(0))
+      .toBeVisible()
+      .withTimeout(2000);
+  });
+
+  it('should support progress alias for activity-indicator', async () => {
+    // Both should work identically
+    await element(by.text('Actions')).tap();
```
noomorph: If those elements don't exist, is there any sense to tap on Actions, though?

detox/detox.d.ts
```diff
@@ -1094,6 +1094,8 @@ declare global {
         interface NativeElement extends NativeElementActions {
         }
 
+        type SemanticMatchingTypes = 'image' | 'input-field' | 'text' | 'button' | 'scrollview' | 'list' | 'switch' | 'slider' | 'picker' | 'activity-indicator' | 'progress';
```
noomorph, line 1159: Maybe `SemanticElementType`? In any case, using plural form for `type` is rather on incorrect side of spectrum, plus `semantic matching type` feels like PR description / implementation detail, not the essence of this type per se.

detox/src/android/matchers/native.js
```diff
@@ -29,9 +36,29 @@ class IdMatcher extends NativeMatcher {
 }
 
 class TypeMatcher extends NativeMatcher {
-  constructor(value) {
+  constructor(typeOrSemanticType) {
     super();
-    this._call = invoke.callDirectly(DetoxMatcherApi.matcherForClass(value));
+    if (semanticTypes.includes(typeOrSemanticType)) {
```
noomorph: I would advise also to leave some metadata like:

```
this._rawType = typeOrSemanticType
```

So that this property travels as JSON to Client (Web socket).
This should help our Allure automatic steps creation to infer human-readable types.

🙏

detox/detox.d.ts
markdevocht: I totally agree with you, no need to 'argue' about this

detox/src/matchers/semanticTypes.js
```diff
@@ -0,0 +1,137 @@
...
+
+  // Button elements
+  'button': {
+    ios: ['UIButton', 'RCTTouchableOpacity', 'RCTTouchableHighlight', 'RCTTouchableWithoutFeedback'],
+    android: ['android.widget.Button', 'android.widget.ImageButton']
+  },
+
+  // Scroll containers - The UITableView inherits from scrollview so it could also show up here...
+  'scrollview': {
+    ios: ['RCTScrollView', 'UIScrollView'],
```
noomorph: @markdevocht looks like we may want to add `RCTScrollViewComponentView`, also... per @ygreking 's report (RN77)

detox/src/android/matchers/native.js
markdevocht: good point!

detox/src/matchers/semanticTypes.js
markdevocht: Added!

detox/src/matchers/semanticTypes.js
markdevocht: logical it makes sense, and since you're a lot more experienced in JS I will do as you command 👍

detox/test/e2e/27.semantic-types.test.js
markdevocht: changed the test all-together.

detox/src/android/matchers/native.js
markdevocht: also added now.

detox/ios/DetoxSync
noomorph, line 1: @markdevocht do you need to change this DetoxSync file?

detox/src/ios/expectTwo.js
```diff
@@ -5,6 +5,24 @@ const fs = require('fs-extra');
 const _ = require('lodash');
 
 
+const semanticTypes = require('../matchers/semanticTypes');
+
+// Functions for semantic type predicate creation
```
noomorph:
```suggestion
```

detox/src/android/matchers/native.test.js
```diff
@@ -0,0 +1,84 @@
+// @ts-nocheck
+// Mock the semanticTypes module before importing anything that depends on it
```
noomorph:
```suggestion

```

detox/src/ios/expectTwo.test.js
```diff
@@ -793,6 +793,49 @@ describe('expectTwo', () => {
...
+
+    it(`should parse correct JSON for semantic type 'image' using by.type()`, async () => {
+      const testCall = await e.element(e.by.type('image')).tap();
+      const jsonOutput = {
+        invocation: {
+          type: 'action',
+          action: 'tap',
+          predicate: {
+            type: 'or',
+            predicates: [
```
noomorph, line 822: I just noticed that you don't assert `rawType`, is there a good reason?

detox/test/src/app.js
```diff
@@ -71,6 +71,7 @@ export default class example extends Component {
         {this.renderScreenButton('Language', Screens.LanguageScreen)}
         {this.renderScreenButton('Sanity', Screens.SanityScreen)}
         {this.renderScreenButton('Matchers', Screens.MatchersScreen)}
+        {this.renderScreenButton('Semantic Types', Screens.SemanticTypesScreen)}
```
noomorph: Be careful regarding screen resolutions - we don't have scroll in the main menu - hence, "too many list items" may break some tests which needs buttons which are already are hidden below the phone frame. Just saying.

noomorph (approved):

detox/src/ios/expectTwo.js
markdevocht: Done

detox/src/android/matchers/native.test.js
markdevocht: Done

detox/src/ios/expectTwo.test.js
markdevocht: added it now :(

detox/ios/DetoxSync
markdevocht: I revert it now.

detox/test/src/app.js
markdevocht: Good point! I added a scrollview to the main page and updated the last 3 tests to scroll to the bottom to reach their buttons. That way we can keep adding more features in the future.

markdevocht:
Created an App icon for the iOS app to make it more recognisable 
image

detox/test/e2e/detox.config.js
```diff
@@ -107,18 +107,9 @@ const config = {
...
-      utilBinaryPaths: ["e2e/util-binary/detoxbutler-1.1.0-aosp-release.apk"],
-      systemUI: {
-        extends: 'genymotion',
-        pointerLocationBar: 'show',
-        touches: 'show',
-        navigationMode: '3-button',
-        statusBar: {
-          clock: '1948',
-        },
+        avdName: 'Pixel_3a_API_35'
```
noomorph: Was that intended, @markdevocht ?

detox/detox.d.ts
noomorph: wait, so SemanticElementType or SemanticElementTypes? did you decide to go with plurals?

detox/test/e2e/detox.config.js
markdevocht: reverted now

noomorph (approved):

Issue #4749, opened 2025-04-02
Detox iOS18 compatibility


d4vidi:
### Describe your idea

Based on a quick, shallow check, seems there are some gaps, though minimal - see #4748. Those are the failing tests:

Consistent across all permutations:
- ✗ _ios_ User Notifications Init from push notification
- ✗ _ios_ WebView CORS (inner frame) detoxDisableWebKitSecurity should find element in cross-origin frame when `detoxDisableWebKitSecurity` is `true`

Failed once:
- ✗ Device _ios_ launchApp in a different language // In 76 new-arch
- ✗ _ios_ Overlay alert window when shown should not be able to tap on elements // In 73 old-arch
---
### Tasks
- [ ] Make tests pass (fix tests / app / Detox)
- [ ] Run consistently on CI

Done list:
- [x] _ios_ WebView CORS (inner frame) detoxDisableWebKitSecurity should find element in cross-origin frame when `detoxDisableWebKitSecurity` is `true`
- [ ] _ios_ User Notifications Init from push notification

manu-abrantes:
The same is happening here. Any updates?

markdevocht:
@d4vidi can you review my PR on this again? I had to roll back the changes you requested because of a crash that it caused.
https://github.com/wix/Detox/pull/4816

d4vidi:
Sure assign to me for the current iteration plz

> @d4vidi can you review my PR on this again? I had to roll back the changes you requested because of a crash that it caused.
> https://github.com/wix/Detox/pull/4816

PR #4816, opened 2025-08-12
Detox iOS18 compatibility - #4749

markdevocht:
Fix for  ios WebView CORS (inner frame) detoxDisableWebKitSecurity should find element in cross-origin frame when detoxDisableWebKitSecurity is true

Needed some new swizzling for iOS 18

### DIFF
--- a/detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
+++ b/detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
@@ -40,16 +40,16 @@ void DTXPreferencesSetWebSecurityEnabled(WKPreferences* prefs, bool enabled) {
     
     if (@available(iOS 18.0, *)) {
         unsigned int ivarCount;
-        Ivar *ivars = class_copyIvarList([WKPreferences class], &ivarCount);
-        if (ivars) {
+        Ivar *ivarList = class_copyIvarList([WKPreferences class], &ivarCount);
+        if (ivarList) {
             for (unsigned int i = 0; i < ivarCount; i++) {
-                const char *ivarName = ivar_getName(ivars[i]);
+                const char *ivarName = ivar_getName(ivarList[i]);
                 if (ivarName && strcmp(ivarName, "_preferences") == 0) {
-                    ivar = ivars[i];
+                    ivar = ivarList[i];
                     break;
                 }
             }
-            free(ivars);
+            free(ivarList);
         }
     } else {
         ivar = class_getInstanceVariable([WKPreferences class], "_preferences");


detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
```diff
@@ -17,60 +25,117 @@ @implementation DTXFakeWKPreferencesRef
 @end
 
 /// Set web-security policy for WebKit (e.g. CORS restriction).
-///
-/// @note Since we can't access the `WKPreferencesSetWebSecurityEnabled` directly with
```
d4vidi, line 21: Wouldn't you rather keep this explanation?

detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
```diff
@@ -17,60 +25,117 @@ @implementation DTXFakeWKPreferencesRef
...
+        
+	} else {
+		DTXFakeWKPreferencesRef* fakeRef = [DTXFakeWKPreferencesRef new];
+		
+		Ivar ivar = class_getInstanceVariable([WKPreferences class], "_preferences");
+		void* realPreferences = (void*)(((uintptr_t)prefs) + ivar_getOffset(ivar));
+		fakeRef._apiObject = realPreferences;
+		
+		WKPreferencesSetWebSecurityEnabled((__bridge WKPreferencesRef)fakeRef, enabled);
+	}
```
d4vidi: slightly on the (styling) nit-picky side here but since `WKPreferencesSetWebSecurityEnabled` is called in both cases, wouldn't it be nicer to implement a `getWKPrefsRef()` kind of function and call `WKPreferencesSetWebSecurityEnabled` with the result?

detox/test/e2e/detox.config.js
```diff
@@ -99,7 +99,7 @@ const config = {
       headless: Boolean(process.env.CI),
       device: {
         type: 'iPhone 15 Pro Max',
-        os: '17.0.1'
+        os: '18.0.0'
```
d4vidi: guess we also need this in `demo-react-native/detox.config.js`

detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
markdevocht: np. updated it to a function.

detox/test/ios/UI/OverlayMessageView.swift
```diff
@@ -8,7 +8,14 @@ import UIKit
...
     private(set) var message: String
-    private let displayDuration: TimeInterval = 2.0
+    
+    private func getDisplayDuration() -> TimeInterval {
+        if #available(iOS 18, *) {
+            return 3.0
+        } else {
+            return 2.0
+        }
+    }
```
d4vidi, line 18: maybe use `3.0` in all versions?

detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.h
```diff
@@ -9,6 +9,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface WKWebViewConfiguration (Detox)
 
+- (BOOL)shouldDisableWebKitSecurity;
```
d4vidi, line 12: 👍🏻

detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
```diff
@@ -16,61 +28,121 @@ @interface DTXFakeWKPreferencesRef: NSObject
...
 ///  purposes on MacOS only. So there's no guarantee that it will work in the future.
 void DTXPreferencesSetWebSecurityEnabled(WKPreferences* prefs, bool enabled) {
-	DTXFakeWKPreferencesRef* fakeRef = [DTXFakeWKPreferencesRef new];
-
-	Ivar ivar = class_getInstanceVariable([WKPreferences class], "_preferences");
-	void* realPreferences = (void*)(((uintptr_t)prefs) + ivar_getOffset(ivar));
-	fakeRef._apiObject = realPreferences;
-
-	WKPreferencesSetWebSecurityEnabled(fakeRef, enabled);
+    if (!prefs) return;
```
d4vidi: Isn't this soft-failing approach dangerous here?

detox/test/ios/UI/OverlayMessageView.swift
markdevocht: I tried that but that doesn't work in iOS 17 or lower, it fails the tests because of that 1 second that the dialog is showing too long.

detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
markdevocht: I don't see it as dangerous, it's actually a fail safe, if (!prefs) basically means if (prefs == nil) in objc.

detox/ios/Detox/Invocation/WKWebViewConfiguration+Detox.m
```diff
@@ -16,61 +28,130 @@ @interface DTXFakeWKPreferencesRef: NSObject
...
+        Ivar *ivars = class_copyIvarList([WKPreferences class], &ivarCount);
+        if (ivars) {
+            for (unsigned int i = 0; i < ivarCount; i++) {
+                const char *ivarName = ivar_getName(ivars[i]);
+                if (ivarName && strcmp(ivarName, "_preferences") == 0) {
+                    ivar = ivars[i];
+                    break;
+                }
+            }
+            free(ivars);
```
d4vidi: - not a fan of `break`, unless commonly used, please avoid using it; is there a functional-programming alternative to this `for` loop? (e.g. array.some())
- `ivars` and `ivar` is really close and a bit confusing

other than that - looks ok 😄

markdevocht: this is true: "not a fan of break, unless commonly used, please avoid using it; is there a functional-programming alternative to this for loop? (e.g. array.some())" BUT in Objc when dealing with low level variable access there is no filter or something alike for it. :disappointed:

As for the ivar and ivars, ok, I'll update one of them. Which one? that's a surprise...

Issue #4767, opened 2025-05-21
device.resetAppState([bundleId])


noomorph:
### Proposal

Introduce a method:

```ts
await device.resetAppState(optionalBundleId);
```

This will:

* **iOS (simulator)**:
  Use `xcrun simctl get_app_container` to locate the installed `.app`, back it up, uninstall the app, then reinstall from the backup — all without triggering a Detox build.

* **Android (emulator)**:
  Either:

  * Use `adb shell pm clear <package>` if sufficient
  * Or: back up the APK path via `adb shell pm path`, uninstall, and reinstall with `adb install`, avoiding any Detox-managed build artifacts.

This enables full app state reset **without rebuild**. Faster, cleaner, reusable in both CI and local dev.

noomorph:
# Important

Okay, I have one more consideration here.

It's best to be able to pass an array of bundle ids, and also if there are multiple apps in the current configuration - make it possible to reset all of them (`true` boolean flag? `ALL` string ?) - because otherwise you risk to have some credentials stuck in one of the adjacent apps from the same vendor.

PR #4819, opened 2025-08-20
feat: device.resetAppState([...bundleIds])

noomorph:
## Description

- This pull request addresses the issue described here: #4767

In this pull request, I have implemented a new `device.resetAppState([...bundleIds])` method that provides a more efficient way to reset app state compared to the traditional uninstall/reinstall cycle. This feature addresses performance concerns when running tests that require clean app state between test runs.

- **Android**: Uses `pm clear` command to clear app data without uninstalling the app
- **iOS**: Implements a backup/restore mechanism which works without need to have app binaries available

**Usage Examples:**
```js
// Reset current app state
await device.resetAppState();

// Reset specific app state
await device.resetAppState('com.example.app');

// Reset multiple apps
await device.resetAppState('com.app1', 'com.app2');
```

---

> _For API changes:_

### DIFF
--- a/detox/detox.d.ts
+++ b/detox/detox.d.ts
@@ -779,7 +779,7 @@ declare global {
              * On Android, this command clears the app's data using the `pm clear` command,
              * effectively resetting the app to its initial installed state without uninstalling it.
              *
-             * On iOS, Detox uses a fallback mechanism - it backs up, deletes and installs the app again.
+             * On iOS, Detox uses a fallback mechanism - it backs up, deletes and installs the app from cache.
              * This process ensures the app is returned to a clean state.
              *
              * @param bundleIds Optional bundle IDs to reset. If none provided, resets the currently selected app.
--- a/docs/api/device.md
+++ b/docs/api/device.md
@@ -163,7 +163,7 @@ Read more in [here](../guide/mocking-user-activity.md). Go back to subsection 1 
 
 #### 6. `resetAppState`—Reset App State Before Launching
 
-Resets the app’s state and launches without necesarily performing a full uninstall/reinstall cycle.
+Resets the app’s state and launches without necessarily performing a full uninstall/reinstall cycle.
 
 - Android: clears app data using `pm clear`, which is typically faster than reinstalling.
 - iOS: uninstalls and reinstalls the app to restore a clean state.


detox/src/devices/runtime/drivers/ios/SimulatorDriver.test.js
```diff
@@ -256,9 +270,30 @@ describe('IOS simulator driver', () => {
...
     });
 
-    it('fails to match a face by passing to AppleSimUtils', async () => {
+    it('fails to match a finger by passing to AppleSimUtils', async () => {
       await uut.unmatchFinger();
       expect(applesimutils.unmatchBiometric).toHaveBeenCalledWith(udid, 'Finger');
     });
   });
+
+  describe('.resetAppState', () => {
```
d4vidi, line 301: well written

detox/src/devices/runtime/drivers/ios/SimulatorDriver.test.js
```diff
@@ -25,6 +29,16 @@ describe('IOS simulator driver', () => {
     const AppleSimUtils = jest.createMockFromModule('../../../common/drivers/ios/tools/AppleSimUtils');
     applesimutils = new AppleSimUtils();
 
+    // Mock AppStateResetShim
+    AppStateResetShimMock = jest.requireMock('./AppStateResetShim');
+    appStateResetShimInstance = {
```
d4vidi: IIRC this should work the same way
```suggestion
    appStateResetShimInstance = new AppStateResetShimMock()
```

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
+const os = require('os');
+const path = require('path');
+
+const fs = require('fs-extra');
+
+const DetoxRuntimeError = require('../../../../errors/DetoxRuntimeError');
+const log = require('../../../../utils/logger').child({ cat: 'device' });
+
+class AppStateResetShim {
```
d4vidi: So just to be sure, the reason this is a Shim and not a plain Helper / delegate / ... is that it compensates over the fact you can't really reset apps on iOS? If so, it kind of makes sense but would make even more sense if this would have been something you can do today but haven't always been able to 🤔 

`*thud*` (the noise of me falling into the naming rabbit hole)

detox/src/devices/runtime/RuntimeDevice.js
```diff
@@ -121,8 +122,13 @@ class RuntimeDevice {
 
     if (params.delete) {
       await this.terminateApp(bundleId);
-      await this.uninstallApp();
-      await this.installApp();
+
+      if (this._behaviorConfig.optimizeReinstall) {
```
d4vidi: so this is kind of a means to allow for a force-everyone-into-doing-this yet somehow get-out-of-jail-free card situation? 😅

This goes to say: I _do_ understand the power behind this, on the one hand - but on the other, the `optimizeInstall` flag itself is too implicit to my taste to begin with, and it's usage here makes things even more so: Every fairly explicit usage of the `delete: true` arg by a detox user would in fact be replaced by something else under the hood. I'd rather not be such a user when things go wrong.

Maybe I'm paranoid, but at least try to think whether perhaps there could be a way we can allow for a more verbose and straightforward usage of this technique?

detox/detox.d.ts
```diff
@@ -773,6 +780,28 @@ declare global {
...
+            /**
+             * Resets the app state by clearing app data and restoring it to a clean state.
+             *
+             * On Android, this command clears the app's data using the `pm clear` command,
+             * effectively resetting the app to its initial installed state without uninstalling it.
+             *
+             * On iOS, Detox uses a shim to back up, delete, and restore the app's data.
+             * This process ensures the app is returned to a clean state.
+             *
+             * @param bundleIds Optional bundle IDs to reset. If none provided, resets the current app.
```
d4vidi:
```suggestion
             * @param bundleIds Optional bundle IDs to reset. If none provided, resets the currently selected app.
```

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+  constructor(applesimutils) {
+    this.applesimutils = applesimutils;
+    this.tempDir = null;
+    this.backedUpApps = new Map(); // bundleId -> tempPath
+  }
+
+  async backupApp(udid, bundleId) {
+    if (!this.tempDir) {
```
d4vidi: maybe do this only after getAppContainer succeeds?

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+    try {
+      for (const bundleId of bundleIds) {
+        await this.backupApp(udid, bundleId);
+      }
+
+      for (const bundleId of bundleIds) {
+        await this.restoreApp(udid, bundleId);
+      }
+    } finally {
+      await this.cleanup();
```
d4vidi: This shouldn't be here, or should it? Why back up (`applesimutils.getAppContainer`) in multiple calls to resetMultipleApps if the "container" has already been cached? can it be done in a more teardown global scope? (I know its tricky but have to ask...)

detox/src/devices/runtime/drivers/ios/SimulatorDriver.test.js
```diff
@@ -256,9 +270,30 @@ describe('IOS simulator driver', () => {
       expect(applesimutils.matchBiometric).toHaveBeenCalledWith(udid, 'Finger');
     });
 
-    it('fails to match a face by passing to AppleSimUtils', async () => {
+    it('fails to match a finger by passing to AppleSimUtils', async () => {
```
d4vidi: 👍🏻
(lol)

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+    await this.applesimutils.uninstall(udid, bundleId);
+
+    // Reinstall from backup
+    await this.applesimutils.install(udid, tempAppPath);
+  }
+
+  async cleanup() {
+    if (this.tempDir) {
+      try {
+        await fs.remove(this.tempDir);
```
d4vidi: maybe put this as the last call? just to be safe

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+      this.tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'detox-apps-'));
+    }
+
+    const appContainerPath = await this.applesimutils.getAppContainer(udid, bundleId);
+    if (!appContainerPath) {
+      throw new DetoxRuntimeError(`Could not find app container for bundle ID: ${bundleId}`);
+    }
+
+    const tempAppPath = path.join(this.tempDir, `${bundleId}-${Date.now()}`);
+    await fs.copy(appContainerPath, tempAppPath);
```
d4vidi: I wonder, if it's the same exact `.app` file - can't we use the container path as-is?

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
d4vidi, line 1: Hey cursor, please remove all the fluff (i.e. comments) :-)

I think I have this as a cursor rule, might push it to the repo

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+    };
+
+    shim = new AppStateResetShim(applesimutils);
+  });
+
+  afterEach(async () => {
+    // Clean up any temp directories created during tests
+    if (tempDir && await fs.pathExists(tempDir)) {
+      await fs.remove(tempDir);
+    }
```
d4vidi: oh, damn. well... can you use fsext.js instead and then mock the filesystem layer, as we conventionally do?

d4vidi: btw that could make a good addition to some cursor rules

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+    it('should create temp directory and backup app', async () => {
+      const udid = 'test-udid';
+      const bundleId = 'com.test.app';
+      const mockAppPath = __dirname; // Use current directory as mock app path
+
+      applesimutils.getAppContainer.mockResolvedValue(mockAppPath);
+
+      const result = await shim.backupApp(udid, bundleId);
+
+      expect(result).toContain('detox-apps-');
```
d4vidi: `toEqual`

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+      const bundleId1 = 'com.test.app1';
+      const bundleId2 = 'com.test.app2';
+      const mockAppPath = __dirname;
+
+      applesimutils.getAppContainer.mockResolvedValue(mockAppPath);
+
+      const result1 = await shim.backupApp(udid, bundleId1);
+      const result2 = await shim.backupApp(udid, bundleId2);
+
+      expect(path.dirname(result1)).toBe(path.dirname(result2));
```
d4vidi: `toEqual`

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  async backupApp(udid, bundleId) {
+    if (!this.tempDir) {
+      this.tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'detox-apps-'));
+    }
+
+    const appContainerPath = await this.applesimutils.getAppContainer(udid, bundleId);
+    if (!appContainerPath) {
+      throw new DetoxRuntimeError(`Could not find app container for bundle ID: ${bundleId}`);
```
d4vidi: I doubt any user could make sense of this, how about rephrasing to something more actionable?

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+    await fs.copy(appContainerPath, tempAppPath);
+    this.backedUpApps.set(bundleId, tempAppPath);
+
+    return tempAppPath;
+  }
+
+  async restoreApp(udid, bundleId) {
+    const tempAppPath = this.backedUpApps.get(bundleId);
+    if (!tempAppPath) {
+      throw new DetoxRuntimeError(`No backup found for bundle ID: ${bundleId}`);
```
d4vidi: this message is better than backupApp()'s but can also be slightly improved IMO

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
```diff
@@ -0,0 +1,73 @@
...
+class AppStateResetShim {
+  constructor(applesimutils) {
+    this.applesimutils = applesimutils;
+    this.tempDir = null;
+    this.backedUpApps = new Map(); // bundleId -> tempPath
+  }
+
+  async backupApp(udid, bundleId) {
+    if (!this.tempDir) {
+      this.tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'detox-apps-'));
```
d4vidi: can make usage of tempfile.js or at least `tmp` here?

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+      applesimutils.getAppContainer.mockResolvedValue(mockAppPath);
+
+      // First backup the app
+      await shim.backupApp(udid, bundleId);
+
+      // Then restore it
+      await shim.restoreApp(udid, bundleId);
+
+      expect(applesimutils.uninstall).toHaveBeenCalledWith(udid, bundleId);
+      expect(applesimutils.install).toHaveBeenCalledWith(udid, expect.any(String));
```
d4vidi: `expect.any(String)` is not good enough here, as we need to make sure the right path is used

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+      await expect(shim.backupApp('test-udid', 'com.test.app'))
+        .rejects.toThrow('Could not find app container for bundle ID: com.test.app');
+    });
+  });
+
+  describe('restoreApp', () => {
+    it('should uninstall and reinstall app from backup', async () => {
+      const udid = 'test-udid';
+      const bundleId = 'com.test.app';
+      const mockAppPath = __dirname;
```
d4vidi: Better to use a mocked path here, this doesn't feel very safe

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+
+      // Verify the backup was stored
+      expect(shim.backedUpApps.get(bundleId)).toBe(result);
+    });
+
+    it('should reuse existing temp directory for multiple apps', async () => {
+      const udid = 'test-udid';
+      const bundleId1 = 'com.test.app1';
+      const bundleId2 = 'com.test.app2';
+      const mockAppPath = __dirname;
```
d4vidi: should be defined as common consts. having this in each test is too noisy IMO

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+      const udid = 'test-udid';
+      const bundleIds = ['com.app1', 'com.app2'];
+      const mockAppPath = __dirname;
+
+      applesimutils.getAppContainer.mockResolvedValue(mockAppPath);
+
+      // Spy on the methods and disable their actual implementations
+      const backupSpy = jest.spyOn(shim, 'backupApp').mockResolvedValue();
+      const restoreSpy = jest.spyOn(shim, 'restoreApp').mockResolvedValue();
+      const cleanupSpy = jest.spyOn(shim, 'cleanup').mockResolvedValue();
```
d4vidi: This is too white-box if you ask me. The test should verify that the function does what it needs to do as a bottom line. Instead, this just checks there are some internal function calls...

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+    it('should cleanup even if backup fails', async () => {
+      const udid = 'test-udid';
+      const bundleIds = ['com.app1'];
+      const mockAppPath = __dirname;
+
+      applesimutils.getAppContainer.mockResolvedValue(mockAppPath);
+
+      // Mock backupApp to fail
+      const backupSpy = jest.spyOn(shim, 'backupApp').mockRejectedValue(new Error('Backup failed'));
+      const cleanupSpy = jest.spyOn(shim, 'cleanup').mockResolvedValue();
```
d4vidi: again, too white-box

detox/src/devices/runtime/drivers/ios/AppStateResetShim.test.js
```diff
@@ -0,0 +1,213 @@
...
+      const udid = 'test-udid';
+      const bundleIds = ['com.app1'];
+      const mockAppPath = __dirname;
+
+      applesimutils.getAppContainer.mockResolvedValue(mockAppPath);
+
+      // Mock backupApp to succeed but restoreApp to fail
+      const backupSpy = jest.spyOn(shim, 'backupApp').mockResolvedValue();
+      const restoreSpy = jest.spyOn(shim, 'restoreApp').mockRejectedValue(new Error('Restore failed'));
+      const cleanupSpy = jest.spyOn(shim, 'cleanup').mockResolvedValue();
```
d4vidi: again, too white-box

d4vidi (changes requested):
Summary:
1. I don't think the global `optimizeReinstall` flag is a good approach.
2. Generated unit test code quality is too poor. Preferably that can be fixed using good-enough cursor rules, but otherwise please let's try to avoid allowing AI with leaving us like that.
3. Sporadic suggestions

detox/src/devices/runtime/RuntimeDevice.js
noomorph: Okay, I'll `params.resetAppState` instead.

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
noomorph: good call.

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
noomorph: Indeed. Hmm... this should be better:

```js
throw new DetoxRuntimeError({
  message: `App with bundle ID '${bundleId}' is not installed on this device (${udid}). Please install the app first before attempting to reset its state.`,
  hint: `To check apps installed on the device, use: xcrun simctl listapps ${udid}`,
});
```

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
noomorph: Well, if I understand correctly what you're asking, then **no**. After the uninstallation, the container path becomes invalid. I cannot know what was the original app path and I must always doubt whether the app is identical to the simulator's app contents or not. The paths of installed apps on simulator change with each reinstall, so I absolutely need to call `applesimutils.getAppContainer` and to copy the directory elsewhere.

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
noomorph: So far, I see that the best thing would be to rename it to `AppStateResetFallback`. 

Indeed, `-Shim` naming presumes that some functionality is available in newer versions and can be emulated (shimmed) in the older ones. It is not 100% accurate — we have no idea whether such functionality will appear or not. :]

detox/src/devices/runtime/drivers/ios/AppStateResetShim.js
noomorph: Sure, good idea.

erdemkiiliic:
@noomorph Hey man, First of all, great work. While checking, I noticed something:

```
await device.launchApp({
  resetAppState: true,
  newInstance: true,
  permissions: {
    userTracking: 'YES',
    notifications: 'YES',
    camera: 'YES',
    location: 'always',
    microphone: 'YES',
    contacts: 'YES',
  },
  launchArgs,
});
```
with this usage, it resets the behavior as expected, but I see that the permissions cannot be re-granted. Similarly, when using { delete: true }, I see that the app starts with the permissions granted. I also saw this note in the documentation:
‘Similar to {@link Detox.DeviceLaunchAppConfig.delete | { delete: true }}, but instead of uninstalling and installing the app,’

However, I wanted to point out that it doesn’t seem to work in the same way. Wouldn’t it be better if the behavior were made similar to the delete option? Thanks.

PR #4825, opened 2025-09-01
CI: Use Genymotion ARM64 only for latest RN build job

d4vidi:
## Description

Optimize the usage of ARM64 genymotion devices in our CI, until it becomes more available as a resource.
Consider this an improvement over #4818


noomorph (approved):
Looks legit

PR #4844, opened 2025-10-03
fix(android): grant all permissions after resetting app state

noomorph:
## Description

- This pull request addresses the issue introduced in: #4819

In this pull request, I have added extra calls to adb shell `pm grant --all-permissions` to imitate our `adb install -g` calls for normal case.


d4vidi (commented):
This is all really cool so it makes it more difficult to be breaking the bad news -

In where Detox for Android is right now, we just always install apps with all permissions granted in advance (see the `-g` flag in the [ADB install function](https://github.com/wix/Detox/blob/master/detox/src/devices/common/drivers/android/exec/ADB.js#L122)) -

Image

So it's really great we can allow for permissions granularity within a reasonable effort but unfortunately, first we're going to have to undo the full-permissions as a default (thus addressing a years outstanding long feature request).

I suggest that for this specific scenario, we'd just stick to just granting all permissions right after the reset, keeping things as seamless as possible.

erdemkiiliic:
Is there an update on this issue? @noomorph

noomorph:
Mmm, I suggest this approach, @d4vidi

- **If app config has a permissions map:**
  Do **not** use `-g` with `adb install`.
  Instead, explicitly grant each permission using `adb shell pm grant`.
- **If app config has no permissions:**
  Use `-g` (grant all permissions upon install).
  In `resetAppState`, run `adb shell pm grant --all-permissions` as a fallback.

Speaking Bash:

```bash
if [ -n "$APP_PERMISSIONS_MAP" ]; then
  # Install without -g
  adb install "$APK_PATH"
  # Grant each permission explicitly
  for perm in $APP_PERMISSIONS_MAP; do
    adb shell pm grant "$PACKAGE_NAME" "$perm"
  done
else
  # Install with -g
  adb install -g "$APK_PATH"
  # Fallback: grant all permissions in resetAppState
  adb shell pm grant "$PACKAGE_NAME" --all-permissions
fi
```

- We'll have compatibility with the previous logic
- Officially we won't recommend it as a a way to go, because click on system dialog may be a bit challenging (ui automator perhaps?)
- If someone needs `device.resetAppState()` with older device, they'll have to do some legwork and add permissions as a map

PR #4837, opened 2025-09-24
Support React Native 0.81

Maschga:
## Background
I need React Native `0.81.x` support for Expo SDK 54.

## Description
- Based on #4812

In this pull request, I have added support for React Native version `0.81.x`.

---

> _For features/enhancements:_
 - [ ] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files.

> _For API changes:_
 - [ ] I have made the necessary changes in the [types index](https://github.com/wix/Detox/blob/master/detox/index.d.ts) file.

PR #4848, opened 2025-10-09
fix: windows

Maschga:
## Description

With this setup, running commands like `npm run XXX` works on Windows.
If support for Windows is not desired, then the PR can be closed.

DONE:
- [x] `require.resolve()` -> `require()`
- [x] `.gradlew` -> `gradlew`
- [x] running scripts with `sh` prefix
- [x] `;` -> `&&`
- [x] `cd -` -> `cd ..`

TODO:
- [ ] testing on Linux


noomorph:
Hi, @Maschga , which flows did not work for you - what practical issues for yourself as a contributor you were trying to fix? I assume most of the work was around launching test project app, right? Could you elaborate a bit more, please?

P.S. Also: cmd? powershell? wsl? wsl2? cygwin? mingw? I mean to clarify for which environment you tried to solve the problems... 🤔  IIRC, there is a gazillion of ways to cook node.js project on Windows, so some specificity won't hurt.

Maschga:
This PR is related to https://github.com/wix/Detox/pull/4837#issuecomment-3371364711.
Basically every workflow including the `./gradlew` command did not work.
In addition, running `npm run test:android-debug` / `npm run e2e:android-debug` resulted in a `ERR_UNSUPPORTED_ESM_URL_SCHEME` error.

I use Windows 11 with Git Bash with Node.js `v22.20.0`.
Please let me know, if you need further information.

gosha212:
@Maschga But this way it will stop working on linux / mac

Maschga:
What exactly doesn't work? Which error messages do you get?
Maybe we can find a good solution.

gosha212:
all executables on mac / linux should be executed with ./xxx

Maschga:
Got a nice fix for Windows user:
Running
```
npm config set script-shell "C:\Program Files\Git\git-bash.exe"
```
sets the npm-shell to Git Bash (assuming Git is installed) in the user directory (`~/.npmrc`).
Everytime I run `npm run XXX` it opens a Git Bash Shell in which the script is executed.

PR #4863, opened 2025-10-29
SystemUI configuration on init, for Android

d4vidi:
## Description

In this pull request, I have introduced the ability to set up system UI via Detox's config.

TODO
- [ ] (Maybe) Separate keyboard from systemUI → systemBehavior
- [x] Documentation 
- [x] Test coverage
- [x] Override vars using `null`
- [x] Init cache optimization
- [ ] Revisit setStatusBar API (Backlogged: #4870)
- [x] unit tests over systemUI CFG helper

---

> _For API changes:_


detox/src/devices/allocation/drivers/android/AndroidAllocDriver.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * @typedef {import('../AllocationDriverBase').AllocationDriverBase} AllocationDriverBase
+ * @typedef {import('../../../common/drivers/DeviceCookie').DeviceCookie} DeviceCookie
+ */
+
+const log = require('../../../../utils/logger').child({ __filename });
+const sleep = require('../../../../utils/sleep');
+
+const DeviceInitCache = require('./utils/DeviceInitCache');
```
noomorph, line 9: This looks like should be an inner state of AndroidAllocDriver, no?

detox/src/devices/allocation/drivers/android/utils/SystemUICfgHelper.js
```diff
@@ -0,0 +1,180 @@
...
+const presets = require('./systemUICfgPresets');
+
+const batteryPrecents = {
+  full: 100,
+  half: 50,
+  low: 20,
+};
+
+const navigationModes = {
+//  '2-button': 'com.android.internal.systemui.navbar.twobutton',
```
noomorph, line 20: Can we have some context why this is commented out? Even if with good intention, it is bad to leave without explanation why you thought it's better to leave this information inside the code...

detox/src/devices/allocation/drivers/android/AndroidAllocDriver.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * @param {DeviceCookie & { adbName: string }} deviceCookie
+   * @param {Detox.DetoxSharedAndroidDriverConfig} deviceConfig
+   * @private
+   */
+  async _setupSystemUI(deviceCookie, deviceConfig) {
+    const { adbName } = deviceCookie;
+    const adbWrapper = {
```
noomorph: mmmm.... looks a bit weird, but oh..kay 😬 😄

detox/src/devices/allocation/drivers/AllocationDriverBase.d.ts
```diff
@@ -8,7 +8,7 @@ export interface DeallocOptions {
 export interface AllocationDriverBase {
   init?(): Promise<void>;
   allocate(deviceConfig: any): Promise<DeviceCookie>;
-  postAllocate?(deviceCookie: DeviceCookie): Promise<DeviceCookie | void>;
+  postAllocate?(deviceCookie: DeviceCookie, configs?: { deviceConfig: Detox.DetoxDeviceConfig }): Promise<DeviceCookie | void>;
```
noomorph, line 11: Shall we make it public or we can actually keep that internal?

detox/src/devices/allocation/drivers/android/AndroidAllocDriver.js
d4vidi: Sure, can do 👍🏻

detox/src/devices/allocation/drivers/android/utils/SystemUICfgHelper.js
d4vidi: It's been specified in the types - https://github.com/wix/Detox/pull/4863/files#diff-4644014ea4b16de519e21260b0d8a11d9ff6fbbcc5f3ccc88dc3a48ff9743738R379
But sure I'll add it here too, why not 😄

detox/src/devices/allocation/drivers/android/AndroidAllocDriver.js
d4vidi: Yeah it definitely ought to be helper-internal.

detox/src/devices/allocation/drivers/AllocationDriverBase.d.ts
d4vidi: Well we pass it in right now no matter where the driver comes from, so why not? 🤔

d4vidi:
> A few nitpicks, but overall I have a positive impression. 😊 See my previous comments.

np. Thanks, it was actually important. I'm on it.

Issue #4874, opened 2025-11-20
WebSocket error when native overlays appear over WebView during test execution


cunhabruno:
### Did you test using the latest Detox?

- [x] I have tested this issue on the latest Detox release and it still reproduces.

### Did your test fail and you're not sure why?

- [x] I have read the troubleshooting guide and it didn't help me.

### What happened?

When running E2E tests that interact with WebView elements and then trigger native modals/drawers to open, Detox throws an unexpected WebSocket error that crashes the test. for example:

```
Unexpected error on an attempt to handle the response received over the web socket.

Unexpected message received over the web socket: invokeResult

The payload was:
{"messageId":9,"type":"invokeResult","params":{}}
```

Steps to Reproduce

1. Execute a WebView action (e.g., scroll, waitfor) that triggers asynchronous operations
2. The action causes a native modal/drawer to open over the WebView
3. WebView operations complete and send invokeResult messages back through WebSocket
4. Detox has already switched context to native layer
5. WebSocket handler rejects the invokeResult message as unexpected

Expected Behavior
Detox should gracefully handle invokeResult messages that arrive after the test context has switched from WebView to native layer

Actual Behavior
The test crashes with a `DetoxRuntimeError` because the WebSocket message handler doesn't expect invokeResult messages when there's no corresponding inflight promise.

Environment

- Detox version: 20.41.2
- Platform: iOS
- Test framework: Jest
- React Native version: 0.77.3

### In what environment did this happen?

Detox version: 
React Native version: 
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 
Device model: 
iOS version: 
macOS version: 
Xcode version: 
Test-runner (select one): jest / other

### Detox logs

<details><summary>Detox logs</summary>

```
paste logs here!
```

</details>

### Device logs

<details><summary>Device logs</summary>

```
paste your device.log here!
```

</details>

noomorph:
@cunhabruno could you show us an example of such payload and some more context, please? I wonder whether this problem may require a bit different fix at all than #4875 suggests...

dacudo:
This is still an issue. I see it in an iOS test where I tap a WebElement then a NativeElement then a WebElement. These steps are done a few times in the test and it fails on the third time. The same test passes on Android.

noomorph:
@dacudo any chance you could contribute some test that can help us to reproduce this in detox/test/e2e suite, please 🙏  ?

dacudo:
[...]

PR #4875, opened 2025-11-21
feat(session): add ignoreUnexpectedMessages configuration option

cunhabruno:
## Description

- This pull request addresses the issue described here: #4874 

Add configuration option to allow users to control whether Detox should throw or log warnings when receiving unexpected WebSocket messages. This is useful for applications with complex view hierarchies (e.g., React Native apps with WebViews) where legitimate messages might be received in unexpected contexts.

---

> _For API changes:_

### DIFF
--- a/detox/src/client/AsyncWebSocket.js
+++ b/detox/src/client/AsyncWebSocket.js
@@ -15,7 +15,7 @@ const DEFAULT_SEND_OPTIONS = {
 };
 
 class AsyncWebSocket {
-  constructor({ url, ignoreUnexpectedMessages }) {
+  constructor({ url, ignoreUnexpectedMessages = false }) {
     this._url = url;
     this._ws = null;
     this._eventCallbacks = {};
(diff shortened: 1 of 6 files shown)


detox/src/client/AsyncWebSocket.js
```diff
@@ -15,14 +15,16 @@ const DEFAULT_SEND_OPTIONS = {
...
+  constructor(url, options = {}) {
     this._url = url;
     this._ws = null;
     this._eventCallbacks = {};
     this._messageIdCounter = 0;
     this._opening = null;
     this._closing = null;
     this._abortedMessageIds = new Set();
+    this._ignoreUnexpectedMessages = options.ignoreUnexpectedMessages ?? 
+      (process.env.DETOX_IGNORE_UNEXPECTED_WS_MESSAGES === 'true');
```
noomorph: we have collectCliConfig for this, we don't use process.env directly.

noomorph (approved):

noomorph:
@d4vidi can you take a look here? It seems that I cannot merge :D so your review is welcome.

d4vidi:
@noomorph I've removed myself from the reviewers list, how about now?

PR #4881, opened 2025-12-12
build: move to yarn 4

noomorph:
## Description

Internal change. Moving from `npm` and `lerna` to `yarn`.


scripts/utils/publishNewVersion.js
```diff
@@ -44,19 +46,52 @@ function projectSetup() {
...
-    log('SKIP NPM is set: Lerna-publishing without publishing to NPM');
+
+  // Bump versions in all packages
+  const preid = versionType.includes("pre") ? `--preid ${npmTag}` : '';
+
+  // Update version in root package.json
+  exec.execSync(`yarn version ${versionType} ${preid}`);
+
+  // Update versions in workspace packages
+  const packages = ['detox', 'detox-cli'];
```
d4vidi: This is a bit too implicit to my taste, and is bound to go out of sync.
Is there a way this can be done differently in order to maintain a single source of truth (lerna.json style, or a custom property inside the various package.json files)?

.gitignore
```diff
@@ -232,6 +240,8 @@ detox/test/artifacts
 
 Detox-android/
 Detox-ios-src.tbz
+Detox-ios-framework.tbz
```
dev-john-nguyen, line 243: When I upgraded to 20.46.1, I'm getting this error below. I'm thinking it could be this `.gitignore`.
```
Error opening archive: Failed to open '~node_modules/detox/Detox-ios-framework.tbz'
```

Reverting back to 20.46.0 works though!

noomorph: @dev-john-nguyen thanks for the heads up. Check 20.46.3 please

PR #4894, opened 2025-12-30
iOS 26 compatibility

markdevocht:
Detox iOS 26 compatibility + setting simulators to iOS 26.0.1

### DIFF
--- a/detox/test/e2e/06.device.test.js
+++ b/detox/test/e2e/06.device.test.js
@@ -1,7 +1,6 @@
 describe('Device', () => {
   it('reloadReactNative - should tap successfully', async () => {
     await device.reloadReactNative();
-    await waitFor(element(by.text('Sanity'))).toBeVisible();
     await element(by.text('Sanity')).tap();
     await element(by.text('Say Hello')).tap();
     await expect(element(by.text('Hello!!!'))).toBeVisible();
@@ -9,7 +8,6 @@ describe('Device', () => {
 
   it('relaunchApp - should tap successfully', async () => {
     await device.relaunchApp();
-    await waitFor(element(by.text('Sanity'))).toBeVisible();
     await element(by.text('Sanity')).tap();
     await element(by.text('Say Hello')).tap();
     await expect(element(by.text('Hello!!!'))).toBeVisible();
@@ -17,7 +15,6 @@ describe('Device', () => {
 
   it('relaunchApp({delete: true}) - should tap successfully', async () => {
     await device.relaunchApp({delete: true});
-    await waitFor(element(by.text('Sanity'))).toBeVisible();
     await element(by.text('Sanity')).tap();
     await element(by.text('Say Hello')).tap();
     await expect(element(by.text('Hello!!!'))).toBeVisible();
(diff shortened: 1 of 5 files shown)


detox/test/e2e/06.device.test.js
```diff
@@ -1,20 +1,23 @@
 describe('Device', () => {
   it('reloadReactNative - should tap successfully', async () => {
     await device.reloadReactNative();
+    await waitFor(element(by.text('Sanity'))).toBeVisible().withTimeout(2000);
```
noomorph: What happened here? Is there a problem with synchronization?

scripts/upload_artifact.sh
```diff
@@ -44,9 +44,9 @@ generate_allure_report() {
   local ROOT_DIR=$1
 
   cd $ROOT_DIR
-  if [ -d "allure-results" ]; then
+  if [ -d "artifacts" ]; then
```
noomorph: Perhaps, this one needs to be reverted.

noomorph: Looks like a leftover from our experiments.

detox/test/e2e/06.device.test.js
markdevocht: The problem is not with synchronization but the operating system's hookup change. It takes longer for Detox at startup to get hooked. So we had to have the notifications show 2 seconds longer (which is still lower than the timeout of a notification in iOS) and because of these notifications covering some of the top labels on the main screen we also needed to introduce a timeout. We already had some logic like this in the device.test.js `await device.launchApp({newInstance: true, languageAndLocale});,` but that was set to 1000 milliseconds with the description of "// iOS toast is hiding the element".

scripts/upload_artifact.sh
markdevocht: ok, reverted :)

detox/test/ios/AppDelegate.m
```diff
@@ -121,7 +126,9 @@ - (void)handleChangeScreen:(NSNotification *)notification {
 }
 
 - (void)applicationDidBecomeActive:(NSNotification *)notification {
-    [self showOverlayMessageWithMessage:@"Active"];
+    if (demoShowActiveNotification) {
+        [self showOverlayMessageWithMessage:@"Active"];
```
noomorph, line 130: So, if now, the "Active" notification is not a problem, why did not you remove those "waitFor" workarounds? 🤔

scripts/upload_artifact.sh
```diff
@@ -46,7 +46,7 @@ generate_allure_report() {
   cd $ROOT_DIR
   if [ -d "allure-results" ]; then
     set -x
-    allure generate && upload_to_surge allure-report "Detox Allure Report"
+    allure generate artifacts && upload_to_surge allure-report "Detox Allure Report"
```
noomorph: Looks like a leftover (non-reverted something) from my previous attempts to help you with artifacts on Buildkite.

markdevocht: reverted

noomorph (approved):
You have my seal of approval:

PR #4900, opened 2026-01-06
Android: Split (google) emulators across different adb servers

d4vidi:
## Description

This pull request aims to greatly increase stability in Android (google) emulator devices in a multi-devices environment. It addresses the following problem (no issue reported so I'm writing it  here) -

### Problem

With local emulators being shifted back into the main focus, we've recently started noticing on numerous environments (local, CI) that with Detox running over more than 1 emulator concurrently, the ADB server gets "lost". Namely it often cannot handle concurrently executed ADB commands, thus resulting in frequent ADB command failures. This phenomenon has become more noticeable with the recent introduction of the automated device status-bar configuration feature.

### Solution

The ADB ↔ emulator framework supports multiple ADB servers and so this PR introduces a shift to an architecture by which each dispatched emulator gets its own ADB server, listening on a dedicated port.

Initializing an ADB server with a port can be done using either the `-P <port>` argument or the `ANDROID_ADB_SERVER_PORT` env var.

Starting up an emulator while specifying a custom ADB server port can be done strictly via the `ANDROID_ADB_SERVER_PORT` env var.

This solution prefers the `-P` arg when possible - since it's more verbose, but otherwise resorts to the env-var. This takes into account the fact emulator allocation is performed by the primary context - which is fundamentally a single process, and the runtime execution of tests commands is performed by worker - each running in its own process.

### Debts

- [ ] **Strict port allocation.** For now, we work inside the open-range starting from the default ADB server port (5037) + 1 (i.e. 5038). We implicitly assume the 5037+ ports are either free or strictly used by adb servers.
- [ ] Apply this technique to **Geny-cloud and attached** (real) Android devices?
  Note: Didn't do this because I haven't looked into whether there exists a (clean?) way to connect those devices onto specific servers (i.e. specify the port).
- [x] Protect using a dedicated "toggle" since this has breaking potential (in some envs)

---

> _For API changes:_


detox/src/devices/allocation/drivers/android/emulator/FreeEmulatorFinder.test.js
```diff
@@ -18,24 +16,20 @@ describe('FreeEmulatorFinder', () => {
     mockDeviceRegistry.getTakenDevicesSync.mockImplementation(() => fakeDeviceList);
 
     const FreeEmulatorFinder = require('./FreeEmulatorFinder');
+    const mockAdb = /** @type {any} */ ({});
```
d4vidi: remove

detox/src/devices/allocation/drivers/android/emulator/EmulatorAllocDriver.js
```diff
@@ -146,6 +171,41 @@ class EmulatorAllocDriver extends AndroidAllocDriver {
...
+  /**
+   * @param {boolean} useSeparateAdbServers
+   * @returns {Promise<number[]>}
+   * @private
+   */
+  async _getRunningAdbServers(useSeparateAdbServers = true) {
+    const ports = [this._adb.defaultServerPort];
+
+    if (useSeparateAdbServers) {
+        for (let port = this._adb.defaultServerPort + 1; await isPortTaken(port); port++) {
```
noomorph: What's the guarantee that the port is taken by an **ADB server** actually?

detox/src/devices/allocation/drivers/android/emulator/EmulatorAllocDriver.js
```diff
@@ -146,6 +171,41 @@ class EmulatorAllocDriver extends AndroidAllocDriver {
...
+    return ports;
+  }
+
+  _getFreeAdbServerPort(currentDevices, useSeparateAdbServers) {
+    if (!useSeparateAdbServers) {
+      return this._adb.defaultServerPort;
+    }
+    
+    const maxPortDevice = _.maxBy(currentDevices, 'adbServerPort');
+    return _.get(maxPortDevice, 'adbServerPort', this._adb.defaultServerPort) + 1;
```
noomorph: Do you check that it is actually free?

detox/src/devices/allocation/drivers/android/emulator/EmulatorAllocDriver.js
noomorph: Can't we just enumerate ports from `adbPortRegistry`?

noomorph (changes requested):
I feel a bit uneasy about this change. That `process.env.*` mutation to enable third party reporters – I see where you are coming from, but it is, to say the least, very puzzling situation... The tradeoff is real - one worker = one device at time, but yes, we can allow to accumulate a bit more tech debt. I'm more puzzled that after 4-5 years of using Android emulators outside gmsaas we never caught issues like that... if the flakiness appears specifically near device allocation (status bar tweaks), did you double check that there's no way something else is missing? 

Overall, the PR is straightforward enough - you augmented device cookie, now you use also adb server port. The port allocation seems fishy to me in the current implementation - on CI scale, the risk is high enough that something will occupy presumably an ADB'ish port, and are you well equipped to handle this case? 🤔 

The port allocation logic seems to be suspicious enough that I'd ask you to stop and look, please. 🙏

d4vidi:
@noomorph thanks for the review! I'd rather we talk in person but in any case here is where I was coming from with this solution:

- Yes indeed, the port allocation solution is as naive as it gets. Nevertheless, we've always run under the assumption that port 5037 is free and there were no issues with that; As a matter of fact, this port is used not just by detox users, but by any Android / react-native developer out there that has ever run an emulator. So this puts that assumption on the very safe side IMO. That is why I was hesitant to try to introduce a more robust soltuion (but applied of a protection toggle nonetheless).
- That said, the alternatives I did have in mind are in these 2 directions:
  - Random port
  - Scanning for free ports within a constant range (e.g. 5038-5067 - thus also implicitly limiting the amount of concurrent adb servers)
- process.env -- well, I'm not overly fond of it myself but I didn't have a better idea. I'm not sure it's that bad.

duckdum:
Hey @d4vidi  — we're running into exactly the issue this PR addresses. We parallelize Android Detox tests with 3 emulators (--maxWorkers 3) and see frequent random failures caused by ADB contention:

```
adb: error: listener 'tcp:61045' not found
Detox can't seem to connect to the test app(s)!
```

The shared ADB server intermittently drops adb reverse port mappings, which causes unhandled promise rejections in ADB.reverseRemove() that get attributed to running tests by Jest-circus — failing tests that are otherwise fine. We traced the full code path and opened a small patch for the symptom (#4927), but this PR would fix the root cause.

Is there a plan to move this forward? Happy to help test it on our CI if that's useful.

d4vidi:
Hey @duckdum. I'm no longer with the project. Perhaps the current maintainers could pick this up @noomorph @markdevocht @yedidyak

PR #4912, opened 2026-02-22
fix(ios): respect atIndex in getAttributes

### LINKED ISSUE #4633: atIndex() has no effect over getAttributes()
### What happened?

The idiomatic way of getting the attributes of a specific element (e.g. at index 1) from a matching list is:
```js
const matches = await element(by.type('android.widget.TextView')).getAttributes());
const element = matches[1];
```

But the following technique is also valid, technically:
```js
const element = await element(by.type('android.widget.TextView')).atIndex(1).getAttributes();
```

Seems the latter works on Android, but not iOS. Instead, `atIndex()` is ignored and the entire list of elements is returned.

The API on iOS should behave the same way as on Android.

- [ ] I have followed the instructions under [Identifying which synchronization mechanism causes us to wait too much](https://wix.github.io/Detox/docs/troubleshooting/synchronization#identifying-which-synchronization-mechanism-causes-us-to-wait-too-much).

_No response_

Detox version: 
React Native version: 
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 
Device model: 
iOS version: 
macOS version: 
Xcode version: 
Test-runner (select one): jest / other

### Detox logs

N/A

### Device logs

N/A

### More data, please!

_No response_

vazidmansuri005:
## Description

- Resolves #4633
- `element(...).atIndex(N).getAttributes()` on iOS was ignoring the `atIndex()` modifier and returning attributes for all matching elements instead of just the one at the specified index.

### Root cause

The `attributes` computed property in `Element.swift` never checked `self.index`. When multiple elements matched the predicate, it unconditionally mapped over all views and returned them in an `elements` array — regardless of whether `atIndex()` was specified.

### Fix

Added index-aware logic to the `attributes` property, matching the pattern already used by the `view` property (which correctly handles `self.index`).

| Scenario | Before | After |
|----------|--------|-------|
| `atIndex(N).getAttributes()` with multiple matches | Returns `{ elements: [...] }` (ignores index) | Returns single element attributes at index N |
| `getAttributes()` with single match | Returns single element attributes | No change |
| `getAttributes()` without `atIndex`, multiple matches | Returns `{ elements: [...] }` | No change |

### Tests

- Added e2e tests in `33.attributes.test.js` for `atIndex` + `getAttributes` on iOS (both legacy and new arch)
- Tests verify that when `atIndex` is used, the result is a single element object (no `elements` array wrapper)


vazidmansuri005:
Hi @d4vidi @noomorph 👋 
Could you please review this PR when you get a chance? Thanks!

PR #4927, opened 2026-04-01
fix(android): handle missing adb reverse mapping in reverseRemove

duckdum:
## Description

This pull request addresses an issue with parallel Android emulator test execution where `ADB.reverseRemove()` throws an unhandled error that causes random test failures.

### Problem

When running Detox tests with multiple Android emulators (`--maxWorkers 2`+), tests randomly fail with:

```
Command failed: "adb" -s emulator-5554 reverse --remove tcp:61045
adb: error: listener 'tcp:61045' not found
```

followed by:

```
Detox can't seem to connect to the test app(s)\!
```

### Root Cause

With multiple emulators sharing a single ADB server, the server occasionally drops the `adb reverse` port mapping on one emulator. When this happens:

1. The instrumentation process loses its connection to the Detox test server and terminates
2. The `close` event on the child process fires, triggering `Instrumentation._onTerminated()` → `MonitoredInstrumentation._onInstrumentationTerminated()` → the `AndroidDriver` termination function
3. The termination function calls `ADB.reverseRemove()`, which tries `adb reverse --remove tcp:<port>` — but the mapping is already gone, so it throws `"listener not found"`
4. Since the `close` event callback is `async` but never `await`ed, this becomes an **unhandled promise rejection**
5. Jest-circus captures unhandled rejections via `process.on('unhandledRejection')` and attributes them to whichever test is currently running — **causing that test to fail** even though the error is unrelated to test logic

The emulator itself recovers fine on the next `device.launchApp()` call (which re-creates the reverse mapping). The only damage is the collateral test failure from the unhandled rejection.

### Fix

Catch the `"not found"` error in `ADB.reverseRemove()` since removing an already-absent mapping is a benign no-op. All other errors are still thrown.

This is consistent with how other ADB methods in this file handle expected failure cases (e.g., `getState()` catching "device not found", `grantAllPermissions()` catching "no permission specified", `getFileSize()` handling "No such file or directory").

### Relation to #4900

[PR #4900](https://github.com/wix/Detox/pull/4900) addresses the broader ADB instability by giving each emulator its own ADB server. This PR is complementary — it provides resilience at the `reverseRemove` level regardless of the ADB server architecture. Even with per-emulator ADB servers, a mapping could still be absent during cleanup if the emulator crashed, so this defensive check is valuable in either case.

---

- N/A (bug fix)

> _For API changes:_
- N/A (internal behavior change only)


specialturtle:
Forge has reproduced this on Android API 35 with Detox 20.51.3: all product assertions passed, then async instrumentation teardown called reverseRemove after the mapping had already disappeared. ADB returned listener tcp:<port> not found and the rejected cleanup escaped Jest, failing the otherwise-passing run. Failure evidence: https://github.com/Turtleprojects/Forge/pull/2118#issuecomment-4952623054

Our temporary compatibility guard suppresses only the exact listener-shaped ADB error (adb: error: listener tcp:<numeric port> not found) and rethrows every other ADB failure, with regression cases for success, the known missing-listener result, and unrelated failures. Could this upstream fix use the same listener-specific match and add a regression test? The current message.includes(not found) can also hide unrelated command/device/path failures. Once an equivalent fix is released, Forge will adopt it, run three exact-head API 35 soaks, and remove the guard.

PR #4930, opened 2026-04-14
Replace execa with tinyexec

stianjensen:
## Description

Drop execa (large dependency tree) in favor of tinyexec ^0.3.0 (zero dependencies, dual CJS/ESM). Uses args-tokenizer to split command strings since tinyexec expects (command, args[]) rather than a single command string.

tinyexec 1.x is ESM-only, so I opted to use the 0.3.x line for now (since my understanding is Detox still supports node versions that don't support require(esm)).

---

> _For API changes:_


noomorph:
Superseded by #4956, which retains your commit in history — it'll land to `master` on merge. Thanks for the contribution @stianjensen!

stianjensen:
Thanks! Just happy to see it landing 😄

Issue #3198, opened 2022-01-31
Any plans to add support for Android bio enrolment? 


jamesmcn1:
We've recently had a bug in production which was related to bio authentication which this only affect android users. We have a suite of bio auth detox tests for iOS but having this for android aswell would be amazing and would have probs prevented this bug from going out into the wild.

Any plans to add support for Android bio enrollment in the future?

d4vidi:
Hey there! I'm accepting this as a feature request. We don't have the resources to introduce this any time soon, but we'll try to sort something out. Of course, we highly appreciate contributions.

PR #4934, opened 2026-04-21
feat(android): add biometrics support for Android emulators

limpbrains:
## Description

- This pull request addresses the issue described here: #3198

In this pull request, I have added Android emulator support for Detox's biometrics API. Existing iOS
behavior is unchanged. Reviewed and tested end-to-end on a real emulator before submission.

- `device.setBiometricEnrollment(bool)` and `device.matchFinger()` / `unmatchFinger()` now work on
Android emulators (AVDs), using the Virtual Fingerprint HAL.
- `device.matchFace()` / `unmatchFace()` work on Android emulators via the Virtual Face HAL, opt-in
with `setBiometricEnrollment(true, { androidFace: true })`. The face path reboots the emulator on
first activation (~20s); subsequent calls short-circuit (~3s).
- Attached devices and Genymotion Cloud throw clear `"only supported on Android emulators (AVDs)"`
errors with actionable hints.
- Requires a userdebug AVD (standard Android Studio images).

## Test plan

- Unit: 3217 tests pass; new coverage across `ADB.test.js`, `AndroidDriver.test.js`, new `EmulatorDriver.test.js`, `RuntimeDevice.test.js`.
- E2E: 8/8 pass on a real Pixel 3a API 34 userdebug AVD (`detox/test/e2e/14.android-biometrics.test.js`), covering both fingerprint and face paths including idempotent re-enable.
- Backwards compatible — existing `setBiometricEnrollment(bool)` callers and all iOS behavior unchanged.

---

- [x] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files — `docs/api/device.md`.

- [x] I have made the necessary changes in the [types index](https://github.com/wix/Detox/blob/master/detox/index.d.ts) file — `detox/detox.d.ts` (new `BiometricEnrollmentOptions` type + refreshed JSDoc; `detox/index.d.ts` is a pass-through wrapper so no direct edit was needed).

---

_Code developed by Claude_

### DIFF
--- a/detox/detox.d.ts
+++ b/detox/detox.d.ts
@@ -697,6 +697,23 @@ declare global {
         type Digit = 0 | DigitWithoutZero;
         type BatteryLevel = `${Digit}` | `${DigitWithoutZero}${Digit}` | "100";
 
+        /**
+         * Options for {@link Device.setBiometricEnrollment}.
+         *
+         * On iOS, these options have no effect: iOS simulators enroll both Face ID and Touch ID together.
+         * On Android, the default modality is fingerprint (Virtual Fingerprint HAL); pass `androidFace: true`
+         * to instead enroll the Virtual Face HAL — note this requires a userdebug AVD and incurs an emulator
+         * reboot (~30s) due to the Virtual Face HAL's feature-flag + sensor-props setup.
+         */
+        interface BiometricEnrollmentOptions {
+            /**
+             * Android emulator only. When `true`, `setBiometricEnrollment` sets up the Virtual Face HAL
+             * (enables the feature flag, sets sensor type/strength, reboots the emulator, then enrolls a
+             * virtual face). When `false` or omitted, the Virtual Fingerprint HAL path is used (no reboot).
+             */
+            androidFace?: boolean;
+        }
+
         interface Device {
             /**
              * Holds the environment-unique ID of the device, namely, the adb ID on Android (e.g. emulator-5554) and the Mac-global simulator UDID on iOS -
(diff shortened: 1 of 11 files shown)


limpbrains:
### Questions for maintainers

A few design decisions I'd like to double-check before merging — happy to revise any of these:

1. **AVD-only scope.** Physical devices and Genymotion Cloud throw `"only supported on Android emulators (AVDs)"`. This is because the Virtual Fingerprint / Face HAL and `adb emu finger` console commands don't exist on real devices, and Genycloud's adb tunnel doesn't reliably pass through them either. Is AVD-only the right scope, or is there a real-device path (e.g., via `cmd fingerprint enroll` on rooted Android) you'd want supported too?

2. **`{ androidFace: true }` option name.** I chose an explicit Android-prefixed flag to make the reboot cost and platform scope obvious at the call site. Would you prefer something more neutral like `{ modality: 'face' }` (ignored on iOS since iOS enrolls both), or the current Android-scoped shape?

3. **Face HAL reboot + `launchApp` requirement.** The first-time face enable on a fresh AVD reboots the emulator (Virtual Face HAL needs the feature flag + secure setting + sensor props applied before boot). The user must call `device.launchApp({ newInstance: true })` afterward. Subsequent calls short-circuit. Is this UX acceptable, or would you want `EmulatorDriver` to re-launch the app transparently?

4. **`device_config set_sync_disabled_for_tests persistent`.** Needed so the face feature flag survives the reboot. It's a global AVD-state mutation that `setBiometricEnrollment(false, {androidFace: true})` doesn't revert — documented in `docs/api/device.md`. Should the disable path reset it, or is the documented caveat enough?

5. **Test-app UI coverage.** The e2e tests validate the Detox-API layer only (commands issue successfully, errors thrown where expected). There's no native `BiometricPrompt` screen in `detox/test` to verify app-observable effects. Would you like that added in this PR, or is it a reasonable follow-up?

cc @noomorph

PR #4935, opened 2026-04-23
chore: replace glob with tinyglobby

stianjensen:
## Description

glob pulls in a heavy dependency tree (inflight, once, minimatch, etc.). tinyglobby is a modern, lightweight alternative with minimal dependencies (fdir + picomatch) and ships dual CJS/ESM exports.

Detox only uses basic glob patterns (*, simple extensions) with a cwd option, so the migration is a straightforward drop-in replacement with no behavioral differences for our use cases.

- Replace glob with tinyglobby in temporaryPath.js, DetoxLogFinalizer test, and e2e artifactUtils
- Remove glob from dependencies


noomorph:
Hi, I'll see if we can take these dependency PRs next week.

aleksanb:
Can this be reconsidered for merge?

Issue #4937, opened 2026-04-30
ADB commands fail when device serial contains parentheses (mDNS/Wi-Fi)


yuki2006:
## Description

When an Android device is connected via Wi-Fi/mDNS, the ADB device serial can contain parentheses, e.g.:

```
adb-5721009297-Rq3U4s (2)._adb-tls-connect._tcp
```

This causes all Detox tests to fail with:

```
/bin/sh: -c: line 0: syntax error near unexpected token `('
```

## Root Cause

In `src/devices/common/drivers/android/exec/ADB.js` line 391, the device serial is interpolated without quoting:

```js
const serial = `${deviceId ? `-s ${deviceId}` : ''}`;
const cmd = `"${this.adbBin}" ${serial} ${params}`;
```

The shell interprets `(` as a subshell start, causing a parse error.

Note: `adbCmdSpawned` (line 400) is **not** affected because it passes arguments as an array to `spawn`, bypassing shell interpretation.

## Fix

Quote the device serial:

```js
const serial = `${deviceId ? `-s "${deviceId}"` : ''}`;
```

## Environment

- Detox: 20.47.0
- Device: Physical Android connected via Wi-Fi (mDNS)
- ADB serial: `adb-5721009297-Rq3U4s (2)._adb-tls-connect._tcp`
- Host: macOS

PR #4950, opened 2026-05-18
fix(android): quote adb device serials

omribz156:
## Description

- This pull request addresses the issue described here: #4937

In this pull request, I have quoted the device serial used by shell-built ADB commands. This keeps mDNS/Wi-Fi serials with characters like parentheses from being parsed by the shell, while leaving the spawned ADB paths unchanged because they already pass the serial as an argument array.

## Verification

- `corepack yarn workspace detox jest src/devices/common/drivers/android/exec/ADB.test.js --runInBand`
- `corepack yarn workspace detox eslint src/devices/common/drivers/android/exec/ADB.js src/devices/common/drivers/android/exec/ADB.test.js`
- `git diff --check`

---

> _For API changes:_

This was implemented with Codex assistance, with the patch kept to the shell-command serial quoting path and reviewed against the existing ADB unit tests.

### DIFF
--- a/detox/src/devices/common/drivers/android/exec/ADB.js
+++ b/detox/src/devices/common/drivers/android/exec/ADB.js
@@ -388,7 +388,7 @@ class ADB {
   }
 
   async adbCmd(deviceId, params, options = {}) {
-    const serial = `${deviceId ? `-s ${deviceId}` : ''}`;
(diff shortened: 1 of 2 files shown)


noomorph (approved):
Good catch and a well-scoped fix — the quoted serial handles all the common special characters that trip up Wi-Fi/mDNS serials, and the spawned paths are correctly left untouched. Thanks for the contribution!

PR #4951, opened 2026-05-20
fix(types): allow regex in toHaveText

omribz156:
## Description

- This pull request addresses the issue described here: #4914

This updates the native `toHaveText` TypeScript declaration to accept `RegExp` as well as `string`. The Android expectation path already has runtime coverage for `toHaveText(/text/)`, so this keeps the public typings aligned with existing behavior and adds type-test coverage for both global and module import styles.

---

 - [ ] I have added/updated the relevant references in the [documentation](https://github.com/wix/Detox/tree/master/docs) files. Not applicable: this is a typings parity fix for already-covered runtime behavior.

> _For API changes:_

## Verification

- `yarn install --immutable --mode=skip-build` (completed with existing peer dependency warnings)
- `yarn --cwd detox/test test:types`
- `yarn --cwd detox/test eslint types/detox-global-tests.ts types/detox-module-tests.ts`
- `yarn --cwd detox jest src/android/AndroidExpect.test.js --runInBand`
- `git diff --check`

This was implemented with Codex assistance, with the patch kept focused and manually reviewed against the existing matcher/type path.

### DIFF
--- a/detox/detox.d.ts
+++ b/detox/detox.d.ts
@@ -1516,9 +1516,10 @@ declare global {
             /**
              * In React Native apps, expect UI component of type <Text> to have text.
              * In native iOS apps, expect UI elements of type UIButton, UILabel, UITextField or UITextViewIn to have inputText with text.
-             * @example await expect(element(by.id('mainTitle'))).toHaveText('Welcome back!);
+             * @example await expect(element(by.id('mainTitle'))).toHaveText('Welcome back!');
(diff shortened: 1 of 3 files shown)


noomorph:
Thanks for the contribution! The type change is correct for Android, but `toHaveText` with regex wouldn't have worked on iOS as-is — I had to add the native implementation there too. Superseded by #4955 which includes the full cross-platform support.

